PR #98, opened 2014-12-15
Card devexp_416 - Support arbitrary images for STI builders

soltysh:
This PR contains:
- [x] Added type for sti script names.
- [x] Removed the need for sti-helper. The untarring is done on the fly when calling assemble script. 
- [x] Added additional parameter that can steer the location of untarred contents.

Still to be done:
- [x] Testing

Nonetheless comments to the build process are welcome since this part is ready.
//cc @bparees @csrwng @smarterclayton


pkg/sti/executor.go
```diff
@@ -119,23 +130,121 @@ func (h *requestHandler) execute(command string) error {
...
+	// goroutine to stream container's error
+	go func(reader io.Reader) {
+		scanner := bufio.NewScanner(reader)
+		for scanner.Scan() {
+			text := scanner.Text()
+			if glog.V(1) {
+				glog.Errorf(text)
+			}
+			if h.errorChecker != nil && h.errorChecker.wasExpectedError(text) {
+				expectedError += text + "; "
```
csrwng: If we find an error that matches, do we need to keep appending additional errors to the expectedError string?

bparees: the idea was to report as much information to the user as possible so they don't just fix one error and immediately hit another.

csrwng: Got it... the concern was that we would create a really large string if we have a very long stderr output. Maybe we then limit it by length? (ie. if len(expectedError) < SizeLimit && h.errorChecker != nil && ... )

bparees: @soltysh and I felt fairly optimistic that this wouldn't be an issue, but sure an upper bound couldn't hut.

soltysh: @csrwng will add, let's say 1kB limit, is that ok?

csrwng: sounds good

pkg/sti/build.go
```diff
@@ -73,40 +78,59 @@ func (b *Builder) Build() (*Result, error) {
...
+		}
+	} else if err != nil {
 		return nil, err
 	}
 
 	return bh.Result(), nil
 }
 
-func (h *buildHandler) PostExecute(containerID string, cmd []string) error {
+func (h *buildHandler) wasExpectedError(text string) bool {
```
smarterclayton: This is horribly hacky - please add a comment detailing why this method exists and try to limit what people use it for.

pkg/sti/api/script.go
```diff
@@ -0,0 +1,20 @@
+package api
+
+// Script defines script names used by STI process.
+type Script string
```
mfojtik: why is this needed and you can't use just string?

pkg/sti/build.go
```diff
@@ -53,18 +57,19 @@ func newBuildHandler(req *Request) (*buildHandler, error) {
...
 // An error represents a failure performing the build rather than a failure
 // of the build itself.  Callers should check the Success field of the result
 // to determine whether a build succeeded or not.
-func (b *Builder) Build() (*Result, error) {
+func (b *Builder) Build() (*api.Result, error) {
 	bh := b.handler
 	defer bh.cleanup()
 
-	err := bh.setup([]string{"assemble", "run"}, []string{"save-artifacts"})
+	err := bh.setup([]api.Script{api.Assemble, api.Run}, []api.Script{api.SaveArtifacts})
```
mfojtik: to me it is more readable `api.AssembleScript` and `api.RunScript`

pkg/sti/build.go
```diff
@@ -73,40 +78,59 @@ func (b *Builder) Build() (*Result, error) {
...
+		glog.Warningf("Image %s does not have tar! Performing additional build to add the scripts and sources.",
+			bh.Request().BaseImage)
+		if err := bh.build(); err != nil {
+			return nil, err
+		}
+		glog.V(2).Infof("Building %s using sti-enabled image", bh.Request().Tag)
+		if err := bh.execute(api.Assemble); err != nil {
+			return nil, err
+		}
+	} else if err != nil {
```
mfojtik: this should be a switch or type switch

pkg/sti/docker/docker.go
```diff
@@ -129,26 +148,34 @@ func (d *stiDocker) RemoveContainer(id string) error {
...
-func (d *stiDocker) GetDefaultScriptsURL(name string) (string, error) {
-	imageMetadata, err := d.CheckAndPull(name)
-	if err != nil {
-		return "", err
-	}
-	defaultScriptsURL := ""
-	env := append(imageMetadata.ContainerConfig.Env, imageMetadata.Config.Env...)
+// getVariable gets environment variable's value from the image metadata
+func (d *stiDocker) getVariable(image *docker.Image, name string) string {
+	envName := fmt.Sprintf("%s=", name)
```
mfojtik: why not just `envName := name + "="`

pkg/sti/docker/docker.go
```diff
@@ -166,11 +193,40 @@ func (d *stiDocker) RunContainer(opts RunContainerOptions) (err error) {
...
 	} else {
-		cmd = append(cmd, opts.Command)
+		// for internal scripts we can have separate path for scripts and untar operation destination
+		scriptsURL := opts.ScriptsURL
+		if len(scriptsURL) == 0 {
+			scriptsURL = d.getVariable(imageMetadata, ScriptsURL)
+		}
+		commandBaseDir = strings.TrimPrefix(scriptsURL, "image://")
+		glog.V(2).Infof("Base directory for STI scripts is '%s'", commandBaseDir)
+		glog.V(2).Infof("Untarring destination is '%s'", tarDestination)
```
mfojtik: merge these two into one

pkg/sti/errors/errors.go
```diff
@@ -159,12 +161,12 @@ func NewDefaultScriptsURLError(err error) error {
...
+func NewContainerError(name string, code int, expected string) error {
 	return ContainerError{
-		Message:    fmt.Sprintf("non-zero (%d) exit code from %s", code, name),
-		Details:    err,
-		ErrorCode:  ErrStiContainer,
-		Suggestion: "check the container logs for more information on the failure",
-		ExitCode:   code,
+		Message:       fmt.Sprintf("non-zero (%d) exit code from %s", code, name),
+		ExpectedError: expected,
+		ErrorCode:     ErrStiContainer,
```
mfojtik: ErrSTIContainer

pkg/sti/executor.go
```diff
@@ -119,23 +130,121 @@ func (h *requestHandler) execute(command string) error {
...
+	return err
 }
 
-func (h *requestHandler) PostExecute(containerID string, cmd []string) (err error) {
+func (h *requestHandler) build() error {
+	// create Dockerfile
+	buffer := bytes.Buffer{}
+	location := h.request.Location
+	if len(location) == 0 {
+		location = "/tmp"
```
mfojtik: make this a const (`defaultLocation`)

pkg/sti/api/script.go
soltysh: This was done to remove all those strings from the code, of course I could switch all the string into constants, but this is more readable IMHO.

pkg/sti/build.go
soltysh: The type is already called `Script` so I don't see any value in renaming those vars.

pkg/sti/build.go
soltysh: Nah, I've tried writing this with type switch and it looks ugly, I prefer this solution if you don't mind, more readable.

pkg/sti/build.go
soltysh: Unfortunately current logic hides the from the build part the docker internals,  my initial idea here was to log the entire stderr and return that to user with `ContainerError`. But @csrwng complained about that idea, so I came up with this solution, but frankly said I'm partially satisfied with this. So any suggestions to make it better are welcome. In the mean time I've added necessary comment to that method.

pkg/sti/executor.go
soltysh: Done.

soltysh:
Updated the PR, waiting to hear once more from you guys.

soltysh:
Addressed @mfojtik's [comment](https://github.com/openshift/ruby-20-centos/pull/7#discussion_r22587383). Need @mfojtik and @csrwng signoff to merge this.

pkg/sti/api/script.go
mfojtik: I don't think this will improve readability and personally I don't like masking a 'string' to 'Script' type as it makes me search how the 'Script' type is defined. I would rather prefer to have generic string and then have constants for the script names (which seems more idiomatic to me). @smarterclayton thoughts?

csrwng:
LGTM

soltysh:
Rebased on top of #103 waiting for tests to pass.

soltysh:
With @mfojtik approval [merge]

PR #104, opened 2015-01-09
Added STI image generator

mfojtik:
This is an attempt to provide a generator for new STI images. As an author of a new STI image, I don't want to spend time creating and copying content to STI scripts / etc but I want to have a boilerplate I can start work with.

```
mkdir new-image
sti create openshift/new-image ./new_image
```

This will create `.sti` folder with sample STI scripts, Dockerfile, `test/run` script and Makefile.

### DIFF
--- a/cmd/sti/main.go
+++ b/cmd/sti/main.go
@@ -92,6 +92,24 @@ func newCmdBuild(req *api.Request) *cobra.Command {
 	return buildCmd
 }
 
+func newCmdCreate() *cobra.Command {
+	return &cobra.Command{
+		Use:   "create <imageName> <destination>",
+		Short: "Bootstrap a new STI image repository",
+		Long:  "Bootstrap a new STI image with given imageName inside the destination directory",
+		Run: func(cmd *cobra.Command, args []string) {
+			if len(args) < 2 {
+				cmd.Help()
+				os.Exit(1)
+			}
+			b := sti.NewCreate(args[0], args[1])
+			b.AddSTIScripts()
+			b.AddDockerfile()
+			b.AddTests()
+		},
+	}
+}
+
 func newCmdUsage(req *api.Request) *cobra.Command {
 	usageCmd := &cobra.Command{
 		Use:   "usage <image>",
@@ -169,6 +187,7 @@ func main() {
 	stiCmd.AddCommand(newCmdVersion())
 	stiCmd.AddCommand(newCmdBuild(req))
 	stiCmd.AddCommand(newCmdUsage(req))
+	stiCmd.AddCommand(newCmdCreate())
 	setupGlog(stiCmd.PersistentFlags())
 
 	err := stiCmd.Execute()
--- a/pkg/sti/create/templates/docker.go
+++ b/pkg/sti/create/templates/docker.go
@@ -1,0 +1,45 @@
+package templates
+
+const Dockerfile = `
+# {{.ImageName}}
+FROM centos:centos7
+
+# TODO: Install required packages:
+#
+# RUN yum install -y --enablerepo=centosplus epel-release <packages>
+
+
+# Add STI usage script
+ADD ./.sti/bin/usage /usage
+
+# You can add any other required files/configurations here
+# ADD conf /app-root/conf
+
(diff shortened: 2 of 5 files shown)


mfojtik:
@bparees @soltysh thoughts ?

csrwng:
@mfojtik - my initial reaction is that I really like this. We should make it as easy as possible to create STI-enabled images and having a set of boiler-plate scripts that at least give you an idea what each script should do is awesome.

pkg/sti/create/templates/docker.go
```diff
@@ -0,0 +1,42 @@
...
+ENV HOME     /app-root
+ENV PATH     $HOME/bin:$PATH
+
+WORKDIR     /app-root/src
+USER        appuser
+
+# TODO: Specify the default port your application is running on
+EXPOSE 8080
+
+CMD ["/opt/ruby/bin/sti-helper"]
```
csrwng: should this be sti-helper?

pkg/sti/create/templates/scripts.go
```diff
@@ -0,0 +1,82 @@
...
+	echo "Restoring build artifacts"
+	mv /tmp/artifacts/* $HOME/.
+fi
+
+echo "---> Installing application source"
+mkdir -p ${APP_RUNTIME_DIR}
+cp -Rf ${APP_SRC_DIR}/* ${APP_RUNTIME_DIR}/
+
+pushd "$APP_RUNTIME_DIR/${APP_ROOT}" >/dev/null
+echo "---> Building your Ruby application from source"
```
csrwng: Probably should not mention Ruby here

pkg/sti/create/templates/docker.go
mfojtik: @soltysh told me it will go away ;-) so I will remove it once that time comes... (this is copy from the ruby-20-centos image)

pkg/sti/create/templates/docker.go
```diff
@@ -0,0 +1,42 @@
...
+
+# Add STI usage script
+ADD ./.sti/bin/usage /usage
+
+# You can add any other required files/configurations here
+# ADD conf /app-root/conf
+
+# TODO: Define the URL from where the default STI script will be fetched:
+#
+# ENV STI_SCRIPTS_URL https://raw.githubusercontent.com/<repository>/master/.sti/bin
```
soltysh: I'd suggest including here both of the vars we expect from images, meaning `STI_SCRIPTS_URL` and `STI_LOCATION`.

pkg/sti/create/templates/docker.go
```diff
@@ -0,0 +1,42 @@
...
+
+# TODO: Create the application runtime user account and app-root directory:
+RUN mkdir -p /app-root/{src} && \
```
soltysh: You can remove braces from `src`.

pkg/sti/create/templates/docker.go
soltysh: Change that to `CMD ["/usage"]`.

pkg/sti/create/templates/scripts.go
soltysh: Remove ruby :wink:

pkg/sti/create/templates/docker.go
mfojtik: done.

cmd/sti/main.go
```diff
@@ -92,6 +92,24 @@ func newCmdBuild(req *api.Request) *cobra.Command {
 	return buildCmd
 }
 
+func newCmdCreate() *cobra.Command {
+	return &cobra.Command{
+		Use:   "create <image> <source>",
```
soltysh: Change:
* image -> imageName
* source -> [destination] directory
Improve long descriptions with what is what.

soltysh:
I'd also propose putting into all the templates link to https://github.com/openshift/source-to-image/blob/master/docs/builder_image.md

mfojtik:
@soltysh thanks! all comments addressed.

soltysh:
LGTM.

bparees:
lgtm.

mfojtik:
[merge]

PR #107, opened 2015-01-12
fix #106 build with source from local file system

### LINKED ISSUE #106: Building an image without sti-helper from a file system source fails silently
`sti build` using source from the local filesystem and an image without `sti-helper` fails silently when copying the source to `upload/src`, as the upload directory does not exist yet.

The error output from `cp` is suppressed, even with `--loglevel=3`.

accursoft:
and handle errors from shell commands.

Note that this doesn't address the regular output of shell commands, which still need to be integrated into the logging framework.

### DIFF
--- a/pkg/sti/executor.go
+++ b/pkg/sti/executor.go
@@ -274,7 +274,8 @@ func (h *requestHandler) cleanup() {
 }
 
 func (h *requestHandler) fetchSource() error {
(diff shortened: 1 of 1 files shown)


csrwng:
LGTM, thx for the PR!
@bparees can you pls review and merge?

bparees:
lgtm.

PR #109, opened 2015-01-13
Fixed handling STI scripts which are placed just inside the sources and used with arbitrary image

soltysh:
@bparees review?

### DIFF
--- a/cmd/sti/main.go
+++ b/cmd/sti/main.go
@@ -162,8 +162,8 @@ func checkErr(err error) {
 		if e.Details != nil {
 			glog.V(1).Infof("Details: %v", e.Details)
 		}
-		glog.Error("If the problem persists reach us on freenode #openshift " +
-			"or fill an issue at https://github.com/openshift/source-to-image/issues " +
+		glog.Error("If the problem persists consult the docs at https://github.com/openshift/source-to-image/tree/master/docs." +
+			"Eventually reach us on freenode #openshift or file an issue at https://github.com/openshift/source-to-image/issues " +
 			"providing us with a log from your build using --loglevel=3")
 		os.Exit(e.ErrorCode)
 	} else {
--- a/pkg/sti/script/installer.go
+++ b/pkg/sti/script/installer.go
@@ -65,11 +65,12 @@ func (i *installer) DownloadAndInstall(scripts []api.Script, workingDir string, 
 	if !download {
 		return false, nil
 	}
-
 	for _, script := range scripts {
 		scriptPath := i.handler.getPath(script, workingDir)
 		if required && scriptPath == "" {
 			return false, errors.NewScriptDownloadError(script, nil)
+		} else if scriptPath == "" {
+			continue
 		}
 		if err := i.handler.install(scriptPath, workingDir); err != nil {
 			return false, err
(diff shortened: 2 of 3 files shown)


pkg/sti/script/installer.go
```diff
@@ -108,6 +109,15 @@ func (s *handler) download(scripts []api.Script, workingDir string) (bool, error
 		}
 	}
 
+	// check for the scripts inside the sources
+	scriptsDir := filepath.Join(workingDir, "/upload/src/.sti/bin")
+	if s.fs.Exists(scriptsDir) {
+		for _, script := range scripts {
+			downloads[script] <- true
```
bparees: don't you need to actually check that each of these files exists, not just the .sti/bin dir, when setting download to true?

mfojtik: we should have test case for this

soltysh: Yep, your right guys, thx :blush:

soltysh:
@bparees @mfojtik fixed your comments, although no TC yet for that one. I'm thinking about some more generic testing, current solution for integration uses external repo, which I don't have access. I'd rather replace that with test repo like we do with test http server. But that we'll be part of a separate PR.

cmd/sti/main.go
```diff
@@ -162,8 +162,8 @@ func checkErr(err error) {
 		if e.Details != nil {
 			glog.V(1).Infof("Details: %v", e.Details)
 		}
-		glog.Error("If the problem persists reach us on freenode #openshift " +
-			"or fill an issue at https://github.com/openshift/source-to-image/issues " +
+		glog.Error("If the problem persists consult the docs at https://github.com/openshift/source-to-image/tree/master/docs." +
+			"Eventually reach us on freenode #openshift or fill an issue at https://github.com/openshift/source-to-image/issues " +
```
bparees: You can also reach us on freenode....   or *file* an issue at.........

pkg/sti/script/installer.go
```diff
@@ -108,6 +109,20 @@ func (s *handler) download(scripts []api.Script, workingDir string) (bool, error
...
 
+	// check for the scripts inside the sources
+	scriptsDir := filepath.Join(workingDir, "/upload/src/.sti/bin")
+	if s.fs.Exists(scriptsDir) {
+		for _, script := range scripts {
+			file := filepath.Join(scriptsDir, string(script))
+			if s.fs.Exists(file) {
+				downloads[script] <- true
+			} else {
+				errs[script] <- errors.NewScriptDownloadError(script, nil)
```
bparees: it's not necessarily an error if it's not there, right?
1) it could be an optional script
2) as i recall we union all the sources of scripts together, so you might get just the "run" script from the src dir, and "assemble" comes from the default scripts url.

soltysh: This is analogical to what we do in `downloadAsync` (see https://github.com/openshift/source-to-image/blob/master/pkg/sti/script/installer.go#L95-L109), where we'll complain about errors if the file cannot be found. The result is then processed in [executor.go](https://github.com/openshift/source-to-image/blob/master/pkg/sti/executor.go#L104-L111) where depending if the scripts are required or not, fails or warns a user.

cmd/sti/main.go
soltysh: Fixed.

pkg/sti/script/installer.go
mfojtik: @soltysh that seems like a lot of magic to me, but we can improve that later

bparees: it doesn't look like the downloadAsync code reports an error for missing scripts, it just returns if !download

if "download" can be true when the download failed because the file didn't exist, then this code needs commenting.

soltysh: Yeah, you're right, although looking at further logic, it's the fact of being downloaded or not (https://github.com/openshift/source-to-image/blob/master/pkg/sti/script/installer.go#L135) decides about the final error. But I'll remove the `else` part as it's not needed there.

soltysh:
Fixed the merge conflict.

bparees:
lgtm.

mfojtik:
lgtm [merge]

PR #110, opened 2015-01-14
Store the STI build command line to disk

mfojtik:
This saves you from typing all parameters for `sti build` command. It stores them when you run the command for the first time and then you can just run `sti build` (without arguments).

It stores the options into `.stifile` (and attempts to read then from this location).

This also enable us to ship the `.stifile` in application GIT repositories, so if you want to build the application image, you don't need to search for the STI builder image name.

UPDATE:

You have to enable this behavior by using the `--use-config` flag.

### DIFF
--- a/cmd/sti/main.go
+++ b/cmd/sti/main.go
@@ -12,6 +12,7 @@ import (
 
 	"github.com/openshift/source-to-image/pkg/sti"
 	"github.com/openshift/source-to-image/pkg/sti/api"
+	"github.com/openshift/source-to-image/pkg/sti/config"
 	"github.com/openshift/source-to-image/pkg/sti/errors"
 	"github.com/openshift/source-to-image/pkg/sti/version"
 )
@@ -42,6 +43,10 @@ func dockerSocket() string {
 	return "unix:///var/run/docker.sock"
 }
 
+func validRequest(r *api.Request) bool {
+	return !(len(r.Source) == 0 || len(r.BaseImage) == 0 || len(r.Tag) == 0)
+}
+
 func newCmdVersion() *cobra.Command {
 	return &cobra.Command{
 		Use:   "version",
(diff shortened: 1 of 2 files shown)


mfojtik:
@soltysh @bparees thoughts?

cmd/sti/main.go
```diff
@@ -59,14 +67,23 @@ func newCmdBuild(req *api.Request) *cobra.Command {
 		Short: "Build a new image",
 		Long:  "Build a new Docker image named <tag> from a source repository and base image.",
 		Run: func(cmd *cobra.Command, args []string) {
-			if len(args) < 3 {
+			// Attempt to restore the build command from the configuration file
+			config.Restore(req, cmd)
```
bparees: doesn't this make it impossible to override stored args for the various flags like nocache, scriptsurl, etc?

mfojtik: no that flags could be override if you specify them (there is a check if the flag value changed, in that case what user specified on command line takes priority and the JSON file is updated.

bparees:
I like the concept, but i'm worried about the implementation.  At a minimum we need to add some informational messages to the user like "Using config from...." and show the values or something, otherwise i think users are going to be very surprised by this behavior.

also as i mentioned in the comments, it looks like this is going to save all the extraneous args (ie other than the "big 3") the first time, and then you have no way to override them after that because they get restored immediately.

cmd/sti/main.go
```diff
@@ -59,14 +67,24 @@ func newCmdBuild(req *api.Request) *cobra.Command {
...
+
+			// If user specifies the arguments, then we override the stored ones
+			if len(args) >= 3 {
+				req.Source = args[0]
+				req.BaseImage = args[1]
+				req.Tag = args[2]
+			}
+
+			if !validRequest(req) {
```
soltysh: I don't see any value here for having a one-liner if backed as a separate method. I'd suggest copying the if from `validRequest` here.

cmd/sti/main.go
bparees: yup, sorry I didn't parse that restore code carefully enough.

mfojtik:
After IRC discussion with @bparees I reworked this PR so it is not automatically storing/loading the configuration but you have to use `--use-config` option:

```
$ sti build file://$(pwd) openshift/ruby-20-centos myapp --clean --use-config # -> this will create .stifile
$ sti build --use-config # -> this will read from .stifile
```

We might enable this feature by default in future.

cmd/sti/main.go
mfojtik: I still prefer to have method for this as the validation is pretty long line and does not explicitly say what it is doing. I revamped the method to just return the bool instead of having if, hope that is OK ;-)

soltysh:
LGTM

mfojtik:
@bparees ok with merge?

cmd/sti/main.go
```diff
@@ -54,19 +59,35 @@ func newCmdVersion() *cobra.Command {
...
 		Short: "Build a new image",
 		Long:  "Build a new Docker image named <tag> from a source repository and base image.",
 		Run: func(cmd *cobra.Command, args []string) {
-			if len(args) < 3 {
+			// Attempt to restore the build command from the configuration file
+			if useConfig {
+				config.Restore(req, cmd)
+			}
+
+			// If user specifies the arguments, then we override the stored ones
```
bparees: would it make more sense to do this in Restore?  You've already got logic in restore that handles overriding stored parameters with the values passed on the CLI

mfojtik: yeah, I was trying to make Restore/Save same in terms of parameters, this will require I pass the args to the Restore :-) But it makes sense

mfojtik: Actually I don't think it should be part of Restore() as the `use-config` will have to be parsed in Restore as well and by default the user-config is disabled. In that case setting the args will be hidden in Restore()...

mfojtik:
[merge]

PR #112, opened 2015-01-20
Bug 1182464 - added info from failing container

soltysh:
@csrwng could you please have a look, since I'm modifying the expected error behavior slightly, I wanted to get rid of the hack that @smarterclayton didn't like, in favor of passing the error output (which is limited to 1024) up the stack.

### DIFF
--- a/pkg/sti/build.go
+++ b/pkg/sti/build.go
@@ -57,7 +57,6 @@ func newBuildHandler(req *api.Request) (*buildHandler, error) {
 		callbackInvoker: util.NewCallbackInvoker(),
 	}
 	rh.postExecutor = bh
-	rh.errorChecker = bh
 	return bh, nil
 }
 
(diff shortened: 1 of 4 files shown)


csrwng:
LGTM

pkg/sti/build.go
```diff
@@ -105,6 +104,8 @@ func (b *Builder) Build() (*api.Result, error) {
 		if err := bh.execute(api.Assemble); err != nil {
 			return nil, err
 		}
+	} else if ok && err != nil {
+		return nil, errors.NewAssembleError(bh.Request().Tag, e.Output, err)
```
mfojtik: this is getting a bit hairy, I would refactor this to use switch or re-organize the if's

pkg/sti/errors/errors.go
```diff
@@ -84,15 +84,26 @@ func NewScriptDownloadError(name api.Script, err error) error {
 
 // NewSaveArtifactsError returns a new error which indicates there was a problem
 // calling save-artifacts script
-func NewSaveArtifactsError(name string, err error) error {
+func NewSaveArtifactsError(name, output string, err error) error {
 	return Error{
-		Message:    fmt.Sprintf("saving artifacts for %s failed", name),
+		Message:    fmt.Sprintf("saving artifacts for %s failed:\n%s", name, output),
```
mfojtik: the new line will make grepping this super-difficult ;-)

pkg/sti/errors/errors.go
```diff
@@ -100,7 +111,7 @@ func NewBuildError(name string, err error) error {
 		Message:    fmt.Sprintf("building %s failed", name),
 		Details:    err,
 		ErrorCode:  ErrBuild,
-		Suggestion: "check the assemble script for errors",
+		Suggestion: "check the build output for errors",
```
mfojtik: maybe suggest how to do that? (increase log-level?)

pkg/sti/errors/errors.go
soltysh: Not necessarily and if so we'll change it later :bowtie:

pkg/sti/errors/errors.go
soltysh: That's already happening, see https://github.com/openshift/source-to-image/blob/master/cmd/sti/main.go#L189-L191

pkg/sti/build.go
soltysh: Fixed with switch type.

soltysh:
@mfojtik addressed your comments, how's that?

mfojtik:
@soltysh LGTM

soltysh:
[merge]

PR #115, opened 2015-01-27
[DO NOT MERGE] Make run script optional

### LINKED ISSUE #114: Make run script optional
Currently the `run` script is required. I would like it to have optional. This would let it reuse the `CMD` command which was specified in the image we could possibly extend by creating the builder image.

Imagine the `jboss/wildfly` image. This is upstream image has already defined a `CMD` command. I don't see a reason to duplicate it just to satisfy the requirement of having the `run` command. In my case I would just repeat what I have already defined, but this is unnecessary + adds another file and possibly could confuse people.

Without a `run` script, the `CMD` or `ENTRYPOINT` from the extended image should be used as-is. In case the run script exist - it should override those.

goldmann:
This makes the `run` script optional. In case the run script is not provided, the `CMD` used in the builder image will be reused.

Fixes #114.

**Warning**
This is work in progress. Currently it always uses the `CMD` set in the builder image and ignores the provided `run` script.

### DIFF
--- a/pkg/sti/build.go
+++ b/pkg/sti/build.go
@@ -69,7 +69,7 @@ func (b *Builder) Build() (*api.Result, error) {
 	defer bh.cleanup()
 
 	glog.Infof("Building %s", bh.Request().Tag)
-	err := bh.setup([]api.Script{api.Assemble, api.Run}, []api.Script{api.SaveArtifacts})
+	err := bh.setup([]api.Script{api.Assemble}, []api.Script{api.Run, api.SaveArtifacts})
 	if err != nil {
 		return nil, err
 	}
@@ -150,9 +150,16 @@ func (h *buildHandler) PostExecute(containerID string, location string) error {
 		}
 	}
 
-	cmd := []string{}
+	image, err := h.docker.GetBuilderImage(containerID)
+
+	if err != nil {
+		return errors.NewBuildError(h.request.Tag, err)
+	}
+
+	// cmd := []string{}
 	opts := docker.CommitContainerOptions{
-		Command:     append(cmd, filepath.Join(location, string(api.Run))),
+		Command:     image.Config.Cmd,
+		//append(cmd, filepath.Join(location, string(api.Run))),
 		Env:         h.generateConfigEnv(),
 		ContainerID: containerID,
 		Repository:  h.request.Tag,
--- a/pkg/sti/docker/docker.go
+++ b/pkg/sti/docker/docker.go
@@ -28,6 +28,7 @@ type Docker interface {
 	GetScriptsURL(name string) (string, error)
 	RunContainer(opts RunContainerOptions) error
 	GetImageID(name string) (string, error)
+	GetBuilderImage(name string) (*docker.Image, error)
 	CommitContainer(opts CommitContainerOptions) (string, error)
 	RemoveImage(name string) error
 	PullImage(name string) error
@@ -40,6 +41,7 @@ type Docker interface {
 type Client interface {
 	RemoveImage(name string) error
 	InspectImage(name string) (*docker.Image, error)
+	InspectContainer(name string) (*docker.Container, error)
 	PullImage(opts docker.PullImageOptions, auth docker.AuthConfiguration) error
 	CreateContainer(opts docker.CreateContainerOptions) (*docker.Container, error)
 	AttachToContainer(opts docker.AttachToContainerOptions) error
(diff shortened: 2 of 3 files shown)


goldmann:
Hey @soltysh - since I do not know the STI codebase - any idea how to make STI check if there is the run script provided or not? I was looking for some simple mechanism that is already built into STI, but it seems there is nothing like this. Can you confirm this? Feel free to add notes to the code and/or extend it to make it work like it should :)

pkg/sti/build.go
```diff
@@ -150,9 +150,16 @@ func (h *buildHandler) PostExecute(containerID string, location string) error {
...
+
+	if err != nil {
+		return errors.NewBuildError(h.request.Tag, err)
+	}
+
+	// cmd := []string{}
 	opts := docker.CommitContainerOptions{
-		Command:     append(cmd, filepath.Join(location, string(api.Run))),
+		Command:     image.Config.Cmd,
+		//append(cmd, filepath.Join(location, string(api.Run))),
```
goldmann: This part needs to be changed and use `run` script if provided.

soltysh:
@goldmann have a look at [build.go#determineIncremental](https://github.com/openshift/source-to-image/blob/master/pkg/sti/build.go#L196-L201) where we check for `save-artifacts` existence. The tricky part here is that when scripts are embedded into the image (eg. `STI_SCRIPTS_LOCATION=image:///opt/sti/bin`) then there's no way of checking its existence other than actually running the image. In our case (`save-artifacts`) we just assume its existence, and we'll warn the user if it's not there. In all other cases (meaning when you downloaded scripts) the situation is easier, you just do `h.fs.Exists`. That means there's no such method for checking that, since we've done it only once so far (for `save-artifacts`), so you're welcome to create one :smile:

pkg/sti/build.go
```diff
@@ -150,9 +150,16 @@ func (h *buildHandler) PostExecute(containerID string, location string) error {
 		}
 	}
 
-	cmd := []string{}
+	image, err := h.docker.GetBuilderImage(containerID)
```
soltysh: There's no need to reach container, just invoke it with the name of the builder image:
```
image, err := h.docker.GetBuilderImage(h.request.BaseImage)
```
and the `GetBuilderImage` method should just access the `InspectImage` method. There's no need to add new ones.

pkg/sti/docker/docker.go
```diff
@@ -40,6 +41,7 @@ type Docker interface {
 type Client interface {
 	RemoveImage(name string) error
 	InspectImage(name string) (*docker.Image, error)
+	InspectContainer(name string) (*docker.Container, error)
```
soltysh: No need for that method, see my previous comment.

pkg/sti/docker/docker.go
```diff
@@ -338,6 +340,23 @@ func (d *stiDocker) GetImageID(name string) (string, error) {
 	return image.ID, nil
 }
 
+// This function resolves the original (ie. not created by STI) parent image
+// and returns it for the container specified as the input.
+func (d *stiDocker) GetBuilderImage(name string) (*docker.Image, error) {
+	container, err := d.client.InspectContainer(name)
```
soltysh: Just return `Inspectimage`, see my previous comment on it.

goldmann:
Thanks a lot for the comments, I'll look at it in a bit!

pkg/sti/build.go
goldmann: I wasn't aware of `h.request.BaseImage` :)

soltysh:
@goldmann talking to @bparees we've come up with solution for checking for the `run` script while using the embedded scripts. You need to extend the build process in such a way that it should check for the `run` script existence. This should be done while having the container running, because doing it again would be a waste of time generally and we'd like to avoid that.

soltysh:
Have a look [here](https://github.com/openshift/source-to-image/blob/master/pkg/sti/docker/docker.go#L227-L229) where we invoke untarring and assemble in one step, I'm guessing you'd have to come up with some smart idea to check for run existence in this step :wink:

mfojtik:
@soltysh @goldmann what is the status of this? is this resolved already?

soltysh:
@mfojtik it's not, I've trelloed that to cover that https://trello.com/c/oDj441HE/571-make-run-script-for-sti-optional

goldmann:
I had no time to work on this PR more, I'm pretty sure it's badly outdated right now.

soltysh:
I'm hoping to work on that after GA.

mfojtik:
@soltysh I wish we can have labels on issues ;-) @danmcp is this something you can enable for us?

soltysh:
:+1: on that idea

danmcp:
@soltysh @mfojtik There isn't a way to separate push access and the label capability.  So we only have a couple of options.  Either:

1) Give everyone push access.  Or more people at least.
or
2) Write a bot that can label and unlabel on request.

I am fine with doing 2).  I am not sure 1) is feasible.

soltysh:
I'm for 2 if it's the simplest solution.

mfojtik:
@goldmann @soltysh I think we can close this now?

soltysh:
:+1: pretty outdated as of now.

goldmann:
Although the PR is outdated, the request is still valid. I guess I'll not be able to finish this PR myself :(

soltysh:
There's an issue (#114) that tracks that request.

goldmann:
Right, I forgot about it, closing this PR in that case.

PR #118, opened 2015-02-02
Initialize Flags on the spot

kargakis:
Initialize Flags alongside the rest initialization of Config so that the
underlying map reference will point to an initialized map and not nil
which can cause runtime panics if someone tries to write to it in between
the Config initialization and the [current](https://github.com/openshift/source-to-image/blob/9fad846d68808d43a815072db4d075cbe5012c05/pkg/sti/config/config.go#L32) map initialization
(blog.golang.org/go-maps-in-action)


mfojtik:
lgtm

[merge]

PR #119, opened 2015-02-04
bug_1185730

csrwng:
https://bugzilla.redhat.com/show_bug.cgi?id=1185730

### DIFF
--- a/pkg/sti/docker/docker.go
+++ b/pkg/sti/docker/docker.go
@@ -33,6 +33,7 @@ type Docker interface {
 	PullImage(name string) error
 	CheckAndPull(name string) (*docker.Image, error)
 	BuildImage(opts BuildImageOptions) error
+	GetImageUser(name string) (string, error)
 }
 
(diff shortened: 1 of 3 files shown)


csrwng:
@bparees @soltysh - this is a way to fix this issue. It's definitely not the only way... I'd like to hear what you think.

bparees:
looks like a nice clean solution to me.

csrwng:
Changed COPY to ADD, and added a TODO to take into consideration images that may not have chown in them.

csrwng:
@smarterclayton @bparees pls review

bparees:
so we're broken for images that do not contain tar and do use a non-root user, for now?

csrwng:
correct. With this fix we will not be broken as long as the image contains 'chown'

bparees:
lgtm.

csrwng:
merge?

smarterclayton:
[merge]

PR #120, opened 2015-02-05
parseEnvs: Return error when no env was found...

kargakis:
...and uppercase environment variables.


kargakis:
On a second thought on this, should it be considered an error if no environment variables were parsed by `sti`? I have followed all possible paths and found no problem since the underlying map has to be [checked](https://github.com/openshift/source-to-image/blob/1a6d78779bb8d45d40d75cc27d8551089afc5f02/pkg/sti/executor.go#L125) before used. I am closing this and if anyone thinks otherwise, feel free to re-open it.

PR #124, opened 2015-02-10
Refactor to support more build strategies

mfojtik:
This is an initial attempt to make the STI codebase more extensible and include support for different build strategies. This PR does:

* Remove '/sti/' from the package name. Everything now lives inside the 'pkg/' folder just like Origin
* Add 'pkg/build/strategies' with STI and OnBuild strategies
* This code will decide what strategy to use based on the baseImage name, this will be changed in future (as long I we get OnBuild back from InspectImage()
* The ONBUILD strategy now fully support STI `assemble` and `run` script. The assemble is ran as the last instruction. Both scripts are optional, if you don't have them (either in application repo || as CLI parameter) the build will continue and try to guess the entrypoint.
* A lot of changes to interfaces to make the code more extensible and re-usable. For example the ONBUILD is using 'STI' to call Prepare() (so it downloads the code same way as STI). Avoids unnecessary code duplication)
* A lot of small refactoring to make all this works

Basically this is a base for further hacking on different build methods. We should refactor the STI build strategy to be more 'flexible' and get rid of the hairy interface we have there.

**Testing**:

To try the ONBUILD sti builder you can run this command:

```
$ sti build --loglevel=5 git://github.com/mfojtik/sinatra-app-example.git ruby:2.1-onbuild testapp
$ docker run testapp
```

This is the sample output from the command above: https://gist.github.com/mfojtik/3fb2a79f44d27b5f0c76#file-gistfile1-txt-L60


mfojtik:
@soltysh @bparees PTAL, this is work in progress (especially the ONBUILD stuff)

mfojtik:
UPDATE: The last commit added 'entrypoint' guessing for ONBUILD images. Basically just look up if there are "launchers" (./exec, ./run, ./start, ./execute) and set them as the CMD.

The names for launchers are trying to match: https://github.com/gliderlabs/herokuish#entrypoints

pkg/build/strategies/strategies.go
```diff
@@ -0,0 +1,22 @@
...
+)
+
+func GetBuilder(request *api.Request) (build.Builder, error) {
+	// TODO: The ONBUILD instructions should be available in ImageInspect call,
+	// however they are not in the current go-dockerclient version.
+	// Right now we assume that if the base image has '-onbuild' in the tag, then
+	// we do OnBuild build.
+	if strings.Contains(request.BaseImage, "-onbuild") {
+		return onbuild.NewOnBuild(request)
+	} else {
```
kargakis: You don't need the else clause here

mfojtik: @kargakis this is just WIP, still refactoring ;-)

mfojtik:
UPDATE: I'm refactoring the interfaces to make the code more pluggable and also make the tests working again.

pkg/build/strategies/strategies.go
mfojtik: @kargakis problems solved! thanks!

mfojtik:
FYI, I need to fixup the unit tests to use new interfaces. This should be done tomorrow. Stay tuned :)

pkg/build/strategies/strategies.go
kargakis: :+1:

mfojtik:
@bparees @soltysh review pls

cmd/sti/main.go
```diff
@@ -10,11 +10,13 @@ import (
...
 	"github.com/spf13/pflag"
 
-	"github.com/openshift/source-to-image/pkg/sti"
-	"github.com/openshift/source-to-image/pkg/sti/api"
-	"github.com/openshift/source-to-image/pkg/sti/config"
-	"github.com/openshift/source-to-image/pkg/sti/errors"
-	"github.com/openshift/source-to-image/pkg/sti/version"
+	"github.com/openshift/source-to-image/pkg/api"
+	"github.com/openshift/source-to-image/pkg/build/strategies"
+	"github.com/openshift/source-to-image/pkg/build/strategies/sti"
```
soltysh: Does it make sense to have `build/strategies`? I'd propose having just `build` or `strategies` package, where the first makes more sense IMHO than the latter.

cmd/sti/main.go
```diff
@@ -126,7 +130,7 @@ func newCmdCreate() *cobra.Command {
 				cmd.Help()
 				os.Exit(1)
 			}
-			b := sti.NewCreate(args[0], args[1])
+			b := create.NewCreate(args[0], args[1])
```
soltysh: We should come up with some unified way of getting objects, here we're using `create.NewCreate`, whereas a couple of lines before you do `strategies.GetBuilder`. I'd propose to use the [origin convention](https://github.com/openshift/origin/blob/master/pkg/cmd/server/origin/master.go#L242-L249), so in above cases you'd have:
`strategies.New()`, `create.New()`, etc.

cmd/sti/main.go
```diff
@@ -150,7 +154,7 @@ func newCmdUsage(req *api.Request) *cobra.Command {
 			checkErr(err)
 			req.Environment = envs
 
-			uh, err := sti.NewUsage(req)
+			uh, err := sti.NewSTIUsage(req)
```
soltysh: If it's `sti` package, why naming it `NewSTIUsage`?

pkg/build/strategies/onbuild/onbuild.go
```diff
@@ -0,0 +1,133 @@
...
+	"github.com/openshift/source-to-image/pkg/git"
+	"github.com/openshift/source-to-image/pkg/tar"
+	"github.com/openshift/source-to-image/pkg/util"
+)
+
+// OnBuild strategy executes the simple Docker build in case the image does not
+// support STI scripts but has ONBUILD instructions recorded.
+type OnBuild struct {
+	docker  docker.Docker
+	git     git.Git
```
soltysh: I'm still thinking about interfacing VCS access as well, but that can be left until we'll actually support something more than just git, I guess.

pkg/build/strategies/onbuild/onbuild.go
```diff
@@ -0,0 +1,133 @@
...
+	fs      util.FileSystem
+	tar     tar.Tar
```
soltysh: What do you think about squashing the two into one, since tar operation are FileSystem related?

pkg/build/strategies/onbuild/onbuild.go
```diff
@@ -0,0 +1,133 @@
...
+	source  build.Downloader
+	garbage build.Cleaner
+}
+
+// NewOnBuild returns a new instance of OnBuild builder
+func NewOnBuild(request *api.Request) (*OnBuild, error) {
+	dockerHandler, err := docker.NewDocker(request.DockerSocket)
```
soltysh: Suggestion: `docker.NewDocker` -> `docker.New`

pkg/build/strategies/onbuild/onbuild.go
```diff
@@ -0,0 +1,133 @@
...
+	if err != nil {
+		return nil, err
+	}
+	return b.fs.Open(tarFileName)
+}
+
+// Build executes the ONBUILD kind of build
+func (b *OnBuild) Build(request *api.Request) (*api.Result, error) {
+	glog.V(2).Info("Preparing the source code for build")
+	if err := b.Prepare(request); err != nil {
```
soltysh: If `Prepare` is called only inside `build` and `usage` why not making this a private method? What's the reason in exposing this?

pkg/build/strategies/sti/sti.go
```diff
@@ -0,0 +1,479 @@
...
+	// we're assuming save-artifacts to exists for embedded scripts (if not we'll
+	// warn a user upon container failure and proceed with clean build)
+	// for external save-artifacts - check its existence
+	saveArtifactsExists := !request.ExternalOptionalScripts ||
+		b.fs.Exists(filepath.Join(request.WorkingDir, "upload", "scripts", string(api.SaveArtifacts)))
+	request.Incremental = previousImageExists && saveArtifactsExists
+	return nil
+}
+
+func (b *STI) Save(request *api.Request) (err error) {
```
soltysh: I'd stay with `SaveArtifacts` name, as plain `Save` might be misleading.

pkg/build/strategies/strategies.go
```diff
@@ -0,0 +1,23 @@
...
+	"github.com/openshift/source-to-image/pkg/build/strategies/onbuild"
+	"github.com/openshift/source-to-image/pkg/build/strategies/sti"
+)
+
+func GetBuilder(request *api.Request) (build.Builder, error) {
+	// TODO: The ONBUILD instructions should be available in ImageInspect call,
+	// however they are not in the current go-dockerclient version.
+	// Right now we assume that if the base image has '-onbuild' in the tag, then
+	// we do OnBuild build.
+	if strings.Contains(request.BaseImage, "-onbuild") {
```
soltysh: That's very hacky :stuck_out_tongue_winking_eye:

mfojtik: once we can consume onbuild instruction from docker inspect in go-dockerclient, this will go away

cmd/sti/main.go
mfojtik: I agree with create.New() but I don't like strategies.New() as we are not just returning a strategy but we are making decisions about what strategy is going to be returned. The object returned is a Builder, so GetBuilder kinda makes sense to me.

pkg/build/strategies/onbuild/onbuild.go
mfojtik: there is already an interface for that called Downloader. Right now git.Clone implements Download(), in future we might have more SCM that implements this method. There is already a TODO in the code for that. Check the pkg/git/clone.go

pkg/build/strategies/onbuild/onbuild.go
mfojtik: this need to be mocked for unit tests, I don't want to make this PR even more complex, so if you feel it should be squashed fill an issue.

pkg/build/strategies/onbuild/onbuild.go
mfojtik: What is the reason for not doing that? I don't see any reason why we need to hide this method. It can be useful for other builders as well (I want a builder that use STI as Preparer)... Make methods private cause nobody can consume them and it also makes the code less extensible.

cmd/sti/main.go
mfojtik: I like to be explicit, so I would argue for "build/strategies/sti".

mfojtik:
@smarterclayton PTAL

cmd/sti/main.go
bparees: +1.  i'm always a fan of verbosity.

pkg/api/types.go
```diff
@@ -57,6 +57,8 @@ type Request struct {
 
 	// LayeredBuild describes if this is build which layered scripts and sources on top of BaseImage.
 	LayeredBuild bool
+
+	InstallDestination string
```
bparees: comment please...installdestination for what?

mfojtik: fixed.

pkg/build/interfaces.go
```diff
@@ -0,0 +1,50 @@
...
+
+// Builder is the interface that provides basic methods all implementation
+// should have.
+// Build method executes the build based on Request and returns the Result.
+type Builder interface {
+	Build(*api.Request) (*api.Result, error)
+}
+
+// Preparer provides the Prepare method for builders that need to prepare source
+// code before it gets passed to the build. All builders must comfort this
```
bparees: comfort?  implement?

bparees: also if builders must implement it, why not make it part of the Builder interface?

pkg/build/interfaces.go
```diff
@@ -0,0 +1,50 @@
...
+// the 'incremental' build workflow.
+// The Determine method checks if the artifacts from the previous build exists
+// and if they can be used in the current build.
+// The Save method stores the artifacts for the next build.
+type IncrementalBuilder interface {
+	Determine(*api.Request) error
+	Save(*api.Request) error
+}
+
+// ScriptsHandler provides interface for downloading the source code and scripts
```
bparees: it downloads the source too?  or it pulls the scripts from the source if present?

pkg/build/interfaces.go
```diff
@@ -0,0 +1,50 @@
...
+// and also for executing them.
+type ScriptsHandler interface {
+	Execute(api.Script, *api.Request) error
+}
+
+// Downloader provides methods for downloading the source code
```
bparees: if this is really specific to downloading source (vs scripts) i'd make that clearer in the name.

pkg/build/interfaces.go
```diff
@@ -0,0 +1,50 @@
...
+type Downloader interface {
+	Download(*api.Request) error
+}
+
+// DockerBuilder interface represents the generic Docker Builder and the Build
+// method is used to execute the Docker build.
```
bparees: so STI is going to do standard "docker build" builds now?  i don't think it should.
if this is in support of the cases where we have to do a docker build to layer the source in, or to handle ONBUILD commands, then that makes sense but the comment should be clearer about the intent if so.

pkg/build/interfaces.go
mfojtik: @bparees the OnBuild uses this to call the Prepare method from STI. Making it as part of Builder interface will cause that I will have to implement Build() methods as well.

pkg/build/interfaces.go
mfojtik: no, that is just a leftover from refactoring. it only execute the scripts

pkg/build/interfaces.go
mfojtik: scripts are downloaded elsewhere, just a bad comment. fixing it

pkg/build/interfaces.go
mfojtik: The DockerBuild is the minimal builder that just builds the Docker image (see sti.go:441)

pkg/build/strategies/sti/sti.go
```diff
@@ -0,0 +1,486 @@
...
+	}
+
+	// The sources are downloaded using the GIT downloader.
+	// TODO: Add more SCM in future.
+	b.source = &git.Clone{b.git, b.fs}
+
+	b.garbage = &build.DefaultCleaner{b.fs, b.docker}
+
+	// Set interfaces
+	b.preparer = &b
```
smarterclayton: Why expose these interfaces like this?  So you can test it?  The fact that you are keeping references to yourself to do this indicates there's a missing refactor here (which you don't have to solve now, but I want to point it out so at least it's in your mind).  

Also, you can nuke about ten of the & by using using `&STI` up above.

mfojtik: @smarterclayton yeah testing and re-using some of them in the onbuild strategy (preparer), I don't want to do full refactor atm, just have something useful to base the onbuild strategy on.

smarterclayton: That's fine as long as you have a direction.

----- Original Message -----

pkg/build/strategies/onbuild/onbuild.go
```diff
@@ -0,0 +1,135 @@
...
+	}
+	// FIXME: This assumes that the WORKDIR is set to the application source root
+	//        directory.
+	buffer.WriteString(fmt.Sprintf(`CMD ["./%s"]`+"\n", entrypoint))
+	return b.fs.WriteFile(filepath.Join(uploadDir, "Dockerfile"), buffer.Bytes())
+}
+
+// hasAssembleScript checks if the the assemble script is available
+func (b *OnBuild) hasAssembleScript(request *api.Request) bool {
+	assemblePath := filepath.Join(request.WorkingDir, "upload", "src", "assemble")
```
bparees: shouldn't src be scripts here?

mfojtik: No, the assemble is copied into upload/src (see the InstallDestination) to be available in the root of the application source. This is copying the pattern herokiush is using. Since we don't know into which directory the application is installed in the container, we need to assume the WORKDIR is the application root directory. So the 'run' and 'assemble' needs to be there. Also the assemble must use $CWD as the application dir.

pkg/build/strategies/strategies.go
```diff
@@ -0,0 +1,25 @@
...
+		return onbuild.New(request)
+	}
+
+	// The default builder is STI builder
+	return sti.New(request)
```
bparees: these are both STI builds, i don't like the implication in this naming convention that things doing ONBUILD are somehow not STI builds.

pkg/create/create.go
```diff
@@ -16,7 +16,7 @@ type Bootstrap struct {
 
 // NewCreate returns a new bootstrap for giben image name and destination
```
bparees: s/giben/given/  (since you're in here...)

pkg/build/strategies/strategies.go
mfojtik: @bparees yeah, they are both STI builds, but different strategies. I can rename this methods to `GetStrategy`. Now the OnBuild can consume STI scripts it is even more clear.

mfojtik: already renamed and better godoc provided to express this.

pkg/build/strategies/sti/sti.go
```diff
@@ -0,0 +1,486 @@
...
+func (b *STI) handleContainerError(cerr errors.ContainerError) (*api.Result, error) {
+	if wasExpectedError(cerr.Output) {
+		glog.Warningf("Image %s does not have tar! Performing additional build to add the scripts and sources.",
+			b.request.BaseImage)
+		if _, err := b.builder.Build(b.request); err != nil {
+			return nil, err
+		}
+		glog.V(2).Infof("Building %s using sti-enabled image", b.request.Tag)
+		if err := b.scripts.Execute(api.Assemble, b.request); err != nil {
+			glog.V(2).Infof("FOOOO\n")
```
bparees: remove this :)

bparees:
one final nit and then i think it lgtm.

it does still bother me that we're using CMD instead of ENTRYPOINT which means images that set an ENTRYPOINT other than "sh" are not going to work correctly, but i think that's a limitation of the dockclient package?

mfojtik:
@bparees I switched to ENTRYPOINT https://github.com/mfojtik/source-to-image/commit/4aa474443d9bcab1c544d8732c29e1a063a82347

bparees:
@mfojtik I mean in the "normal" STI flow cases.

mfojtik:
yeah, that seems like a limitation of dockerclient. I can have look on that tomorrow.

mfojtik:
@soltysh can you take a second review for this? i want 2 lgtm for merging ;-)

mfojtik:
@soltysh @bparees last addon... I think the 'layered' build inside STI should be treated as a strategy. It actually nicely plays with the interfaces I put it (you can use different build strategies to construct a new one...) PTAL. That is just a refactoring nit, it does not affect the functionality. The code in sti.go is now much cleaner and the tests are splitted (one for STI and one for Layered)

bparees:
@mfojtik I like the latest layered build refactor.  still lgtm.

soltysh:
LGTM :+1:

mfojtik:
:rage1: [merge] !

mfojtik:
[merge]

PR #127, opened 2015-02-13
Return imageID instead of image name in onbuild

mfojtik:


mfojtik:
just a cosmetic fix... [merge]

PR #128, opened 2015-02-16
Fix typos

rhcarvalho:
@mfojtik asked, so here we go!

[typokiller](https://github.com/rhcarvalho/typokiller) found very few typos here, awesome!

### DIFF
--- a/pkg/build/strategies/sti/sti_test.go
+++ b/pkg/build/strategies/sti/sti_test.go
@@ -149,17 +149,17 @@ func TestBuild(t *testing.T) {
 			t.Errorf("Unexpected optional scripts requested: %#v", fh.SetupOptional)
 		}
 
-		// Verify that determineIncremental was called
+		// Verify that Determine was called
 		if !fh.DetermineIncrementalCalled {
 			t.Errorf("Determine incremental was not called.")
 		}
 
-		// Verify that saveartifacts was called for an incremental build
+		// Verify that Save was called for an incremental build
 		if incremental && !fh.SaveArtifactsCalled {
-			t.Errorf("SaveArtifacts was not called for an incremental build")
+			t.Errorf("Save artifacts was not called for an incremental build")
 		}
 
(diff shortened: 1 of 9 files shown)


mfojtik:
awesome! thanks!

[merge]

pkg/build/strategies/sti/sti_test.go
```diff
@@ -149,17 +149,17 @@ func TestBuild(t *testing.T) {
 			t.Errorf("Unexpected optional scripts requested: %#v", fh.SetupOptional)
 		}
 
-		// Verify that determineIncremental was called
+		// Verify that Determine was called
```
soltysh: This is bad change, I prefer the stay with the old one.

pkg/build/strategies/sti/sti_test.go
```diff
@@ -149,17 +149,17 @@ func TestBuild(t *testing.T) {
...
 		if !fh.DetermineIncrementalCalled {
 			t.Errorf("Determine incremental was not called.")
 		}
 
-		// Verify that saveartifacts was called for an incremental build
+		// Verify that Save was called for an incremental build
```
soltysh: Same here.

rhcarvalho: I based the changes in the fact that the interface for [`IncrementalBuilder`](https://github.com/openshift/source-to-image/blob/d2c8661cf024ad774210a218b03717b2673d53e1/pkg/build/interfaces.go#L29-L32) is:
```go
type IncrementalBuilder interface {
	Determine(*api.Request) error
	Save(*api.Request) error
}
```

pkg/build/strategies/sti/sti_test.go
mfojtik: determineIncremental is no more

soltysh: Yeah, both of you're right, I can't get used to your refactor :wink:

PR #129, opened 2015-02-17
Refactoring installer to return more information about the download proess

soltysh:
@mfojtik this is first run (works but still needs work esp. in places marked `TODO`), let me know what you think about it.


pkg/script/installer.go
```diff
@@ -9,170 +9,144 @@ import (
...
-
-type handler struct {
-	docker     docker.Docker
 	image      string
 	scriptsURL string
-	destDir    string
+	docker     docker.Docker
 	downloader util.Downloader
 	fs         util.FileSystem
+	result     map[api.Script]*InstallResult
```
csrwng: Just my opinion... I would think it'd be better to keep the result as a return value and not make it part of the handler. It would make it clearer inside the Install function where the result comes from.

pkg/util/download.go
```diff
@@ -23,22 +23,11 @@ type schemeReader interface {
...
-
-// ImageReader just returns information the URL is from inside the image
-type ImageReader struct{}
-
 type downloader struct {
 	schemeReaders map[string]schemeReader
 }
 
-// NewDownloader creates an instance of the default Downloader implementation
+// New creates an instance of the default Downloader implementation
```
csrwng: I guess the intent is to rename the function below to 'New' ?

pkg/util/download.go
```diff
@@ -51,6 +40,50 @@ func NewDownloader() Downloader {
 	}
 }
 
+// DownloadFile downloads the file pointed to by URL into local targetFile
```
csrwng: s/DownloadFile/Download/

pkg/build/strategies/sti/sti.go
```diff
@@ -318,11 +314,12 @@ func (b *STI) Execute(command api.Script, request *api.Request) error {
...
-		Stdout:          outWriter,
-		Stderr:          errWriter,
-		PullImage:       request.ForcePull,
-		ExternalScripts: request.ExternalRequiredScripts,
+		Image:     request.BaseImage,
+		Stdout:    outWriter,
+		Stderr:    errWriter,
+		PullImage: request.ForcePull,
+		// TODO fix this value
+		ExternalScripts: true,
```
bparees: what's going on here?

pkg/errors/errors.go
```diff
@@ -71,16 +71,16 @@ func NewPullImageError(name string, err error) error {
...
-// downloading a script
-func NewScriptDownloadError(name api.Script, err error) error {
-	return Error{
-		Message:    fmt.Sprintf("%s script download failed", name),
-		Details:    err,
-		ErrorCode:  ErrScriptDownload,
-		Suggestion: "provide URL with STI scripts with -s flag or check the image if it contains STI_SCRIPTS_URL variable set",
-	}
-}
+// // NewScriptDownloadError returns a new error which indicates there was a problem
```
mfojtik: remove this instead of commenting, we have git history

pkg/scripts/download.go
```diff
@@ -14,7 +14,7 @@ import (
 
 // Downloader downloads the specified URL to the target file location
 type Downloader interface {
-	DownloadFile(url *url.URL, targetFile string) (bool, error)
+	Download(url *url.URL, target string) (bool, error)
```
mfojtik: I still don't like returing an bool from Download. What is the point of doing that? If the download was successful then the err is nil. If not, then err should contain details about what failed. We should get rid of using bools as a return value checks  and use it only in methods where it make sense.

pkg/scripts/download.go
```diff
@@ -51,6 +40,50 @@ func NewDownloader() Downloader {
 	}
 }
 
+// DownloadFile downloads the file pointed to by URL into local targetFile
+// Returns information a boolean flag informing whether any download/copy operation
+// happened and an error if there was a problem during that operation
+func (d *downloader) Download(url *url.URL, targetFile string) (bool, error) {
+	sr := d.schemeReaders[url.Scheme]
```
mfojtik: use full names for variables. I hate seeing vars like 's', 'sr', 'gh' and so on. Having a proper (even long) variable name is much better than this.

pkg/scripts/download.go
```diff
@@ -51,6 +40,50 @@ func NewDownloader() Downloader {
...
+		glog.Errorf("No URL handler found for %s", url.String())
+		return false, stierr.NewURLHandlerError(url.String())
+	}
+	if sr.IsFromImage() {
+		glog.V(2).Infof("Using image internal scripts from: %s", url.String())
+		return false, nil
+	}
+	reader, err := sr.Read(url)
+	if err != nil {
+		glog.Warningf("Unable to download %s (%s)", url.String(), err)
```
mfojtik: Why is the warning here? If I have optional script and it fails to download, I don't want to see any warnings. Move the warning to place where it does make sense.

pkg/scripts/download.go
```diff
@@ -51,6 +40,50 @@ func NewDownloader() Downloader {
...
+		return false, err
+	}
+	defer reader.Close()
+
+	out, err := os.Create(targetFile)
+	defer out.Close()
+
+	if err != nil {
+		defer os.Remove(targetFile)
```
mfojtik: I don't see the point of doing this in defer if you are going to return in next line.

pkg/scripts/download.go
```diff
@@ -51,6 +40,50 @@ func NewDownloader() Downloader {
...
+		glog.Errorf("Unable to create target file %s (%s)", targetFile, err)
+		return false, err
+	}
+
+	if _, err = io.Copy(out, reader); err != nil {
+		defer os.Remove(targetFile)
```
mfojtik: same here

pkg/scripts/download.go
```diff
@@ -51,6 +40,50 @@ func NewDownloader() Downloader {
...
+		glog.Warningf("Skipping file %s due to error copying from source: %s", targetFile, err)
+	}
+
+	glog.V(2).Infof("Downloaded '%s'", url.String())
+	return true, nil
+}
+
+// HttpURLReader retrieves a response from a given http(s) URL
+type HttpURLReader struct {
+	httpGet func(url string) (*http.Response, error)
```
mfojtik: httpGetFunc

mfojtik: Get

pkg/scripts/install.go
```diff
@@ -0,0 +1,173 @@
...
+	Installed bool
+
+	// Error last error encountered during install operation
+	Error error
+}
+
+// Installer interface is responsible for installing scripts needed to run the build
+type Installer interface {
+	InstallRequired(scripts []api.Script, dstDir string) (map[api.Script]*InstallResult, error)
+	InstallOptional(scripts []api.Script, dstDir string) map[api.Script]*InstallResult
```
mfojtik: +1

pkg/scripts/install.go
```diff
@@ -0,0 +1,173 @@
...
+		url, err := url.Parse(scriptsURL + "/" + string(script))
+		if err != nil {
+			result.Error = err
+			continue
+		}
+
+		target := filepath.Join(dstDir, string(script))
+		download, err := i.downloader.Download(url, target)
+		// make sure to have appropriate file permissions for downloaded scripts
+		if download && err == nil {
```
mfojtik: this should be done in the downloader.Download() method.

pkg/scripts/install.go
```diff
@@ -0,0 +1,173 @@
...
+			if err = i.fs.Chmod(target, 0700); err != nil {
+				result.Error = err
+			}
+		}
+		// save information about download & error only if there's no  previous information
+		if !result.Downloaded {
+			result.Downloaded = download
+		}
+		if result.Error == nil {
```
mfojtik: why not just assigning result.Error = err, why is 'nil' a special case?

pkg/scripts/install.go
```diff
@@ -0,0 +1,173 @@
...
+			result.Error = err
+		}
+	}
+}
+
+func (i *installer) install(results map[api.Script]*InstallResult, dstDir string) {
+	for script, result := range results {
+		// ignore not downloaded scripts
```
mfojtik: add better comment, like: "Skip installation for already downloaded scripts"

pkg/scripts/install.go
```diff
@@ -0,0 +1,173 @@
...
+			result.Installed = true
+			continue
+		}
+
+		scriptPath := i.getScriptPath(script, dstDir)
+		if scriptPath == "" {
+			result.Error = errors.NewInstallError(script)
+			continue
+		}
+		if err := i.fs.Rename(scriptPath, filepath.Join(dstDir, "upload/scripts", string(script))); err != nil {
```
mfojtik: nuke this if:

```
result.Error := i.fs.Rename(scriptPath, filepath.Join(dstDir, "upload/scripts", string(script)))
if result.Error != nil {
  continue
}
```

pkg/scripts/install.go
```diff
@@ -0,0 +1,173 @@
...
+}
+
+// getScriptPath checks scripts retrived from different sources in the following order:
+// - script found at the --scripts URL
+// - script found in the application source .sti/bin directory
+// - script found at the default image URL
+// and returns first match.
+func (i *installer) getScriptPath(script api.Script, dir string) string {
+	scriptLocationsOrder := map[api.Location]string{
+		api.DownloadedScripts: "user provided url",
```
mfojtik: If this is here only to make V(2) log output nicer, I would rather nuke it.

pkg/api/script.go
```diff
@@ -3,6 +3,9 @@ package api
 // Script defines script names used by STI process.
 type Script string
 
+// Location defines script locations used by STI process.
+type Location string
```
mfojtik: nuke this ;-) I think this is really a bad practice to mask the string into special type....

pkg/build/strategies/sti/sti.go
```diff
@@ -156,7 +156,7 @@ func (b *STI) Prepare(request *api.Request) error {
 
 	// Setup working directories
 	for _, v := range workingDirs {
-		if err := b.fs.MkdirAll(filepath.Join(request.WorkingDir, v)); err != nil {
+		if err := b.fs.MkdirAll(filepath.Join(request.WorkingDir, string(v))); err != nil {
```
mfojtik: if you nuke Location, you don't have to do string()

pkg/build/strategies/sti/sti.go
```diff
@@ -169,15 +169,11 @@ func (b *STI) Prepare(request *api.Request) error {
...
-		b.requiredScripts, request.WorkingDir, true); err != nil {
+	if _, err := b.installer.InstallRequired(b.requiredScripts, request.WorkingDir); err != nil {
 		return err
 	}
-
-	if request.ExternalOptionalScripts, err = b.installer.DownloadAndInstall(
-		b.optionalScripts, request.WorkingDir, false); err != nil {
-		glog.Warningf("Failed downloading optional scripts: %v", err)
-	}
+	b.installer.InstallOptional(b.optionalScripts, request.WorkingDir)
```
mfojtik: i would iterate over result and log the warnings about failed optional scripts download under V(2)

pkg/scripts/download.go
```diff
@@ -51,6 +40,50 @@ func NewDownloader() Downloader {
 	}
 }
 
+// DownloadFile downloads the file pointed to by URL into local targetFile
```
mfojtik: godoc

pkg/scripts/download.go
mfojtik: also nuke the boolean, see my other comment

pkg/scripts/install.go
```diff
@@ -0,0 +1,173 @@
...
+
+	"github.com/openshift/source-to-image/pkg/api"
+	"github.com/openshift/source-to-image/pkg/docker"
+	"github.com/openshift/source-to-image/pkg/errors"
+	"github.com/openshift/source-to-image/pkg/util"
+)
+
+// InstallResult structure describes the result of install operation
+type InstallResult struct {
+
```
mfojtik: you can also put the Name here and then you don't need to return the map. You can just iterate over result and get the result based on the name.

pkg/scripts/install.go
```diff
@@ -0,0 +1,173 @@
...
+// InstallOptional downloads and installs a set of scripts into dstDir, the result is a
+// map of scripts with detailed information about each of the scripts install process
+func (i *installer) InstallOptional(scripts []api.Script, dstDir string) map[api.Script]*InstallResult {
+	return i.run(scripts, dstDir)
+}
+
+func (i *installer) run(scripts []api.Script, dstDir string) map[api.Script]*InstallResult {
+	results := make(map[api.Script]*InstallResult)
+	if len(scripts) == 0 {
```
mfojtik: move this check into InstallOptional / InstallRequired

pkg/scripts/install.go
```diff
@@ -0,0 +1,173 @@
...
+		if !result.Downloaded {
+			result.Downloaded = download
+		}
+		if result.Error == nil {
+			result.Error = err
+		}
+	}
+}
+
+func (i *installer) install(results map[api.Script]*InstallResult, dstDir string) {
```
mfojtik: this seems wrong :-) you installed using install results...

pkg/build/strategies/sti/sti.go
soltysh: Here, based on `ExternalScripts` flag docker knows if to look for scripts inside the image or not. Hence I'm reworking this part I'll be clearing this all out.

pkg/errors/errors.go
soltysh: Will do once I'm done with refactoring :smile:

pkg/scripts/download.go
soltysh: Done.

pkg/scripts/install.go
soltysh: Actually, I'll move this to installer, that'll be the best option IMHO.

pkg/scripts/install.go
soltysh: This is the actual install part, we're picking a script and we're moving it to `upload` directory. Without it you won't get any scripts uploaded to builder image.

pkg/scripts/install.go
soltysh: why? to have it twice?

soltysh:
@mfojtik this the final proposition of the refactor, lemme know what you think about it. I'll be working on tests since I've changed a couple of things.

pkg/api/script.go
```diff
@@ -12,6 +12,17 @@ const (
 	SaveArtifacts Script = "save-artifacts"
 	// Usage i the name of the script responsible for printing the builder image's short info.
 	Usage Script = "usage"
+
+	// UserScripts is the location of scripts downloaded from user provided URL (-s flag).
+	UserScripts string = "downloads/scripts"
```
mfojtik: you don't need to define type in const if you assigning a string.

pkg/api/types.go
```diff
@@ -83,3 +83,23 @@ type Result struct {
 	// ImageID describes resulting image ID.
 	ImageID string
 }
+
+// InstallResult structure describes the result of install operation
+type InstallResult struct {
+
+	// Name describes which script this result refers to
+	Script Script
```
mfojtik: Rename this to Name or fix the godoc ;-) I'm ok with both

pkg/build/strategies/sti/sti.go
```diff
@@ -169,14 +169,15 @@ func (b *STI) Prepare(request *api.Request) error {
...
+	required, err := b.installer.InstallRequired(b.requiredScripts, request.WorkingDir)
+	if err != nil {
 		return err
 	}
-
-	if request.ExternalOptionalScripts, err = b.installer.DownloadAndInstall(
-		b.optionalScripts, request.WorkingDir, false); err != nil {
-		glog.Warningf("Failed downloading optional scripts: %v", err)
+	optional := b.installer.InstallOptional(b.optionalScripts, request.WorkingDir)
+	if glog.V(1) {
```
mfojtik: just a nit, I would put a newline here because the `if` is not related to the line above.

pkg/errors/errors.go
```diff
@@ -10,15 +10,17 @@ import (
 const (
 	ErrInspectImage int = 1 + iota
```
mfojtik: you don't have to do this now, but I think the correct name for all there errors should be:

InspectImageError
PullImageError
...etc

pkg/scripts/download.go
```diff
@@ -51,21 +38,59 @@ func NewDownloader() Downloader {
...
+		}
+		return err
+	}
+	defer reader.Close()
+
+	out, err := os.Create(targetFile)
+	defer out.Close()
+
+	if err != nil {
+		os.Remove(targetFile)
```
mfojtik: if you were not able to create the targetFile, why you are removing it? ;-)

pkg/scripts/download.go
```diff
@@ -51,21 +38,59 @@ func NewDownloader() Downloader {
...
-func (h *HttpURLReader) IsFromImage() bool {
-	return false
+// HttpURLReader retrieves a response from a given http(s) URL
+type HttpURLReader struct {
+	Get func(url string) (*http.Response, error)
+}
+
+// NewHTTPReader creates an instance of the HttpURLReader
+func NewHTTPReader() schemeReader {
+	reader := &HttpURLReader{}
```
mfojtik:
```
return &HttpURLReader{Get: http.Get}
```

pkg/scripts/install.go
```diff
@@ -0,0 +1,184 @@
...
+	return i.run(scripts, dstDir)
+}
+
+type downloadResult struct {
+	location string
+	err      error
+}
+
+func (i *installer) run(scripts []api.Script, dstDir string) []api.InstallResult {
+	if len(scripts) == 0 {
```
mfojtik: this check should be moved to InstallRequired/Optional?

pkg/scripts/install.go
```diff
@@ -0,0 +1,184 @@
...
+				continue
+			}
+			result.URL = downloadResult.location
+			// if location results are erroneous we store error in result object
+			// and continue searching other locations
+			if downloadResult.err != nil {
+				// one exception is when error contains information about scripts being inside the image
+				if e, ok := downloadResult.err.(errors.Error); ok && e.ErrorCode == errors.ErrScriptsInsideImage {
+					// in which case update result object and break further searching
+					result.Error = nil
```
mfojtik: can you create new api.InstallResult here? instead of doing these assignments? It will squash this 3 lines into one and the same on line 175

pkg/scripts/install.go
```diff
@@ -0,0 +1,184 @@
...
+		defaultResults = i.download(defaultURL, scripts, filepath.Join(dstDir, api.DefaultScripts))
+	} else {
+		glog.Warningf("Image %s does not contain STI_SCRIPTS_URL defined!", i.image)
+	}
+
+	return i.install(scripts, userResults, sourceResults, defaultResults, dstDir)
+}
+
+func (i *installer) download(scriptsURL string, scripts []api.Script, dstDir string) map[api.Script]*downloadResult {
+	result := make(map[api.Script]*downloadResult)
```
mfojtik: set the result size as you know what is the len(script).

mfojtik:
@soltysh minor nits, otherwise LGTM, good job!

pkg/scripts/install.go
```diff
@@ -0,0 +1,184 @@
...
+
+	var userResults, sourceResults, defaultResults map[api.Script]*downloadResult
+
+	// get scripts from user provided URL
+	if i.scriptsURL != "" {
+		userResults = i.download(i.scriptsURL, scripts, filepath.Join(dstDir, api.UserScripts))
+	}
+
+	// get scripts from source
+	sourceResults = make(map[api.Script]*downloadResult)
```
mfojtik: add len(scripts) here.

pkg/scripts/install.go
soltysh: It would complicate code a little bit, since I'm always returning the last error for the user as an information.

soltysh:
Now that the PR is fully ready, I'll rebase it for clearer reading :wink:

mfojtik:
LGTM now

soltysh:
Ready for final review @mfojtik @bparees @csrwng 
[test] in the mean time

pkg/build/strategies/sti/sti.go
```diff
@@ -253,7 +260,7 @@ func (b *STI) Determine(request *api.Request) (err error) {
 	// we're assuming save-artifacts to exists for embedded scripts (if not we'll
 	// warn a user upon container failure and proceed with clean build)
 	// for external save-artifacts - check its existence
-	saveArtifactsExists := request.ExternalOptionalScripts ||
+	saveArtifactsExists := !b.externalScripts[api.SaveArtifacts] ||
```
bparees: I still don't understand why the first half of this check is needed?  isn't it always possible save-artifacts exists inside the image, regardless of what was downloaded?

pkg/errors/errors.go
bparees: +1

pkg/build/strategies/sti/sti.go
soltysh: Since I reworked the installer, the return information from installer precisely states if files was downloaded or not and the origin of the file. Since I'll be reworking this code as part of #130 I'll think about simplifying this, because I don't like the complexity of this part.

soltysh:
Fixed gofmt for `pkg/scripts/install_test.go`.

pkg/api/script.go
```diff
@@ -12,6 +12,17 @@ const (
...
+	// UserScripts is the location of scripts downloaded from user provided URL (-s flag).
+	UserScripts = "downloads/scripts"
+	// DefaultScripts is the location of scripts downloaded from default location (STI_SCRIPTS_URL environment variable).
+	DefaultScripts = "downloads/defaultScripts"
+	// SourceScripts is the location of scripts downloaded with application sources.
+	SourceScripts = "upload/src/.sti/bin"
+	// UploadScripts is the location of scripts that will be downloaded to the image during STI build.
+	UploadScripts = "upload/scripts"
+	// Source is the location of application sources.
+	Source string = "upload/src"
```
csrwng: the type for Source should be implied, no need to specify 'string'

pkg/api/script.go
```diff
@@ -12,6 +12,17 @@ const (
 	SaveArtifacts Script = "save-artifacts"
 	// Usage i the name of the script responsible for printing the builder image's short info.
 	Usage Script = "usage"
+
```
csrwng: For clarity, it may be useful to separate these constants in 2 blocks... one with 'Script' type constants and the other one for paths.

pkg/api/script.go
```diff
@@ -12,6 +12,17 @@ const (
...
+	// UserScripts is the location of scripts downloaded from user provided URL (-s flag).
+	UserScripts = "downloads/scripts"
+	// DefaultScripts is the location of scripts downloaded from default location (STI_SCRIPTS_URL environment variable).
+	DefaultScripts = "downloads/defaultScripts"
+	// SourceScripts is the location of scripts downloaded with application sources.
+	SourceScripts = "upload/src/.sti/bin"
+	// UploadScripts is the location of scripts that will be downloaded to the image during STI build.
```
csrwng: s/downloaded/uploaded/

soltysh:
@bparees does this change https://github.com/soltysh/source-to-image/commit/75b399b9ef232026e7329c8fe2456955d9962cdc buys you?

pkg/api/script.go
csrwng: Also, it may make it clearer in the rest of the code to suffix these with 'Dir' or something like that so you know what the constant is referring to.

soltysh:
Rebased and fixed @csrwng comments.

pkg/scripts/install.go
```diff
@@ -0,0 +1,184 @@
...
+			sourceResults[script].err = errors.NewDownloadError(file, -1)
+		}
+	}
+
+	// get scripts from default URL
+	defaultURL, err := i.docker.GetScriptsURL(i.image)
+	if err == nil && defaultURL != "" {
+		defaultResults = i.download(defaultURL, scripts, filepath.Join(dstDir, api.DefaultScripts))
+	} else {
+		glog.Warningf("Image %s does not contain STI_SCRIPTS_URL defined!", i.image)
```
csrwng: Maybe it's more accurate to say that you can't get the STI_SCRIPTS_URL from the image (may be defined, but some other error occurred retrieving it)

pkg/scripts/install.go
csrwng: or you could define a method on result like .Set(x,y,z)

pkg/scripts/install.go
soltysh: I'll remove that entirely, since `GetScriptsURL` already has that kind of warning, so it doesn't make sense to repeat it.

pkg/scripts/install.go
soltysh: I don't define methods on types mostly because I was told by you know who :wink: not to do that.

bparees:
@soltysh looks ok to me, but you'd best make sure @csrwng is happy.

csrwng:
LGTM

csrwng:
@bparees merge?

bparees:
[merge]

PR #130, opened 2015-02-17
Card devexp_426 - Enable retrieving images for incremental builds from external repo.

soltysh:
This will be waiting for #129 to be merged first.

### DIFF
--- a/cmd/sti/main.go
+++ b/cmd/sti/main.go
@@ -106,7 +106,7 @@ func newCmdBuild(req *api.Request) *cobra.Command {
 		},
 	}
 
-	buildCmd.Flags().BoolVar(&(req.Clean), "clean", false, "Perform a clean build")
+	buildCmd.Flags().BoolVar(&(req.Clean), "clean", true, "Perform a clean build")
 	buildCmd.Flags().BoolVar(&(req.RemovePreviousImage), "rm", false, "Remove the previous image during incremental builds")
 	buildCmd.Flags().StringP("env", "e", "", "Specify an environment var NAME=VALUE,NAME2=VALUE2,...")
 	buildCmd.Flags().StringVarP(&(req.Ref), "ref", "r", "", "Specify a ref to check-out")
--- a/pkg/build/interfaces.go
+++ b/pkg/build/interfaces.go
@@ -23,11 +23,11 @@ type Cleaner interface {
 
 // IncrementalBuilder provides methods that is used for builders that implements
 // the 'incremental' build workflow.
-// The Determine method checks if the artifacts from the previous build exists
+// The Exists method checks if the artifacts from the previous build exists
 // and if they can be used in the current build.
 // The Save method stores the artifacts for the next build.
 type IncrementalBuilder interface {
-	Determine(*api.Request) error
+	Exists(*api.Request) bool
 	Save(*api.Request) error
 }
 
--- a/pkg/build/strategies/sti/sti.go
+++ b/pkg/build/strategies/sti/sti.go
@@ -53,12 +53,12 @@ type STI struct {
 	installedScripts map[api.Script]bool
 
 	// Interfaces
-	preparer    build.Preparer
-	incremental build.IncrementalBuilder
-	scripts     build.ScriptsHandler
-	source      build.Downloader
-	garbage     build.Cleaner
-	layered     build.Builder
+	preparer  build.Preparer
+	artifacts build.IncrementalBuilder
+	scripts   build.ScriptsHandler
+	source    build.Downloader
+	garbage   build.Cleaner
+	layered   build.Builder
 }
 
 // New returns the instance of STI builder strategy for the given request.
@@ -94,7 +94,7 @@ func New(req *api.Request) (*STI, error) {
 
 	// Set interfaces
 	b.preparer = b
-	b.incremental = b
+	b.artifacts = b
 	b.scripts = b
 	b.postExecutor = b
 	return b, err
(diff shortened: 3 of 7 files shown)


pkg/build/strategies/sti/sti.go
```diff
@@ -250,6 +248,15 @@ func (b *STI) Determine(request *api.Request) (err error) {
 		return
 	}
 
+	// if image does not exist in local registry try searching for it
+	if !previousImageExists && request.ForcePull {
```
bparees: why are you checking forcepull here?  forcepull is to force pulling the sti base image, right?  we shouldn't overload it's meaning.

also according to the trello card i think the behavior should be:
1) default to clean build (do not pull previous image)
2) otherwise you always need to pull the previous image (even if it exists, it could be stale)

soltysh: You're right about the `forcePull` flag. But I'm not sure what you mean by the later. That's true I should be downloading the image even if previous exists (might be older version), but I'm not sure what you mean by 1?

pkg/build/strategies/sti/sti.go
```diff
@@ -115,7 +115,7 @@ func (b *STI) Build(request *api.Request) (*api.Result, error) {
 	if request.Incremental {
 		glog.V(1).Infof("Existing image for tag %s detected for incremental build.", request.Tag)
 	} else {
-		glog.V(1).Infof("Clean build will be performed")
+		glog.V(1).Infof("Clean build will be performed.")
```
mfojtik: is the dot needed? I can see that we are not putting dot into other messages, we should try to make the output consistent (even if this is V(1))

pkg/build/strategies/sti/sti.go
```diff
@@ -250,6 +248,15 @@ func (b *STI) Determine(request *api.Request) (err error) {
 		return
 	}
 
+	// if image does not exist in local registry try searching for it
+	if !previousImageExists && request.ForcePull {
+		image, err := b.docker.PullImage(request.Tag)
+		if err != nil {
+			glog.Warningf("Unable to find %s image!", request.Tag)
```
mfojtik: This should be V(1) at least.

pkg/build/strategies/sti/sti.go
soltysh: I'll make it consistent then by removing dots :wink: as for now we have both.

soltysh:
I rebased this PR on top of #129, but relevant commit is https://github.com/soltysh/source-to-image/commit/a69921498d847c4c2c462dad3d9e049b7ff550c3
@bparees opinions?

[test] in the meantime.

pkg/build/strategies/sti/sti.go
```diff
@@ -237,27 +243,18 @@ func (b *STI) PostExecute(containerID string, location string) error {
...
 
-	// we're assuming save-artifacts to exist for embedded scripts (if not we'll
-	// warn a user upon container failure and proceed with a clean build)
-	// for external save-artifacts - check its existence
-	saveArtifactsExists := request.ExternalOptionalScripts ||
-		b.fs.Exists(filepath.Join(request.WorkingDir, "upload", "scripts", string(api.SaveArtifacts)))
-
-	request.Incremental = previousImageExists && saveArtifactsExists
-	return nil
+	return previousImageExists && b.installedScripts[api.SaveArtifacts]
```
bparees: so if i have an image that includes a save-artifacts script, and i do not provide an external save-artifacts script, this will always return false, won't it?  because installedScripts[api.SaveArtifacts] will never be true.  So we'll never attempt an incremental build for that image, and thus never invoke its built in save-artifacts.

Am I misunderstanding something?

bparees: (this is my only remaining concern w/ this PR)

soltysh: I'm guessing you mean you have an image with `STI_SCRIPTS_URL` to to eg. `image:///myscripts/bin`. Then if you carefully look through the `install.go` you'll see that:
1. I'll invoke download for `defaultURL` see [here](https://github.com/soltysh/source-to-image/commit/10a9d2e91d5c299b8c0cad287aa3e261ab3c3b0d#diff-24a010a7140e82378231d9b71a31792eR96)
2. The return value for `save-artifacts` will be `NewScriptsInsideImageError` see [here](https://github.com/soltysh/source-to-image/commit/48e76c8aa2399abe587a7dc31d2fb48c8791e25b#diff-0fbae86cd401240768b81a29c45e3d62R116)
3. During installing I'll mark this script as not downloaded but installed see [here](https://github.com/soltysh/source-to-image/commit/10a9d2e91d5c299b8c0cad287aa3e261ab3c3b0d#diff-24a010a7140e82378231d9b71a31792eR149)

Thus you'll get exactly what you want :smile: I hope that clarifies this

bparees: awesome, thanks.  lgtm.

soltysh:
Rebased on top of recent changes, waiting for tests results.

cmd/sti/main.go
```diff
@@ -106,7 +106,7 @@ func newCmdBuild(req *api.Request) *cobra.Command {
 		},
 	}
 
-	buildCmd.Flags().BoolVar(&(req.Clean), "clean", false, "Perform a clean build")
+	buildCmd.Flags().BoolVar(&(req.Clean), "clean", true, "Perform a clean build")
```
mfojtik: why is clean a default now?

pkg/build/interfaces.go
```diff
@@ -27,7 +27,7 @@ type Cleaner interface {
 // and if they can be used in the current build.
 // The Save method stores the artifacts for the next build.
 type IncrementalBuilder interface {
-	Determine(*api.Request) error
+	Determine(*api.Request) bool
```
mfojtik: `IsIncremental(*api.Request) bool` ?

mfojtik:
LGTM (just one small style suggestion and question)...

pkg/build/interfaces.go
soltysh: Yep, that makes sense, doing it right away.

soltysh:
Addressed @mfojtik request for renaming `Determine` -> `IsIncremental`, waiting for test resuslts and merging. Future requests will be solved in separate PRs :stuck_out_tongue_closed_eyes:

soltysh:
One last (:wink:) improvment per @mfojtik request `Determine` -> `Exists` and STI internal variable renamed from `incremental` -> `artifacts`.

mfojtik:
LGTM, thanks for fixing this @soltysh !

soltysh:
[merge]

PR #137, opened 2015-02-20
Add Go 1.4 to travis and sync the test-go with origin

mfojtik:


mfojtik:
@soltysh PTAL

(I added golint there as an optional test, just want to see if the script itself is working might be useful in future)

soltysh:
LGTM, but I'd rather not to have golint yet, since I know we're not ready for it :wink:

mfojtik:
@soltysh my intent was to just exercise it so we know it 'works' ;-) (have to be fixed for travis, but we can do that later).

mfojtik:
[merge]

mfojtik:
[merge]

.travis.yml
```diff
@@ -2,13 +2,17 @@ language: go
...
-  - go get code.google.com/p/go.tools/cmd/cover
+  - go get golang.org/x/tools/cmd/cover
+  - go get -u github.com/golang/lint/golint
+  - ./hack/verify-gofmt.sh
+  - ./hack/verify-golint.sh || true
   - ./hack/build-go.sh
 
 script:
-  - ./hack/test-go.sh
+  - KUBE_RACE="-race" KUBE_COVER="-cover -covermode=atomic" ./hack/test-go.sh "" -p=4
```
bparees: seems odd our scripts use KUBE env variables....

soltysh: Actually we're not, I've renamed all of them to be `STI_*`, PR with fix to above https://github.com/openshift/source-to-image/pull/141

PR #142, opened 2015-02-23
Card devexp_426 - Renaming Clean to Incremental to better match current behaviour, where Incremental needs to be forced

soltysh:
After talking to @mfojtik we've decided to rename the `Clean` variable to `Incremental` to better match current behavior. This came up when doing https://github.com/openshift/origin/pull/1101.

/cc @bparees


mfojtik:
@soltysh LGTM, it makes more sense to me now as the 'clean' is now always true.

bparees:
+1 [merge]

PR #143, opened 2015-02-24
Renaming ruby-20-centos to new ruby-20-centos7

jhadvig:
This is a part of work on https://trello.com/c/UJG1mPZf/364-3-v3-ruby-image-image-beta2
Also last step before we cant get rid of the ruby-20-centos repository


jhadvig:
@soltysh PTAL

mfojtik:
LGTM [merge]

PR #145, opened 2015-02-24
Add support for .sti/environment file

mfojtik:


mfojtik:
fyi, this works for both ONBUILD and STI strategies

soltysh:
LGTM

README.md
```diff
@@ -40,6 +40,17 @@ image:
 Additionally for the best user experience and optimized sti operation we suggest image
 to have `/bin/sh` and tar command inside.
 
+Users can also set extra environment variables in the application source code,
+which are passed to the build and the `assemble` script consumes them. These
```
bparees: make it clear these env variables also end up existing in the output image.

pkg/scripts/environment.go
```diff
@@ -0,0 +1,73 @@
...
+
+	f, err := os.Open(envPath)
+	if err != nil {
+		return nil, errors.New("unable to read .sti/environment file")
+	}
+	defer f.Close()
+
+	result := []Environment{}
+
+	scanner := bufio.NewScanner(f)
```
bparees: there's really no built in properties-style parsing library for go?

pkg/build/strategies/sti/sti.go
```diff
@@ -296,6 +296,13 @@ func (b *STI) Save(request *api.Request) (err error) {
 func (b *STI) Execute(command string, request *api.Request) error {
 	glog.V(2).Infof("Using image name %s", request.BaseImage)
 
+	env, err := scripts.GetEnvironment(request)
+	if err != nil {
+		glog.V(1).Infof("No .sti/environment provided (%v)", err)
+	}
+
+	buildEnv := append(scripts.ConvertEnvironment(env), b.generateConfigEnv()...)
```
bparees: what happens for layered builds (the case where we don't have tar)?

mfojtik: good catch, we should do the same as for on build. I will add that tomorrow.

pkg/scripts/environment.go
mfojtik: I will check stdlib, but I'm sceptical

mfojtik:
@bparees fixed. the layered build will get the env vars same way as regular STI build. I was looking at stdlib for property file parser, but found nothing (I found some 3th party libs, but I think they are overkill for this)

mfojtik:
[test]

bparees:
lgtm.  [merge]

PR #151, opened 2015-03-02
Fixed copy operation upon working with local application source

soltysh:
@mfojtik PTAL


mfojtik:
LGTM (as long as contextDir works ;-)

soltysh:
It does, while testing it I've found this small nits in sti-ruby https://github.com/openshift/sti-ruby/pull/10

[merge]

PR #152, opened 2015-03-09
Fix fs.Copy() to handle copying STI scripts

mfojtik:


mfojtik:
Should fix https://bugzilla.redhat.com/show_bug.cgi?id=1198888

@bparees @soltysh PTAL

pkg/util/fs.go
```diff
@@ -76,12 +76,13 @@ func (h *fs) Exists(file string) bool {
...
-	info, err := os.Stat(sourcePath)
-	if err != nil {
+	if _, err := os.Stat(sourcePath); err != nil {
 		return err
 	}
 
-	if !info.IsDir() {
+	info, err := os.Stat(targetPath)
+
+	if err != nil || (info != nil && !info.IsDir()) {
```
soltysh: I've just tested it, there's no need to check info for nillnes, either there's an error or you'll get that info.

soltysh:
Other than this small nit, LGTM.

pkg/util/fs.go
mfojtik: I got panic if I don't check for nil here and the err is not nil

soltysh:
[merge]

PR #153, opened 2015-03-11
Bug 1196967 - fixed version information

soltysh:
Fixed some leftovers in the release process, especially the binary version which was trying to modify old package `pkg/sti/version` instead of `pkg/version`.

@bparees PTAL


bparees:
lgtm. [merge]

PR #160, opened 2015-03-26
Add --assembleOutput=true to sti build command

mfojtik:
With this `sti build` always streams the output from the `assemble` script to console (so users can see the build progress/warning/non-fatal errors without bumping loglevel).

Additionally I switched 'git' into quite mode (you don't really need to see the git clone command, only errors when the git fails). Also the git command output is now streamed via glog() to have consistent output.

The `assembleOutput=false` will make the output super-silent. Nothing is returned to users except errors.

### DIFF
--- a/pkg/build/strategies/sti/sti.go
+++ b/pkg/build/strategies/sti/sti.go
@@ -108,7 +108,7 @@ func New(req *api.Request) (*STI, error) {
 func (b *STI) Build(request *api.Request) (*api.Result, error) {
 	defer b.garbage.Cleanup(request)
 
-	glog.Infof("Building %s", request.Tag)
+	glog.V(1).Infof("Building %s", request.Tag)
 	if err := b.preparer.Prepare(request); err != nil {
 		return nil, err
 	}
@@ -232,10 +232,10 @@ func (b *STI) PostExecute(containerID string, location string) error {
 	b.result.ImageID = imageID
 
 	if len(b.request.Tag) > 0 {
-		glog.Infof("Successfully built %s", b.request.Tag)
+		glog.V(1).Infof("Successfully built %s", b.request.Tag)
 		glog.V(1).Infof("Tagged %s as %s", imageID, b.request.Tag)
 	} else {
-		glog.Infof("Successfully built %s", imageID)
+		glog.V(1).Infof("Successfully built %s", imageID)
 	}
 
 	if b.incremental && b.request.RemovePreviousImage && previousImageID != "" {
@@ -367,7 +367,7 @@ func (b *STI) Execute(command string, request *api.Request) error {
 				}
 				break
 			}
-			if glog.V(2) || command == api.Usage {
+			if glog.V(2) || request.AssembleOutput == true || command == api.Usage {
 				glog.Info(text)
 			}
 		}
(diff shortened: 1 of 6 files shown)


mfojtik:
@bparees @soltysh

pkg/build/strategies/sti/sti.go
```diff
@@ -367,7 +367,7 @@ func (b *STI) Execute(command string, request *api.Request) error {
 				}
 				break
 			}
-			if glog.V(2) || command == api.Usage {
+			if glog.V(2) || request.AssembleOutput == true || command == api.Usage {
```
bparees: shouldn't you also confirm that the command being run is assemble? As this is written you're going to get output from other Execution steps than just assemble, when assembleoutput is true, right?  eg save-artifacts

bparees:
also i'm not sure i love the approach of adding a specific flag for this instead of just making assemble output be at a particular loglevel that can be set, but I guess i can see the argument for doing it this way.

soltysh:
The thing with this is, @mfojtik wants to see just assemble output, which is logged at level=2, but at this level we're also printing more detailed information about the STI process itself. Maybe then we should change those loglevels, where:
0 - no output
1 - just assemble (actually all docker output)
2 - basic sti info (previous 1)
3 - detailed sti info (previous 2 w/o docker output)
4 - tar output

soltysh:
I'm relatively happy with this change, I do understand reasoning, but on the other hand if breaks the log levels idea that should be consistent.

mfojtik:
I don't think that assemble output should be logged under loglevel 2, it should be logged always. It is the most useful thing that STI provides to debug problems or see the actual build. Other informations from STI are less relevant for the end users IMHO.

mfojtik:
iow. I don't want to need to set --loglevel=1 just to see assemble output, I want to see it always (and have option to silence it if I'm not interested in it (or I want to make my scripts silent for whatever reason).

soltysh:
Whereas I sometimes want to have clean output :smile: but that can be
achieved with `sti build &>/dev/null` as well. So I'm sold, but for
redefining the build logs not for additional flag.

mfojtik:
@soltysh there is reason why tools like 'git' have '--quite' option instead of telling you to use &>/dev/null (piping?).

mfojtik:
@soltysh what I'm trying to sold here is the idea that output from assemble should not be treated just like a special log level you set to see, but something you should see always (or never).

bparees:
@mfojtik yeah, i get it and i'm ok with it, just bothers my sensibilities a little that we already have a tool for this job (log level).  As i said it would require reworking the loglevels to get exactly this behavior so it's probably not the right answer to do that.

lgtm.

mfojtik:
@bparees sure, fyi this is part of the build log cleanup work to make the errors more visible to users... with this change users will see assemble output and nothing else. With loglevel=1 I will have to tweak the loglevel in build containers ;-) which will be a bit messy

bparees:
@mfojtik yup, this should fulfil that card i think, plus your other PR

soltysh:
:+1:

mfojtik:
[merge]

PR #161, opened 2015-03-26
Change --assembleOutput to --quiet

mfojtik:


soltysh:
Thanks for doing it :heart: LGTM
[merge]

bparees:
still wondering what this will mean for when we are executing save-artifacts.  seems like that output would be shown now, by default.

soltysh:
Yes it does since we'll be showing all containers output. But by default
we're not invoking so that I wouldn't worry for now at least :wink:

PR #167, opened 2015-04-10
Fixed script path when invoking PostExecute.

soltysh:
@bparees this is initial take to fix the problem with custom `assemble` script inside source repo and `run` pointing to non existent file. This is not ready yet as I need to address  the tests.

### DIFF
--- a/pkg/build/strategies/layered/layered.go
+++ b/pkg/build/strategies/layered/layered.go
@@ -141,8 +141,6 @@ func (b *Layered) Build(request *api.Request) (*api.Result, error) {
 	b.request.BaseImage = newBaseImage
 	// the scripts are inside the image
 	b.request.ScriptsURL = "image://" + filepath.Join(getLocation(request), "scripts")
-	// the source is also inside the image
-	b.request.Location = filepath.Join(getLocation(request), "src")
 
 	glog.V(2).Infof("Building %s using sti-enabled image", b.request.Tag)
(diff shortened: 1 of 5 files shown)


bparees:
will this work for save-artifacts too?

soltysh:
@bparees it will, because the `save-artifacts` is run using the same method as `assemble` (see [here](https://github.com/openshift/source-to-image/blob/master/pkg/build/strategies/sti/sti.go#L298) and [here](https://github.com/openshift/source-to-image/blob/master/pkg/docker/docker.go#L228)). Besides I just double checked it :wink:

soltysh:
@bparees it's ready for your review. The source of the problem was we have a logic around the destination of scripts in `docker.go` but we were missing that in `sti.go#PostExecute`, that's basically what this patch includes. In the mean time
[test]

bparees:
lgtm.  fixes https://github.com/openshift/source-to-image/issues/173

@thoraxe fyi.. after this goes in we'll have to bring it into origin but then overriding run/assemble independently will work.

[merge]

soltysh:
Strange error, as if problem with our vagrant plugin.
re-[merge]

thoraxe:
@bparees step forward!

Do we have a plan for "additive" scripts? EG not having to override the entire script with a copy of the original?

bparees:
We're still carrying an open issue for that.

Ben Parees | OpenShift

soltysh:
[merge] 

again, as previous one wasn't picked up

Issue #169, opened 2015-04-15
README.md still makes reference to --clean


cianclarke:
https://github.com/openshift/source-to-image/blame/master/README.md#L92
    
    If a save-artifacts script exists, a prior image already exists, and the --clean option is not used, the workflow is as follows:
    
I'm not sure how this line should read, as the operation seems to have been flipped too? --clean=true is --incremental=false - but should probably update README. Would be good to also

bparees:
thanks, clarified it here:
https://github.com/openshift/source-to-image/pull/170

PR #170, opened 2015-04-15
fix reference to --clean flag

bparees:


bparees:
fixes https://github.com/openshift/source-to-image/issues/169

PR #178, opened 2015-05-06
Fixed the PostExecute command and added tests to cover that part

soltysh:
@bparees @rhcarvalho this fix should solve your problems. Sorry for that  :disappointed:

[test]

### DIFF
--- a/pkg/build/strategies/sti/sti.go
+++ b/pkg/build/strategies/sti/sti.go
@@ -221,10 +221,10 @@ func (b *STI) PostExecute(containerID, location string) error {
 
 	buildEnv := append(scripts.ConvertEnvironment(env), b.generateConfigEnv()...)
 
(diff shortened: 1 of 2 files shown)


bparees:
hmm:
$ make clean build

$ sti build git://github.com/openshift/ruby-hello-world openshift/ruby-20-centos7 test-ruby-app

$ docker run test-ruby-appexec: "/tmp/scripts/run": stat /tmp/scripts/run: no such file or directory
FATA[0002] Error response from daemon: Cannot start container 68a71466712b239804e37380ac21a6e010211798aef6a96ea85438ae60387065: [8] System error: exec: "/tmp/scripts/run": stat /tmp/scripts/run: no such file or directory

bparees:
(that was with your changes, to be clear)

bparees:
@soltysh hm, well i'm not sure what's up with that test but i tried your fix on a devenv and it does seem to work, so.. lgtm.
[merge]

PR #182, opened 2015-05-11
merge env variables when adding them

bparees:


bparees:
@soltysh ptal.

bparees:
[test]

pkg/scripts/environment.go
```diff
@@ -66,8 +66,15 @@ func ConvertEnvironment(env []Environment) (result []string) {
 
 // ConvertEnvironmentToDocker converts the []Environment into Dockerfile format
 func ConvertEnvironmentToDocker(env []Environment) (result string) {
+	first := true
```
soltysh: Why introducing additional variable, if you can use index in the for loop.

pkg/scripts/environment.go
```diff
@@ -66,8 +66,15 @@ func ConvertEnvironment(env []Environment) (result []string) {
...
 	for _, e := range env {
-		result += fmt.Sprintf("ENV %s %s\n", e.Name, e.Value)
+		if first {
+			result += fmt.Sprintf("ENV %s=\"%s\"", e.Name, e.Value)
+			first = false
+		} else {
+			result += fmt.Sprintf(" \\\n\t%s=\"%s\"", e.Name, e.Value)
+		}
 	}
```
soltysh: Besides to avoid that if inside the loop, you could create first the `ENV` line and then do `for _, e := range env[1:]` (don't forget to check the length of `env` before doing that :wink: )

soltysh:
@bparees two small nits and one big one, please split the PR into Godeps changes and your actual changes, like we do that in origin.

pkg/scripts/environment.go
bparees: good point. done.

pkg/scripts/environment.go
bparees: nice optimization but i'm going to rely on branch prediction, given that adding the length check seems at least equally burdensome in terms of the code complexity.

bparees:
@soltysh commits split.  i wasn't clear on how to describe this particular commit since it's not an update of an existing dep, it's a new dep.

soltysh:
It's ok.
LGTM [merge]

soltysh:
I've seen that error once, if it repeats I need to have a closer look at it. For now 
re-[merge].

PR #184, opened 2015-05-14
Add human-friendly api.Request printer

mfojtik:
Produces output:

```console
$ sti build --loglevel=2 -e "FOO=BAR,TEST=BAR" foo bar test 

Base Image:		bar
Source:			foo
Output Image Tag:	test
Environment:		FOO=BAR,TEST=BAR
Incremental Build:	disabled
Remove Old Build:	disabled
Force Pull:		enabled
Quiet:			disabled
Layered Build:		disabled
Docker Endpoint:	unix:///var/run/docker.sock
Docker Pull Config:	/home/mfojtik/.dockercfg
Docker Pull User:	mfojtik
```

(the fields are actually aligned, GH messed this up)

### DIFF
--- a/pkg/api/printer.go
+++ b/pkg/api/printer.go
@@ -1,0 +1,89 @@
+package api
+
+import (
+	"bytes"
+	"fmt"
+	"io"
+	"os"
+	"strings"
+	"text/tabwriter"
+)
+
+func (r *Request) PrintObj() string {
+	out, err := tabbedString(func(out io.Writer) error {
+		fmt.Fprintf(out, "Builder Image:\t%s\n", r.BaseImage)
+		fmt.Fprintf(out, "Source:\t%s\n", r.Source)
+		if len(r.Ref) > 0 {
+			fmt.Fprintf(out, "Source Ref:\t%s\n", r.Ref)
+		}
+		if len(r.ContextDir) > 0 {
+			fmt.Fprintf(out, "Context Directory:\t%s\n", r.ContextDir)
+		}
+		fmt.Fprintf(out, "Output Image Tag:\t%s\n", r.Tag)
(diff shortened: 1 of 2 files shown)


pkg/api/printer.go
```diff
@@ -0,0 +1,89 @@
...
+	}
+	return out
+}
+
+func printEnv(out io.Writer, env map[string]string) {
+	result := []string{}
+	for k, v := range env {
+		result = append(result, fmt.Sprintf("%s=%s", k, v))
+	}
+	if len(result) == 0 {
```
soltysh: I'd move that to the beginning and check env length acutally, fail fast, right?

pkg/api/printer.go
```diff
@@ -0,0 +1,89 @@
...
+	"fmt"
+	"io"
+	"os"
+	"strings"
+	"text/tabwriter"
+)
+
+func (r *Request) PrintObj() string {
+	out, err := tabbedString(func(out io.Writer) error {
+		fmt.Fprintf(out, "Base Image:\t%s\n", r.BaseImage)
```
soltysh: In all our docs we call that the builder image, I'm aware of the var naming, still I'd say to keep with the Builder name.

soltysh:
Two small nits, but LGTM.

pkg/api/printer.go
bparees: +1

pkg/api/printer.go
bparees: +1

bparees:
agree w/ @soltysh's comments and otherwise lgtm.

mfojtik:
Thanks guys, all comments fixed. [merge]

PR #186, opened 2015-05-15
Add Security section into README

mfojtik:


soltysh:
LGTM [merge]

strzibny:
I must say that `$` signs are confusing in README but if you say that it can be run under plain user in some cases, I am fine with it.

PR #193, opened 2015-05-18
Fix package names in release Dockerfile

mfojtik:


mfojtik:
@bparees saves the release \o/

bparees:
what did i do?
[merge]

bparees:
@smarterclayton apparently this fixed the cross compile issue for STI.  I don't know the logic behind it, perhaps @mfojtik can explain why the amd64 packages work.

bparees:
ah, found the IRC discussion:
(10:55:02 AM) vbatts: mfojtik: i don't think this build/release image has all the bootstrapping needed
(10:55:33 AM) vbatts: mfojtik: the build is failing on darwin/amd64, but the golang-pkg-darwin-amd64.noarch rpm is not installed

Issue #199, opened 2015-05-25
UID isolation


soltysh:
Per @smarterclayton comment on my recent PR, I'm opening issue to verify if following comment is still valid (from https://github.com/openshift/source-to-image/blob/master/test/integration/images/sti-fake/Dockerfile)
```
RUN mkdir -p /sti-fake/src && \
# FIXME: The 'default' user has conflicting UID (1000). This should be fixed in
#        how source-to-image handles UID's during isolation.
#
    deluser default
```

soltysh:
I've tested this without this line as well as with other images, namely `openshift/python-33-centos7` the is no problem. I'm guessing this is a leftover from the time we were doing bind-mounts to copy the sources of the application. Since we currently stream it with tar, the resulting files are created with permissions of the user defined in the dockerfile. @bparees @smarterclayton @csrwng @mfojtik correct me if I'm wrong, if so I suggest removing this comment as it's not valid anymore.

bparees:
@soltysh sounds plausible to me.

csrwng:
@soltysh I'm fine with removing that line. Doesn't seem like it's needed.

soltysh:
Closing per #179.

Issue #179, opened 2015-05-06
sti hangs in docker pull if tag is invalid


bparees:
run STI and provide no tag on the image name, for an image repo where there is no "latest" tag, result is STI hangs forever in "pulling image"... seems like it must not be handling an error correctly:

$ docker pull ce-registry.usersys.redhat.com/jboss-webserver3/tomcat8-openshift
FATA[0005] Tag latest not found in repository ce-registry.usersys.redhat.com/jboss-webserver3/tomcat8-openshift

$ sti build https://github.com/bparees/session-app ce-registry.usersys.redhat.com/jboss-webserver3/tomcat8-openshift badout --loglevel=5
I0506 18:54:38.631805 07628 docker.go:173] Pulling image ce-registry.usersys.redhat.com/jboss-webserver3/tomcat8-openshift

(never errors/returns)

Note that if you provide an actual invalid tag, it does immediately error out.  Not sure it's actually valid to have a repo with no "latest" tag in it?

$ sti build https://github.com/bparees/session-app ce-registry.usersys.redhat.com/jboss-webserver3/tomcat8-openshift:badtag badout --loglevel=5
I0506 18:55:08.986159 08143 docker.go:173] Pulling image ce-registry.usersys.redhat.com/jboss-webserver3/tomcat8-openshift:badtag
I0506 18:55:12.403179 08143 docker.go:177] An error was received from the PullImage call: Tag badtag not found in repository ce-registry.usersys.redhat.com/jboss-webserver3/tomcat8-openshift

bparees:
@soltysh fyi.

soltysh:
@bparees addressed in #202, please close.

bparees:
@soltysh your changes don't address the issue I reported which was the case where :latest was specified on the ref, but there was no :latest tag in the registry.  your changes just add a :latest tag if no tag is specified.

that said, this seems to be working now (as in it immediately returns an error due to the invalid tag), so maybe docker fixed their registry:

$  sti build https://github.com/bparees/session-app ce-registry.usersys.redhat.com/jboss-webserver3/tomcat8-openshift badout --loglevel=5
I0527 20:24:16.810998 30901 util.go:34] Using bparees@redhat.com credentials for pulling ce-registry.usersys.redhat.com/jboss-webserver3/tomcat8-openshift

Builder Image:		ce-registry.usersys.redhat.com/jboss-webserver3/tomcat8-openshift
Source:			https://github.com/bparees/session-app
[...]

PR #202, opened 2015-05-26
Issue179 - fix sti hangs when using non-existent latest

soltysh:
This PR mainly addresses #179, but it also contains removals mentioned in #199.
@bparees @csrwng PTAL

### DIFF
--- a/pkg/docker/docker.go
+++ b/pkg/docker/docker.go
@@ -22,6 +22,7 @@ const (
 	LocationLabel   = "io.openshift.sti.location"
 
 	DefaultLocation = "/tmp"
+	DefaultTag      = "latest"
 )
 
 // Docker is the interface between STI and the Docker client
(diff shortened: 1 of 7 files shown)


pkg/docker/docker.go
```diff
@@ -22,6 +23,7 @@ const (
 	LocationLabel   = "io.openshift.sti.location"
 
 	DefaultLocation = "/tmp"
+	DefaultTag      = ":latest"
```
mfojtik: remove ':'

pkg/docker/docker.go
```diff
@@ -178,6 +180,10 @@ func (d *stiDocker) CheckAndPull(name string) (image *docker.Image, err error) {
 
 // PullImage pulls an image into the local registry
 func (d *stiDocker) PullImage(name string) (image *docker.Image, err error) {
+	_, tag := parsers.ParseRepositoryTag(name)
+	if len(tag) == 0 {
+		name += DefaultTag
```
mfojtik: use strings.Join()

csrwng:
Should we also not do the same logic for running a container?

soltysh:
@mfojtik addressed your comments. @csrwng you're right, I've added that to all our methods where I though it'd make sense. PTAL again.

csrwng:
LGTM

soltysh:
[merge]

Issue #203, opened 2015-05-27
Add ONBUILD integration tests


soltysh:
Currently we don't have **ANY** integration tests for onbuild strategy. We should have a couple at least!

ewolinetz:
It also looks like when we do a `make test` we do not run our integration tests along with the other tests.  

[centos@ip-10-155-13-188 source-to-image]$ make test
hack/test-go.sh  
. . .
?   	github.com/openshift/source-to-image/test/integration	[no test files]

mfojtik:
@ewolinetz @soltysh @bparees we can close this now?

soltysh:
I think so.

Issue #156, opened 2015-03-20
Send output to stdout, not stderr


goldmann:
Currently everything is sent to `stderr`. The console output should be sent to `stdout` instead. Only errors should be sent to `stderr`.

soltysh:
Yeah, I'd like that too, see [here](https://github.com/openshift/source-to-image/blob/master/cmd/sti/main.go#L181-L183). I guess I need to work on a PR to glog.

smarterclayton:
glog won't accept PRs.  We probably need to switch to a different library in our code that expects to run in these contexts (glog isn't really intended for the "display output to end users" use case).  A good starting point would be implementing the core static glog API against log (including V) as a drop in package replacement for glog.

soltysh:
@smarterclayton maybe before implementing that, we could spent some time and investigate other logging libraries like https://github.com/Sirupsen/logrus or https://github.com/op/go-loggin or others (?), I remember the first being mentioned upstream.

smarterclayton:
All of those bring more complexity which we don't have time for.  I also don't want to move away from glog on the server..

soltysh:
Also see [this comment from origin](https://github.com/openshift/origin/issues/1647#issuecomment-100244784).

bparees:
@smarterclayton we any closer to moving away from glog or otherwise gaining this capability in a logging tool?

bparees:
@jim-minter we've moved to using a glog wrapper in s2i which should make this possible now (if it hasn't already fixed this problem)

jim-minter:
@bparees 

Is the general intention here:
relevant container output (stdout, stderr) typically goes to s2i stdout
arbitrary s2i logging goes to s2i stderr
?

Are logging prefixes wanted or not wanted on the container output bit?

bparees:
@jim-minter 
logging prefix on container output behavior should stay the same.  I believe the current behavior is you don't get logging prefixes on it unless you specify over a certain loglevel (because if you specify a high enough loglevel, we punt all logging to glog).

as for the general intention, in an ideal world:
s2i info/debug goes to stdout
s2i warn/error goes to stderr
container stdout goes to stdout
container stderr goes to stderr

geoberle:
what is the status on this?
i can see that build output in openshift 3.10 is showing up in aggregated logging as loglevel "err".

acrozes:
On OpenShift 3.11, I find some errors like this on: 
```
    "message": "Push successful\n",
    "level": "err",
[...]

PR #204, opened 2015-05-27
Request -> Config refactor + additional cleanups.

### LINKED ISSUE #100: Hide private `Request` fields
Since the extraction of `types.go` into their own package `api` there a couple of fields unnecessarily public, this includes:
* `WorkingDir`
* `Incremental`
* `ExternalRequiredScripts`
* `ExternalOptionalScripts`

### LINKED ISSUE #144: Add better mechanism for `api.Request` validation
Currently `api.Request` object is validated only [here](https://github.com/openshift/source-to-image/blob/master/cmd/sti/main.go#L70) and in all functions creating those objects. As long as STI is used as standalone this is OK, but when you incorporate STI as part of [origin](https://github.com/openshift/origin/blob/master/pkg/build/builder/sti.go) then the validation does not happen, but it should.

soltysh:
I've added method for validating Config object - should fix #144. I've also reviewed and removed unnecessary fields from Config, which was also renamed from Request (fixed #100).

@bparees @mfojtik @csrwng PTAL, that's the last refactor I'd like to make before updating the origin's version. Maybe we could name it 1.0 release? Any thoughts one that are welcome as well.


pkg/api/types.go
```diff
@@ -59,14 +59,9 @@ type Request struct {
...
 	WorkingDir string
 
-	// LayeredBuild describes if this is build which layered scripts and sources on top of BaseImage.
+	// LayeredBuild describes if this is build which layered scripts and sources on top of BuilderImage.
 	LayeredBuild bool
 
-	// InstallDestination allows to override the default destination of the STI
-	// scripts. It allows to place the scripts into application root directory
-	// (see ONBUILD strategy). The default value is "upload/scripts".
-	InstallDestination string
```
mfojtik: pkg/build/strategies/sti/sti.go
71:	inst := script.NewInstaller(req.BaseImage, req.ScriptsURL, req.InstallDestination, docker)

This was originally used here, for some reason it is not anymore. Is there a reason why?

pkg/api/validation/validation.go
```diff
@@ -0,0 +1,44 @@
...
+)
+
+// ValidationError is an implementation of the 'error' interface, which represents an error of validation.
+type ValidationError struct {
+	Type  ValidationErrorType
+	Field string
+}
+
+func (v ValidationError) Error() string {
+	return fmt.Sprintf("%s: %s", v.Field, v.Type)
```
mfojtik: I prefer `return strings.Join([]string{v.Field, v.Type}, ":")` but I think both ways should have equally similar memory footprint ;-)

pkg/build/strategies/onbuild/onbuild.go
```diff
@@ -47,8 +47,7 @@ func New(request *api.Request) (*OnBuild, error) {
 		tar:    tar.New(),
 	}
 	// Use STI Prepare() and download the 'run' script optionally.
-	request.InstallDestination = "upload/src"
```
mfojtik: how this works now?

pkg/build/strategies/sti/sti.go
```diff
@@ -108,37 +108,37 @@ func New(req *api.Request) (*STI, error) {
...
 	} else {
 		glog.V(1).Infof("Clean build will be performed")
 	}
 
-	glog.V(2).Infof("Performing source build from %s", request.Source)
+	glog.V(2).Infof("Performing source build from %s", config.Source)
 	if b.incremental {
-		if err := b.artifacts.Save(request); err != nil {
+		if err := b.artifacts.Save(config); err != nil {
 			glog.Warningf("Error saving previous build artifacts: %v", err)
```
mfojtik: Can you join these two errors together?

pkg/build/strategies/sti/sti.go
```diff
@@ -108,37 +108,37 @@ func New(req *api.Request) (*STI, error) {
...
 			glog.Warning("Clean build will be performed!")
 		}
 	}
 
-	glog.V(1).Infof("Building %s", request.Tag)
-	if err := b.scripts.Execute(api.Assemble, request); err != nil {
+	glog.V(1).Infof("Building %s", config.Tag)
```
mfojtik: Since we inherit the log level from OpenShift, we should be consistent on what we log and under what log level. I think this belongs to V(2) as many other things. You don't have to solve this here, but we should have issue to track this.

pkg/build/strategies/sti/sti.go
```diff
@@ -234,33 +234,33 @@ func (b *STI) PostExecute(containerID, location string) error {
...
 
 	b.result.Success = true
 	b.result.ImageID = imageID
 
-	if len(b.request.Tag) > 0 {
-		glog.V(1).Infof("Successfully built %s", b.request.Tag)
-		glog.V(1).Infof("Tagged %s as %s", imageID, b.request.Tag)
+	if len(b.config.Tag) > 0 {
+		glog.V(1).Infof("Successfully built %s", b.config.Tag)
+		glog.V(1).Infof("Tagged %s as %s", imageID, b.config.Tag)
```
mfojtik: Same here, collapse these two. You can use "\n" if you really need two lines but I think we can just say:

"Successfully build %s and tagged as %s"

mfojtik:
@soltysh few nits but I really like this refactoring. Good job!

pkg/api/types.go
soltysh: It's not being used anymore from what I looked into. And the problem this variable was trying to address is solved differently from what I tested. Currently [this function](https://github.com/openshift/source-to-image/blob/master/pkg/build/strategies/onbuild/onbuild.go#L148) copies STI scripts to main app dir which works nicely. If there are other usecases that this was addressing we don't have them written down in any reproducable way (eg. integration tests). That's why I've also opened #203 to track these.

pkg/api/validation/validation.go
soltysh: Sprintf is somehow more readable in IMHO.

pkg/build/strategies/onbuild/onbuild.go
soltysh: See [this comment](https://github.com/openshift/source-to-image/pull/204#discussion_r31238157)

pkg/build/strategies/sti/sti.go
soltysh: I remember you ask about it every time I modify code around here :wink:. The idea is to have first information about what went wrong and then specify readably that we'll be doing a clean build.

pkg/build/strategies/sti/sti.go
soltysh: The logging subsystem needs rework, there's already open issue #156.

pkg/build/strategies/sti/sti.go
soltysh: Done.

soltysh:
@mfojtik answered all of your questions/doubts, fixed one. @csrwng can I have your LGTM on it as well. I'd like to have it merged today to update it in origin later.

cmd/sti/main.go
```diff
@@ -100,57 +98,57 @@ func newCmdBuild(req *api.Request) *cobra.Command {
...
+				cfg.Source = args[0]
+				cfg.BuilderImage = args[1]
 				if len(args) >= 3 {
-					req.Tag = args[2]
+					cfg.Tag = args[2]
 				}
 			}
 
-			if !validRequest(req) {
+			if len(validation.ValidateConfig(cfg)) != 0 {
```
csrwng: I think it'd be ideal if we moved some of this logic to a function in a different package (anything that doesn't have to do with the environment or command line args) so we can reuse it from origin. Otherwise we'd have to repeat the same code in origin (validating the config, creating the build strategy, etc)

pkg/api/describe/describer.go
```diff
@@ -0,0 +1,92 @@
...
+	result := []string{}
+	for k, v := range env {
+		result = append(result, fmt.Sprintf("%s=%s", k, v))
+	}
+	fmt.Fprintf(out, "Environment:\t%s\n", strings.Join(result, ","))
+}
+
+func printBool(b bool) string {
+	if b {
+		return "\033[1menabled\033[0m"
```
csrwng: Not sure if it's overkill, but maybe we could use something like: https://github.com/fatih/color to display stuff in bold. Then it can be turned off with a command-line switch if desired.

csrwng:
Apart from comments above, LGTM

cmd/sti/main.go
soltysh: That was the initial goal, to have validation method available to be run in origin (see #144). But I agree we can revisit this method once again to do that. I've created an issue to track that #206. If you don't mind, I'll do that as a follow-up. I don't want to get this PR more hairy than it's already is :wink:

pkg/build/strategies/sti/sti.go
mfojtik: @soltysh I don't think that is right, why we can't just say that in one sentence? more lines, more work to parse.

soltysh: If I include the information about clean build in the first sentence it might get lost along the error. It's intentionally separate line to *mark* we will do a clean build for the user.

pkg/api/describe/describer.go
soltysh: We can leave that for after :wink:

soltysh:
@csrwng since that last comment was on commit and it's lost, I've added additional validation for DockerConfig.Endpoint, as this what we use in origin, at least now and it makes sense to have required.

pkg/build/strategies/sti/sti.go
mfojtik: "Clean build will be performed as there as an error saving previous build artifacts: %v", err

mfojtik:
LGTM

soltysh:
[merge]

PR #207, opened 2015-05-29
Issue 205 - renamed io.openshift.sti.* label to shorter io.s2i.*

soltysh:
Address #205.

@bparees @mfojtik PTAL

PR #208, opened 2015-05-29
Allow passing DNS servers and search prefixes to build container

csrwng:
Allows origin code to pass in DNS lookup info to container launched by STI

### DIFF
--- a/pkg/docker/docker.go
+++ b/pkg/docker/docker.go
@@ -88,6 +88,8 @@ type RunContainerOptions struct {
 	Stderr          io.Writer
 	OnStart         func() error
 	PostExec        PostExecutor
+	DNS             []string
+	DNSSearch       []string
 }
 
 // CommitContainerOptions are options passed in to the CommitContainer method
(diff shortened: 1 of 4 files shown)


csrwng:
@mfojtik @soltysh ptal
This is the first part of the issue here: https://bugzilla.redhat.com/show_bug.cgi?id=1225410
@danmcp fyi

mfojtik:
LGTM

csrwng:
[test]

csrwng:
[merge]

pkg/docker/docker.go
```diff
@@ -346,9 +348,16 @@ func (d *stiDocker) RunContainer(opts RunContainerOptions) (err error) {
 	if opts.Stdout != nil {
 		config.AttachStdout = true
 	}
+	var hostConfig *docker.HostConfig
+	if opts.DNS != nil || opts.DNSSearch != nil {
```
bparees: what's the docker behavior if only DNSSearch but not DNS is set here?  (ie DNS ends up getting set to nil)

smarterclayton:
Receiving reports container DNS is broken after this change

bparees:
has this change even been applied to origin?  i'm not seeing a bump in origin for STI since may 15th.

soltysh:
Nope, I'm waiting for #207 and updating it.

csrwng:
We don't need this change anymore since we're going the documentation route. Should I submit a PR to reverse it?

soltysh:
Yeah, if it's not needed anymore please remove it. I'll merge it and I'll
update origin with latest version.

rassie:
@csrwng: which change to the documentation do you refer to? I'm encountering https://github.com/openshift/origin/issues/2598 and this PR is supposed to be the solution.

csrwng:
@rassie instead of having the s2i builder pass the nameserver and search entries to the container it launches, simply add the openshift master as a nameserver to the node's resolv.conf

smarterclayton:
Unfortunately the search entries aren't passed along.  This means that namespace local searches don't work.  We probably need to "pass on" the correct DNS config (which may be different than the builder) to s2i run

PR #210, opened 2015-06-02
Moved sti scripts docs under builder section where it belongs more.

soltysh:
This update is to match upstream changes https://github.com/openshift/openshift-docs/pull/433

@bparees PTAL


bparees:
lgtm. [merge]

PR #215, opened 2015-06-19
Add more metadata into output image

mfojtik:
This PR will make s2i store more metadata about the source code the image was built from.  This will help debugging problems as you can later, in runtime identify what revision of the app is the container running.

### DIFF
--- a/pkg/api/types.go
+++ b/pkg/api/types.go
@@ -2,8 +2,20 @@ package api
 
 import docker "github.com/fsouza/go-dockerclient"
 
+const (
+	DefaultNamespace    = "io.openshift.s2i."
+	KubernetesNamespace = "io.k8s."
+)
+
 // Config contains essential fields for performing build.
 type Config struct {
+	// DisplayName is a result image display-name label. This defaults to the
+	// output image name.
+	DisplayName string
+
+	// Description is a result image description label. The default is no
+	// description.
+	Description string
 
 	// BuilderImage describes which image is used for building the result images.
 	BuilderImage string
@@ -120,4 +132,37 @@ type InstallResult struct {
 
 	// Error describes last error encountered during install operation
 	Error error
+}
+
+// SourceInfo stores information about the source code
+type SourceInfo struct {
+	// Ref represents a commit SHA-1, valid GIT branch name or a GIT tag
+	// The output image will contain this information as 'io.openshift.build.commit.ref' label.
+	Ref string
+
+	// CommitID represents an arbitrary extended object reference in GIT as SHA-1
+	// The output image will contain this information as 'io.openshift.build.commit.id' label.
+	CommitID string
+
+	// Date contains a date when the committer created the commit.
+	// The output image will contain this information as 'io.openshift.build.commit.date' label.
+	Date string
+
+	// Author contains information about the committer name and email address.
+	// The output image will contain this information as 'io.openshift.build.commit.author' label.
+	Author string
+
+	// Message represents the first 80 characters from the commit message.
+	// The output image will contain this information as 'io.openshift.build.commit.message' label.
+	Message string
+
+	// Location contains a valid URL to the original repository.
+	// The output image will contain this information as 'io.openshift.build.source-location' label.
+	Location string
+
+	// ContextDir contains path inside the Location directory that
+	// contains the application source code.
+	// The output image will contain this information as 'io.openshift.build.source-context-dir'
+	// label.
+	ContextDir string
 }
--- a/pkg/build/strategies/sti/sti.go
+++ b/pkg/build/strategies/sti/sti.go
@@ -52,6 +52,7 @@ type STI struct {
 	installedScripts map[string]bool
 	scriptsURL       map[string]string
 	incremental      bool
+	sourceInfo       *api.SourceInfo
 
 	// Interfaces
 	preparer  build.Preparer
@@ -165,7 +166,7 @@ func (b *STI) Prepare(config *api.Config) error {
 
 	// fetch sources, for theirs .sti/bin might contain sti scripts
 	if len(config.Source) > 0 {
-		if err = b.source.Download(config); err != nil {
+		if b.sourceInfo, err = b.source.Download(config); err != nil {
 			return err
 		}
 	}
@@ -232,6 +233,7 @@ func (b *STI) PostExecute(containerID, location string) error {
 		Env:         buildEnv,
 		ContainerID: containerID,
 		Repository:  b.config.Tag,
+		Labels:      util.GenerateOutputImageLabels(b.sourceInfo, b.config),
 	}
 
 	imageID, err := b.docker.CommitContainer(opts)
--- a/pkg/git/clone.go
+++ b/pkg/git/clone.go
@@ -16,8 +16,9 @@ type Clone struct {
 
 // Download downloads the application source code from the GIT repository
 // and checkout the Ref specified in the config.
-func (c *Clone) Download(config *api.Config) error {
+func (c *Clone) Download(config *api.Config) (*api.SourceInfo, error) {
 	targetSourceDir := filepath.Join(config.WorkingDir, "upload", "src")
+	var info *api.SourceInfo
 
 	if c.ValidCloneSpec(config.Source) {
 		if len(config.ContextDir) > 0 {
@@ -26,15 +27,14 @@ func (c *Clone) Download(config *api.Config) error {
 		glog.V(2).Infof("Cloning into %s", targetSourceDir)
 		if err := c.Clone(config.Source, targetSourceDir); err != nil {
 			glog.V(1).Infof("Git clone failed: %+v", err)
-			return err
+			return nil, err
 		}
 
-		if config.Ref != "" {
-			glog.V(1).Infof("Checking out ref %s", config.Ref)
-
+		if len(config.Ref) > 0 {
 			if err := c.Checkout(targetSourceDir, config.Ref); err != nil {
-				return err
+				return nil, err
 			}
+			glog.V(1).Infof("Checked out %q", config.Ref)
 		}
 
 		if len(config.ContextDir) > 0 {
@@ -44,17 +44,28 @@ func (c *Clone) Download(config *api.Config) error {
 			path := filepath.Join(targetSourceDir, config.ContextDir) + string(filepath.Separator) + "."
 			err := c.Copy(path, originalTargetDir)
 			if err != nil {
-				return err
+				return nil, err
 			}
+			info = c.GetInfo(targetSourceDir)
 			c.RemoveDirectory(targetSourceDir)
+		} else {
+			info = c.GetInfo(targetSourceDir)
 		}
 
-		return nil
+		return info, nil
 	}
 	// we want to copy entire dir contents, thus we need to use dir/. construct
 	path := filepath.Join(config.Source, config.ContextDir) + string(filepath.Separator) + "."
 	if !c.Exists(path) {
-		return errors.NewSourcePathError(path)
+		return nil, errors.NewSourcePathError(path)
 	}
-	return c.Copy(path, targetSourceDir)
+	if err := c.Copy(path, targetSourceDir); err != nil {
+		return nil, err
+	}
+
+	// When building from a local directory (not using GIT clone spec scheme) we
+	// skip gathering informations about the source as there is no guarantee that
+	// the folder is a GIT repository or it requires context-dir to be set.
+	glog.Warning("You are using <source> location that is not valid GIT repository. The source code information will not be stored into the output image. Use this image only for local testing and development.")
+	return nil, nil
 }
--- a/pkg/git/git.go
+++ b/pkg/git/git.go
@@ -5,10 +5,12 @@ import (
 	"io"
 	"net/url"
 	"os"
+	"os/exec"
 	"regexp"
 	"strings"
 
 	"github.com/golang/glog"
+	"github.com/openshift/source-to-image/pkg/api"
 	"github.com/openshift/source-to-image/pkg/util"
 )
 
@@ -17,6 +19,7 @@ type Git interface {
 	ValidCloneSpec(source string) bool
 	Clone(source, target string) error
 	Checkout(repo, ref string) error
+	GetInfo(string) *api.SourceInfo
 }
 
 // New returns a new instance of the default implementation of the Git interface
@@ -88,6 +91,28 @@ func (h *stiGit) Checkout(repo, ref string) error {
 		Dir:    repo,
 	}
 	return h.runner.RunWithOptions(opts, "git", "checkout", ref)
+}
+
+// GetInfo retrieves the informations about the source code and commit
+func (h *stiGit) GetInfo(repo string) *api.SourceInfo {
+	git := func(arg ...string) string {
+		command := exec.Command("git", arg...)
+		command.Dir = repo
+		out, err := command.CombinedOutput()
+		if err != nil {
+			glog.V(1).Infof("Error executing 'git %#v': %s (%v)", arg, out, err)
+			return ""
+		}
+		return strings.TrimSpace(string(out))
+	}
+	return &api.SourceInfo{
+		Location: git("config", "--get", "remote.origin.url"),
+		Ref:      git("rev-parse", "--abbrev-ref", "HEAD"),
+		CommitID: git("rev-parse", "--short", "--verify", "HEAD"),
+		Author:   git("--no-pager", "show", "-s", "--format=%an <%ae>", "HEAD"),
+		Date:     git("--no-pager", "show", "-s", "--format=%ad", "HEAD"),
+		Message:  git("--no-pager", "show", "-s", "--format=%<(80,trunc)%s", "HEAD"),
+	}
 }
 
 func pipeToLog(reader io.Reader, log func(...interface{})) {
--- a/pkg/test/git.go
+++ b/pkg/test/git.go
@@ -1,5 +1,7 @@
 package test
 
+import "github.com/openshift/source-to-image/pkg/api"
+
 // FakeGit provides a fake GIT
 type FakeGit struct {
 	ValidCloneSpecSource string
@@ -32,4 +34,12 @@ func (f *FakeGit) Checkout(repo, ref string) error {
 	f.CheckoutRepo = repo
 	f.CheckoutRef = ref
 	return f.CheckoutError
+}
+
+func (f *FakeGit) GetInfo(repo string) *api.SourceInfo {
+	return &api.SourceInfo{
+		Ref:      "master",
+		CommitID: "1bf4f04",
+		Location: "file:///foo",
+	}
 }
--- a/pkg/util/labels.go
+++ b/pkg/util/labels.go
@@ -1,0 +1,49 @@
+package util
+
+import (
+	"strings"
+
+	"github.com/golang/glog"
+	"github.com/openshift/source-to-image/pkg/api"
+)
+
+// GenerateOutputImageLabels generate the labels based on the source repository
+// informations.
+func GenerateOutputImageLabels(info *api.SourceInfo, config *api.Config) map[string]string {
+	result := map[string]string{}
+
+	if len(config.Description) > 0 {
+		result[api.KubernetesNamespace+"description"] = config.Description
+	}
+
+	if len(config.DisplayName) > 0 {
+		result[api.KubernetesNamespace+"display-name"] = config.DisplayName
+	} else {
+		result[api.KubernetesNamespace+"display-name"] = config.Tag
+	}
+
+	if info == nil {
+		glog.V(3).Infof("Unable to fetch source informations, the output image labels will not be set")
+		return result
+	}
+
+	addBuildLabel(result, "image", config.BuilderImage)
+	addBuildLabel(result, "commit.author", info.Author)
+	addBuildLabel(result, "commit.date", info.Date)
+	addBuildLabel(result, "commit.id", info.CommitID)
+	addBuildLabel(result, "commit.ref", info.Ref)
+	addBuildLabel(result, "commit.message", info.Message)
+	addBuildLabel(result, "source-location", info.Location)
+	addBuildLabel(result, "source-context-dir", config.ContextDir)
+
+	return result
+}
+
+// addBuildLabel adds a new "io.openshift.s2i.build.*" label into map when the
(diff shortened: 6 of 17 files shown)


mfojtik:
@bparees @soltysh PTAL (this is not urgent, just good to have for investigating problems)

bparees:
@mfojtik https://trello.com/c/cd1S05XE/596-add-commit-author-info-labels-to-built-images

please make sure whatever you're doing here aligns with the needs of that card.

mfojtik:
yes i can add all of them

pkg/api/types.go
```diff
@@ -121,3 +121,26 @@ type InstallResult struct {
...
+
+// SourceInfo stores information about the source code
+type SourceInfo struct {
+	// Store the current GIT ref which might be a branch or commit id
+	// This information is available in the image as 'io.openshift.build.commit.ref' LABEL
+	Ref string
+
+	// The GIT commit id
+	// This information is available in the image as 'io.openshift.build.commit.id' LABEL
+	ID string
```
soltysh: Pls rename this to to `CommitID`, since `ID` is too vague IMHO. I assume the actual output data will match the name of this field, right?

pkg/build/strategies/sti/sti.go
```diff
@@ -218,6 +219,17 @@ func (b *STI) PostExecute(containerID, location string) error {
 
 	buildEnv := append(scripts.ConvertEnvironment(env), b.generateConfigEnv()...)
 
+	labels := map[string]string{}
+	if b.sourceInfo != nil {
+		labels = map[string]string{
+			"io.openshift.build.commit.author":     b.sourceInfo.Author,
```
soltysh: I've talked with @bparees about that, we've agreed to have the prefix configurable. By default it'll be `io.s2i...` and in OpenShift it'll be `io.openshift...`

pkg/build/strategies/sti/sti.go
```diff
@@ -218,6 +219,17 @@ func (b *STI) PostExecute(containerID, location string) error {
...
+			"io.openshift.build.commit.date":       b.sourceInfo.Date,
+			"io.openshift.build.commit.id":         b.sourceInfo.ID,
+			"io.openshift.build.commit.ref":        b.sourceInfo.Ref,
+			"io.openshift.build.source-repository": b.sourceInfo.Location,
```
soltysh: I'm a little bit concerned that source-repository here isn't the right naming, since you can build from local directory, which is not a source-repository (btw. how should we handle relative vs absolute paths then? always expand to absolute path?) One other issue is context directory, that is missing in here and it **has** to be here imho.

pkg/build/strategies/sti/sti.go
mfojtik: i don't want to have this configurable because it makes things much harder for no good reasons.

mfojtik:
Updated this PR to include the LABELs:

```console
I0622 16:00:28.329682 20941 docker.go:471] Committing container with config: {Hostname: Domainname: User: Memory:0 MemorySwap:0 CPUShares:0 CPUSet: AttachStdin:false AttachStdout:false AttachStderr:false PortSpecs:[] ExposedPorts:map[] Tty:false OpenStdin:false StdinOnce:false Env:[] Cmd:[/usr/local/sti/run] DNS:[] Image: Volumes:map[] VolumesFrom: WorkingDir: Entrypoint:[] NetworkDisabled:false SecurityOpts:[] OnBuild:[] Labels:map[io.openshift.build.commit.author:Michal Fojtik <mfojtik@redhat.com> io.openshift.build.commit.date:Thu May 7 17:14:28 2015 +0200 io.openshift.build.commit.id:master io.openshift.build.commit.ref:7ee9858 io.openshift.build.source-repository:file:///home/mfojtik/sandbox/ruby]}
```

pkg/build/strategies/sti/sti.go
mfojtik: @soltysh local directory **must be** source repository

mfojtik:
So the current set of labels is this:
```json
        "Labels": {
            "io.openshift.build.commit.author": "Michal Fojtik \u003cmfojtik@redhat.com\u003e",
            "io.openshift.build.commit.date": "Thu May 7 17:14:28 2015 +0200",
            "io.openshift.build.commit.id": "master",
            "io.openshift.build.commit.ref": "7ee9858",
            "io.openshift.build.source-repository": "file:///home/mfojtik/sandbox/ruby",
            "io.s2i.scripts-url": "image:///usr/local/sti",
            "k8s.io/description": "Platform for building and running Ruby 2.0 applications",
            "k8s.io/display-name": "Ruby 2.0",
            "openshift.io/expose-services": "8080:http",
            "openshift.io/tags": "builder,ruby,ruby20"
        },

```

pkg/build/strategies/sti/sti.go
soltysh: k

mfojtik:
[test]

No idea why the Travis failed, I can build this just fine locally.

pkg/api/types.go
bparees: +1 commitid.

bparees: behavior for these fieids:

ref is a commit.  commitid will match ref
ref is a branch.  commitid will be a commit in that branch
ref is a tag.  commitid will be the commit pointed to by the tag.

pkg/api/types.go
```diff
@@ -121,3 +121,26 @@ type InstallResult struct {
...
+	// This information is available in the image as 'io.openshift.build.source' LABEL
+	Location string
+
+	// The date of the commit
+	// This information is available in the image as 'io.openshift.build.commit.date' LABEL
+	Date string
+
+	// Author of the commit
+	// This information is available in the image as 'io.openshift.build.commit.author' LABEL
+	Author string
```
bparees: add the commit comment as a field too.

pkg/build/strategies/sti/sti.go
bparees: still, calling it source-path or source-location would future proof us.

soltysh: Ok, still context directory is something we need here.

bparees: true.

pkg/api/types.go
mfojtik: @bparees the commit message can be pretty long... are you sure you want to have it as label? I mean if you have source location and GIT ref, then you can just fetch the commit and get the message... Alternatively we can truncate the message to some reasonable length, or just take the first line.

mfojtik:
@bparees @soltysh all comments addressed

mfojtik:
@bparees @soltysh the only thing that bothers me here is the label format inconsistency... The final image labels looks like this now:

```json
            "io.openshift.build.commit.author": "Michal Fojtik \u003cmfojtik@redhat.com\u003e",
            "io.openshift.build.commit.date": "Thu May 7 17:14:28 2015 +0200",
            "io.openshift.build.commit.id": "master",
            "io.openshift.build.commit.ref": "7ee9858",
            "io.openshift.build.source-location": "file:///home/mfojtik/sandbox/ruby",
            "io.s2i.scripts-url": "image:///usr/local/sti",
            "k8s.io/description": "Platform for building and running Ruby 2.0 applications",
            "k8s.io/display-name": "Ruby 2.0",
            "openshift.io/expose-services": "8080:http",
            "openshift.io/tags": "builder,ruby,ruby20"
```

I think we have to make this consistent, either in OpenShift STI images (if it is not too late for that) or in STI.

mfojtik:
@bparees @soltysh also note that the 'display-name' is not quite correct here... the displayName should be the application name and not the builder image name.

soltysh:
Isn't display-name inherited from the builder itself?

pkg/api/types.go
```diff
@@ -121,3 +121,32 @@ type InstallResult struct {
...
+	// The output image will contain this information as 'io.openshift.build.commit.date' label.
+	Date string
+
+	// Author contains information about the committer name and email address.
+	// The output image will contain this information as 'io.openshift.build.commit.author' label.
+	Author string
+
+	// SourceLocation contains a valid URL to the original repository.
+	// The output image will contain this information as 'io.openshift.build.source-location' label.
+	SourceLocation string
```
soltysh: If the struct is called `SourceInfo`, having this field named just `Location` would be enough, the same applies to `SourceContextDir` -> `ContextDir`.

soltysh:
A few problems:
1) I did:
```
sti build https://github.com/openshift/sti-python.git --context-dir=3.3/test/setup-test-app/ openshift/python-33-centos7 python-sample-app
```
and got in result:
```
            "io.openshift.build.commit.author": "",
            "io.openshift.build.commit.date": "",
            "io.openshift.build.commit.id": "",
            "io.openshift.build.commit.ref": "",
            "io.openshift.build.source-context-dir": "3.3/test/setup-test-app/",
            "io.openshift.build.source-location": "",
            "io.s2i.scripts-url": "image:///usr/local/sti",
            "k8s.io/description": "Platform for building and running Python 3.3 applications",
            "k8s.io/display-name": "Python 3.3",
            "openshift.io/expose-services": "8080:http",
            "openshift.io/tags": "builder,python,python33"
```

I didn't specify any ref, nor commit id, still I'd expect both to contain master and latest commit id respectively, along with author and date. Why the source-location is also empty?


2) I also did this:
```
sti build /tmp/xxx openshift/python-33-centos7 python-sample-app
```
where `/tmp/xxx` contains the files from above context dir and got no info about build. Which is OK, but also means we don't require source to be in any repository, which you told me we do. See [this line](https://github.com/openshift/source-to-image/blob/master/pkg/git/clone.go#L59) which handles copying sources if no valid git clone is specified [this check](https://github.com/openshift/source-to-image/blob/master/pkg/git/clone.go#L22).


3) Finally I did this:
```
sti build 3.3/test/setup-test-app/ openshift/python-33-centos7 python-sample-app
```
from directory where [sti-python](https://github.com/openshift/sti-python) is cloned, and again got no info about build, but I'd expect it should be smart enough to get them or at least try to get them, IMHO.


Points 1 and 3 are the ones that worries me.

mfojtik:
@soltysh good catch... the 1) and 3) cases we remove the source directory so we can't execute git commands there... I fixed this to gather the info before we remove stuff.

bparees:
@mfojtik @soltysh 
yeah short of introducing a new "--name" flag to S2I, the display-name is going to be inherited...  maybe it's worth tagging "app" onto the end of the existing display name, to at least identify it a little, pending the decision to introduce such a flag?  (and to have a default behavior if the user doesn't specify the flag?)

mfojtik:
@bparees @soltysh I will do what you are proposing but also would be nice to offer '--name' and '--description'. If they are set then they override the builder image tags.

soltysh:
How about taking for the app name the image name (the last part after / sign, without the tag part of course) if none is specified?

mfojtik:
@soltysh we are talking about the 'display name' here. Your proposal is more a 'name' of the build config :-)

pkg/api/types.go
mfojtik: @soltysh my intent was to match them with the label names

bparees:
@mfojtik no i think @soltysh is saying:

    $ sti build https://myrepo.github.com mybuilderimage myorg/myappimage

should result in the display name being "myappimage"

but i have to say that seems a little pointless... you already know the image name so if you're going to make the display name be the image name, you don't need the display name field to be set at all.

soltysh:
@bparees +1, was about to write that down :wink:

soltysh:
@bparees well, either we'll clear the display-name when nothing is passed, which is necessary because we don't want to have Ruby 2.0 there, or give there some meaningful name, in which case falling back to output image was quite natural imho. Besides the image name is one thing and diplay-name something different, unless the logic will take the image name if no display-name is provided in that case, that makes sense.

pkg/api/types.go
mfojtik: fixed.

bparees:
@mfojtik I can't find where my comment went but for commit message first sentence is fine with me.

mfojtik:
@bparees i'm saying that 'myappimage' is not a good 'display name' but a good 'name' (eg. for buildConfig)

bparees:
@mfojtik for the commit message, search for \n or 80 chars, whichever comes first.
if you hit 80 chars, take the first 77 and append "..."

bparees:
i'm fine with either defaulting to the imagename for display-name, or taking the existing display-name and appending "app"  (keep in mind not all builder images will have a display-name label either)

mfojtik:
(the label discussion is moved here: https://github.com/openshift/origin/issues/3414)

mfojtik:
@bparees @soltysh ready for next round of reviews.... summary of changes:

* Added `--application-name` and `--description` options to `sti build`. They are both optional and the `--application-name` defaults to output image name
* Added `io.openshift.s2i.build.commit.message` (truncated to 80 chars)
* Added ` source-context-dir` label (when the contextDir is set)
* When building from random folder, we attempt to read the metadata anyway (what if the folder is GIT repo after all? ;-)... when we fail, we tell it
* More plumbing around namespaces

pkg/test/git.go
```diff
@@ -33,3 +35,11 @@ func (f *FakeGit) Checkout(repo, ref string) error {
 	f.CheckoutRef = ref
 	return f.CheckoutError
 }
+
+func (f *FakeGit) GetInfo(repo string) *api.SourceInfo {
+	return &api.SourceInfo{
+		Ref:      "master",
+		CommitID: "1bf4f04",
+		Location: "file;//foo",
```
rhcarvalho: `file:///foo`?

pkg/git/git.go
```diff
@@ -90,6 +93,28 @@ func (h *stiGit) Checkout(repo, ref string) error {
...
+		out, err := command.CombinedOutput()
+		if err != nil {
+			glog.V(1).Infof("Error executing 'git %#v': %s (%v)", arg, out, err)
+			return ""
+		}
+		return strings.TrimSpace(string(out))
+	}
+	return &api.SourceInfo{
+		Location: git("config", "--get", "remote.origin.url"),
+		CommitID: git("rev-parse", "--abbrev-ref", "HEAD"),
```
soltysh: This returns the ref actually, and the command in ref returns commitID, you need to switch them.

pkg/api/types.go
```diff
@@ -2,8 +2,20 @@ package api
...
 
+const (
+	DefaultNamespace    = "io.openshift.s2i."
+	KubernetesNamespace = "k8s.io/"
+)
+
 // Config contains essential fields for performing build.
 type Config struct {
+	// DisplayName is a result image display-name label. This defaults to the
+	// result image name.
```
soltysh: s/result/output

pkg/api/types.go
```diff
@@ -121,3 +133,36 @@ type InstallResult struct {
 	// Error describes last error encountered during install operation
 	Error error
 }
+
+// SourceInfo stores information about the source code
+type SourceInfo struct {
+	// Ref represents a commit SHA-1, valid GIT branch name or a GIT tag
+	// The output image will contain this information as 'io.openshift.build.commit.ref' label.
```
soltysh: Actually now it's `io.openshift.s2i.build.commit.ref`, rewrite it to say `DefaultNamespace + 'commit.ref'`.
This also applies to all other `SourceInfo` fields.

pkg/git/clone.go
```diff
@@ -44,17 +44,30 @@ func (c *Clone) Download(config *api.Config) error {
...
-		return nil
+		return info, nil
 	}
 	// we want to copy entire dir contents, thus we need to use dir/. construct
 	path := filepath.Join(config.Source, config.ContextDir) + string(filepath.Separator) + "."
 	if !c.Exists(path) {
-		return errors.NewSourcePathError(path)
+		return nil, errors.NewSourcePathError(path)
+	}
+	if err := c.Copy(path, targetSourceDir); err != nil {
```
soltysh: I'd try getting info before doing copy, because if I do (my 3rd test case):
```
sti build 3.3/test/setup-test-app/ openshift/python-33-centos7 python-sample-app
```
From inside sti-python this could return valid info, assuming you do that in the source (represented by `path` here), but after copying all the git info is lost.

soltysh: Otherwise, you will never be able to get decent info about that.

soltysh:
@mfojtik a couple of small nits :wink: and you're almost done.

pkg/api/types.go
mfojtik: @soltysh I hate referring to const in comments ;-)

pkg/git/clone.go
mfojtik: @soltysh when using non-git repo or just a subfolder without contextDir then we should not gather **any** info at all and don't set any labels.

mfojtik:
@soltysh thanks for reviewing this, I fixed all comments (except the two). I will fix the label comments as follow up when we decide what labels format we are going to use.

pkg/git/clone.go
soltysh: Spoke on IRC, I'm ok with current solution.

pkg/api/types.go
soltysh: Your choice :stuck_out_tongue:

soltysh:
LGTM, will wait with merge for @bparees approval

pkg/api/types.go
```diff
@@ -2,8 +2,20 @@ package api
 
 import docker "github.com/fsouza/go-dockerclient"
 
+const (
+	DefaultNamespace    = "io.openshift.s2i."
+	KubernetesNamespace = "k8s.io/"
```
bparees: thought @smarterclayton said k8s was using docker style labels for their images (not their annotations of course)?

pkg/git/clone.go
```diff
@@ -26,14 +27,13 @@ func (c *Clone) Download(config *api.Config) error {
...
-			return err
+			return nil, err
 		}
 
 		if config.Ref != "" {
 			glog.V(1).Infof("Checking out ref %s", config.Ref)
-
 			if err := c.Checkout(targetSourceDir, config.Ref); err != nil {
-				return err
+				return c.GetInfo(targetSourceDir), err
```
bparees: errr...can't i have a context dir and a ref?

pkg/git/git.go
```diff
@@ -90,6 +93,28 @@ func (h *stiGit) Checkout(repo, ref string) error {
...
+		}
+		return strings.TrimSpace(string(out))
+	}
+	return &api.SourceInfo{
+		Location: git("config", "--get", "remote.origin.url"),
+		Ref:      git("rev-parse", "--abbrev-ref", "HEAD"),
+		CommitID: git("rev-parse", "--short", "--verify", "HEAD"),
+		Author:   git("--no-pager", "show", "-s", "--format=%an <%ae>", "HEAD"),
+		Date:     git("--no-pager", "show", "-s", "--format=%ad", "HEAD"),
+		Message:  git("--no-pager", "show", "-s", "--format=%<(80,trunc)%s", "HEAD"),
```
bparees: no "..." if it exceeds 80 chars?  boo.

pkg/util/labels.go
```diff
@@ -0,0 +1,40 @@
...
+		glog.V(5).Infof("No source information!")
+		return result
+	}
+
+	result[api.DefaultNamespace+"build.commit.author"] = info.Author
+	result[api.DefaultNamespace+"build.commit.date"] = info.Date
+	result[api.DefaultNamespace+"build.commit.id"] = info.CommitID
+	result[api.DefaultNamespace+"build.commit.ref"] = info.Ref
+	result[api.DefaultNamespace+"build.commit.message"] = info.Message
+	result[api.DefaultNamespace+"build.source-location"] = info.Location
```
bparees: might be better to check that you have a value for each field.  in the future we might support an SCM that has a subset of these fields.

bparees:
@mfojtik couple comments, one major concern which i realize wasn't introduced by you but would be good to fix here.

pkg/api/types.go
mfojtik: cool, so it is `io.k8s.` ?

bparees: ". Kube upstream uses docker suggested labels in the Kubelet (io.foo.bar)"

so yes, I guess so.

pkg/git/git.go
mfojtik: the git truncate should add that suffix (at least I saw it on my machine ;-)

pkg/git/clone.go
mfojtik: yeah, seems like not... that is a bug in our code I guess.

mfojtik: no, we return only if the checkout ref fails ;-) ufffff (and I just removed the GetInfo() if that happens)

mfojtik:
@bparees updated, the current set of labels for the output image built by S2I now looks like this:

```json
"Labels": {
    "io.k8s.description": "Example description",
    "io.k8s.display-name": "My cool Ruby app",
    "io.openshift.s2i.build.commit.author": "Michal Fojtik \u003cmfojtik@redhat.com\u003e",
    "io.openshift.s2i.build.commit.date": "Wed Jun 24 13:36:57 2015 +0200",
    "io.openshift.s2i.build.commit.id": "bcbefa6",
    "io.openshift.s2i.build.commit.message": "initial",
    "io.openshift.s2i.build.commit.ref": "master-foo",
    "io.openshift.s2i.build.source-location": "file:///home/mfojtik/sandbox/ruby",
    "io.s2i.scripts-url": "image:///usr/local/sti",
    "k8s.io/description": "Platform for building and running Ruby 2.0 applications",
    "k8s.io/display-name": "Ruby 2.0",
    "openshift.io/expose-services": "8080:http",
    "openshift.io/tags": "builder,ruby,ruby20"
},
```

soltysh:
Still LGTM, nice work @mfojtik!

pkg/git/git.go
bparees: ok

bparees:
@mfojtik I guess we then need to go back and  add "io.openshift.tags" and "io.openshift.expose-services" to our images so we can start to deprecate the "openshift.io" ones?

we can still convert those to "openshift.io/tags" when/if we make them annotations on the imagestreamtag.

bparees:
@mfojtik think you missed this comment?

    might be better to check that you have a value for each field. in the future we might support an SCM that has a subset of these fields.

bparees:
comment was not missed, I didn't read the addLabel function.
[merge]

PR #220, opened 2015-06-26
Add support for new scripts-url label format

mfojtik:


mfojtik:
@soltysh @bparees PTAL

pkg/docker/docker.go
```diff
@@ -18,8 +18,8 @@ const (
 	ScriptsURLEnvironment = "STI_SCRIPTS_URL"
 	LocationEnvironment   = "STI_LOCATION"
 
-	ScriptsURLLabel  = "io.s2i.scripts-url"
```
soltysh: I'd leave this as a constant here, at least it's easier for reading, at most adding depecated comment, which should be also next to accompanying envs.

soltysh:
Small nit, other than that LGTM.

pkg/docker/docker.go
mfojtik: I don't like to leave a deprecated constant here with value that we claim as 'deprecated' here. I think it will be more confusing... I'm OK with leaving a comment on top of this new value saying that this is a new value and the old one is deprecated.

soltysh: k

mfojtik: fixed.

soltysh:
[merge]

mfojtik:
[merge] one more time

PR #222, opened 2015-06-26
Fix documentation to mention new label

mfojtik:


mfojtik:
(added to origin PR iinstead...)

PR #225, opened 2015-06-26
Rewording and updating README.md

ewolinetz:
@soltysh @bparees PTAL

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,9 +4,9 @@
 [![Travis](https://travis-ci.org/openshift/source-to-image.svg?branch=master)](https://travis-ci.org/openshift/source-to-image)
 
 
-Source-to-image (`sti`) is a tool for building reproducible docker images. `sti` produces
-ready-to-run images by injecting source code into a docker image and *assembling*
-a new docker image which incorporates the builder image and built source, and is ready to use
+Source-to-image (`sti`) is a tool for building reproducible Docker images. `sti` produces
+ready-to-run images by injecting source code into a Docker image and *assembling*
+a new Docker image which incorporates the builder image and built source.  The result is then ready to use
 with `docker run`. `sti` supports incremental builds which re-use previously downloaded
 dependencies, previously built artifacts, etc.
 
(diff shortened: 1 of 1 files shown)


bparees:
lgtm.
[merge]

README.md
```diff
@@ -19,17 +19,17 @@ Want to just get started now? Check out the [instructions](#getting-started).
 
 1. Simplify the process of application source + builder image -> usable image for most use cases (the
    80%)
-1. Define and implement a workflow for incremental build that eventually uses only docker
+1. Define and implement a workflow for incremental builds that eventually uses only Docker
```
soltysh: builds that eventually use~~s~~ only

README.md
```diff
@@ -71,15 +71,15 @@ For detailed description of the requirements and the scripts along with examples
...
+Docker image with ONBUILD instructions and choosing the OnBuild strategy. This
 strategy will trigger all ONBUILD instructions and execute the assemble script
 (if it exists) as the last instruction.
 
 Since the ONBUILD images usually don't provide any entrypoint, in order to use
-this build strategy you have to provide it. You can either include the 'run',
+this build strategy you will have to provide one. You can either include the 'run',
 'start' or 'execute' script in your application source root folder or you can
-specify a valid STI scripts URL and the 'run' script will be fetched and set as
+specify a valid STI script URL and the 'run' script will be fetched and set as
```
soltysh: Actually the scripts should be in plural here, we are always referring to multiple scripts (at least 2 are required: `run` and `assemble` but additional 2 are possible).

README.md
bparees: uses is correct... the noun "uses" applies to is "workflow" and not "builds", i believe.  or at least that argument can be made..it's a confusing sentence :)

README.md
bparees: eh, again debatable.  it's a single script url.  what is located at that url can be one or more scripts.

just like a "source url" contains sources but we don't call it a sources url.

soltysh: ok

README.md
soltysh: Right, after looking once again I see how that applies to workflow in which case it's correct. Sorry for the confusion.

PR #226, opened 2015-06-29
Applied comments from @csrwng to create

soltysh:
@mfojtik ptal

### DIFF
--- a/pkg/create/templates/docker.go
+++ b/pkg/create/templates/docker.go
@@ -11,8 +11,8 @@ FROM openshift/base-centos7
 # ENV BUILDER_VERSION 1.0
 
 # TODO: Set labels used in OpenShift to describe the builder image
-#LABEL k8s.io/description="Platform for building xyz" \
-#      k8s.io/display-name="builder x.y.z" \
+#LABEL io.k8s.description="Platform for building xyz" \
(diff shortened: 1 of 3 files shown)


pkg/create/templates/docker.go
```diff
@@ -11,8 +11,8 @@ FROM openshift/base-centos7
 # ENV BUILDER_VERSION 1.0
 
 # TODO: Set labels used in OpenShift to describe the builder image
-#LABEL k8s.io/description="Platform for building xyz" \
-#      k8s.io/display-name="builder x.y.z" \
+#LABEL io.k8s.description="Platform for building xyz" \
+#      io.k8s.display-name="builder x.y.z" \
```
bparees: hmm... I thought k8s was still using kis.io/display-name, was this confirmed somewhere?

soltysh: I'm using what @mfojtik is using in his latest changes, eg. https://github.com/openshift/sti-nodejs/pull/51

mfojtik: for annotations we use 'k8s.io/*' and for LABELS (which are not annotations) we follow Docker 'io.k8s.*'

bparees:
lgtm.  [merge]

PR #227, opened 2015-06-30
Additional changes to the create command, to produce valid builder image contents.

soltysh:
@mfojtik ptal

### DIFF
--- a/pkg/create/templates/test.go
+++ b/pkg/create/templates/test.go
@@ -47,6 +47,12 @@ prepare() {
     echo "ERROR: The image ${IMAGE_NAME} must exist before this script is executed."
     exit 1
   fi
+  # sti build requires the application is a valid 'GIT' repository
+  pushd ${test_dir}/test-app >/dev/null
+  git init
+  git config user.email "build@localhost" && git config user.name "builder"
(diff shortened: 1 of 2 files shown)


pkg/create/templates/test.go
```diff
@@ -39,7 +39,7 @@ container_ip() {
 }
 
 run_sti_build() {
-  sti build ${sti_args} file://${test_dir}/test-app ${IMAGE_NAME} ${IMAGE_NAME}-testapp
+  sti build ${sti_args} ${test_dir}/test-app ${IMAGE_NAME} ${IMAGE_NAME}-testapp
```
mfojtik: add file:/// back, it is the preferred way. The other one will give you fat warning.

rhcarvalho: Hey, I see there's no more `git init` step in `prepare()` ... that means that actually using `file://` produce a bad test runner, one that won't work

mfojtik: we should use --context-dir in that case...

soltysh:
@mfojtik reverted the git init step in `test/run`.

mfojtik:
LGTM

soltysh:
[merge]

PR #229, opened 2015-07-02
Adding test cases for ONBUILD strategy builds

ewolinetz:
@soltysh @bparees PTAL - *begins* to address issue #203

### DIFF
--- a/hack/build-test-images.sh
+++ b/hack/build-test-images.sh
@@ -27,3 +27,4 @@ buildimage sti_test/sti-fake-user test/integration/images/sti-fake-user
 buildimage sti_test/sti-fake-scripts test/integration/images/sti-fake-scripts
 buildimage sti_test/sti-fake-scripts-no-save-artifacts test/integration/images/sti-fake-scripts-no-save-artifacts
 buildimage sti_test/sti-fake-no-tar test/integration/images/sti-fake-no-tar
+buildimage sti_test/sti-fake-onbuild test/integration/images/sti-fake-onbuild
--- a/test/integration/images/sti-fake-onbuild/Dockerfile
+++ b/test/integration/images/sti-fake-onbuild/Dockerfile
@@ -1,0 +1,11 @@
+FROM sti_test/sti-fake
+
+ONBUILD RUN touch /sti-fake/src/onbuild
+
+# The ONBUILD strategy only works with the application source dir so we need
+# to manually specify to copy to another location.
+#
+# This is a little hack-ish given that we know our assemble script requires files to be in /tmp/src
+# we will copy there, and also set our WORKDIR to be the same location so it has access to the scripts
+ONBUILD COPY . /tmp/src/
+WORKDIR /tmp/src
--- a/test/integration/integration_test.go
+++ b/test/integration/integration_test.go
@@ -29,6 +29,7 @@ const (
 	FakeImageScripts                = "sti_test/sti-fake-scripts"
 	FakeImageScriptsNoSaveArtifacts = "sti_test/sti-fake-scripts-no-save-artifacts"
 	FakeImageNoTar                  = "sti_test/sti-fake-no-tar"
+	FakeImageOnBuild                = "sti_test/sti-fake-onbuild"
 
 	TagCleanBuild                             = "test/sti-fake-app"
 	TagCleanBuildUser                         = "test/sti-fake-app-user"
@@ -38,6 +39,11 @@ const (
 	TagIncrementalBuildScripts                = "test/sti-incremental-app-scripts"
 	TagIncrementalBuildScriptsNoSaveArtifacts = "test/sti-incremental-app-scripts-no-save-artifacts"
 	TagCleanLayeredBuildNoTar                 = "test/sti-fake-no-tar"
+	TagCleanBuildOnBuild                      = "test/sti-fake-app-onbuild"
+	TagIncrementalBuildOnBuild                = "test/sti-incremental-app-onbuild"
+	TagCleanBuildOnBuildNoName                = "test/sti-fake-app-onbuild-noname"
+	TagCleanBuildNoName                       = "test/sti-fake-app-noname"
+	TagCleanLayeredBuildNoTarNoName           = "test/sti-fake-no-tar-noname"
 
 	// Need to serve the scripts from local host so any potential changes to the
 	// scripts are made available for integration testing.
@@ -130,35 +136,51 @@ func waitForHTTPReady() error {
 
 // Test a clean build.  The simplest case.
 func TestCleanBuild(t *testing.T) {
-	integration(t).exerciseCleanBuild(TagCleanBuild, false, FakeBuilderImage, "")
+	integration(t).exerciseCleanBuild(TagCleanBuild, false, FakeBuilderImage, "", true, true)
 }
 
 func TestCleanBuildUser(t *testing.T) {
-	integration(t).exerciseCleanBuild(TagCleanBuildUser, false, FakeUserImage, "")
+	integration(t).exerciseCleanBuild(TagCleanBuildUser, false, FakeUserImage, "", true, true)
 }
 
 func TestCleanBuildFileScriptsURL(t *testing.T) {
-	integration(t).exerciseCleanBuild(TagCleanBuild, false, FakeBuilderImage, FakeScriptsFileURL)
+	integration(t).exerciseCleanBuild(TagCleanBuild, false, FakeBuilderImage, FakeScriptsFileURL, true, true)
 }
 
 func TestCleanBuildHttpScriptsURL(t *testing.T) {
-	integration(t).exerciseCleanBuild(TagCleanBuild, false, FakeBuilderImage, FakeScriptsHttpURL)
+	integration(t).exerciseCleanBuild(TagCleanBuild, false, FakeBuilderImage, FakeScriptsHttpURL, true, true)
 }
 
 func TestCleanBuildScripts(t *testing.T) {
-	integration(t).exerciseCleanBuild(TagCleanBuildScripts, false, FakeImageScripts, "")
+	integration(t).exerciseCleanBuild(TagCleanBuildScripts, false, FakeImageScripts, "", true, true)
 }
 
 func TestLayeredBuildNoTar(t *testing.T) {
-	integration(t).exerciseCleanBuild(TagCleanLayeredBuildNoTar, false, FakeImageNoTar, FakeScriptsFileURL)
+	integration(t).exerciseCleanBuild(TagCleanLayeredBuildNoTar, false, FakeImageNoTar, FakeScriptsFileURL, false, true)
 }
 
 // Test that a build config with a callbackURL will invoke HTTP endpoint
 func TestCleanBuildCallbackInvoked(t *testing.T) {
-	integration(t).exerciseCleanBuild(TagCleanBuild, true, FakeBuilderImage, "")
+	integration(t).exerciseCleanBuild(TagCleanBuild, true, FakeBuilderImage, "", true, true)
 }
 
-func (i *integrationTest) exerciseCleanBuild(tag string, verifyCallback bool, imageName string, scriptsURL string) {
+func TestCleanBuildOnBuild(t *testing.T) {
+	integration(t).exerciseCleanBuild(TagCleanBuildOnBuild, false, FakeImageOnBuild, "", true, true)
+}
+
+func TestCleanBuildOnBuildNoName(t *testing.T) {
+	integration(t).exerciseCleanBuild(TagCleanBuildOnBuildNoName, false, FakeImageOnBuild, "", false, false)
+}
+
+func TestCleanBuildNoName(t *testing.T) {
+	integration(t).exerciseCleanBuild(TagCleanBuildNoName, false, FakeBuilderImage, "", true, false)
+}
+
+func TestLayeredBuildNoTarNoName(t *testing.T) {
+	integration(t).exerciseCleanBuild(TagCleanLayeredBuildNoTarNoName, false, FakeImageNoTar, FakeScriptsFileURL, false, false)
+}
+
+func (i *integrationTest) exerciseCleanBuild(tag string, verifyCallback bool, imageName string, scriptsURL string, expectImageName bool, setTag bool) {
 	t := i.t
 	callbackURL := ""
 	callbackInvoked := false
@@ -188,11 +210,18 @@ func (i *integrationTest) exerciseCleanBuild(tag string, verifyCallback bool, im
 		callbackURL = ts.URL
 	}
 
+	var buildTag string
+	if setTag {
+		buildTag = tag
+	} else {
+		buildTag = ""
+	}
+
 	config := &api.Config{
 		DockerConfig: dockerConfig(),
 		BuilderImage: imageName,
 		Source:       TestSource,
-		Tag:          tag,
+		Tag:          buildTag,
 		Incremental:  false,
 		CallbackURL:  callbackURL,
 		ScriptsURL:   scriptsURL}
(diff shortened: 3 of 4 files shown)


mfojtik:
LGTM [merge]

Good job Eric!

hack/build-test-images.sh
```diff
@@ -27,3 +27,5 @@ buildimage sti_test/sti-fake-user test/integration/images/sti-fake-user
 buildimage sti_test/sti-fake-scripts test/integration/images/sti-fake-scripts
 buildimage sti_test/sti-fake-scripts-no-save-artifacts test/integration/images/sti-fake-scripts-no-save-artifacts
 buildimage sti_test/sti-fake-no-tar test/integration/images/sti-fake-no-tar
+buildimage sti_test/sti-fake-onbuild-source test/integration/images/sti-fake-onbuild-source
+buildimage sti_test/sti-fake-onbuild test/integration/images/sti-fake-onbuild
```
bparees: i don't see the purpose of this image?  and i'd just call the other one "sti-fake-onbuild"  instead of "sti-fake-onbuild-source" (it is the builder image which uses ONBUILD commands)

test/integration/integration_test.go
```diff
@@ -29,6 +29,7 @@ const (
 	FakeImageScripts                = "sti_test/sti-fake-scripts"
 	FakeImageScriptsNoSaveArtifacts = "sti_test/sti-fake-scripts-no-save-artifacts"
 	FakeImageNoTar                  = "sti_test/sti-fake-no-tar"
+	FakeImageOnBuild                = "sti_test/sti-fake-onbuild"
```
bparees: I would have expected this to be the "sti_test/sti-fake-onbuild-source" image (prior to my suggested renaming).

the idea of ONBUILD type builds is that you've got a builder image which contains ONBUILD instructions that should be run as part of the STI build (which is accomplished by performing a "docker build" using the builder image as the FROM)

test/integration/integration_test.go
```diff
@@ -38,6 +39,8 @@ const (
 	TagIncrementalBuildScripts                = "test/sti-incremental-app-scripts"
 	TagIncrementalBuildScriptsNoSaveArtifacts = "test/sti-incremental-app-scripts-no-save-artifacts"
 	TagCleanLayeredBuildNoTar                 = "test/sti-fake-no-tar"
+	TagCleanBuildOnBuild                      = "test/sti-onbuild-trigger"
+	TagOnBuildTrigger                         = "test/sti-file-check-onbuild-trigger"
```
bparees: confused by these names.  what's "trigger" here?

test/integration/integration_test.go
```diff
@@ -303,6 +310,74 @@ func (i *integrationTest) exerciseIncrementalBuild(tag, imageName string, remove
...
+	if callbackHasValidJSON != verifyCallback {
+		t.Fatalf("Sti build did not invoke callback with valid json message")
+	}
+
+	i.checkForImage(tag)
+	containerID := i.createContainer(tag)
+	defer i.removeContainer(containerID)
+	i.checkBasicBuildState(containerID, resp.WorkingDir)
+
+	i.fileExists(containerID, "/sti-fake/src/onbuild")
```
bparees: rather than create a whole new exercise* function just for this(? did i miss some other delta?), i would augment the existing exercise* functions to enable them to validate the output image by checking for file presence.

You shouldn't really need special test code to test an ONBUILD image, you should just be able to confirm that the existing tests pass when the input image contains ONBUILD instructions (and confirm the ONBUILD instructions were executed)

bparees:
removed @soltysh's merge tag as i have some questions on this.  sorry

soltysh:
@ewolinetz thanks for picking it up, but I agree with all the comments @bparees made. Additionally you have some error regarding missing `FakeScriptsFileUrl` variable in jenkins.

hack/build-test-images.sh
ewolinetz: my thought was since ONBUILD doesn't take effect until another image uses it as a FROM we would need to build one as a source first and then we can test to make sure the ONBUILD was triggered with this image..

test/integration/integration_test.go
ewolinetz: I can rephrase this, I used "trigger" as the Docker documentation references the ONBUILD instruction allows for any build instruction to be registered as a "trigger", so this is referring to the ONBUILD being triggered.

test/integration/integration_test.go
ewolinetz: that makes a lot of sense to do - will refactor one of the existing fuctions to do this

hack/build-test-images.sh
bparees: You need one image that contains ONBUILD in its dockerfile.

You then need to use that image as the builder image for an STI invocation.  STI automatically determines that the builder image contains ONBUILD and performs a "pre-build" step in which it docker builds an intermediary image to trigger the ONBUILD instructions.

So again, you just need the one builder image, and then test it like any other builder image, verifying that in the output image, it's clear the ONBUILD instruction was run.

test/integration/integration_test.go
bparees: again I think this goes away, you just need a TagCleanBuildOnBuild image and, if you want, a TagIncrementalBuildOnBuild image.  Those being the 2 output images you're going to produce with an sti build invocation that uses the "onbuild" builder image.

test/integration/integration_test.go
ewolinetz: I had misunderstood how the integration_test had worked, addressed this by removing the unnecessary image (the one without the ONBUILD instruction)

ewolinetz:
@bparees I ran into some odd cases with the ONBUILD strategy from sti, is there a better way to do the OnBuild Dockerfile?  There were a lot of files that weren't copied/set up like the other tests

bparees:
@ewolinetz not following, can you give me an example of what you're seeing?

ewolinetz:
@bparees I was able to work through issues with save-artifact tests failing and cleaned up the sti-fake-onbuild/Dockerfile

test/integration/images/sti-fake-onbuild/Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM sti_test/sti-fake
+
+ONBUILD RUN touch /sti-fake/src/onbuild
+
+RUN touch /sti-fake/src/index.html
```
bparees: still not understanding why this is needed?  the build(assemble script) should create it.

test/integration/integration_test.go
```diff
@@ -38,6 +39,8 @@ const (
 	TagIncrementalBuildScripts                = "test/sti-incremental-app-scripts"
 	TagIncrementalBuildScriptsNoSaveArtifacts = "test/sti-incremental-app-scripts-no-save-artifacts"
 	TagCleanLayeredBuildNoTar                 = "test/sti-fake-no-tar"
+	TagCleanBuildOnBuild                      = "test/sti-onbuild-trigger"
+	TagOnBuild                                = "test/sti-file-check-onbuild-trigger"
```
bparees: these names ("sti-file-check-onbuild-trigger") seem odd.  why not just "test/sti-onbuid-output" and "test/sti-clean-onbuild-output"?

Also in keeping with the others, i'd expect:
TagCleanBuildOnBuild
TagIncrementalBuildOnBuild
to be the variable names.

test/integration/images/sti-fake-onbuild/Dockerfile
ewolinetz: @soltysh is there any reason that an OnBuild strategy might not have the assemble script available to it when trying to build?  http://pastebin.test.redhat.com/295467

test/integration/images/sti-fake-onbuild/Dockerfile
```diff
@@ -0,0 +1,9 @@
...
+
+ADD scripts/.sti/bin/ /tmp/src/
```
soltysh: First of all if you're adding those scripts inside the image, you should point S2I where to look for them with:
`LABEL io.openshift.s2i.scripts-url=image:///tmp/scripts/`. Or not to copy them here since the base image (`sti_test/sti-fake`) has `LABEL io.openshift.s2i.scripts-url=http://127.0.0.1:23456/.sti/bin`, which means the build will download the scripts from there.

test/integration/images/sti-fake-onbuild/Dockerfile
soltysh: Agreed with @bparees, looking into why the `assemble` script is not picked up though.

test/integration/images/sti-fake-onbuild/Dockerfile
soltysh: Almost forgot, make sure above `ADD` command corresponds to `io.openshift.s2i.scripts-url` label.

ewolinetz: Agreed, this should ideally only consist of the FROM line and the ONBUILD line.  I thought I had removed this...

test/integration/images/sti-fake-onbuild/Dockerfile
soltysh: @ewolinetz your problem is first line of the [`assemble`](https://github.com/openshift/source-to-image/blob/master/test/integration/scripts/.sti/bin/assemble#L3) script, which is:
```
cp -Rf /tmp/src /sti-fake/
```
You don't need to satisfy that operation, so just change your RUN command into:
```
RUN touch /sti-fake/src/index.html && \
    mkdir -p /tmp/src
```

bparees: huh?  that sounds backwards.  we should not be creating index.html in a dockerfile RUN command.

ewolinetz: @soltysh if you look at the pastebin you can see that it is able to copy the sources specified in the LABEL but it never moves them to /tmp like it does for the top test (sti-fake).  I also notice that when it `tar`s up the source it does so without them being nested under a src directory.  (Compare lines 27 and 69)

ewolinetz:
after speaking with @soltysh we thought that since OnBuild works out of the app source dir we should also have an ONBUILD COPY instruction

bparees:
At the end of the day we need to be sure the test is checking two things:
1) that the ONBUILD instruction runs (touches some file)
2) that the assemble script runs (touches some *other* file)

is that test now doing that, or is it only doing #1?

ewolinetz:
We are now doing both 1 and 2.

# Executing 2 build triggers
Trigger 0, RUN touch /sti-fake/src/onbuild
Step 0 : RUN touch /sti-fake/src/onbuild
 ---> Running in 22ab5e21e54d
Trigger 1, COPY . /tmp/src/
Step 0 : COPY . /tmp/src/
 ---> 8595b1d7c51c
Removing intermediate container 22ab5e21e54d
Removing intermediate container 434460dbbbba
Step 1 : RUN sh assemble
 ---> Running in 8e1a2d584a9b
 ---> b9de13440b73
...

bparees:
@soltysh I have been operating under the assumption that the onbuild flow is:
1) perform a docker build of the builder image (to trigger the onbuild instructions)
2) run that output image using the assemble script as the entry point

so that onbuild builds are bascially identical to sti builds, except that they do a pre-build.  but it sounds like it's not that straight forward?

ewolinetz:
Per @soltysh 's request in PR #237 I started to update the integration_test.go file for cases where we do not receive back an ImageID from our build.  I'm in the process of also creating an OnBuild test case in addition to our current tests to cover this case

ewolinetz:
in trying to create an ONBUILD test that also does not return an ImageID I'm once again running into the issue that we are not able to execute the assemble script once we add an ONBUILD instruction.  In this case, once I add an ONBUILD to also copy the test fails because an ImageID is returned.

bparees:
we need to get to the bottom of why onbuild is using a different flow than the normal sti invocation.

soltysh:
@bparees the flow is following:
1. Run docker build, which will invoke all onbuild operations. Additionally it'll also execute `assemble` if one was provided, as a last step of that build (see [here](https://github.com/openshift/source-to-image/blob/master/pkg/build/strategies/onbuild/onbuild.go#L141-L143)).
2. Commit the result of 1 with provided entrypoint (here's some logic @mfojtik will be better explaining but in short words, it's looking for [some executable](https://github.com/openshift/source-to-image/blob/master/pkg/build/strategies/onbuild/entrypoint.go#L21)).

@ewolinetz make sure to rebase on top of @mfojtik changes from #237, where he fixed onbuild to allow not passing output image.

In the end everything is working the way it supposed to, imho. We just don't have that path covered in tests nor in docs.

ewolinetz:
@soltysh so am I correct in assuming that we could only test the changes from #237 with an OnBuild strategy?  (e.g. is it possible for STI to produce a build response that does not contain an ImageID?)

soltysh:
We handle that part in strategy spcific code, so it needs to be tested
twice iirc. ImageID will always be there because that tbe actual result of
a build - an image. The only thing is wether we commit it under user
provided name or not.

ewolinetz:
would it make sense to see that if we did not provide a tag in the build config that we would still receive back a ImageID from STI?  That was what I was seeing with my tests - if I specified not to send a tag, and when I checked if I received an ID, I did.  Which was not what I was expecting (as OnBuild will not return an ID if I do not send a tag).  I can add another test if that is the correct outcome.

soltysh:
I suggest to look through the code and compare how plain s2i vs onbuild
does that.  I'm currently not in front of computer and I won't reach one
till Sunday ;-) But generally speaking I suggest both strategies should
work the same way. I'm fine with either solution as long as it's applied in
both places.  Or maybe it's something that should be handled one level
higher in common place, this would simplify our code.

ewolinetz:
going through the code i've found that for `Build` sti always returns an ImageID, OnBuild will only return one if you provide one, and layered never returns one.  I have added tests for each of these scenarios in addition to the OnBuild clean & integration tests.

test/integration/integration_test.go
```diff
@@ -214,34 +243,52 @@ func (i *integrationTest) exerciseCleanBuild(tag string, verifyCallback bool, im
 		t.Fatalf("Sti build did not invoke callback with valid json message")
 	}
 
-	i.checkForImage(tag)
-	containerID := i.createContainer(tag)
-	defer i.removeContainer(containerID)
-	i.checkBasicBuildState(containerID, resp.WorkingDir)
+	if expectImageName {
```
soltysh: Please rewrite this if as:
```
if expectImageName && resp.ImageID == "" {
    ...
} else if !expectImageName && resp.ImageID != "" {
    ...
}
```
I just can't look at nested ifs :wink:

soltysh:
One small nit, but LGTM. While waiting for fix:
[test]

soltysh:
Thanks for doing this!!!

test/integration/integration_test.go
```diff
@@ -214,34 +243,52 @@ func (i *integrationTest) exerciseCleanBuild(tag string, verifyCallback bool, im
...
+		if resp.ImageID == "" {
+			t.Fatalf("Sti build did not receive an imageID in response")
```
mfojtik: S2I

test/integration/integration_test.go
mfojtik: you can also revert the logic and check `if ! expectedImageName`. In that case just move the `if setTag` there and return. 

Then you just get one less if. Then `Fatalf` actually 'exit' so there is no need for `else` there. So this can look like this:

```go
if setTag {
  # ...
}

if ! expectedImageName && len(resp.ImageID) > 0 {
  t.Fatalf("...")
}

if len(resp.ImageID) == 0 {
  t.Fatalf("...")
}

```

test/integration/integration_test.go
ewolinetz: I figured it'd still be sti since we're talking about the command `sti build`.  I'll update these lines and any other instances of "sti" to be "s2i"

ewolinetz:
addressed nit and reworded "STI" to "S2I"

test/integration/images/sti-fake-onbuild/Dockerfile
```diff
@@ -0,0 +1,11 @@
+FROM sti_test/sti-fake
+
+ONBUILD RUN touch /sti-fake/src/onbuild
+
+# The ONBUILD strategy only works with the application source dir so we need
+# to manually specify to copy to another location.
+#
+# This is a little hack-ish given that we know our assemble script requires files to be in /tmp/src
+# we will copy there, and also set our WORKDIR to be the same location so it has access to the scripts
+ONBUILD COPY . /tmp/src/
```
bparees: i'm sure i've missed some of the conversation, but i'm still confused as to why this is necessary.

soltysh: By default all onbuild images copy current directory when building (eg. [python:onbuild](https://github.com/docker-library/python/blob/7663560df7547e69d13b1b548675502f4e0917d1/3.4/onbuild/Dockerfile#L9)). Aside from the application sources current directory contains also `assemble` and `run` scripts, we need to make sure they are copied into the image, because if you have a look at generated `Dockerfile` you'll notice that right after `FROM` line we put `RUN sh assemble` [line](https://github.com/openshift/source-to-image/blob/master/pkg/build/strategies/onbuild/onbuild.go#L168-L170). The reason why we are copying it to `/tmp/src/` is we set WORKDIR to be `/tmp/src/` in the next line. I hope that helps :smiley:

soltysh:
For me it's still LGTM, @bparees leaving the actual merge for you, as you are having some doubts still here :wink:

test/integration/images/sti-fake-onbuild/Dockerfile
bparees: per our IRC discussion i'm accepting temporary defeat here.  but we should remove the S2I-OnBuild assumption that requires the images contain a COPY command of this form.  We should just inject the source into the image ourselves, like we do in the normal s2i flow.

bparees:
[merge]
thanks for the perseverance @ewolinetz :)

soltysh:
@ewolinetz :rose: :beer: :cake: :heart: or all for your work!

soltysh:
Strange error has happened, openshift-vagrant failed us...
re-[merge]

PR #230, opened 2015-07-03
Added fallback to use local image when Builder's PullImage fails

soltysh:
@bparees ptal

### DIFF
--- a/pkg/build/strategies/strategies.go
+++ b/pkg/build/strategies/strategies.go
@@ -2,6 +2,7 @@ package strategies
 
 import (
 	dockerclient "github.com/fsouza/go-dockerclient"
+	"github.com/golang/glog"
 	"github.com/openshift/source-to-image/pkg/api"
 	"github.com/openshift/source-to-image/pkg/build"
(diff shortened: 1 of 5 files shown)


pkg/build/strategies/strategies.go
```diff
@@ -37,8 +38,12 @@ func GetBuilderImage(config *api.Config) (*docker.PullResult, error) {
 	var image *dockerclient.Image
 	if config.ForcePull {
 		image, err = d.PullImage(config.BuilderImage)
+		if err != nil {
+			glog.Warningf("An error occurred when pulling %s: %v. Falling back to use local instead.", config.BuilderImage, err)
```
bparees: Instead of "falling back to use local image" this should say "attempting to use local image" since the d.CheckImage could then also fail if the image doesn't exist locally at all.

soltysh:
@bparees addressed your comment.

bparees:
lgtm [merge]

soltysh:
This is not the first time, that when merging changes for some reason our test images are not being built. It would be good to have a closer look into it.
re-[merge]

PR #232, opened 2015-07-06
Fixed some leftovers from moving to new label scheme including: docs,getDestination method and template for Dockerfile.

soltysh:
@mfojtik @bparees ptal


bparees:
lgtm. [merge]

PR #233, opened 2015-07-06
issue 133 turn on loglevel 5 when -v passed into test script

gabemontero:
@bparees @soltysh I have a pull request for the changes needed to enable STI loggging when -v is passed into the test-integration.sh

### DIFF
--- a/test/integration/integration_test.go
+++ b/test/integration/integration_test.go
@@ -5,6 +5,8 @@ package integration
 import (
 	"bytes"
 	"encoding/json"
+	"flag"
+	"fmt"
 	"io/ioutil"
 	"net/http"
 	"net/http/httptest"
@@ -15,7 +17,7 @@ import (
 	"time"
 
 	"github.com/fsouza/go-dockerclient"
-
+	"github.com/golang/glog"
 	"github.com/openshift/source-to-image/pkg/api"
 	"github.com/openshift/source-to-image/pkg/build/strategies"
 )
@@ -81,6 +83,15 @@ func dockerClient(config *api.DockerConfig) (*docker.Client, error) {
 	return docker.NewClient(config.Endpoint)
 }
 
+func getLogLevel() (level int) {
+	for level = 5; level >= 0; level-- {
+		if glog.V(glog.Level(level)) == true {
+			break
+		}
+	}
+	return
+}
+
 // setup sets up integration tests
 func (i *integrationTest) setup() {
 	i.dockerClient, _ = dockerClient(dockerConfig())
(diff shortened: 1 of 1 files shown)


test/integration/integration_test.go
```diff
@@ -102,6 +113,25 @@ func (i *integrationTest) setup() {
 		}
 		i.setupComplete = true
 	}
+
+	from := flag.CommandLine
+	if vflag := from.Lookup("v"); vflag != nil {
+		fmt.Fprint(os.Stdout, "integration_test.go: found -v flag from starter bash script with initial glog level ", getLogLevel(), "\n")
```
soltysh: Somehow, I prefer `Fprintf` instead of `Fprint`, in that case the output seems to be more readable imho.

test/integration/integration_test.go
```diff
@@ -43,7 +45,7 @@ const (
 	// scripts are made available for integration testing.
 	//
 	// Port 23456 must match the port used in the fake image Dockerfiles
-	FakeScriptsHttpURL = "http://127.0.0.1:23456/sti-fake/.sti/bin"
+	FakeScriptsHTTPURL = "http://127.0.0.1:23456/sti-fake/.sti/bin"
```
soltysh: We intentionally left Http not uppercased for readability. Too much uppercase hurts eyes :wink:, just fyi.

test/integration/integration_test.go
```diff
@@ -102,6 +113,25 @@ func (i *integrationTest) setup() {
...
+
+		// the thing here is that we are looking for the bash -v passed into test-integration.sh (with no value),
+		// but for glog (https://github.com/golang/glog/blob/master/glog.go), one specifies
+		// the logging level with -v=# (i.e. -v=0 or -v=3 or -v=5).
+		// so, for the changes stemming from issue 133, we 'reuse' the bash -v, and set the highest glog level.
+		// (if you look at STI's main.go, and setupGlog, it essentially maps glog's -v to --loglevel for use by the sti command)
+		//NOTE - passing --loglevel or -v=5 into test-integration.sh does not work
+		if getLogLevel() != 5 {
+			vflag.Value.Set("5")
+			fmt.Fprint(os.Stdout, "integration_test.go: glog level now ", getLogLevel(), "\n")
```
soltysh: Same here with `Fprintf`.

test/integration/integration_test.go
gabemontero: OK ... I had changed it as a result of golint output.  If it is everyone's preference to buck the golint recommendation, I can switch it.

test/integration/integration_test.go
gabemontero: OK ... I so far have not noticed the difference, and I liked the ability to choose between os.Stdout and os.Stderr.  I'll ponder/consider, pending additional reasons to switch.

test/integration/integration_test.go
soltysh: You don't have to strictly follow golint, sometimes it's good to take readability in mind, otherwise you may end up with too much uppercased acronyms.

soltysh:
Two small nits, other than that LGTM.

test/integration/integration_test.go
bparees: i'll defer to @soltysh's eyes on this one, but HTTP would probably have been my instinct too :)

soltysh: :wink:

gabemontero:
I've incorporated Maciej's comments.

test/integration/integration_test.go
```diff
@@ -81,6 +83,15 @@ func dockerClient(config *api.DockerConfig) (*docker.Client, error) {
 	return docker.NewClient(config.Endpoint)
 }
 
+func getLogLevel() (level int) {
+	for level = 5; level >= 0; level-- {
+		if glog.V(glog.Level(level)) == true {
+			break
```
bparees: could just return instead of break

test/integration/integration_test.go
```diff
@@ -102,6 +113,25 @@ func (i *integrationTest) setup() {
 		}
 		i.setupComplete = true
 	}
+
+	from := flag.CommandLine
+	if vflag := from.Lookup("v"); vflag != nil {
+		fmt.Fprintf(os.Stdout, "integration_test.go: found -v flag from starter bash script with initial glog level %d\n", getLogLevel())
```
bparees: any reason not to use glog to output this?

gabemontero: it was just a function on not getting glog to work initially when I was developing ... I can switch it.

gabemontero: Actually, seem to be hitting a bit of a chicken / egg situation with the first print.
Neither
glog.V(0).Infof("integration_test.go: found -v flag from starter bash script with initial glog level %d", getLogLevel())
no glog.Infof ....
print that first message.  However, when I then set the -v param, the glogs work.

Would you rather I just not print until I set the logging level, or can I use fmt.Fprintf for that first print?

bparees: you can just leave it.

bparees:
lgtm. [merge]

Issue #235, opened 2015-07-09
Optional output image does not work for onbuild strategy


soltysh:
If I run:
```
$ sti build git://github.com/pmorie/simple-html sti_test/sti-fake-onbuild
```
The `sti_test/sti-fake-onbuild` is one of the integration images introduced by @ewolinetz in #229, notice that I **didn't** specify the output image name, because it's optional, right :wink:, I'm getting:

```
$ sti build git://github.com/pmorie/simple-html sti_test/sti-fake-onbuild --loglevel=3

Builder Image:          sti_test/sti-fake-onbuild
Source:                 git://github.com/pmorie/simple-html
Output Image Tag:
Incremental Build:      enabled
Remove Old Build:       disabled
Force Pull:             enabled
Quiet:                  disabled
Layered Build:          disabled
Docker Endpoint:        unix:///var/run/docker.sock
Docker Pull Config:     /home/maszulik/.dockercfg
Docker Pull User:       soltysh

I0709 10:42:56.508916 07135 docker.go:216] Pulling image sti_test/sti-fake-onbuild:latest
I0709 10:43:06.175273 07135 docker.go:219] An error was received from the PullImage call: Error: image sti_test/sti-fake-onbuild:latest not found
W0709 10:43:06.175324 07135 strategies.go:42] An error occurred when pulling sti_test/sti-fake-onbuild: unable to get sti_test/sti-fake-onbuild:latest. Attempting to use local image.
I0709 10:43:06.178121 07135 onbuild.go:73] Preparing the source code for build
I0709 10:43:06.178242 07135 clone.go:27] Cloning into /tmp/sti352390512/upload/src
I0709 10:43:07.246956 07135 docker.go:199] Image sti_test/sti-fake-onbuild:latest available locally
I0709 10:43:07.246983 07135 docker.go:305] Image contains io.openshift.s2i.scripts-url set to 'https://github.com/openshift/source-to-image/blob/master/test/integration/scripts/.sti/bin/'
I0709 10:43:07.248605 07135 docker.go:199] Image sti_test/sti-fake-onbuild:latest available locally
I0709 10:43:07.248617 07135 docker.go:305] Image contains io.openshift.s2i.scripts-url set to 'https://github.com/openshift/source-to-image/blob/master/test/integration/scripts/.sti/bin/'
[...]

PR #237, opened 2015-07-09
Do not attempt to get image ID when no output image specified

mfojtik:


mfojtik:
@soltysh @bparees PTAL

mfojtik:
Fixes #235

soltysh:
LGTM [merge]
@ewolinetz can you add int tests for those cases as well, I'm wondering if we do have them for plain s2i builds...

ewolinetz:
@soltysh I can look into that, I am wondering if we should hold off on adding more until the current questions on the PR are answered?

PR #238, opened 2015-07-10
Add flag to prevent builders from running as root

csrwng:
If the --no-root flag is specified, the build will fail if the builder does not specify a USER with a numeric id that is not 0.

### DIFF
--- a/pkg/build/strategies/onbuild/onbuild_test.go
+++ b/pkg/build/strategies/onbuild/onbuild_test.go
@@ -101,6 +101,85 @@ func TestCreateDockerfileWithAssemble(t *testing.T) {
 	}
 }
 
+func TestCheckNoRoot(t *testing.T) {
+	tests := []struct {
+		name      string
+		noRoot    bool
+		user      string
+		onbuild   []string
+		expectErr bool
+	}{
+		{
+			name:      "NoRoot is not set",
+			noRoot:    false,
+			user:      "root",
+			onbuild:   []string{},
+			expectErr: false,
+		},
+		{
+			name:      "NoRoot is set, non-numeric user",
+			noRoot:    true,
+			user:      "default",
+			onbuild:   []string{},
+			expectErr: true,
+		},
+		{
+			name:      "NoRoot is set, user 0",
+			noRoot:    true,
+			user:      "0",
+			onbuild:   []string{},
+			expectErr: true,
+		},
+		{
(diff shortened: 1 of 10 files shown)


csrwng:
@bparees @mfojtik ptal

pkg/build/strategies/onbuild/onbuild_test.go
```diff
@@ -101,6 +101,78 @@ func TestCreateDockerfileWithAssemble(t *testing.T) {
...
+			user:      "200",
+			onbuild:   []string{"USER 500", "VOLUME /data"},
+			expectErr: false,
+		},
+		{
+			name:      "NoRoot is set, numeric user, onbuild user 0",
+			noRoot:    true,
+			user:      "200",
+			onbuild:   []string{"RUN echo \"hello world\"", "USER 0"},
+			expectErr: true,
```
bparees: add a basic case for "user: 0"?

pkg/docker/docker.go
```diff
@@ -176,12 +177,19 @@ func (d *stiDocker) GetImageUser(name string) (string, error) {
 // IsImageOnBuild provides information about whether the Docker image has
 // OnBuild instruction recorded in the Image Config.
 func (d *stiDocker) IsImageOnBuild(name string) bool {
+	onbuild, err := d.GetOnBuild(name)
+	return err == nil && len(onbuild) > 0
+}
+
+// IsImageOnBuild provides information about whether the Docker image has
```
bparees: GetOnBuild.

pkg/docker/docker.go
```diff
@@ -176,12 +177,19 @@ func (d *stiDocker) GetImageUser(name string) (string, error) {
...
+// OnBuild instruction recorded in the Image Config.
+func (d *stiDocker) GetOnBuild(name string) ([]string, error) {
 	name = getImageName(name)
 	image, err := d.client.InspectImage(name)
 	if err != nil {
-		return false
+		return nil, err
 	}
-	return len(image.Config.OnBuild) > 0
+	return image.Config.OnBuild, nil
```
bparees: This returns the "ONBUILD USER root" command from the original base image as "USER root" in the array?

csrwng: Yes

bparees:
lgtm [merge]

PR #239, opened 2015-07-10
Add error for no-root condition

csrwng:
Displays an error when no-root condition is not met


csrwng:
@bparees I realized that my error wasn't getting displayed when running without a --loglevel > 1. I am now returning a proper error so it gets displayed.

bparees:
lgtm.

PR #243, opened 2015-07-20
Revert "Add flag to prevent builders from running as root"

csrwng:
This reverts commits 31c87068865edcafe754e0a83c67f2903d5000d6 and 5d118bd5f36bac6d9dc1aa80b3cb81992ea42b31.

The no-root flag will be re-implemented as a flag specifying which users to allow


csrwng:
@bparees ptal

csrwng:
[test]

bparees:
lgtm [merge]

PR #245, opened 2015-07-24
Rename /opt/openshift to /opt/app-root

vbehar:
following the same change in sti-base : https://github.com/openshift/sti-base/pull/58


bparees:
lgtm.  thanks!
[merge]

@soltysh @mfojtik fyi.

PR #247, opened 2015-07-24
Remove empty line at the top of the generated scripts

vbehar:
Otherwise, when running the image with "docker run ...", we have an error :
    Cannot start container [...]: exec format error

Because docker "looks at the first few bytes to determine the type of the file"

See https://github.com/docker/docker/issues/10668 for more


bparees:
lgtm.    thanks @vbehar!
[merge]

soltysh:
re-[merge]

Issue #197, opened 2015-05-22
Ignoring files


Maxim-Filimonov:
During investigation of a very big image sizes produced by STI I have found that STI ignores completely .gitignore or .dockerignore and puts every folder into the container it is building. 
We have some temporarily files which are ignored in both git and dockerignore with significant weight(about 125MB) adding quite a lot of unnecessary data to production image. I cannot find a way to make STI ignore those files. Would be good if STI can read .dockerignore or .gitignore to get this files and do not put them inside the container.

bparees:
STI does a git clone of your repo, so if you have files that are in your .gitignore, why are they in your repo in the first place?

soltysh:
Is this a continuation of the problem you described in #196 or something different?

Maxim-Filimonov:
@bparees we don't use git clone. As we are running sti on CI we just use 'sti build .' inside checked out folder. Files there are artifacts from running tests inside the same folder and mongodb journal from tests again. With docker directly we just added those files to .dockerignore while I can't wait the wait to do it with sti.
@soltysh it is different problem

bparees:
Though it's it's not an ideal solution, for now you could make a custom assemble script that deletes the files you don't care about as part of the "build". 

Ben Parees | OpenShift

Maxim-Filimonov:
@bparees yeah i'm gonna do exactly that for `tmp/cache` as it's not needed for running app. But it still takes 2 minutes to transfer all those files inside the image before the build. I'll try to hack some PR over weekend

bparees:
Cool, just a thought as you're working on it....we'd probably prefer to introduce a new ".stiignore" file over using existing files so we don't get unintended behavior.  

If you do want to use add a patch to respect existing .dockerignore/.gitignore files i think it would be best to do it as a non-default flag that must be passed to STI, again to avoid unexpected behavior for users in which files don't get included.

soltysh:
@bparees since under the covers we do actually docker build, I don't mind us reusing .dockerignore. Especially if somebody already has one. One other options would be to treat those files (.stiignore and .dockerignore) equally, meaning the same format and behavior.

bparees:
We don't always do a docker build under the covers. Only under certain circumstances right? 

Ben Parees | OpenShift

soltysh:
Right, that was a shortcut in my head :wink:

Maxim-Filimonov:
[...]

Issue #206, opened 2015-05-29
Review code in cmd/sti/main.go#newCmdBuild to extract what might be useful to have in origin.


soltysh:
Per [this comment](https://github.com/openshift/source-to-image/pull/204#discussion_r31258661), we should review the code in cmd/sti/main.go#newCmdBuild and extract as much as we can from there into separate package that we'll be able to reduce the amount of duplication in code in S2I and origin.

coreydaley:
@bparees 
The only code that I see in newCmdBuild that MIGHT be useful to factor out into something reusable is https://github.com/openshift/source-to-image/blob/master/pkg/cmd/cli/cmd/build.go#L100-L123

So, with that being said:
- Do you see other code that would be useful?
- Is it worth it?
- Is this github issue still relevant?

bparees:
not worth it.

Issue #149, opened 2015-02-27
sti build hangs


iocanel:
I am using sti from the command line and in most cases, I notice that it "hungs".

When I am using log level 3 I see that the build output always hungs at exactly the same line.

What is strange though is that when I check the docker logs, I can see that the assemble script has successfully finished and also that the container has exited gracefully. The target image of course is never created.

The last couple of days I've made so many attempts and only two of them were successful (I haven't done anything different).

bparees:
Can you give us steps to recreate?  what STI image, what STI scripts, what commit level of STI, etc?

(also what docker version)

iocanel:
I've been using the https://github.com/fabric8io/base-sti base image. The sti scripts are part of the image and they are referenced using their github url. 

My docker is at 1.4.1 and I'll try to upgrade and see if it makes any difference.

Here's an example of how I use the command:

sti build https://github.com/fabric8io/example-camel-cdi.git fabric8/base-sti fabric8/camel-example --loglevel=3

bparees:
yup, I was able to reproduce this just running your sti command.  and I do see the container has exited after running assemble successfully.

Given that the sti output hangs partway through the assemble script output, i'm guessing we're screwing up/overfilling a buffer somewhere.  @soltysh can you take a look?

iocanel:
The behaviour that I see makes me feel that the issue is related with the readers of the containers output and error. It seems like on of them stops reading and fails to close preventing the method to return? (limited understanding of golang here).

bparees:
in the meantime, could you edit your assemble script to redirect the maven output to /dev/null?  that might let you make progress...

iocanel:
It does seem to make a difference.

soltysh:
Looking...

iocanel:
Using the latest master, I still have the same issue.

soltysh:
@iocanel can you please check this once more. I just did a fresh build of STI ([this revision](https://github.com/openshift/source-to-image/commit/aded45154b8b3bd2297d7643bde6176b4d71f01f)) and the command:
```
sti build https://github.com/fabric8io/example-camel-cdi.git fabric8/base-sti fabric8/camel-example --loglevel=3
```
worked without any problem resulting in `fabric8/camel-example` image. The only problem I hit was running that result image ended in:
```
$ docker run fabric8/camel-example
/tmp/scripts/run: line 3: deploy-and-start: command not found
```
[...]

PR #248, opened 2015-07-24
issue197: implement .dockerignore style function via .s2iignore file …

gabemontero:
…and code changes, with updated doc and tests; investigation and work around for hang with very quick git clones (apparent Go bug)

@bparees @soltysh Gentlemen - the pull request for issue 197 - .s2iignore file.  Also has a work around for what I believe is a Go issue for hang during git clone that I've see pretty consistently with the super small sinatra ruby example that clones very quickly.  I also have some analysis in comments just below the workaround.

Fixed https://github.com/openshift/source-to-image/issues/197


README.md
```diff
@@ -40,6 +40,19 @@ image:
...
+expression that best captures a set of files and directories you want removed.  You can also cite
+any specific files you want to remain if they are captured by earlier regular expressions in the `.s2iignore` file.
+The capabilities and syntax of the `.s2iignore` mirror the capabilities of the `.dockerignore`
+file described [in the Docker documentation](https://docs.docker.com/reference/builder/#dockerignore-file) at the time this
+feature was introduced.
+
+The s2i builder does not directly reference `.dockerignore` or other similar ignore files to avoid any possible unintended behavior,
+where if the capability of those mechanisms change in the future in ways to make them
+incompatible with the intent and purpose of s2i and what it provides for generating images to
+run in OpenShift.
```
bparees: +1

pkg/build/ignore/docker/dockign.go
```diff
@@ -0,0 +1,116 @@
...
+}
+
+func New(req *api.Config) *DOCKIGN {
+	d := &DOCKIGN{
+		Config: req,
+	}
+	return d
+}
+
+func (b *DOCKIGN) Ignore(config *api.Config) {
```
bparees: there's no reason for this to be a method on a struct if you're going to take the only field of the struct as an arg anyway....  either make it a first class function and drop the struct, or have it take no args.

What you really probably want to do here is define an interface with a single Ignore() method so you can provide a mock implementation for testing.  And again you can either take an arg on that method, or you can simply have the implementation be defined on a struct that includes the Config object....but the former makes more sense to me.  You're really just defining an interface for a utility function and the implementation of the utility function doesn't need to hold any state.

on top of that i'd split this out into two methods:
1) get the list of files to ignore
2) remove files in a list from a directory

pkg/build/ignore/docker/dockign.go
```diff
@@ -0,0 +1,116 @@
...
+	path := filepath.Join(config.WorkingDir, "upload", "src", api.DefaultIgnoreFile)
```
bparees: if this path isn't being well defined in a constant somewhere, we probably ought to.  i'm a little worried if we have code all over the place doing a Join(workdir,"upload","src") to figure out where the source is after git clone.

pkg/build/ignore/docker/dockign.go
```diff
@@ -0,0 +1,116 @@
...
+	file, err := os.Open(path)
+	if err != nil {
+		glog.V(4).Infof("Ignore processing, problem opening %s because of %v\n", path, err)
```
bparees: need to break out this error handling...if the file doesn't exist, that's fine, but if something else went wrong we need to warn the user and abort the build, otherwise we're going to build something that includes files they did not intend.

pkg/build/ignore/docker/dockign.go
```diff
@@ -0,0 +1,116 @@
...
+
+	// so, to duplicate the .dockerignore capabilities (https://docs.docker.com/reference/builder/#dockerignore-file)
+	// we have a flow that follows:
+	/*
+
+		0) First note, .dockerignore rules are NOT recursive (unlike .gitignore) .. you have to list subdir explicitly
+		1) Read in the exclusion patterns
+		2) Skip over comments (noted by #)
+		3) note overrides (via exclamation sign i.e. !) and reinstate files (don't remove) as needed
+		4) leverage Glob matching to build list, as .dockeringore is documented as following filepath.Match / filepath.Glob
```
bparees: s/.dockeringore/.dockerignore/

pkg/build/ignore/docker/dockign.go
```diff
@@ -0,0 +1,116 @@
...
+
+	*/
+
+	filesToDel := make(map[string]string)
+	scanner := bufio.NewScanner(file)
+	for scanner.Scan() {
+		filespec := scanner.Text()
+		glog.V(4).Infof(".s2iignore lists a file spec of %s \n", filespec)
+
+		if strings.HasPrefix(filespec, "#") {
```
bparees: how are lines with a leading space handled by dockerignore?  i assume we should be stripping prefix whitespace before doing any line processing.

pkg/build/ignore/docker/dockign.go
```diff
@@ -0,0 +1,116 @@
...
+			continue
+		}
+		if strings.HasPrefix(filespec, "!") {
+			//remove any existing files to del that the override covers
+			// and patterns later on that undo this take precedence
+
+			// first, remove ! ... note, replace ! with */ did not have
+			// exected effect with filepath.Match
```
bparees: exected->expected

pkg/build/ignore/docker/dockign.go
```diff
@@ -0,0 +1,116 @@
...
+			filespec = strings.Replace(filespec, "!", "", 1)
+
+			// iterate through and determine ones to leave in
+			dontDel := make([]string, 0)
+			for candidate, _ := range filesToDel {
+				// strip base for file name to file name compare
+				// NOTE, it would appear from git/docker ignore doc that you don't
+				// do something like !*/*/file.ext
```
bparees: so ! arguments only apply to flie basenames, you can't specify a full directory?

eg i can't specify "!foo/bar/myfile"?

and if i specify "!myfile" it will not ignore (will keep) any file named "myfile" in any path?

pkg/build/ignore/docker/dockign.go
```diff
@@ -0,0 +1,116 @@
...
+			continue
+		}
+
+		globspec := filepath.Join(config.WorkingDir, "upload", "src", filespec)
+		glog.V(4).Infof("using globspec %s \n", globspec)
+		list, gerr := filepath.Glob(globspec)
+		if gerr != nil {
+			glog.V(4).Infof("Glob failed with %v \n", gerr)
+		} else {
+			for _, globresult := range list {
```
bparees: filesToDel=append(filesToDel,list)
https://golang.org/pkg/builtin/#append

pkg/build/ignore/docker/dockign.go
```diff
@@ -0,0 +1,116 @@
...
+				glog.V(5).Infof("Glob result %s \n", globresult)
+				filesToDel[globresult] = globresult
+
+			}
+		}
+
+	}
+
+	if err := scanner.Err(); err != nil && err != io.EOF {
+		glog.V(4).Infof("Problem processing .s2iignore %v \n", err)
```
bparees: again this should probably be a fatal error for the build.

pkg/build/ignore/docker/dockign.go
```diff
@@ -0,0 +1,116 @@
...
+	}
+
+	// process complied list of files based on entries from .s2iignore
```
bparees: complied->compiled

pkg/build/ignore/docker/dockign.go
```diff
@@ -0,0 +1,116 @@
...
+	for _, fileToDel := range filesToDel {
+		glog.V(5).Infof("attempting to remove file %s \n", fileToDel)
+		rerr := os.RemoveAll(fileToDel)
+		if rerr != nil {
+			glog.V(5).Infof("error removing file %s because of %v \n", fileToDel, rerr)
```
bparees: fatal.

pkg/build/interfaces.go
```diff
@@ -41,11 +41,20 @@ type Downloader interface {
 	Download(*api.Config) (*api.SourceInfo, error)
 }
 
+// Implement ignore file type processing on source tree
+// NOTE: raised to this level for possible future extensions to
+// support say both .gitignore and .dockerignore level functionality
+// ( currently do .dockerignore)
```
bparees: ah hah. i should have read ahead before commenting.  I still think this bears some restructuring but it's not so far off from what i described.

pkg/build/ignore/docker/dockign.go
```diff
@@ -0,0 +1,116 @@
...
+
+	"github.com/golang/glog"
+	"github.com/openshift/source-to-image/pkg/api"
+)
+
+type DOCKIGN struct {
+	Config *api.Config
+}
+
+func New(req *api.Config) *DOCKIGN {
```
bparees: aside from the refactoring comments below, there's generally no reason to create a New method if you're just initializing the struct like this, since it doesn't really provide any value.

pkg/build/strategies/onbuild/onbuild_test.go
```diff
@@ -7,13 +7,17 @@ import (
...
+	"github.com/openshift/source-to-image/pkg/build"
+	"github.com/openshift/source-to-image/pkg/build/ignore/docker"	
 	"github.com/docker/docker/builder/parser"
 
 	"github.com/openshift/source-to-image/pkg/api"
 	"github.com/openshift/source-to-image/pkg/test"
 )
 
-type fakeSourceHandler struct{}
+type fakeSourceHandler struct{
```
bparees: what's special about ignorer that fakeSourceHandler needs Ignorer populated, but doesn't need Preparer and Downloader populated, for this test?

pkg/git/git.go
```diff
@@ -80,8 +80,165 @@ func (h *stiGit) Clone(source, target string) error {
...
 	go pipeToLog(outReader, glog.Info)
 	go pipeToLog(errReader, glog.Error)
-	return h.runner.RunWithOptions(opts, "git", "clone", "--quiet", "--recursive", source, target)
+	//previously, we were doing the git clone with --quiet, but with the piping of the wrapping cmd stdout/stderr
+	// to glog on separate threads, we were causing a deadlock processing stdout/stderr channels across 4 goroutines.
+	// a SIGQUIT produced goroutine thread dump that shows the deadlock is included  after this method for historical
+	// purposes.
+	// Since it seems contradictory to run with --quiet and also pipe git output to glog, the workaround to the apparent
+	// go issue is the remove the --quiet option.  I also saved the core file produced by the SIGQUIT.
+	// a discussion on whether this is frequent enough or bad enough to open a Go issue on github could be warranted.
```
bparees: hm.  I'm not seeing the actual deadlock/contention point here, I may need you to walk me through this on monday.  it seems like maybe we're just stuck waiting for input that never comes and we need to better handle/recognize when the output (stdout or stderr) stream is done?

and i'm not sure we want to remove --quiet since now we'll be spewing the output of git clone into the logs, right?  (which i guess isn't that bad, but it's less pretty anyway).

but regardless great find/investigation.  Not sure why you are the only one hitting this regularly, but i know @rcernich and @soltysh will be interested in this as they were trying to chase down git clone hangs previously.

pkg/git/git.go
```diff
@@ -80,8 +80,165 @@ func (h *stiGit) Clone(source, target string) error {
...
+	return h.runner.RunWithOptions(opts, "git", "clone", "--recursive", source, target)
 }
+	/*
```
bparees: we'll probably want to remove this from the final commit but worth having for discussion for now.

bparees:
some restructuring suggestions and needs unit tests but the foundation looks sound to me.

pkg/git/git.go
gabemontero: Yeah, I think if still desired a verbal conversation could flush out the diagnosis.  But I just had a bit of an additional break through ...

I was going to note when I dove into your review responses tomorrow that I originally tried not launching the go routines that called pipeToLog to glog since --quiet was specified, and that did NOT do the trick, and so I fell back onto removing --quiet.

But a lightbulb popped on about 15 minutes ago and I realized that attempt was not quite right.  Couldn't wait until Monday to try it and report ;-) ....It is not just the go routines, but the setting of the CommandOpts Stdout and Stderr that is at issue here when --quiet is specified (and hence no output is sent to those pipes).  I mentioned in the big temporary comment I left for this discussion that was at contention too.

I just tried it - when I change the clone method to just

	opts := util.CommandOpts{
	}
	return h.runner.RunWithOptions(opts, "git", "clone", "--quiet", "--recursive", source, target)

Then the git clone of my fork of simple-openshift-sinatra-sti runs perfectly.

So yeah, if you don't want any output from git clone to show up ... i.e. -- quiet, then the whole creation of pipes and redirection of logging make no sense.  

I also tried an experiment with just specifying stderr (in case --quiet does not supress that), and that appears to be working too (though my very first one may have hung, but I may have aborted it too quickly).  I did also retry both stdout and stderr and it still failed.

I'll push either the no pipes and stderr only pipe change (if I continue to see consistent good results) with the other responses to your comments on Monday.

README.md
```diff
@@ -40,6 +40,19 @@ image:
 Additionally for the best user experience and optimized `sti` operation we suggest images
 to have `/bin/sh` and `tar` commands available.
 
+Pruning the contents of the source tree is possible if the user supplies a
+`.s2iignore` file in the root directory of the source repository.  Supply line by line the regular
+expression that best captures a set of files and directories you want removed.  You can also cite
```
soltysh: Maybe I'm picky, but I'd generally stay away from pruning/remove wording, because we are just ignoring/excluding those file from the build process, the actual removal that's happening under the hood does not matter for end user, imho.

README.md
```diff
@@ -61,7 +74,7 @@ See [here](https://github.com/openshift/source-to-image/blob/master/docs/builder
 The `sti build` workflow is:
 
 1. `sti` creates a container based on the build image and passes it a tar file that contains:
-    1. The application source in `src`
+    1. The application source in `src`, excluding any files marked for removal by `.s2iignore`
```
soltysh: Same comment, just excluding.

hack/common.sh
```diff
@@ -103,10 +103,23 @@ sti::build::build_binaries() {
...
-          -ldflags "${version_ldflags}" \
-          "${binaries[@]}"
-      sti::build::unset_platform_envs "${platform}"
+      # have to enable these debug flags for using gdb separately from goflags
+      # because the -gcflags "-N -l" ... i.e. the -N and -l are the setting for -gcflags
+      # not for go itself .. for now we punt on other goflags if developers are
+      # compiling to run with gdb
+      if [[ -v STI_ENABLE_GDB ]]; then
+	  echo "Enabling compiling go with debug to allow gdb usage (i.e. -gcflags \"-N -l\") since STI_ENABLE_GDB is set"
+	  echo "In this case, STI_ENABLE_GDB is set to  ${STI_ENABLE_GDB}"
```
soltysh: Why you're printing it, if you're not using its value?

hack/common.sh
```diff
@@ -103,10 +103,23 @@ sti::build::build_binaries() {
...
+	  go install -gcflags "-N -l"\
+             -ldflags "${version_ldflags}" \
+             "${binaries[@]}"
+	  sti::build::unset_platform_envs "${platform}"
+      else
+	  go install "${goflags[@]:+${goflags[@]}}" \
+             -ldflags "${version_ldflags}" \
+             "${binaries[@]}"
+	  sti::build::unset_platform_envs "${platform}"
+      fi
```
soltysh: I'm not sure about doing this. We generally try to keep consistent with what we have in origin, at least more or less.
Doesn't just setting `STI_GOFLAGS` work here (see [here](https://github.com/openshift/source-to-image/blob/master/hack/common.sh#L77))?

pkg/build/ignore/docker/dockign.go
```diff
@@ -0,0 +1,116 @@
...
+	"io"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/golang/glog"
+	"github.com/openshift/source-to-image/pkg/api"
+)
+
+type DOCKIGN struct {
```
soltysh: Keep naming consistent with rest of the system, so no UPPERCASE please.

soltysh:
Aside from @bparees comments, my biggest dislike is the placing and existence of the `Ignorer`. Generally speaking ignoring files is a job of a `Preparer` interface and the placing of that call in `Prepare` method is ok. But I'd move it to `util` package (`util/ignore.go`) with one `Ignore` function (without struct for it, because it's not needed) accepting just one parameter the directory to be processed. Then from inside the `Prepare` method you'll just call that method passing the `filepath.Join(config.WorkingDir, api.Source)`. This last step will keep the number of places dealing with our directories' names to minimum, I don't like these names being scattered around the entire code base. Lastly, the file name can be part of that method, so no need for `api.DefaultIgnoreFile`, as well.

README.md
bparees: +1 @soltysh, I actually debated suggesting "filtering" instead of "pruning" when I was reviewing this, but I didn't want to get too picky :)

soltysh: I'm in a different team now, I can be picky, besides I'm too far away to get my ass kicked for doing so :stuck_out_tongue_winking_eye:

gabemontero:
OK ... I'll probably start addressing this piece meal, since I think there is a range here from whether I agree or not.  I will say up front that since it was a new file, I somehow forgot to commit / push my unit test.  I just noticed this when doing a `git status`:

Untracked files:
  (use "git add <file>..." to include in what will be committed)

	pkg/build/ignore/docker/dockin_test.go

I'll push that as part of the other changes (I'll hold off initially until the directory / location comment from @soltysh is settled on.

pkg/build/ignore/docker/dockign.go
gabemontero: I explicitly made this UPPERCASE because "STI" was upper case, but in hindsight, that is a special case.  I will address.

hack/common.sh
gabemontero: Setting of STI_GOFLAGS unfortuanately did NOT work as I explained in the above comment.  Thought in looking at that comment, I see I failed to mention STI_GOFLAGS ;-) ... I'll update the comment.

My intent here was to ultimately update origin as well, but I wanted to try some more things within the less active source-to-image repository first.  As it is turning out, I'm having to do my gdb investigation in small spurts as I get new assignments (which still invariably becomes educational and take some time), but wanted to avoid having to copy in and copy out my local copy and hack/common.sh as I found time to revisit gdb.

If I add the additional disclaimer that this will either be a) duplicated in origin as time permits, or b) will remove if gdb proves non-useful for Go dev, are we good then with leaving this change in?

hack/common.sh
gabemontero: I am assuming you mean the 2nd echo ... I can remove that, and just leave the first echo that confirms that since the env is in fact set, we are employing -gcflags ...

README.md
gabemontero: Funny ... I was debating word choice here as well when I landed on "pruning". 

"Filtering" works for me.  Switching to that.

README.md
gabemontero: changed to:
 excluding any files selected by `.s2iignore`

pkg/git/git.go
gabemontero: Yes ... that was the intent.

hack/common.sh
bparees: i'm actually still a little confused by the comment...   
I gather that:
go install -gcflags "-N -1" ${goflags[@]:+${goflags[@]}}
doesn't work because everything after "-gcflags" gets treated as an argument to -gcflags (?)

but what about:
go install ${goflags[@]:+${goflags[@]}} -gcflags "-N -1" 

?

pkg/git/git.go
gabemontero: Piping only stderr made the hang less intermittent, but I in fact did produce a hang that lasted over 5 minutes before I aborted the `sti build ` invocation .... settling on the form in my prior comment for this thread.

pkg/build/ignore/docker/dockign.go
gabemontero: Yeah, these literals were used in clone.go a decent amount.  Part of my duplicating the pattern in dockign.go was to see if that was an oversight in prior reviews, or a toleration because it was in one file or something like that, where the review of this change would potentially flesh out the answer.

That said, the fact that filepath.Join, after the first argument, want to take single elements in a directory path (where the Join handles the separator) could also come into play here.  You would only build constants for "upload" and "src" and "tmp" (used in clone.go), not the combined paths.

I'll move on that approach, but can adjust if this particular comment thread progresses longer.

bparees: yeah I think where I was suggesting we go is something closer to:
filepath.Join(config.WorkingSourceDir,api.DefaultIgnoreFile)

where config.WorkingSourceDir gets built up in a single spot elsewhere.

pkg/build/ignore/docker/dockign.go
gabemontero: So you ack'ed the existence of an interface below.

On the api.Config struct passed in:  I did this because
a) this makes Ignore/Ignorer looking consistent with Build/Builder, Prepare/Preparer, and their associated new Methods
b) if later enhancements call for more information needed off of api.Config, we already have the entire object, and don't need to change the callers to Ignore to pass in more info.  Since we are passing a pointer to the struct, it should not be expensive as a parameter.

On splitting the Ignore method into get the list, remove files from the list:
a) I failed to check in my unit test :-), but it was not terribly difficult 
b) otherwise though, I contemplated the split you described, but the delete part is so trivial in the end from a lines of code perspective, I did not see the value and putting another frame on the stack.

All this said ... let's circle back here after we cross off some of the other items from the review (including getting my unit test checked in this time).

pkg/build/ignore/docker/dockign.go
gabemontero: part of next push

pkg/build/ignore/docker/dockign.go
gabemontero: part of next push

pkg/build/ignore/docker/dockign.go
bparees: Passing the struct in as an arg is fine (and doing it as a ptr definitely makes sense), my primary concern here was more with the fact that you had also defined a struct with a Config field, which isn't needed since you're passing the Config object in to the only method on the struct.  Defining the struct is fine (necessary since you're doing it as an interface and not a first class function, which as you note follows the Prepare pattern and that's fine with me)

So mostly I think this comes down to:
remove the Config field from the struct
remove the New() method

bparees: (and i'll withhold judgement on the splitting of the method into two parts until I see what the unit tests look like :) )

pkg/build/ignore/docker/dockign.go
gabemontero: Ben and I conferred ... we agreed this comment could be dismissed.

pkg/build/ignore/docker/dockign.go
gabemontero: Part of next push

pkg/build/ignore/docker/dockign.go
gabemontero: Part of next push

pkg/build/ignore/docker/dockign.go
gabemontero: part of next push

pkg/build/ignore/docker/dockign.go
gabemontero: part of next push

pkg/build/ignore/docker/dockign.go
gabemontero: gotcha ... will be part of next push

pkg/build/ignore/docker/dockign.go
gabemontero: will be part of next push

pkg/build/strategies/onbuild/onbuild_test.go
gabemontero: OK .. yeah, reworked this.  We are now back to this: `type fakeSourceHandler struct{}`

And `func newFakeOnBuild() *OnBuild ` is back to its prior form.

And then, so fakeSourceHandler properly implements build.SourceHandler, this method is added:

`
func (*fakeSourceHandler) Ignore(r *api.Config) error {
	return nil
}
`
Similar to the Prepare and Download methods for their associated interfaces.

pkg/build/ignore/docker/dockign.go
gabemontero: I realized the flaw in the current implementation after investigating your last question.  The write up on dockerignore is explicit about the search not being recursive.  Only looking at the last element, via the filepath.Base() call, allows a recursive search !!

I have a change in hand for the upcoming push which a) allows directories, b) allows"!foo/bar/myfile" and c) does not incorrectly apply recursion for !myfile

hack/common.sh
gabemontero: it was actually the opposite ...it wanted to treat -N and -l as flags to Go.

bparees: i'm still a little confused.  ignoring the actual script and STI_FLAGS env variable, is this not valid:
go install -gcflags "-N -1" -fooflag -barflag
?

I assume the arg to -gcflags needs to be quoted but given that, can't you set:
STI_FLAGS="-gcflags \"-N -1\""
?

i don't want to hold this PR up on this item, so if for now the simplest thing is "enabling debug means you don't get any other flags" i think we can live with it, but i am curious exactly what the issue is.  Probably something a 30 second conversation tomorrow can clarify.

pkg/build/ignore/docker/dockign.go
gabemontero: So I did a little refactoring and created a "get list of files to ignore method".
However, I expect a disconnect with unit test, as I am in fact temporarily creating files and cleaning them up (where as you pointed to avoiding having stuff on the filesystem)
I of course have a rationale for why actually interfacing with the file system is important here instead of the "fake" pattern (for lack of a better title).  In particular, vetting that the filespec.Glob call is returning what we want I believe is vital.  Faking out other pieces, but avoiding the use of Glob, seems insufficient, but perhaps I still need the veil pulled from my eyes here.

If possible let's stare at the screen together tomorrow once you've had a once over.

gabemontero:
So, I've pushed the next iteration, where I've largely enacted upon the suggestions from Ben and Maciej.

Minimally, where additional back and forth will occur is with:
1) The internals of the Ignore method and the unit tests not interacting with the files system.  I've added an additional response to Ben.  Hoping we can just stare at it together for a few and reconcile.
2) the hack/common.sh change ... either some repo's confirm the issue I saw and the need to not use STI_GOFLAGS, or the disconnect is solved ... also, if we can allow this to only exist in s2i for a bit before moving to origin
3) Maciej's last point on Prepare vs. Ignore, the package / subdir structure.  Acknowledging that "preparation" can include a ignore filter, I basically disagree perhaps at a more fundamental level here, as I started at the Prepare step and pretty quickly did not like the look of it.  I think Prepare is already possibly too broad and needs to be broken up into more precise actions, with the Ignore step presumably being one of them.  I think switching between git and docker ignore functionality and impls will be cleaner with this structure.  I *think* Ben called out in one of his comment an Ignorer interface as well, but there has been a lot of dialogue here, confirmation would be good.

Perhaps a common verbal conversation to hash out the Prepare / Ignore distinction is warranted.

hack/common.sh
soltysh: I got struck by a line from `go help build`, which says:
```
-gccgoflags 'arg list' 
        arguments to pass on each gccgo compiler/linker invocation.
```
and I tried `export STI_GOFLAGS="-gcflags '-N -l'"`, works like a charm, without any modifications.

soltysh:
I'm still not convinced with your argument. All the interfaces in `build` package represent a step in a build process, which IMHO the `Ignorer` does not represent, I'm still seeing this as part of the `Preparer` interface. But if you insist on having it there you'd have to follow the same usage patterns the other build interfaces do (even though the onbuild is actually using sti's Prepare). Iow. each of these interfaces is implemented by actual builder, which means you'd have to move the logic into sti builder actually. 
My longer term goal here is to completely abstract the actual work of the builder, so that the actual build process is done in a couple of steps, eg.:
1. prepare
2. build
3. clean
where each of these steps is represented by an interface and each builder is responsible for implementing those. This way we'll have clean flow of the build process, which currently is not always that obvious and it'll be easier to implement in origin, as stated in #206.

bparees:
@soltysh I need to review @gabemontero 's latest push, but fundamentally I don't think it makes sense for every builder implementation to provide its own implementation of common filesystem manipulation methods (which is basically what Ignore amounts to).

it should be an interface so that each builder impl *can* choose which (existing) Ignore implementation it wants, but they shouldn't be forced to provide their own implementation.

soltysh:
@bparees that's completely fine by me, but then again it deserves it's own package, as `build` is not the right place for it. Besides I not a big fan of `ignore/docker/dockign.go`, I'd prefer having a clean `ignore/ignore.go` :wink: It's our ignore, why suggesting any docker relationship here.

bparees:
@soltysh agreed on the packaging and lack of love for "dockign" as a name.

soltysh:
Thank you, big kofola + vinea for you :heart:

hack/common.sh
bparees: so are you going to change the behavior to accept STI_GOFLAGS args even when debugging?

soltysh: No, I've meant that there's no need to change anything with above. It's just a matter of passing the arguments in the right way.

gabemontero:
Thanks for the feedback folks.

On the name and the docker relationship, we starting by touting we were providing ".dockerignore" level functionality(vs. what is provided with .gitignore for example).  I envisioned even pointing to the spot in the docker docs.  The opener of the issue even noted that .dockerignore functionality was sufficient for now.

Now, on "dockign", and the package names, yes, I was not as confident on those with my first stab here.  And I had the initial preconception that it *had* to go under build.  Good to know that is not the case.

How about "pkg/ignore/docker.go" for the initial ".dockerignore" implementation, and if we ever supply ".gitignore" level functionality, a "pkg/ignore/git.go" ?

thanks again,
gabe

soltysh:
> How about "pkg/ignore/docker.go" for the initial ".dockerignore" implementation, and if we ever supply ".gitignore" level functionality, a "pkg/ignore/git.go" ?

No, you'd name it `docker.go` if it'd support `.dockerignore` file, which it does not. The same for `.gitignore` you'd call it `git.go`. But since this is implementing `.s2iingore` then naming it with `ignore.go` is perfectly fine.

hack/common.sh
bparees: i find it helpful to have an easy enable debug flag rather than having to remember "-gcflags '-N -l'", but i do agree that it should still respect other args provided by the user.

soltysh: Well, I'm not too deep into gdb stuff, but wouldn't be that for one folks these flags are enough, but for other different set of flags is needed? Let me quote Dan :wink: is that 80/20 use case?

gabemontero:
OK - yes, not the file literally, but the level of functionality.  There is still a connection that from my understanding we wanted to be explicit about, but not literally reuse the .dockerignore file to give us *some* protection.

pkg/build/ignore/docker/dockign.go
```diff
@@ -0,0 +1,135 @@
...
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/golang/glog"
+	"github.com/openshift/source-to-image/pkg/api"
+	"fmt"
+)
+
+type DockIgn struct {
```
bparees: type DockerIgnorer struct {}

pkg/build/ignore/docker/dockign.go
```diff
@@ -0,0 +1,135 @@
...
+}
+
+func getListOfFilesToIgnore(config *api.Config) (map[string]string, error) {
+	path := filepath.Join(config.WorkingSourceDir, api.DefaultIgnoreFile)
+	file, err := os.Open(path)
+	if err != nil {
+		if !os.IsNotExist(err) {
+			glog.Errorf("Ignore processing, problem opening %s because of %v\n", path, err)
+			return nil, err
+		} else {
```
bparees: no need for this else block since you're returning, just put the return outside the if.

@smarterclayton i hope you're proud of yourself for what you've turned me into.

soltysh: That's good, why should anyone complain. We're getting better and better with each and every single review and PR :smile:

pkg/build/ignore/docker/dockign_test.go
```diff
@@ -0,0 +1,112 @@
...
+	"testing"
+
+	"github.com/openshift/source-to-image/pkg/api"
+	"os"
+)
+
+func newIgnorer() *DockIgn {
+	i := &DockIgn{
+	}
+	return i
```
bparees: not sure you need this helper method at all, but if you do:
return &DockIgn{}

pkg/build/ignore/docker/dockign_test.go
```diff
@@ -0,0 +1,112 @@
...
+}
+
+func baseTest(t *testing.T, patterns []string, filesToCheck []string) {
+	cleanerr := os.RemoveAll("/tmp/upload")
```
bparees: use a generated temp directory, this is too risky to collide.

pkg/build/ignore/docker/dockign_test.go
```diff
@@ -0,0 +1,112 @@
...
+	if cleanerr != nil {
+		t.Errorf("problem cleaning up %v \n", cleanerr)
+	}
+	
+	c := &api.Config{WorkingDir: "/tmp"}
```
bparees: again generated directory.

pkg/build/ignore/docker/dockign_test.go
```diff
@@ -0,0 +1,112 @@
...
+	dpath := filepath.Join(c.WorkingDir,"upload")
+	derr := os.Mkdir(dpath,0777)	
+	if derr != nil {
+		t.Errorf("Problem creating test dir %s with  %v \n", dpath, derr)
+	}
+	dpath = filepath.Join(dpath,"src")
+	derr = os.Mkdir(dpath,0777)
+	if derr != nil {
+		t.Errorf("Problem creating test dir %s with  %v \n", dpath, derr)
+	}
```
bparees: https://golang.org/pkg/os/#MkdirAll instead of building up the path?

pkg/build/ignore/docker/dockign_test.go
```diff
@@ -0,0 +1,112 @@
...
+	c.WorkingSourceDir = dpath
+	t.Logf("working source dir %s \n", dpath)
+
+	// create s2iignore file
+	ipath := filepath.Join(dpath, api.DefaultIgnoreFile)
+	ifile, ierr := os.Create(ipath)
+	defer ifile.Close()
+	if ierr != nil {
+		t.Errorf("Problem creating .s2iignore at %s with  %v \n",ipath, ierr)
+	}
```
bparees: even better than doing all this might be to create a fixtures dir with the entire directory structure (including .s2iignore) that you want.  then you could just cp -r the fixtures dir and run, rather than programmatically constructing the directory structure.

bparees: though I can see some value in doing it the way you did since you're creating both the file, and the expected result, simultaneously.

pkg/build/ignore/docker/dockign_test.go
```diff
@@ -0,0 +1,112 @@
...
+	baseTest(t, []string{"foo.bar\n"}, []string{"foo.bar"})
+}
+
+func TestWildcardIgnore(t *testing.T) {
+	baseTest(t, []string{"foo.*\n"}, []string{"foo.a", "foo.b"})
+}
+
+func TestExclusion(t *testing.T) {
+	baseTest(t, []string{"foo.*\n", "!foo.a"}, []string{"foo.a","foo.b"})
+}
```
bparees: how about a test for files in subdirectories?
tests to ensure files not mentioned (either as include or excludes) at all are left around?

tests for ordering of !  (if file is mentioned before ! it gets retained, if file is mentioned after !, it is deleted, right?)

! with wildcard?

this is where it might be easer to create a couple test directory structures and just run your code against them, rather than building it programmatically.

pkg/build/strategies/sti/sti_test.go
```diff
@@ -59,6 +62,7 @@ func newFakeSTI(f *FakeSTI) *STI {
 		fs:        &test.FakeFileSystem{},
 		tar:       &test.FakeTar{},
 		preparer:  f,
+		ignorer: i,
```
bparees: &docker.DockIgn{} and get rid of the "i" above.

pkg/git/git.go
```diff
@@ -66,20 +66,20 @@ func (h *stiGit) ValidCloneSpec(source string) bool {
...
+	// the pipeToLog method has been left for now for historical purposes, but if this implemenetation
+	// of git clone holds, we'll want to delete that at some point.
+	
 	opts := util.CommandOpts{
-		Stdout: outWriter,
-		Stderr: errWriter,
 	}
-	go pipeToLog(outReader, glog.Info)
-	go pipeToLog(errReader, glog.Error)
+
```
bparees: so just to confirm... if you invoke sti now with an invalid git clonespec, you still see the error, but it's not wrapped in timestamp/etc by glog?

bparees:
i'm fine with either pkg/ignore/ignore.go or pkg/ignore/docker.go, I guess i have a slight preference for pkg/ignore/ignore.go since it's more a convention that it matches the dockerignore behavior, than an indication that it's somehow related to docker.

and we can always put other implementations in that same file, they don't have to go into their own named files.

gabemontero:
renamed to pkg/ignore/ignore.go in upcoming push.

pkg/build/ignore/docker/dockign.go
gabemontero: will be in next push

pkg/build/ignore/docker/dockign.go
gabemontero: yep, no complaints here - part of next push

pkg/build/ignore/docker/dockign_test.go
gabemontero: got rid of helper, part of next push

pkg/build/ignore/docker/dockign_test.go
gabemontero: part of next push

pkg/build/ignore/docker/dockign_test.go
gabemontero: part of next push

pkg/build/ignore/docker/dockign_test.go
gabemontero: part of next push

pkg/build/ignore/docker/dockign_test.go
gabemontero: I'll leave for now then ... I'm making a note to circle back and at least make sure I'm interpreting your suggestion correctly.

pkg/git/git.go
gabemontero: correct ... given I'm so far the only one on the team seeing the hang, we can still switch back to the old implementation, and I'll just fix it manually when I use sti.

In case I'm correct in sensing your apprehension here ...

pkg/build/strategies/sti/sti_test.go
gabemontero: part of next push

gabemontero:
I'll remove the hack/common.sh change, at least for now.  Next time I'm able to experiment with gdb, I'll see if I see the same error using STI_GOFLAGS, and if I do, will just email @soltysh directly to sort through his successful usage vs. my results if they are still negative.

pkg/build/ignore/docker/dockign.go
smarterclayton: Your tears give me strength

On Jul 28, 2015, at 10:22 AM, Ben Parees <notifications@github.com> wrote:

In pkg/build/ignore/docker/dockign.go
<https://github.com/openshift/source-to-image/pull/248#discussion_r35652714>
:

> +			glog.Errorf("error removing file %s because of %v \n", fileToDel, rerr)
> +			return rerr
> +		}
> +	}
> +
> +	return nil
> +}
> +
> +func getListOfFilesToIgnore(config *api.Config) (map[string]string, error) {
> +	path := filepath.Join(config.WorkingSourceDir, api.DefaultIgnoreFile)
> +	file, err := os.Open(path)
> +	if err != nil {
> +		if !os.IsNotExist(err) {
> +			glog.Errorf("Ignore processing, problem opening %s because of %v\n", path, err)
> +			return nil, err
> +		} else {

no need for this else block since you're returning, just put the return
outside the if.

@smarterclayton <https://github.com/smarterclayton> i hope you're proud of
yourself for what you've turned me into.

—
Reply to this email directly or view it on GitHub
<https://github.com/openshift/source-to-image/pull/248/files#r35652714>.

pkg/build/ignore/docker/dockign_test.go
gabemontero: added tests which map to the additions you noted (if I read your ordering one right, you have it backwards ... I put in a comment prior to the new test which has an example from the docker user manual which hopefully clears it up)

I ended up tweaking baseTest somewhat as a result, but I think the net effect is that it is cleaner than before.  I'm inclined on keeping the overall format, but we can have a quick side bar if need be to discuss further.

gabemontero:
OK ... latest changes based on comments pushed.

pkg/build/ignore/docker/dockign_test.go
bparees: my understanding was that if my file contained:

a*
!ab

then "ab" would be kept/retained/not ignored.

whereas:
!ab
a*

would result in ab being ignored/deleted/not retained.

gabemontero: OK cool that example clears up my interpretation of your take, and that is how the docker user guide explains it.

pkg/build/strategies/sti/sti_test.go
```diff
@@ -59,6 +60,7 @@ func newFakeSTI(f *FakeSTI) *STI {
 		fs:        &test.FakeFileSystem{},
 		tar:       &test.FakeTar{},
 		preparer:  f,
+		ignorer: &ignore.DockerIgnorer{},
```
bparees: this should be aligned, are you gofmt'ing your code?

pkg/git/git.go
bparees: i'm a little apprehensive but of course a bunch of openshift users getting hung builds would be worse.

have you confirmed you're on the latest go and docker versions?

pkg/ignore/ignore.go
```diff
@@ -0,0 +1,128 @@
...
+	"io"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/golang/glog"
+	"github.com/openshift/source-to-image/pkg/api"
+)
+
+type DockerIgnorer struct {
```
bparees: single line

pkg/ignore/ignore_test.go
```diff
@@ -0,0 +1,227 @@
...
+
+			// if here, then file which should still exist has been deleted
+			t.Errorf("file which should exist has been deleted %s \n", fileToCheck)
+			
+		}
+		
+	}
+
+	// clean up test
+	cleanerr := os.RemoveAll(workingDir)
```
bparees: should probably be in a defer

pkg/ignore/ignore_test.go
```diff
@@ -0,0 +1,227 @@
...
+func TestSingleIgnore(t *testing.T) {
+	baseTest(t, []string{"foo.bar\n"}, []string{"foo.bar"}, []string{})
+}
+
+func TestWildcardIgnore(t *testing.T) {
+	baseTest(t, []string{"foo.*\n"}, []string{"foo.a", "foo.b"}, []string{})
+}
+
+func TestExclusion(t *testing.T) {
+	baseTest(t, []string{"foo.*\n", "!foo.a"}, []string{"foo.a","foo.b"}, []string{})
```
bparees: why isn't foo.a kept in this case?

pkg/ignore/ignore_test.go
```diff
@@ -0,0 +1,227 @@
...
+Per the docker user guide, with a docker ignore list of:
+
+    LICENCSE.*
+    !LICENCSE.md
+    *.md
+
+LICENSE.MD will be kept, as *.md overrides !LICENCSE.md
+*/
+func TestExcludeOverride(t *testing.T) {
+	baseTest(t, []string{"LICENCSE.*\n", "!LICENCSE.md\n", "*.md"}, []string{"LICENCSE.foo", "LICENCSE.md"}, []string{"foo.bar"})
```
bparees: you're passing LICENSE.md as a file to be deleted but the comment expects it to be kept...

i think ultimately the deletion is right (*.md wins in the end), so just the comment is wrong.

still not sure you have a test for the reverse order though:
!LICENSE.md
LICENSE.*
result: - LICENSE.md should be deleted.

this case doesn't quite cover it because it's impossible to know if LICENSE.* improperly resulted in deleting LICENSE.md (incorrectly overrode the !LICENCE.md behavior), or *.md properly resulted in deleting LICENCE.md

pkg/build/strategies/sti/sti_test.go
gabemontero: Yes, I have a hook in emacs which runs gofmt whenever I save a Go file.  I'll re-open this file and try to explicitly run gofmt.

gabemontero: Explicitly calling gofmt added a copy of spaces.  I'll push in a bit.

pkg/ignore/ignore.go
gabemontero: in next push

pkg/ignore/ignore_test.go
gabemontero: in next push

pkg/ignore/ignore_test.go
gabemontero: foo.* occurs first in .s2iignore, and removes foo.a and foo.b.  However, !foo.a follows foo.*, and so excludes that ignore/removal, and so it is kept.

pkg/ignore/ignore_test.go
gabemontero: Uggh ... the comment is missing the NOT.  My bad - and I was trying to clarify things ;-)

It doesn't make sense to have an exclusion first before any patterns that delete files, the exclusion has no candidates to add back in, but sure, if the user does something by mistake, let's make sure the system does not bomb.

I'll add the 2 liner you propose, along with the NOT in the comment

pkg/git/git.go
gabemontero: I yum update everything at least once a week ... that takes care of docker.

I've done `go get -u all` maybe 2 or 3 times since I started, but not as frequently, so I just did one now.
Added the glog stuff back to git.go, still hangs.

gabemontero:
ok ... latest push includes responses to Ben from Tuesday evening.

pkg/ignore/ignore_test.go
bparees: right.  but you have foo.a in the delete list, so it is NOT being kept.

bparees: (or rather, the test expects it to be not kept, and i'm assuming this test passes, so it's not being kept)

gabemontero: As it turns out, even with the last refactoring of the test, ignores still don't have to be in the filesToKeep list passed into baseTest(). They were still special cased like previously done.

Some of the relevant trace when you run the test:
--- PASS: TestExclusion (0.00s)
	ignore_test.go:89: storing pattern !foo.a 
	ignore_test.go:97: Noting ignore pattern  foo.a 
	ignore_test.go:150: The file /tmp/sti861017162/upload/src/foo.a exists after Ignore was run 
	ignore_test.go:165: validated ignored file is still present foo.a 

While the end result of the function working is there, I acknowledge it is inconsistent with the filesToDelete / filesToKeep distinction in the function signature.

I'll address.

gabemontero:
ignore_test.go pushed to honor ignores under filesToKeep

pkg/api/types.go
```diff
@@ -5,6 +5,10 @@ import docker "github.com/fsouza/go-dockerclient"
 const (
 	DefaultNamespace    = "io.openshift.s2i."
 	KubernetesNamespace = "io.k8s."
+	DefaultIgnoreFile   = ".s2iignore"
+	DefaultUploadSubdir = "upload"
+	DefaultSrcSubdir = "src"
```
soltysh: Replace above two constants with `Source` from `api/scripts.go` and add there `upload/tmp` as well. And use those in the appropriate places in the code. I don't want to have multiple constants pointing to the same stuff.

pkg/git/clone.go
```diff
@@ -17,12 +17,13 @@ type Clone struct {
...
 func (c *Clone) Download(config *api.Config) (*api.SourceInfo, error) {
-	targetSourceDir := filepath.Join(config.WorkingDir, "upload", "src")
+	targetSourceDir := filepath.Join(config.WorkingDir, api.DefaultUploadSubdir, api.DefaultSrcSubdir)
+	config.WorkingSourceDir = targetSourceDir
 	var info *api.SourceInfo
 
 	if c.ValidCloneSpec(config.Source) {
 		if len(config.ContextDir) > 0 {
-			targetSourceDir = filepath.Join(config.WorkingDir, "upload", "tmp")
+			targetSourceDir = filepath.Join(config.WorkingDir, api.DefaultUploadSubdir, api.DefaultTmpSubdir)
```
soltysh: Here user your new constant added to `api/scripts.go`.

pkg/git/clone.go
```diff
@@ -38,7 +39,7 @@ func (c *Clone) Download(config *api.Config) (*api.SourceInfo, error) {
 		}
 
 		if len(config.ContextDir) > 0 {
-			originalTargetDir := filepath.Join(config.WorkingDir, "upload", "src")
+			originalTargetDir := filepath.Join(config.WorkingDir, api.DefaultUploadSubdir, api.DefaultSrcSubdir)
```
soltysh: Use `Source` from `api/scripts.go`.

pkg/git/clone.go
```diff
@@ -17,12 +17,13 @@ type Clone struct {
 // Download downloads the application source code from the GIT repository
 // and checkout the Ref specified in the config.
 func (c *Clone) Download(config *api.Config) (*api.SourceInfo, error) {
-	targetSourceDir := filepath.Join(config.WorkingDir, "upload", "src")
+	targetSourceDir := filepath.Join(config.WorkingDir, api.DefaultUploadSubdir, api.DefaultSrcSubdir)
```
soltysh: Use `Source` from `api/scripts.go`.

pkg/git/git.go
soltysh: I'm also bit worried about this. I've never had any problems with git hanging, it was rather further down the process the hangs happened, somewhere towards the actual docker run, where we run assemble (iirc from #149). The reason we've redirected output to glog was to have one central place we handle logging. This change of yours breaks that rule, iow. git will write on its own to stderr/stdout which is not desirable. I'd say we need to pay more attention to it, especially that in some time we'll be reworking logging to allow omitting the datetime part from the log see  https://github.com/openshift/origin/issues/1647, and to allow proper redirection of logs (#156).

pkg/ignore/ignore.go
```diff
@@ -0,0 +1,127 @@
...
+	"io"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/golang/glog"
+	"github.com/openshift/source-to-image/pkg/api"
+)
+
+type DockerIgnorer struct {}
```
soltysh: I'm still not in favor of this naming, as I've pointed earlier, I'd call this just `Ignorer`, `DockerIgnorer` I'd reserve for the case where we'd actually support `.dockerignore` file, but it's up to you @bparees .

pkg/ignore/ignore.go
```diff
@@ -0,0 +1,127 @@
...
+		if rerr != nil {
+			glog.Errorf("error removing file %s because of %v \n", fileToDel, rerr)
+			return rerr
+		}
+	}
+
+	return nil
+}
+
+func getListOfFilesToIgnore(config *api.Config) (map[string]string, error) {
```
soltysh: No need to pass the entire `config` structure, if all you care about is `WorkingSourceDir`. Make each method as small as possible, it'll also simplify the testing.

pkg/ignore/ignore.go
```diff
@@ -0,0 +1,127 @@
...
+		glog.V(4).Info(".s2iignore file does not exist")
+		return nil, nil
+	}
+	defer file.Close()
+
+	filesToDel := make(map[string]string)
+	scanner := bufio.NewScanner(file)
+	for scanner.Scan() {
+		filespec := scanner.Text()
+		filespec = strings.Trim(filespec, " ")
```
soltysh: Merge these two lines into:
```
filespec := strings.Trim(scanner.Text(), " ")
```

pkg/ignore/ignore.go
```diff
@@ -0,0 +1,127 @@
...
+		glog.V(4).Infof(".s2iignore lists a file spec of %s \n", filespec)
```
soltysh: I'd move this log after the next if which is omitting comments.

pkg/ignore/ignore_test.go
```diff
@@ -0,0 +1,248 @@
...
+		if getLogLevel() != 5 {
+			vflag.Value.Set("5")
+			t.Logf("integration_test.go: glog level now %d\n", getLogLevel())
+			// FIXME currently glog has only option to redirect output to stderr
+			// the preferred for STI would be to redirect to stdout
+			flag.CommandLine.Set("logtostderr", "true")
+		}
+	}
+        */
+	// end glog enablement
```
soltysh: Don't leave commented out code. Remove it, please.

soltysh:
Two final remarks I have:

1. I'd copy, at least the crucial part, of how the `.s2iignore` will work. We don't want to point them to external URLs. Especially if there's a chance that it might change, then the sentence that it worked at the time of implementation isn't one a user expects.
1. I'd generally split `Ignore` internally into two methods, one gathering files for deletion and 2nd that actually performs deletion. This way it'll be easier for you to write tests, imho.

pkg/api/types.go
gabemontero: Was unaware of the constants in scripts.go.  Thanks for the pointer, will be part of next push.

Quick follow up:  the use of "/" in those constants in scripts.go - are we concerned about running sti on windows?  Using filepath.Join is suppose to take care of "/" vs. "\"

pkg/git/clone.go
gabemontero: part of next push

pkg/git/clone.go
gabemontero: part of next push

pkg/git/clone.go
gabemontero: part of next push

pkg/ignore/ignore.go
gabemontero: @bparees and I discussed this previously in this PR.  I advocated for the `config` structure so the method signature looks more consistent with Build and Prepare, and allows for use of additional fields from `config` in the future without having to change the calls into Ignore.  As of the last discussion, we kept the parameter as is.

pkg/ignore/ignore.go
gabemontero: part of next push

pkg/ignore/ignore.go
gabemontero: OK ... I was originally interested in confirming via trace that "#" were skipped, but yeah, at this point, that is confirmed by other means.  Part of next push.

pkg/ignore/ignore_test.go
gabemontero: Interesting, if a general convention, commented out code can have informative value.  But I admit in this case it was just an ease of use thing, in case I wanted to reuse this logic.  I'll remove as  part of next push.

gabemontero:
push for the in-line code comments from @soltysh this AM done.

Document update will come subsequently.

pkg/api/types.go
soltysh: There's no docker on windows as of now, so for now we're cool :sunglasses:

pkg/ignore/ignore.go
soltysh: Your call.

gabemontero:
push for doc update as motivated by @soltysh done, which removes the explicit citation of .dockerignore as what we are mirroring, and provides the major points on how it works.

README.md
```diff
@@ -40,6 +40,35 @@ image:
...
+Filtering the contents of the source tree is possible if the user supplies a
+`.s2iignore` file in the root directory of the source repository, where `.s2iignore` contains regular
+expressions that capture the set of files and directories you want filtered from the image s2i produces.
+
+Specifically:
+
+1. Specify one rule per line, with each line terminating in `\n`.
+1. Filepaths are appended to the absolute path of the  root of the source tree (either the local directory supplied, or the target destination of the clone of the remote source repository s2i creates).
+1. Wildcards and globbing (file name expansion) leverage Go's `filepath.Match` and `filepath.Glob` functions.
+1. Search is not recursive.  Subdirectory paths must be specified (though wildcards and regular expressions can be used in the subdirectory specifications).
```
soltysh: Provide examples here.

gabemontero: examples added - doc pushed

soltysh:
LGTM

pkg/ignore/ignore.go
bparees: let's leave it, if we come back and add .dockerignore support we'd probably want to refactor this in such a way that it takes the input file as an argument/config value rather than duplicating the whole thing anyway (since the logic would be the same)

pkg/ignore/ignore_test.go
bparees: +1 @gabemontero, i don't think there's a hard and fast rule about always removing commented out code, though i agree it's a fine line that can quickly turn into useless clutter.

pkg/api/types.go
bparees: i've never bothered to learn....  does windows have an equivalent to osx boot2docker?  and when you're using it, is it transparent (eg you get a docker daemon in your host that proxies into the VM)?

if so then we would need to worry about this.

looks like yes: https://docs.docker.com/installation/windows/

but i have no idea if the docker client libraries we are using would work in that environment.  still, probably better to not do it so we don't have a problem later.

gabemontero: OK, I've updated the constants in `scripts.go` to leverage os.PathSeparator and verified in go playground, as well as test-go.sh and test-integration.sh.  Will push in a couple.

gabemontero:
push for path separator with constants done.

bparees:
lgtm
[merge]

gabemontero:
cool deal ... thanks @bparees and @soltysh

gabemontero:
@bparees please [merge] changes as a result from gofmt errors cited by verify-gofmt.sh

bparees:
[merge]

soltysh:
@gabemontero thanks for doing this and patience :wink:

PR #249, opened 2015-07-27
Some fixes for the test script template

vbehar:
A few fixes for the "test" script template : 
- support for running on OSX
- fix incremental build
- more logs

### DIFF
--- a/pkg/create/templates/test.go
+++ b/pkg/create/templates/test.go
@@ -14,14 +14,22 @@ const TestRunScript = `
 #
 IMAGE_NAME=${IMAGE_NAME-{{.ImageName}}-candidate}
 
-test_dir="$(readlink -zf $(dirname "${BASH_SOURCE[0]}"))"
-image_dir=$(readlink -zf ${test_dir}/..)
+# Determining system utility executables (darwin compatibility check)
+READLINK_EXEC="readlink"
+MKTEMP_EXEC="mktemp"
+if (echo "$OSTYPE" | egrep -qs 'darwin'); then
+  ! type -a "greadlink" &>"/dev/null" || READLINK_EXEC="greadlink"
+  ! type -a "gmktemp" &>"/dev/null" || MKTEMP_EXEC="gmktemp"
+fi
+
+test_dir="$($READLINK_EXEC -zf $(dirname "${BASH_SOURCE[0]}"))"
+image_dir=$($READLINK_EXEC -zf ${test_dir}/..)
 scripts_url="file://${image_dir}/.sti/bin"
-cid_file=$(mktemp -u --suffix=.cid)
+cid_file=$($MKTEMP_EXEC -u --suffix=.cid)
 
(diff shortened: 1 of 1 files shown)


pkg/create/templates/test.go
```diff
@@ -106,7 +114,7 @@ test_connection() {
   local result=1
   while [ $attempt -le $max_attempts ]; do
     echo "Sending GET request to http://$(container_ip):${test_port}/"
-    response_code=$(curl -s -w %{http_code} -o /dev/null http://$(container_ip):${test_port}/)
+    response_code=$(docker run --rm -it sequenceiq/alpine-curl curl -s -w %{http_code} -o /dev/null http://$(container_ip):${test_port}/)
```
bparees: this seems a little heavy... is curl not available on osx?

vbehar: yes it is, but on osx if using boot2docker, the container IP is not directly accessible. thus the use of an container, to be in the same network.

vbehar: if you want, I can add a check to ping the IP for example, and use the container to curl only if it is not accessible

bparees: I see.  no i think you can leave it, but leave the original line there, commented out, with a comment about why it's being done that way, so people who don't care can easily revert the behavior.

vbehar: ok

pkg/create/templates/test.go
```diff
@@ -106,7 +114,9 @@ test_connection() {
   local result=1
   while [ $attempt -le $max_attempts ]; do
     echo "Sending GET request to http://$(container_ip):${test_port}/"
-    response_code=$(curl -s -w %{http_code} -o /dev/null http://$(container_ip):${test_port}/)
+    # here, we run the curl inside a container for people running on boot2docker (who can't access the container's IP directly)
+    #response_code=$(curl -s -w %{http_code} -o /dev/null http://$(container_ip):${test_port}/)
+    response_code=$(docker run --rm -it sequenceiq/alpine-curl curl -s -w %{http_code} -o /dev/null http://$(container_ip):${test_port}/)
```
soltysh: I'm a bit worried this introduces yet another dependency `sequenceiq/alpine-curl` which is not desired in tests, imho.

vbehar: I understand (even though it's just a template, so people are free to change it before running it).
what about reverting to the original behaviour, and just displaying a warning to osx users suggesting them to run the curl inside a container if they can't access the container's ip directly ?

soltysh: That will be perfect!

soltysh:
Btw. please squash all your commits into one and after you fix that last issue I'll merge this as is.

vbehar:
ok, fixed and squashed

soltysh:
@vbehar thanks for doing this. LGTM
[merge]

PR #250, opened 2015-07-27
sti -> s2i rebranding

### LINKED ISSUE #244: rename sti binary to s2i
Need to 
- [ ] have build produce an s2i binary instead of sti binary
- [ ] have build produce a symlink sti that points to s2i
- [ ] add logic to the CLI processing so that if the binary is invoked with the "sti" name, print a deprecation warning
- [ ] update remaining doc to say s2i.
- [ ] update image tests to use "s2i" instead of "sti" invocation.

soltysh:
This is on going work to rebrand `sti` to `s2i` CLI.
Fixes #244.


bparees:
thanks for doing this @soltysh!

soltysh:
That covers:
* [x] have build produce an s2i binary instead of sti binary
* [x] have build produce a symlink sti that points to s2i
* [x] add logic to the CLI processing so that if the binary is invoked with the "sti" name, print a deprecation warning
* [x] update remaining doc to say s2i

TDB after this gets merged:
* [ ] update image tests to use "s2i" instead of "sti" invocation.

Future changes (in that particular order):
* [ ] change `.sti/` directory to `.s2i/` directory
* [ ] change `/tmp/sti/` to `/tmp/s2i` or other not using `/tmp` directory
* [ ] change internals to use s2i everywhere

@bparees ptal did I miss something that's needed right now? I think I've changed everything user-facing.

hack/common.sh
```diff
@@ -266,6 +284,16 @@ sti::build::place_bins() {
   )
 }
 
+# sti::build::make_binary_symlinks makes symlinks for the sti
+# binary in _output/local/go/bin
+sti::build::make_binary_symlinks() {
+  if [[ -f "${STI_LOCAL_BINPATH}/s2i" ]]; then
+    for linkname in "${STI_BINARY_SYMLINKS[@]}"; do
+      ln -sf "${STI_LOCAL_BINPATH}/s2i" "${STI_LOCAL_BINPATH}/${linkname}"
```
bparees: no windows checking here?

pkg/create/create.go
```diff
@@ -22,11 +22,11 @@ func New(name, dst string) *Bootstrap {
...
-	os.MkdirAll(b.DestinationDir+"/"+".sti/bin", 0700)
-	b.process(templates.AssembleScript, ".sti/bin/assemble", 0755)
-	b.process(templates.RunScript, ".sti/bin/run", 0755)
-	b.process(templates.UsageScript, ".sti/bin/usage", 0755)
-	b.process(templates.SaveArtifactsScript, ".sti/bin/save-artifacts", 0755)
+	os.MkdirAll(b.DestinationDir+"/"+".s2i/bin", 0700)
+	b.process(templates.AssembleScript, ".s2i/bin/assemble", 0755)
+	b.process(templates.RunScript, ".s2i/bin/run", 0755)
+	b.process(templates.UsageScript, ".s2i/bin/usage", 0755)
+	b.process(templates.SaveArtifactsScript, ".s2i/bin/save-artifacts", 0755)
```
bparees: did you mean to change these?  i wouldn't be totally against changing this path (though we'd have to add backwards compatibility) but it would be a bit painful.

I didn't notice if your edits changed this path in the docs too.

bparees:
couple small comments (looks like you did *not* mean to update the create logic, based on your checklist, so you'll need to fix that).

would like a second pair of eyes but otherwise lgtm.   now I owe *you* a vinea.

pkg/create/create.go
soltysh: Actually these directories are used only inside the `Dockerfile`. Although the name is similar to `.sti/` directory that can be used for passing in s2i scripts from withing application source these serve different purpose. I was debating about whether to change it or not, but in the end I've decided to go with it. Maybe we should change that directory name to be without a dot at the beginning?

hack/common.sh
soltysh: I got nasty here by copying the code used in [origin](https://github.com/openshift/origin/blob/master/hack/common.sh#L336-L342). I'm not even sure how to create windows symlinks, do you think we should care, knowing the fact that docker does not run on windows natively, as of yet at least?

pkg/create/create.go
bparees: this is fine as is then.

hack/common.sh
bparees: i don't particularly care but you did it everywhere else so i was surprised.  last i knew linking in windows required third party tooling (though the filesystem supported it).  but you could always make a copy of the binary instead of invoking ln, if it's windows.

soltysh:
Rebased, @bparees waiting your singoff.

soltysh:
@bparees fixed the links for windows to be copies.

mfojtik:
i would suggest to keep the `sti` binary (as a symlink) around for a while with a deprecation warning to switch to use `s2i`. I don't want to break people that already use `sti` as part of their CI.

soltysh:
@mfojtik that's exactly what we do. When creating release binaries we
produce two binaries named sti and s2i, the former prints exactly that
warning.  When doing local dev we create symlink so that you can use the
old version,  again with deprecation warning. And yes we'll keep it there
for a while :-)

bparees:
lgtm.     ready for merge?

soltysh:
yup, yup [merge]

PR #255, opened 2015-08-10
Adding build labels

jhadvig:
Adding labels that contain information about the build source.
@bparees @soltysh PTAL

This is the first part of the 'Adding build labels' effort.

Here is the second part: https://github.com/openshift/origin/pull/4079

### DIFF
--- a/pkg/api/types.go
+++ b/pkg/api/types.go
@@ -62,6 +62,9 @@ type Config struct {
 	// variables.
 	EnvironmentFile string
 
+	// LabelNamespace provides the namespace under which the labels will be generated.
+	LabelNamespace string
+
 	// CallbackURL is a URL which is called upon successful build to inform about that fact.
 	CallbackURL string
 
--- a/pkg/util/labels.go
+++ b/pkg/util/labels.go
@@ -1,49 +1,63 @@
 package util
 
 import (
-	"strings"
-
 	"github.com/golang/glog"
 	"github.com/openshift/source-to-image/pkg/api"
 )
 
-// GenerateOutputImageLabels generate the labels based on the source repository
-// informations.
+// GenerateOutputImageLabels generate the labels based on the s2i Config
+// and source repository informations.
 func GenerateOutputImageLabels(info *api.SourceInfo, config *api.Config) map[string]string {
-	result := map[string]string{}
+	labels := map[string]string{}
+	namespace := api.DefaultNamespace
+	if len(config.LabelNamespace) > 0 {
+		namespace = config.LabelNamespace
+	}
 
+	labels = GenerateLabelsFromConfig(labels, config, namespace)
+	labels = GenerateLabelsFromSourceInfo(labels, info, namespace)
+	return labels
+}
+
+// GenerateLabelsFromConfig generate the labels based on build s2i Config
+func GenerateLabelsFromConfig(labels map[string]string, config *api.Config, namespace string) map[string]string {
 	if len(config.Description) > 0 {
-		result[api.KubernetesNamespace+"description"] = config.Description
+		labels[api.KubernetesNamespace+"description"] = config.Description
 	}
 
 	if len(config.DisplayName) > 0 {
-		result[api.KubernetesNamespace+"display-name"] = config.DisplayName
+		labels[api.KubernetesNamespace+"display-name"] = config.DisplayName
 	} else {
-		result[api.KubernetesNamespace+"display-name"] = config.Tag
+		labels[api.KubernetesNamespace+"display-name"] = config.Tag
 	}
 
-	if info == nil {
-		glog.V(3).Infof("Unable to fetch source informations, the output image labels will not be set")
-		return result
-	}
-
-	addBuildLabel(result, "image", config.BuilderImage)
-	addBuildLabel(result, "commit.author", info.Author)
-	addBuildLabel(result, "commit.date", info.Date)
-	addBuildLabel(result, "commit.id", info.CommitID)
-	addBuildLabel(result, "commit.ref", info.Ref)
-	addBuildLabel(result, "commit.message", info.Message)
-	addBuildLabel(result, "source-location", info.Location)
-	addBuildLabel(result, "source-context-dir", config.ContextDir)
-
-	return result
+	addBuildLabel(labels, "image", config.BuilderImage, namespace)
+	addBuildLabel(labels, "source-context-dir", config.ContextDir, namespace)
+	return labels
 }
 
-// addBuildLabel adds a new "io.openshift.s2i.build.*" label into map when the
+// GenerateLabelsFromSourceInfo generate the labels based on the source repository
+// informations.
+func GenerateLabelsFromSourceInfo(labels map[string]string, info *api.SourceInfo, namespace string) map[string]string {
+	if info == nil {
+		glog.V(3).Infof("Unable to fetch source informations, the output image labels will not be set")
+		return labels
+	}
+
+	addBuildLabel(labels, "commit.author", info.Author, namespace)
+	addBuildLabel(labels, "commit.date", info.Date, namespace)
+	addBuildLabel(labels, "commit.id", info.CommitID, namespace)
+	addBuildLabel(labels, "commit.ref", info.Ref, namespace)
+	addBuildLabel(labels, "commit.message", info.Message, namespace)
+	addBuildLabel(labels, "source-location", info.Location, namespace)
+	return labels
+}
+
+// addBuildLabel adds a new "*.build.*" label into map when the
 // value of this label is not empty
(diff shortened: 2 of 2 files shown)


jhadvig:
@bparees fyi I've tried to create an unified way how to add the labels for the Docker and S2I build, but after some research and chat with @soltysh and @mfojtik I found out that it cant be done since those two strategies are very differently implemented.

In case of the Docker build the I'm adding the LABELs in similar way how the ENVs are added, appending them to the end of the Dockerfile from which the build is done.

In case of the S2I build the labels are commited at the end of the build.

I know that the initial idea was that OS should be responsible for adding the labels, but thats not possible(as far as I can see it) in case of the S2I build, since the S2I itself is responsible for the build itself. For that reason I put the generating the label logic into the S2I (by refactoring the one that was already there) and share the logic also for the Docker build.

pkg/api/types.go
```diff
@@ -62,6 +62,9 @@ type Config struct {
 	// variables.
 	EnvironmentFile string
 
+	// BuildNamespace provides the namespace under which the labels will be generated.
+	BuildNamespaceLabel string
```
soltysh: I'd call it just `NamespaceLabel` or even just `Namespace`.

pkg/util/labels.go
```diff
@@ -7,43 +7,61 @@ import (
...
+// and source repository informations.
 func GenerateOutputImageLabels(info *api.SourceInfo, config *api.Config) map[string]string {
-	result := map[string]string{}
+	labels := map[string]string{}
+	var buildNamespace string
+	if len(config.BuildNamespaceLabel) > 0 {
+		buildNamespace = config.BuildNamespaceLabel
+	} else {
+		buildNamespace = api.DefaultNamespace
+	}
```
soltysh: You can rewrite this in less verbose fashion:
```
namespace = api.DefaultNamespace
if len(config.Namespace) > 0 {
    namespace = config.Namespace
}
```

pkg/util/labels.go
```diff
@@ -7,43 +7,61 @@ import (
...
+	labels = GenerateLabelsFromConfig(labels, config, buildNamespace)
+	labels = GenerateLabelsFromSourceInfo(labels, info, buildNamespace)
+	return labels
+}
 
+// GenerateLabelsFromConfig generate the labels based on build s2i Config
+func GenerateLabelsFromConfig(labels map[string]string, config *api.Config, buildNamespace string) map[string]string {
 	if len(config.Description) > 0 {
-		result[api.KubernetesNamespace+"description"] = config.Description
+		labels[api.KubernetesNamespace+"description"] = config.Description
```
soltysh: You're not using the `buildNamespace` parameter, you're passing here. Besides shorten it to just `namespace`.

pkg/util/labels.go
```diff
@@ -7,43 +7,61 @@ import (
...
 	}
 
 	if len(config.DisplayName) > 0 {
-		result[api.KubernetesNamespace+"display-name"] = config.DisplayName
+		labels[api.KubernetesNamespace+"display-name"] = config.DisplayName
```
soltysh: Same as above.

pkg/util/labels.go
```diff
@@ -7,43 +7,61 @@ import (
...
 	} else {
-		result[api.KubernetesNamespace+"display-name"] = config.Tag
+		labels[api.KubernetesNamespace+"display-name"] = config.Tag
```
soltysh: Same as above.

pkg/util/labels.go
```diff
@@ -7,43 +7,61 @@ import (
...
+		buildNamespace = api.DefaultNamespace
+	}
+
+	labels = GenerateLabelsFromConfig(labels, config, buildNamespace)
+	labels = GenerateLabelsFromSourceInfo(labels, info, buildNamespace)
+	return labels
+}
 
+// GenerateLabelsFromConfig generate the labels based on build s2i Config
+func GenerateLabelsFromConfig(labels map[string]string, config *api.Config, buildNamespace string) map[string]string {
```
soltysh: I'd suggest using just `namespace`.

pkg/util/labels.go
```diff
@@ -7,43 +7,61 @@ import (
...
 	}
 
+	addBuildLabel(labels, "image", config.BuilderImage, buildNamespace)
+	addBuildLabel(labels, "source-context-dir", config.ContextDir, buildNamespace)
+	return labels
+}
+
+// GenerateLabelsFromSourceInfo generate the labels based on the source repository
+// informations.
+func GenerateLabelsFromSourceInfo(labels map[string]string, info *api.SourceInfo, buildNamespace string) map[string]string {
```
soltysh: Again, just `namespace`.

pkg/api/types.go
rhcarvalho: `LabelNamespace`, and moreover make the comment (one line above) match the name.

pkg/util/labels.go
```diff
@@ -7,43 +7,61 @@ import (
...
-// addBuildLabel adds a new "io.openshift.s2i.build.*" label into map when the
+// addBuildLabel adds a new "*.build.*" label into map when the
 // value of this label is not empty
-func addBuildLabel(to map[string]string, key, value string) {
+func addBuildLabel(to map[string]string, key, value, buildNamespace string) {
 	if len(value) == 0 {
 		return
 	}
-	to[strings.Join([]string{api.DefaultNamespace, "build.", key}, "")] = value
+	to[strings.Join([]string{buildNamespace, "build.", key}, "")] = value
```
rhcarvalho: In this particular case, why do we even need `strings.Join`?!

```go
buildNamespace + "build." + key
```
?

pkg/api/types.go
jhadvig: +1 to the NamespaceLabel

pkg/util/labels.go
jhadvig: This was originally done by @mfojtik, I was just updating the logic. 
Don't have any strong opinion on any of those approaches...

soltysh: I guess it might be performance issue and I'm perfectly ok with current solution.

pkg/api/types.go
rhcarvalho: Hey, the value of that field is the namespace of the label, not the label of the namespace, therefore should be `NamespaceLabel`.

rhcarvalho: If this was about cars and colors, we'd have:

```go
type ManufactureOrder struct {
   CarMake string
   CarColor string
   // ...
}

mo := ManufactureOrder{"Škoda", "blue", ...}
```

`CarColor`, not `ColorCar`, right?

pkg/util/labels.go
rhcarvalho: What? How can `strings.Join([]string{a, b,c})` be more performant than `a + b + c` ?!

pkg/api/types.go
soltysh: Agree with @rhcarvalho

pkg/util/labels.go
rhcarvalho: @jhadvig I know it was not you introducing that pattern. But let's look forward, why not make it simpler, shorter and faster when we see it and can do it? ;-)

https://gist.github.com/rhcarvalho/9ee6179e976dfde01153

rhcarvalho: For the record, `strings.Join`'s approach of using a `[]byte` would be the way to go when we have an arbitrary number of strings / large strings because of garbage produced with intermediate strings. In this particular case we have 1 string literal, and 2 small strings, so that the extra work in not needed.

soltysh: @rhcarvalho http://herman.asia/efficient-string-concatenation-in-go, O(n2) is slower than O(n) imho. In your tests you're just testing times, but not allocations, which makes the difference in the long run. Remember strings in go are immutable, similarly to python.

rhcarvalho: @soltysh I've updated the gist including memory allocation. Again, `strings.Join` loses here. https://gist.github.com/rhcarvalho/9ee6179e976dfde01153

rhcarvalho: @soltysh the blog you linked to is benchmarking concatenating N strings, not 3 strings.

soltysh: Yes, correct it's about concatenating multiple strings.

jhadvig:
@soltysh @bparees @rhcarvalho rebased. PTAL

rhcarvalho:
LGTM

soltysh:
LGTM [merge]

soltysh:
re-[merge]

jhadvig:
@soltysh :+1: will update the origin PR with these changes

bparees:
posthumous lgtm

PR #256, opened 2015-08-10
Add allowed-uids flag to the sti command

csrwng:
Adds a flag --allowed-uids to the sti command that takes a comma-separated list of uid ranges, where a range can be:
1) a single number
2) [from]-[to] where [from] and [to] are integers
3) [from]-  a range with no upper bound
4) -[to]  a range with no lower bound

If this argument is specified, S2I checks whether the user specified in the builder image is 
1) numeric
2) is contained within the specified ranges
(For ONBUILD images the check also applies to USER ONBUILD instructions)

If those conditions are not met, the build fails.

### DIFF
--- a/pkg/build/strategies/strategies.go
+++ b/pkg/build/strategies/strategies.go
@@ -8,6 +8,8 @@ import (
 	"github.com/openshift/source-to-image/pkg/build/strategies/onbuild"
 	"github.com/openshift/source-to-image/pkg/build/strategies/sti"
 	"github.com/openshift/source-to-image/pkg/docker"
+	"github.com/openshift/source-to-image/pkg/errors"
+	userutil "github.com/openshift/source-to-image/pkg/util/user"
 )
 
 // GetStrategy decides what build strategy will be used for the STI build.
@@ -51,5 +53,33 @@ func GetBuilderImage(config *api.Config) (*docker.PullResult, error) {
 	}
 	result.Image = image
 	result.OnBuild = d.IsImageOnBuild(config.BuilderImage)
+
+	if err = checkAllowedUser(d, config, result.OnBuild); err != nil {
+		return nil, err
+	}
+
 	return &result, nil
+}
+
+func checkAllowedUser(d docker.Docker, config *api.Config, isOnbuild bool) error {
+	if config.AllowedUIDs == nil || config.AllowedUIDs.Empty() {
+		return nil
+	}
+	user, err := d.GetImageUser(config.BuilderImage)
+	if err != nil {
+		return err
+	}
+	if !userutil.IsUserAllowed(user, &config.AllowedUIDs) {
+		return errors.NewBuilderUserNotAllowedError(config.BuilderImage, false)
+	}
+	if isOnbuild {
+		cmds, err := d.GetOnBuild(config.BuilderImage)
+		if err != nil {
+			return err
+		}
+		if !userutil.IsOnbuildAllowed(cmds, &config.AllowedUIDs) {
+			return errors.NewBuilderUserNotAllowedError(config.BuilderImage, true)
+		}
+	}
+	return nil
 }
--- a/pkg/build/strategies/strategies_test.go
+++ b/pkg/build/strategies/strategies_test.go
@@ -1,0 +1,101 @@
+package strategies
+
+import (
+	"testing"
+
+	"github.com/openshift/source-to-image/pkg/api"
+	"github.com/openshift/source-to-image/pkg/test"
+	"github.com/openshift/source-to-image/pkg/util/user"
+)
+
+func rangeList(str string) *user.RangeList {
+	l, err := user.ParseRangeList(str)
+	if err != nil {
+		panic(err)
+	}
+	return l
+}
+
+func TestCheckAllowedUser(t *testing.T) {
+	tests := []struct {
+		name        string
+		allowedUIDs *user.RangeList
+		user        string
+		onbuild     []string
+		expectErr   bool
+	}{
+		{
+			name:        "AllowedUIDs is not set",
+			allowedUIDs: rangeList(""),
+			user:        "root",
+			onbuild:     []string{},
+			expectErr:   false,
+		},
+		{
+			name:        "AllowedUIDs is set, non-numeric user",
+			allowedUIDs: rangeList("0"),
+			user:        "default",
+			onbuild:     []string{},
+			expectErr:   true,
+		},
+		{
+			name:        "AllowedUIDs is set, user 0",
+			allowedUIDs: rangeList("1-"),
+			user:        "0",
+			onbuild:     []string{},
+			expectErr:   true,
+		},
+		{
+			name:        "AllowedUIDs is set, numeric user, non-numeric onbuild",
+			allowedUIDs: rangeList("1-10,30-"),
+			user:        "100",
+			onbuild:     []string{"COPY test test", "USER default"},
+			expectErr:   true,
+		},
+		{
+			name:        "AllowedUIDs is set, numeric user, no onbuild user directive",
+			allowedUIDs: rangeList("1-10,30-"),
+			user:        "200",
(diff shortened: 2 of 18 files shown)


csrwng:
@bparees, @gabemontero ptal
@pweil - fyi -- this is the S2I portion of the pull on origin to prevent root on source builds. I've added integration tests to validate that the user is getting checked for both sti and onbuild strategies.

bparees:
lgtm

pkg/build/strategies/onbuild/onbuild.go
```diff
@@ -71,8 +73,34 @@ func (b *OnBuild) SourceTar(config *api.Config) (io.ReadCloser, error) {
...
+	if !userutil.IsOnbuildAllowed(cmds, &config.AllowedUIDs) {
+		return errors.NewBuilderUserNotAllowedError(config.BuilderImage, true)
+	}
+	return nil
+}
+
 // Build executes the ONBUILD kind of build
 func (b *OnBuild) Build(config *api.Config) (*api.Result, error) {
+	// Check whether image user is allowed
+	if err := b.checkAllowedUser(config); err != nil {
```
gabemontero: At both here and in the Build method in sti.go, I believe you could move the checkAllowedUser function and subsequent call to the GetStrategy / GetBuilderImage path in pkg/build/strategies/strategies.go.  If you inspect the image's user there (for either sti or onbuild) and find a problem, you can exit the command sooner than at the Build step (avoiding Prepare for example), as well as address the check in one place.

pkg/build/strategies/sti/sti.go
```diff
@@ -115,6 +116,10 @@ func New(req *api.Config) (*STI, error) {
 func (b *STI) Build(config *api.Config) (*api.Result, error) {
 	defer b.garbage.Cleanup(config)
 
+	if err := b.checkAllowedUser(config); err != nil {
```
gabemontero: Ref my comment in onbuild.go.

pkg/build/strategies/onbuild/onbuild.go
csrwng: I still need a specific check for onbuild (not only the image user, but also the onbuild instructions)... would it work if I add the check for the image user to strategies.go and then still keep a check in the onbuild strategy?

gabemontero: Cesar and I sync'ed up.  Both s2i and onbuild could be covered in strategies.go.  He'll make the final call on whether he  goes with the suggestion or not.

csrwng: Updated the code

csrwng:
[test]

pkg/util/user/range.go
```diff
@@ -0,0 +1,148 @@
...
+		}
+		from, err := strconv.Atoi(parts[0])
+		if err != nil {
+			return nil, ErrParseRange
+		}
+		to, err := strconv.Atoi(parts[1])
+		if err != nil {
+			return nil, ErrParseRange
+		}
+		if to < from {
```
soltysh: You're duplicating that check here and in `NewRange` method.

pkg/util/user/range.go
```diff
@@ -0,0 +1,148 @@
...
+	switch len(parts) {
+	case 1:
+		num, err := strconv.Atoi(parts[0])
+		if err != nil {
+			return nil, ErrParseRange
+		}
+		return NewRange(num, num)
+	case 2:
+		if parts[0] == "" {
+			to, err := strconv.Atoi(parts[1])
```
soltysh: I'm thinking, you're doing the int parsing over and over before each `NewRange*` method, how about move that logic there?

pkg/util/user/range.go
csrwng: thx, will remove. Btw, thank you for still looking at STI code! :-) :+1:

pkg/util/user/range.go
csrwng: do you mean I should change the param type of NewRange to string ?

soltysh: I guess so.

csrwng: I refactored it differently since it still looked kind of ugly with the parsing in every new method. ptal

pkg/build/strategies/strategies.go
```diff
@@ -51,5 +53,33 @@ func GetBuilderImage(config *api.Config) (*docker.PullResult, error) {
...
+	if isOnbuild {
+		cmds, err := d.GetOnBuild(config.BuilderImage)
+		if err != nil {
+			return err
+		}
+		if !userutil.IsOnbuildAllowed(cmds, &config.AllowedUIDs) {
+			return errors.NewBuilderUserNotAllowedError(config.BuilderImage, true)
+		}
+	}
+	return nil
```
gabemontero: Yep - that was what I was talking about.  +1

pkg/build/strategies/strategies_test.go
```diff
@@ -0,0 +1,94 @@
...
+			onbuild:     []string{"USER 500", "VOLUME /data"},
+			expectErr:   false,
+		},
+		{
+			name:        "AllowedUIDs is set, numeric user, onbuild user 0",
+			allowedUIDs: rangeList("1-"),
+			user:        "200",
+			onbuild:     []string{"RUN echo \"hello world\"", "USER 0"},
+			expectErr:   true,
+		},
```
soltysh: I'm missing test case to cover -N when setting range.

pkg/util/user/rangelist_test.go
```diff
@@ -0,0 +1,117 @@
...
+		},
+		{
+			str:         "{1-5}",
+			errExpected: true,
+		},
+		{
+			str:         "1-5-,2-",
+			errExpected: true,
+		},
+	}
```
soltysh: Yet again missing -N test case, imho.

pkg/build/strategies/strategies_test.go
csrwng: I'll add it... this wasn't meant to be an exhaustive test though. I cover those in the util/user package

soltysh: Yeah, I noticed that after scrolling down, still having those here as well doesn't hurt :wink:

pkg/util/user/rangelist_test.go
csrwng: will add

pkg/util/user/range.go
```diff
@@ -0,0 +1,164 @@
...
+	if b.err = err; b.err != nil {
+		return b
+	}
+	if num < 0 {
+		b.err = ErrInvalidRange
+		return b
+	}
+	b.r.to = &num
+	return b
+}
```
soltysh: I'm not too happy about this solution either as it's again duplicating code between `To` and `From` methods. Besides exposing those methods publicly isn't good idea imho. I'd make them private, since you're specifically using them to pass the result of `parseInt`.

pkg/util/user/range.go
```diff
@@ -0,0 +1,164 @@
...
+
+// NewRangeFrom creates a new range with only the lower bound
+func NewRangeFrom(from int) (*Range, error) {
+	return (&rangeBuilder{}).From(from, nil).Range()
+}
+
+func parseInt(str string) (int, error) {
+	num, err := strconv.Atoi(str)
+	if err != nil {
+		return 0, ErrParseRange
```
soltysh: You're kinda swallowing the error here, which might be useful for an end user.

pkg/util/user/range.go
csrwng: you're right... let me try to get rid of the dup code

csrwng: ok, got rid of dup code

csrwng:
@soltysh do you have any additional comments?

soltysh:
I'm good, LGTM. [merge]

Issue #180, opened 2015-05-07
run image as part of running source-to-image


thoraxe:
It came up in conversation that it might be nice if the sti "program" accepted a flag that would: cause the resulting built Docker image to stay running instead of actually being a created image.

There may be cases where I want to test the result of my STI process locally without having to push the Docker image somewhere.

I guess this flag would essentially cause "run" to be executed immediately after "assemble" inside the instance of the builder image.

It's not a huge time savings, but it could potentially make a few other automation tasks easier... maybe?

bparees:
I would actually describe it as immediately running the output image, as opposed to running the run script after the assemble script.  (among other things, that's much easier to implement).

thoraxe:
I guess I don't understand how it's easier.

If there's an env-var for "run immediately", assemble calls run?

I'm probably missing something.

bparees:
what you're proposing would require every single assemble script be updated to respect that env var and invoke run.

in addition, even if you want to run immediately, we're still going to want to commit the output of your assemble script.  so polluting that container by running the run command in it makes that hard/impossible.

it's just cleaner to run the assemble script, commit the result, and then if desired, launch that image.

thoraxe:
Well, the point of this RFE was to improve time and make for more rapid development cycles. Having to stop the result of the build, commit it, and then run the new image takes time. Sure, it's a trivial amount of time, but still time.

You can kind of look at it both ways. Either we "pollute" the running container and then might have to rebuild, or we "pollute" the local user's Docker with lots of old junk images.

The other thing to think about is that, if we're only using this sti tool to "test" before sending code along somewhere else, it's going to end up getting rebuilt anyway.

I don't think I'm being very coherent.

bparees:
there will be junk images either way, since even with your suggestion you'll get a new image everytime you run STI.

you will have fewer junk containers with your approach (half as many) *if* we don't run with --rm

anyway i take your point, but it also seems like there is value in producing a re-runnable image when you build, even if you want to run it immediately initially.  Which means paying the cost of commit.

thoraxe:
Yes, I think there's value in both sides. Sounds like we're on the same page.

soltysh:
One important question is, should we publish the exposed ports to host by default when doing this run? I assume yes, but what if port is already in use, should we increment number to find first free and tell the user? As for the first question I'm positive, the latter I'm quite reluctant for doing so. Thoughts?

thoraxe:
[...]

PR #258, opened 2015-08-12
issue180: launch resulting image in container for diagnostic purposes

gabemontero:
@bparees @soltysh Gentlemen, the PR for issue 180.

Fixed https://github.com/openshift/source-to-image/issues/180

### DIFF
--- a/cmd/sti/main.go
+++ b/cmd/sti/main.go
@@ -25,6 +25,7 @@ import (
 	"github.com/openshift/source-to-image/pkg/create"
 	"github.com/openshift/source-to-image/pkg/docker"
 	"github.com/openshift/source-to-image/pkg/errors"
+	"github.com/openshift/source-to-image/pkg/run"
 	"github.com/openshift/source-to-image/pkg/util"
 	"github.com/openshift/source-to-image/pkg/version"
 )
@@ -167,10 +168,19 @@ func newCmdBuild(cfg *api.Config) *cobra.Command {
 				glog.V(1).Infof(message)
 			}
 
+			if cfg.RunImage {
+				runner, rerr := run.New(cfg)
+				if rerr == nil {
+					rerr = runner.Run(cfg)
+				}
+				checkErr(rerr)
+			}
+
 		},
 	}
 
 	buildCmd.Flags().BoolVarP(&(cfg.Quiet), "quiet", "q", false, "Operate quietly. Suppress all non-error output.")
+	buildCmd.Flags().BoolVar(&(cfg.RunImage), "run", false, "Run resulting image as part of invocation of this command.  You may have to Ctrl-C to exit the sti command invocation and the container launched because this options was specified.")
 	buildCmd.Flags().BoolVar(&(cfg.Incremental), "incremental", false, "Perform an incremental build")
 	buildCmd.Flags().BoolVar(&(cfg.RemovePreviousImage), "rm", false, "Remove the previous image during incremental builds")
 	buildCmd.Flags().StringP("env", "e", "", "Specify an environment var NAME=VALUE,NAME2=VALUE2,...")
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -78,6 +78,7 @@ that image and add them to the tar streamed to the container into `/artifacts`.
 | `-e (--env)`               | Environment variables to be passed to the builder eg. `NAME=VALUE,NAME2=VALUE2,...` |
 | `-E (--environment-file)`               | Specify the path to the file with environment |
 | `--force-pull`             | Always pull the builder image, even if it is present locally (defaults to true) |
+| `--run`             | Launch the resulting image after a successful build. All output from the image is being printed to help determine image's validity. In case of a long running image you will have to Ctrl-C to exit both sti and the running container.  (defaults to false) |
 | `-r (--ref)`               | A branch/tag that the build should use instead of MASTER (applies only to GIT source) |
 | `--rm`                     | Remove the previous image during incremental builds |
 | `--save-temp-dir`          | Save the working directory used for fetching scripts and sources |
--- a/pkg/build/strategies/sti/sti.go
+++ b/pkg/build/strategies/sti/sti.go
@@ -20,11 +20,6 @@ import (
 	"github.com/openshift/source-to-image/pkg/util"
 )
 
-const (
-	// maxErrorOutput is the maximum length of the error output saved for processing
-	maxErrorOutput = 1024
-)
-
 var (
 	// List of directories that needs to be present inside working dir
 	workingDirs = []string{
@@ -325,7 +320,7 @@ func (b *STI) Save(config *api.Config) (err error) {
 		OnStart:         extractFunc,
 	}
 
-	go streamContainerError(errReader, nil, config)
+	go docker.StreamContainerIO(errReader, nil, glog.Error)
 	err = b.docker.RunContainer(opts)
 
 	if e, ok := err.(errors.ContainerError); ok {
@@ -403,32 +398,13 @@ func (b *STI) Execute(command string, config *api.Config) error {
 		}
 	}(outReader)
 
-	go streamContainerError(errReader, &errOutput, config)
+	go docker.StreamContainerIO(errReader, &errOutput, glog.Error)
 
 	err = b.docker.RunContainer(opts)
 	if e, ok := err.(errors.ContainerError); ok {
 		return errors.NewContainerError(config.BuilderImage, e.ErrorCode, errOutput)
 	}
 	return err
-}
-
-func streamContainerError(errStream io.Reader, errOutput *string, config *api.Config) {
-	scanner := bufio.NewReader(errStream)
-	for {
-		text, err := scanner.ReadString('\n')
-		if err != nil {
-			// we're ignoring ErrClosedPipe, as this is information
-			// the docker container ended streaming logs
-			if err != io.ErrClosedPipe && err != io.EOF {
-				glog.Errorf("Error reading docker stderr, %v", err)
-			}
-			break
-		}
-		glog.Error(text)
-		if errOutput != nil && len(*errOutput) < maxErrorOutput {
-			*errOutput += text + "\n"
-		}
-	}
 }
 
 func (b *STI) generateConfigEnv() (configEnv []string) {
--- a/pkg/docker/docker.go
+++ b/pkg/docker/docker.go
@@ -12,6 +12,8 @@ import (
 
 	"github.com/openshift/source-to-image/pkg/api"
 	"github.com/openshift/source-to-image/pkg/errors"
+	"os"
+	"os/signal"
 )
 
 const (
@@ -75,6 +77,7 @@ type Client interface {
 	CommitContainer(opts docker.CommitContainerOptions) (*docker.Image, error)
 	CopyFromContainer(opts docker.CopyFromContainerOptions) error
 	BuildImage(opts docker.BuildImageOptions) error
+	InspectContainer(id string) (*docker.Container, error)
 }
 
 type stiDocker struct {
@@ -106,6 +109,7 @@ type RunContainerOptions struct {
 	Stderr          io.Writer
 	OnStart         func() error
 	PostExec        PostExecutor
+	TargetImage     bool
 }
 
 // CommitContainerOptions are options passed in to the CommitContainer method
@@ -336,26 +340,17 @@ func getDestination(image *docker.Image) string {
 	return DefaultDestination
 }
 
-// RunContainer creates and starts a container using the image specified in the options with the ability
-// to stream input or output
-func (d *stiDocker) RunContainer(opts RunContainerOptions) (err error) {
-	// get info about the specified image
-	image := getImageName(opts.Image)
-	var imageMetadata *docker.Image
-	if opts.PullImage {
-		imageMetadata, err = d.CheckAndPullImage(image)
-	} else {
-		imageMetadata, err = d.client.InspectImage(image)
-	}
-	if err != nil {
-		glog.Errorf("Unable to get image metadata for %s: %v", image, err)
-		return err
+// this funtion simply abstracts out the tar related processing that was originally inline in RunContainer()
+func runContainerTar(opts RunContainerOptions, config docker.Config, imageMetadata *docker.Image) (docker.Config, string) {
+	tarDestination := ""
+	if opts.TargetImage {
+		return config, tarDestination
 	}
 
 	// base directory for all STI commands
 	var commandBaseDir string
 	// untar operation destination directory
-	tarDestination := opts.Destination
+	tarDestination = opts.Destination
 	if len(tarDestination) == 0 {
 		tarDestination = getDestination(imageMetadata)
 	}
@@ -382,31 +377,43 @@ func (d *stiDocker) RunContainer(opts RunContainerOptions) (err error) {
 		cmd = []string{"/bin/sh", "-c", fmt.Sprintf("tar -C %s -xf - && %s",
 			tarDestination, filepath.Join(commandBaseDir, string(opts.Command)))}
 	}
-	config := docker.Config{
-		Image: image,
-		Cmd:   cmd,
+	config.Cmd = cmd
+	return config, tarDestination
+}
+
+// this funtion simply abstracts out the running of the newly produced image when the --run=true option is provided
+func runContainerDockerRun(container *docker.Container, d *stiDocker, image string) {
+	cont, icerr := d.client.InspectContainer(container.ID)
+	liveports := "\n\nPort Bindings:  "
+	if icerr == nil {
+		//Ports is of the follwing type:  map[docker.Port][]docker.PortBinding
+		for port, bindings := range cont.NetworkSettings.Ports {
+			liveports = liveports + "\n  Container Port:  " + port.Port()
+			liveports = liveports + "\n        Protocol:  " + port.Proto()
+			liveports = liveports + "\n        Public Host / Port Mappings:"
+			for _, binding := range bindings {
+				liveports = liveports + "\n            IP: " + binding.HostIP + " Port: " + binding.HostPort
+			}
+		}
+		liveports = liveports + "\n"
 	}
 
-	if opts.Env != nil {
-		config.Env = opts.Env
-	}
-	if opts.Stdin != nil {
-		config.OpenStdin = true
-		config.StdinOnce = true
-	}
-	if opts.Stdout != nil {
-		config.AttachStdout = true
-	}
+	glog.Infof("\n\n\n\n\nThe image %s has been started in container %s as a result of the --run=true option.  The container's stdout/stderr will be redirected to this command's glog output to help you validate its behavior.  You can also inspect the container with docker commands if you like.  If the container is set up to stay running, you will have to Ctrl-C to exit this command, which should also stop the container %s.  This particular invocation attempts to run with the port mappings %+v \n\n\n\n\n", image, container.ID, container.ID, liveports)
 
-	glog.V(2).Infof("Creating container using config: %+v", config)
-	container, err := d.client.CreateContainer(docker.CreateContainerOptions{Name: "", Config: &config})
-	if err != nil {
-		return err
-	}
-	defer d.RemoveContainer(container.ID)
+	signalChan := make(chan os.Signal, 1)
+	cleanupDone := make(chan bool)
+	signal.Notify(signalChan, os.Interrupt)
+	go func() {
+		for range signalChan {
+			glog.V(2).Info("\nReceived an interrupt, stopping services...\n")
+			cleanupDone <- true
+		}
+	}()
+	<-cleanupDone
+}
 
-	glog.V(2).Infof("Attaching to container")
-	attached := make(chan struct{})
+// this funtion simply abstracts out the first phase of attaching to the container that was originally in line with the RunContainer() method
+func runContainerAttachOne(attached chan struct{}, container *docker.Container, opts RunContainerOptions, d *stiDocker) sync.WaitGroup {
 	attachOpts := docker.AttachToContainerOptions{
 		Container: container.ID,
 		Success:   attached,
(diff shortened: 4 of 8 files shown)


bparees:
@thoraxe also fyi.

cmd/sti/main.go
```diff
@@ -167,10 +168,19 @@ func newCmdBuild(cfg *api.Config) *cobra.Command {
...
+					rerr = runner.Run(cfg)
+				}
+				checkErr(rerr)
+			}
+
 		},
 	}
 
 	buildCmd.Flags().BoolVarP(&(cfg.Quiet), "quiet", "q", false, "Operate quietly. Suppress all non-error output.")
+	buildCmd.Flags().BoolVar(&(cfg.RunImage), "run", false, "Run resulting image as part of invocation of this command.  You may have to Ctrl-C to exit the sit command invocation and the container launched because this options was specified.")
```
bparees: s/sit/s2i/

bparees: or maybe sti in this case since you're referring to the actual binary name and we haven't changed it yet.

pkg/docker/docker.go
```diff
@@ -472,17 +479,40 @@ func (d *stiDocker) RunContainer(opts RunContainerOptions) (err error) {
...
-	glog.V(2).Infof("Container exited")
+	if opts.TargetImage {
+		exposedPorts := make([]string, 0)
+		for port := range imageMetadata.Config.ExposedPorts {
+			exposedPorts = append(exposedPorts, port.Port())
+		}
 
-	if exitCode != 0 {
-		return errors.NewContainerError(container.Name, exitCode, "")
+		glog.V(0).Infof("\n\n\n\n\nThe image %s has been started in container %s as a result of the --run=true option.  The container's stdout/stderr will be redirected to this command's glog output to help you validate its behavior.  You can also inspect the container with docker commands if you like.  If the container is set up to stay running, you will have to Ctrl-C to exit this command, which should also stop the container %s.  This particular invocation attempts to run on the exposed ports %+v \n\n\n\n\n", image, container.ID, container.ID, exposedPorts)
```
bparees: I was thinking you'd print the port mappings (ie the host port and what container port it mapped to).  Just the internal port numbers aren't too useful here.

pkg/docker/docker.go
```diff
@@ -352,39 +355,43 @@ func (d *stiDocker) RunContainer(opts RunContainerOptions) (err error) {
...
-		commandBaseDir = filepath.Join(tarDestination, "scripts")
-		glog.V(2).Infof("Both scripts and untarred source will be placed in '%s'", tarDestination)
-	} else {
-		// for internal scripts we can have separate path for scripts and untar operation destination
-		scriptsURL := opts.ScriptsURL
-		if len(scriptsURL) == 0 {
-			scriptsURL = getScriptsURL(imageMetadata)
+
+	tarDestination := ""
+	if len(opts.Command) != 0 {
```
bparees: i wouldn't overload the meaning of an empty opts.Command to mean "running the target image" because we might in the future want to support passing a command to run in the target image.  ie --run --command=<somecommand> instead of only running the default image CMD/ENTRYPOINT.

cmd/sti/main.go
gabemontero: that's right - used `sti` cause referring to binary

pkg/docker/docker.go
gabemontero: yep - good catch - that may have been a leftover from the iterations / permutations around the sleep/wait/signal stuff and how the `sti` command exited.  Switching to a check of the opts.TargetImage boolean.

pkg/docker/docker.go
gabemontero: OK, so I *think* this means in the Ports map at https://github.com/openshift/source-to-image/blob/master/Godeps/_workspace/src/github.com/fsouza/go-dockerclient/container.go#L131
as the PortMapping map above seems to be in reference to some deprecated `docker inspect` output.

Network settings is off of the Container struct:  https://github.com/openshift/source-to-image/blob/master/Godeps/_workspace/src/github.com/fsouza/go-dockerclient/container.go#L221-L235

And we get a ref to the container earlier in this method.

So yep, I'll switch it and give it a whirl.

thoraxe:
@bparees this is way over my head...

cmd/sti/main.go
bparees: still needs to be fixed from sit to sti

bparees:
@thoraxe i just wanted you to see the trouble you cause when you ask for stuff :)

pkg/docker/docker.go
gabemontero: Hmm ... unfortunately, the NetworkSettings are coming up `nil` in my testing.  Perhaps the Container ref returned by CreateContainer is not updated when other operations like StartContainer are done later on.

I could look into exposing https://github.com/openshift/source-to-image/blob/master/Godeps/_workspace/src/github.com/fsouza/go-dockerclient/container.go#L57-L69, which if called after StartContainer, perhaps the APIContainer / APIPort at https://github.com/openshift/source-to-image/blob/master/Godeps/_workspace/src/github.com/fsouza/go-dockerclient/container.go#L32-L52 could get us what you want.

gabemontero: Yep, that seems to have done the trick.  Will be part of upcoming push.

gabemontero:
ok ... push with updates from comments done

cmd/sti/main.go
bparees: still says "sit" and not "sti"

pkg/docker/docker.go
```diff
@@ -472,17 +480,50 @@ func (d *stiDocker) RunContainer(opts RunContainerOptions) (err error) {
...
-	}
-	glog.V(2).Infof("Container exited")
+	if opts.TargetImage {
 
-	if exitCode != 0 {
-		return errors.NewContainerError(container.Name, exitCode, "")
+		lco := docker.ListContainersOptions{
+			All: true,
+		}
+		conts, lcerr := d.client.ListContainers(lco)
```
bparees: the client api doesn't give you a way to just get the container given that you've got the container id?  weird.

pkg/docker/docker.go
```diff
@@ -472,17 +480,50 @@ func (d *stiDocker) RunContainer(opts RunContainerOptions) (err error) {
...
+		if lcerr == nil {
+			for _, cont := range conts {
+				if cont.ID == container.ID {
+					liveports = cont.Ports
+					break
+				}
+			}
+		}
+
+		glog.V(0).Infof("\n\n\n\n\nThe image %s has been started in container %s as a result of the --run=true option.  The container's stdout/stderr will be redirected to this command's glog output to help you validate its behavior.  You can also inspect the container with docker commands if you like.  If the container is set up to stay running, you will have to Ctrl-C to exit this command, which should also stop the container %s.  This particular invocation attempts to run with the port mappings %+v \n\n\n\n\n", image, container.ID, container.ID, liveports)
```
bparees: can you show us a sample of this output?  i'm curious in particular what the port data looks like

bparees:
one nit and a couple Qs and then i think this lgtm, pending what the example output looks like.

i'd like to get @soltysh's sign off too though since he's most familiar w/ some of this code and might be aware of gotchas.

cmd/sti/main.go
gabemontero: Sorry, was distracted with the other comments, missed this one.  Part of next push.

pkg/docker/docker.go
gabemontero: Actually I just found an InspectContainer.  Going to switch to that one, assuming it works.

gabemontero: Nope - InspectContainer and the Container / NetworkSettings / Port / PortBinding only lists:

map[8080/tcp:[]]

The ListContainer / APIContainer / APIPort has more info.  Staying with that.

pkg/docker/docker.go
gabemontero: Here you go:

 [{PrivatePort:8080 PublicPort:0 Type:tcp IP:}] 

While it has an IP field, it is showing up empty, perhaps because we are running against the local docker daemon.  Of course public port 0 is not super informative either.  And cross referencing with netstat -nap was not conclusive.

Don't think we are going to get much better than this though.  I'll push, but I'm thinking this is not worth it, and the user will just want to run docker cmds, netstat, etc. on his own.

bparees: are you sure the ports *are* getting mapped to something on the host?
(easy test:  run s2i with --run and then do a docker ps
here's an example of a container i ran with "docker run -P" so the exposed ports got mapped:
$ docker ps
CONTAINER ID        IMAGE                       COMMAND             CREATED             STATUS              PORTS                     NAMES
961ab6e2203a        openshift/ruby-20-centos7   "/bin/sh"           26 seconds ago      Up 23 seconds       0.0.0.0:32768->8080/tcp   insane_hodgkin

pkg/build/strategies/sti/sti.go
```diff
@@ -403,7 +398,7 @@ func (b *STI) Execute(command string, config *api.Config) error {
 		}
 	}(outReader)
 
-	go streamContainerError(errReader, &errOutput, config)
+	go docker.StreamContainerIO(errReader, &errOutput, glog.Error)
```
soltysh: :8ball:

pkg/build/interfaces.go
```diff
@@ -49,6 +49,12 @@ type Ignorer interface {
 	Ignore(*api.Config) error
 }
 
+// Runner allows the user to actually run the produced imgae as part of the s2i
+// cmd line invocation, to see if it operates as he would expect
+type Runner interface {
+	Run(*api.Config) error
+}
```
soltysh: There's no need to add this interface here. First of all you're not using it anywhere inside the build. Moreover it is not related to build in any way.

pkg/docker/docker.go
```diff
@@ -472,17 +480,50 @@ func (d *stiDocker) RunContainer(opts RunContainerOptions) (err error) {
...
+		if err != nil {
+			return err
+		}
+
+		glog.V(2).Infof("Container exited")
+		if exitCode != 0 {
+			return errors.NewContainerError(container.Name, exitCode, "")
+		}
 	}
+
```
soltysh: I'd suggest splitting this function into smaller parts, it's almost 200 lines long and it's damn hard to follow efficiently, my heads explodes trying to follow the logic in it. Is it possible somehow?

pkg/docker/docker.go
soltysh: Just checked it and I have to complain :disappointed: because the ports are NOT mapped. I can't access localhost:8080 after doing: 
```
sti build https://github.com/openshift/sti-python.git --run --loglevel=3 --context-dir=3.4/test/setup-test-app/ openshift/python-34-centos7 python-sample-app
```
Running `docker run -p 8080:8080 python-sample-app` allows me to access the ports.

soltysh:
Good work @gabemontero, I'll reserve some more time later today or at latest tomorrow morning to analyze it more thoroughly.

pkg/docker/docker.go
gabemontero: Yeah, it was long  before I touched it ;-), but certainly my change only added to the situation.  I'll see what I can do.

gabemontero:
Thanks @soltysh .

pkg/docker/docker.go
soltysh: I know it was long :stuck_out_tongue_winking_eye: but now I found someone lucky enough to fix it :wink: 
But srsly, I've looked at simplifying it a couple of times before, but frankly said I failed, it's too tightly coupled.

pkg/build/interfaces.go
gabemontero: Don't see why being used inside the build or related to the build matters wrt leveraging an interface.  I see the interfaces as a way to outline and define the semantics of major portions of the entire program - concepts that are of interest to the user.

pkg/docker/docker.go
gabemontero: I suspect I'm about to receive some more education on working in the docker world ;-) ... so after looking at `docker help run`, I see that `-p` publishes specific ports, and -P publishes all ports ... assuming that is the mapping you all are referring to, then it seems I need to look at the docker client api and see what triggers the equivalent of -p/-P ...

... do I sound anywhere close ?

pkg/docker/docker.go
gabemontero: LOL - sounds good - it may take a couple of iterations of push/review on our respective parts, but let's see where we land.

pkg/build/interfaces.go
soltysh: Yes, but these interfaces specifically are for builds only (pkg/**build**/interfaces.go). And as long as providing an interface makes sense here you have only one implementation of that interface and you're not using the interface directly anywhere, so generally speaking you don't leverage the power of that interface.

pkg/docker/docker.go
soltysh: I'm completely fine with publishing all the ports (`-P`), it's just I've been always using `-p`, so it's a matter of taste.

pkg/docker/docker.go
soltysh: :+1:

pkg/build/interfaces.go
gabemontero: OK, I see the sub directory point.  It could be moved to an interfaces.go file in another directory.

Use of interfaces can be warranted not just for the current situation, but for the anticipation of needing future implementations in the future.  And if you defer on creating the interface until that need arises, your *initial* implementation often needs some rework to keep it in line with the implicit semantics, whereas explicitly defining them helps scope the situation.

soltysh: There's a bigger plan to abstract the entire flow of s2i build to remove unnecessary code duplication we have in origin. Let's first do that and then think about adding additional interfaces if you don't mind. Besides OS will never leverage that Runner interface in the first place, it's solely for s2i standalone edition.

gabemontero: That is true (abstract / rework the existing flow) ... OK, I'll acquiesce here.

pkg/docker/docker.go
bparees: @gabemontero yes that sounds about right.  The other difference is that with "-p" you can specify the exact mapping (ie "map port 1234 on my host to port 8080 in the container"

with -P docker just allocates host ports randomly.   that's good and bad...  good:  it will (presumably) not pick ports that are already in use by something else.  bad: you have no idea what host port got assigned for a given container port, so you need to go look up the mappings if you want to, for example, access your app via your browser.

that last point is why, if we're going to do -P equivalence, we need to print out the mappings that got assigned, so the user doesn't have to go look it up themselves.

gabemontero: OK, I've got it working.  Now, there are two API's /structs that are accessible from the docker client that display the mappings, each in a slightly different format:

list containers -> APIPort
[{PrivatePort:8080 PublicPort:32768 Type:tcp IP:0.0.0.0}]

inspect container -> Port
map[8080/tcp:[{HostIP:0.0.0.0 HostPort:32769}]] 

I've bounced back and forth a bit on this one, but I think I like the inspect container.

Thoughts?

soltysh: Doesn't matter whichever you choose, as long as we list those information to user clearly, imho.

bparees: +1.  i guess i agree the 2nd one is a little better.  you could of course print out the fields yourself too and format it however you want.

soltysh: That's what I meant with my comment, above is not readable imho. I'd rather see line by line with a port mapping information, clearly standing out from the rest of the text.

gabemontero: Yeah, I was trying to get the formatting for free.  I'll create a formatted string.

gabemontero: Example of the new format:
Port Bindings:  
  Private Port:  8080
        Protocol:  tcp
        Public Host / Port Mappings:
            IP: 0.0.0.0 Port: 32770

gabemontero: note - github comments is removing indentation that I added for the various lines ... private port has 1 level of indentation, protocol and mappings header 2 levels, the ip/ port 3 levels

bparees: i'd probably label Private Port as Container Port instead, otherwise lgtm.

gabemontero:
OK ... push done with changes in response to latest @bparees comments and initial set of @soltysh comments

docs/cli.md
```diff
@@ -78,6 +78,7 @@ that image and add them to the tar streamed to the container into `/artifacts`.
 | `-e (--env)`               | Environment variables to be passed to the builder eg. `NAME=VALUE,NAME2=VALUE2,...` |
 | `-E (--environment-file)`               | Specify the path to the file with environment |
 | `--force-pull`             | Always pull the builder image, even if it is present locally (defaults to true) |
+| `--run`             | Launch the resulting image in a container, where stdout and stderr from the container are piped to the stdout and stderr of sti to help determine the image's validity (along with any docker commands you which to execute on the container provided), and where if the image is long running, you will have to Ctrl-C to exit both sti and the container (defaults to false) |
```
soltysh: I'd suggest to rewrite it like this:
Launch the resulting image after a successful build. All output from the image is being printed to help determine image's validity. In case of a long running image you will have to Ctrl-C to exit both sti and the running container.

pkg/docker/docker.go
```diff
@@ -336,6 +340,142 @@ func getDestination(image *docker.Image) string {
 	return DefaultDestination
 }
 
+func runContainerTar(opts RunContainerOptions, config docker.Config, imageMetadata *docker.Image) (docker.Config, string) {
+	tarDestination := ""
+	if !opts.TargetImage {
```
soltysh: Fail fast, iow:
```
if opts.TargetImage {
    return config, tarDestination
}
// rest of the code but this time without the indentation 
```

pkg/docker/docker.go
```diff
@@ -336,6 +340,142 @@ func getDestination(image *docker.Image) string {
...
+			liveports = liveports + "\n        Protocol:  " + port.Proto()
+			liveports = liveports + "\n        Public Host / Port Mappings:"
+			for _, binding := range bindings {
+				liveports = liveports + "\n            IP: " + binding.HostIP + " Port: " + binding.HostPort
+			}
+		}
+		liveports = liveports + "\n"
+	}
+
+	glog.V(0).Infof("\n\n\n\n\nThe image %s has been started in container %s as a result of the --run=true option.  The container's stdout/stderr will be redirected to this command's glog output to help you validate its behavior.  You can also inspect the container with docker commands if you like.  If the container is set up to stay running, you will have to Ctrl-C to exit this command, which should also stop the container %s.  This particular invocation attempts to run with the port mappings %+v \n\n\n\n\n", image, container.ID, container.ID, liveports)
```
soltysh: No need for V(0), just glog.Infof.

pkg/run/run.go
```diff
@@ -0,0 +1,59 @@
...
+		glog.Errorf("Create of Docker Container client failed with %v \n", ccerr)
+		return nil, ccerr
+	}
+	dr := &DockerRunner{
+		ContainerClient: cc,
+	}
+	return dr, nil
+}
+
+func (b *DockerRunner) Run(config *api.Config) error {
```
soltysh: Missing godoc.

pkg/run/run.go
```diff
@@ -0,0 +1,59 @@
...
+	"github.com/openshift/source-to-image/pkg/docker"
+	"github.com/openshift/source-to-image/pkg/errors"
+	"io"
+)
+
+type DockerRunner struct {
+	ContainerClient docker.Docker
+}
+
+func New(config *api.Config) (*DockerRunner, error) {
```
soltysh: Missing godoc.

pkg/run/run.go
```diff
@@ -0,0 +1,59 @@
...
+
+import (
+	"github.com/golang/glog"
+	"github.com/openshift/source-to-image/pkg/api"
+	"github.com/openshift/source-to-image/pkg/docker"
+	"github.com/openshift/source-to-image/pkg/errors"
+	"io"
+)
+
+type DockerRunner struct {
```
soltysh: Missing godoc.

pkg/docker/util.go
```diff
@@ -38,6 +43,25 @@ func GetImageRegistryAuth(dockerCfg io.Reader, imageName string) client.AuthConf
 	return client.AuthConfiguration{}
 }
 
+func StreamContainerIO(errStream io.Reader, errOutput *string, log func(...interface{})) {
```
soltysh: Missing godoc.

pkg/docker/docker.go
```diff
@@ -336,6 +340,142 @@ func getDestination(image *docker.Image) string {
 	return DefaultDestination
 }
 
+func runContainerTar(opts RunContainerOptions, config docker.Config, imageMetadata *docker.Image) (docker.Config, string) {
```
soltysh: godoc all those new functions, to avoid future problems understanding the flow.

soltysh:
I like it a lot. Thanks for splitting that method, reads better now! A couple of small nits, mostly missing godocs, otherwise LGTM.

docs/cli.md
gabemontero: in next push

pkg/docker/docker.go
gabemontero: in next push

pkg/docker/docker.go
gabemontero: in next push

pkg/run/run.go
gabemontero: in next push

pkg/run/run.go
gabemontero: in next push

pkg/run/run.go
gabemontero: in next push

pkg/docker/util.go
gabemontero: in next push

pkg/docker/docker.go
gabemontero: in next push

gabemontero:
Cool deal - ok, nits, comments/godocs pushed.

bparees:
lgtm [merge]

pkg/docker/util.go
```diff
@@ -38,6 +43,27 @@ func GetImageRegistryAuth(dockerCfg io.Reader, imageName string) client.AuthConf
 	return client.AuthConfiguration{}
 }
 
+// Takes data from the Reader and redirects to the log functin (typically we pass in
+// glog.Error for stderr and glog.Info for stdout
+func StreamContainerIO(errStream io.Reader, errOutput *string, log func(...interface{})) {
```
soltysh: For future, godoc needs to start with the method name you're describing:
```
// StreamContainerIO takes data from the Reader and redirects to the log functin (typically we pass in
// glog.Error for stderr and glog.Info for stdout
func StreamContainerIO(errStream io.Reader, errOutput *string, log func(...interface{})) {
```

pkg/run/run.go
```diff
@@ -0,0 +1,68 @@
...
+	//NOTE, we've seen some Golang level deadlock issues with the streaming of cmd output to
+	// glog, but part of the deadlock seems to have occurred when stdout was "silent"
+	// and produced no data, such as when we would do a git clone with the --quiet option.
+	// We have not seen the hang when the Cmd produces output to stdout.
+
+	go docker.StreamContainerIO(errReader, nil, glog.Error)
+	go docker.StreamContainerIO(outReader, nil, glog.Info)
+	rerr := b.ContainerClient.RunContainer(opts)
+	if e, ok := rerr.(errors.ContainerError); ok {
+		return errors.NewContainerError(config.Tag, e.ErrorCode, errOutput)
```
rhcarvalho: @gabemontero I'm trying to understand this, why do we replace a `ContainerError` with another `ContainerError`?

gabemontero: @rhcarvalho we are updating the contents of the message ... if I recall correctly, the original contents were not as insightful

PR #260, opened 2015-08-13
Adding context-dir to SourceInfo object

jhadvig:
Found our that the SourceInfo.ContextDir field is not filled by the ContextDir from the Config.

@bparees PTAL


bparees:
[merge]

PR #261, opened 2015-08-13
always pull if not present for incremental

bparees:
addresses another part of https://github.com/openshift/origin/issues/4127

@jhadvig @soltysh @mfojtik @gabemontero ptal.

I debated just always pulling the previous image, period.  That would avoid the possibility that someone does an incremental build with a stale previous image and picks up bad dependencies/artifacts, but in the end i decided we have "forcePull" for a reason, so the semantics are:

if forcePull is true, the builder AND the previous image (if incremental) are pulled.
if forcePull is false and the previous image does not exist locally, it is pulled.  
if forcePull is false and the previous image does exist, it is used (might be stale)

### DIFF
--- a/pkg/build/strategies/sti/sti.go
+++ b/pkg/build/strategies/sti/sti.go
@@ -286,7 +286,7 @@ func (b *STI) Exists(config *api.Config) bool {
 
 	// can only do incremental build if runtime image exists, so always pull image
 	previousImageExists, _ := b.docker.IsImageInLocalRegistry(config.Tag)
-	if config.ForcePull {
+	if !previousImageExists || config.ForcePull {
 		if image, _ := b.docker.PullImage(config.Tag); image != nil {
 			previousImageExists = true
 		}


bparees:
[test]

pkg/build/strategies/sti/sti.go
```diff
@@ -286,7 +286,7 @@ func (b *STI) Exists(config *api.Config) bool {
 
 	// can only do incremental build if runtime image exists, so always pull image
 	previousImageExists, _ := b.docker.IsImageInLocalRegistry(config.Tag)
-	if config.ForcePull {
+	if !previousImageExists || config.ForcePull {
```
soltysh: Actually existing previous image might be older than you'd wanted to be. Let me give you an example from OS world:
I'm doing a build (each time incremental) the first lands on node-1, 2nd on node-2, now if the build 3rd will land on node-1 it'll use artifacts from build-1 which is not the latest (build-2 is). That's why it was done so to always pull that image and not to look at `previousImageExists` flag, at least that's how I remembered it when doing it so.

Or we just don't care about that detail because in the worst case we'll re-download the artifacts that changed between build-1 and build-2?

bparees: yeah i discussed that in the description of the pull....

bparees: tl;dr:  that situation is why we have forcePull.

soltysh: :blush: omg, sorry for that.

soltysh:
LGTM [merge]

PR #263, opened 2015-08-15
use separate credentials to pull previous image for incremental builds

bparees:


bparees:
going to do the bump here instead:
https://github.com/openshift/source-to-image/pull/263

bparees:
yet another tidbit of https://github.com/openshift/origin/issues/4127

note that i had to bump the docker client because of issues parsing the .docker/config.json file which is now the default location for docker creds, instead of .dockercfg

@jhadvig @mfojtik @gabemontero @soltysh ptal.

bparees:
[test]

mfojtik:
@bparees LGTM as long as you fix the panic in Travis ;-)

soltysh:
One backward compatibility issue @mfojtik raised and one bigger that I have regarding bumping docker client. You're bumping that to latest available, which is not the one used in origin, hopefully you're not using newer features that will fail in origin, if so then we need update there as well. My general advice when bumping is to keep the deps in sync with origin ones, although origin docker client is way too old :(. In that case good compromise is to use the one from k8s and bump it here and in origin.

bparees:
@soltysh @mfojtik already have a pull to bump origin to the same level: https://github.com/openshift/origin/pull/4179

bparees:
(and k8s https://github.com/kubernetes/kubernetes/pull/12770)

soltysh:
:+1: LGTM then

bparees:
[test]

bparees:
[merge]

PR #264, opened 2015-08-16
Improve log messages for incremental build

nak3:
This PR contains following changes:

- L#290[1]: Remove `b.installedScripts[api.SaveArtifacts]`, since it is 99.9% true. Alternatively add message `build with save-artifacts under the %s", config.Tag, b.scriptsURL["save-artifacts"]`.
- L#278[2]: Move incremental build check in `func Exists()`.

[1] https://github.com/openshift/source-to-image/blob/4902844a6fc2c6869cdf7cc8ca9f269628c1ef4e/pkg/build/strategies/sti/sti.go#L290
[2] https://github.com/openshift/source-to-image/blob/4902844a6fc2c6869cdf7cc8ca9f269628c1ef4e/pkg/build/strategies/sti/sti.go#L278

### DIFF
--- a/pkg/build/strategies/sti/sti.go
+++ b/pkg/build/strategies/sti/sti.go
@@ -115,10 +115,12 @@ func (b *STI) Build(config *api.Config) (*api.Result, error) {
 		return nil, err
 	}
 
-	if b.incremental = b.artifacts.Exists(config); b.incremental {
-		glog.V(1).Infof("Existing image for tag %s detected for incremental build", config.Tag)
+	if !config.Incremental {
+		glog.V(1).Infof("Incremental build is disabled. Clean build will be performed")
+	} else if b.incremental = b.artifacts.Exists(config); b.incremental {
+		glog.V(1).Infof("Existing image for tag %s detected. Perform Incremental build with save-artifacts under the %s", config.Tag, b.scriptsURL["save-artifacts"])
 	} else {
-		glog.V(1).Infof("Clean build will be performed")
+		glog.V(1).Infof("Incremental build is enabled. But existing image for tag %s was not detected. Clean build will be performed", config.Tag)
 	}
 
 	glog.V(2).Infof("Performing source build from %s", config.Source)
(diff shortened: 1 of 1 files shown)


pkg/build/strategies/sti/sti.go
```diff
@@ -287,7 +284,7 @@ func (b *STI) Exists(config *api.Config) bool {
 		}
 	}
 
-	return previousImageExists && b.installedScripts[api.SaveArtifacts]
+	return previousImageExists
```
bparees: I realized recently that this check (installedScripts) appears to always return true because we assume the image may contain a save-artifacts script, which we can't confirm without running the image... but i'm still uncomfortable with removing the check entirely in case that changes in the future.

i'd be ok with moving the check to its own method (eg hasSaveArtifacts()) but i'd rather not remove it entirely.

soltysh: Not always true ;) It'll be always true for images with embedded scripts, because in that case the only option to verify save-artifacts existence is actually running that script. In all other cases it depends on whether that script was downloaded from one of the possible locations (external URL, sources).

pkg/build/strategies/sti/sti.go
```diff
@@ -115,10 +115,12 @@ func (b *STI) Build(config *api.Config) (*api.Result, error) {
 		return nil, err
 	}
 
-	if b.incremental = b.artifacts.Exists(config); b.incremental {
-		glog.V(1).Infof("Existing image for tag %s detected for incremental build", config.Tag)
+	if !config.Incremental {
+		glog.V(1).Infof("Incremental build is disabled. Clean build will be performed")
+	} else if b.incremental = b.artifacts.Exists(config); b.incremental {
+		glog.V(1).Infof("Existing image for tag %s detected. Perform Incremental build with save-artifacts under the %s", config.Tag, b.scriptsURL["save-artifacts"])
```
soltysh: I'm not too happy with those changes, imho the V1 verbosity currently provides enough information about the build process. Especially that the information about whether build has incremental enabled or disabled is already printed at the beginning, see [here](https://github.com/openshift/source-to-image/blob/master/cmd/sti/main.go#L165-L167) where the entire Config is being printed, but at a higher log level - 2. Above changes complicate the code more than needed.

bparees: that info doesn't get printed when we run inside openshift.

soltysh: It is, but again at loglevel=2, see [here](https://github.com/openshift/origin/blob/master/pkg/build/builder/sti.go#L102).

bparees: yeah and ugly, no formatting :(

but ok

soltysh:
@bparees given the recent changes we've introduced I think we can close this PR and if the author will want to pick it up he can always open a new one, wdyt?

nak3:
@soltysh yes, I am ok, if you close this PR. I'm sorry that I didn't update this.

PR #267, opened 2015-08-19
fix syntax error: unexpected range for go1.3.3

mdshuai:
when go version is 1.3.3, compile with error:
[root@openshift-144 source-to-image]# make
hack/build-go.sh
++ Building go targets for linux/amd64: cmd/sti
github.com/openshift/source-to-image/pkg/docker
_output/local/go/src/github.com/openshift/source-to-image/pkg/docker/docker.go:407: syntax error: unexpected range, expecting {
_output/local/go/src/github.com/openshift/source-to-image/pkg/docker/docker.go:410: syntax error: argument to go/defer must be function call
_output/local/go/src/github.com/openshift/source-to-image/pkg/docker/docker.go:412: non-declaration statement outside function body
_output/local/go/src/github.com/openshift/source-to-image/pkg/docker/docker.go:413: syntax error: unexpected }

### DIFF
--- a/pkg/docker/docker.go
+++ b/pkg/docker/docker.go
@@ -404,8 +404,8 @@ func runContainerDockerRun(container *docker.Container, d *stiDocker, image stri
 	cleanupDone := make(chan bool)
 	signal.Notify(signalChan, os.Interrupt)
 	go func() {
-		for _ = range signalChan {
-			glog.V(2).Info("\nReceived an interrupt, stopping services...\n")
+		for signal := range signalChan {
+			glog.V(2).Info("\nReceived signal %s, stopping services...\n", signal)
 			cleanupDone <- true
 		}
 	}()
--- a/pkg/docker/docker.go
+++ b/pkg/docker/docker.go
@@ -404,7 +404,7 @@ func runContainerDockerRun(container *docker.Container, d *stiDocker, image stri
 	cleanupDone := make(chan bool)
 	signal.Notify(signalChan, os.Interrupt)
 	go func() {
-		for range signalChan {
+		for _ = range signalChan {
 			glog.V(2).Info("\nReceived an interrupt, stopping services...\n")
 			cleanupDone <- true
 		}


bparees:
lgtm but will let @gabemontero give the sign off since he wrote the original code and might know something i don't.

gabemontero:
Ah yes - I originally had something like the change proposed here, but then the travis and jenkins jobs failed because hack/verify-gofmt.sh.

In fact, as you can see from the failed travis job, that failure again occurred:
$  ./hack/verify-gofmt.sh
!!! gofmt needs to be run on the following files: 
./pkg/docker/docker.go

Try running 'gofmt -s -d [path]'

The command "./hack/verify-gofmt.sh" failed and exited with 1 during .

So the change from @mdshuai is not going to fly.

I see there is the --tags option for go build, and that it provides #ifdef type capabilities, but @bparees - do we want to go down that path here?

bparees:
Weird that it's a gofmt error, but no we definitely do not want to add
ifdef support for something like this. We don't make any claims about
supporting go 1.3 anyway.

Thanks for the background.

Ben Parees | OpenShift

bparees:
@mdshuai is there a particular need to build this under go 1.3.3 or can we close this?

if you absolutely must fix the 1.3.3 compilation i think you need to change the code to:
`		for _, dummy := range signalChan {`

gabemontero:
And run hack/verify-gofmt.sh after the change.

mdshuai:
@bparees @gabemontero On our ose env(rhel7), the default go version is go1.3.3. Sometime we need compile the s2i binary on those env. There is really some different in go1.4 and go1.3 about "for ... range" usage.
https://golang.org/doc/go1.4#forrange

For the error @gabemontero meet when run hack/verify-gofmt.sh is cause by this.
https://golang.org/cmd/gofmt/#hdr-The_simplify_command

I have update the code for range a channel.After change have run hack/verify-gofmt.sh and no issue.
If you think we don't  want to support for go1.3.3, then you can close this. help review , thanks.

gabemontero:
Igtm

@bparees while I think it should be fine, I'll double check the new --run=true option once this is upstream, make sure it still does what we want with this change

bparees:
@gabemontero you can cherry pick this change and build it locally to test it before we merge.

it also makes me think we should have a test case in s2i that runs on merge for this function, if we're worried about it.

gabemontero:
@bparees  OK, I'll try to cherry pick and see (this will be my first official cherry-pick attempt)

On the test, since the --run=true requires a Ctrl-C to exit the run, and the "verification" is examining the images logs, it gets a little tricky.  But probably I'm missing something that you can educate me on.

bparees:
can't you kill -9 the s2i process and that should kill everything?


On Thu, Aug 20, 2015 at 11:34 AM, Gabe Montero <notifications@github.com>
wrote:

> @bparees <https://github.com/bparees> OK, I'll try to cherry pick and see
> (this will be my first official cherry-pick attempt)
>
> On the test, since the --run=true requires a Ctrl-C to exit the run, and
> the "verification" is examining the images logs, it gets a little tricky.
> But probably I'm missing something that you can educate me on.
>
> —
> <https://github.com/openshift/source-to-image/pull/267#issuecomment-133053189>
> .
>



-- 
Ben Parees | OpenShift

gabemontero:
Sure, from the shell script, run sti in the background, then kill it at
some point, but a) how to determine *when* to kill it (i.e. let it run long
enough) ... perhaps via docker ps and/or docker exec ... ??

bparees:
Redirect the output to a file and monitor the output?

Ben Parees | OpenShift

gabemontero:
Yeah and maybe grep for an expected string or something .... I'll open a separate issue to look into it - thx

gabemontero:
ok ... cherry-picked the change, verified still good with --run=true.  @bparees let's merge

bparees:
[merge]

bparees:
thanks @gabemontero and @mdshuai

Issue #257, opened 2015-08-11
sti rebuild fails with index out of range


dhirajsb:
The sti and sample app are on github, running the following commands causes an index out of range

```shell
rh-wfh-dbokde:camel-hello-world dbokde$ sti build https://github.com/dhirajsb/camel-hello-world dhirajsb/karaf4-sti camel-hello-world-sti
rh-wfh-dbokde:camel-hello-world dbokde$ sti rebuild --force-pull=false camel-hello-world-sti --loglevel=5
I0811 13:42:20.466769 60224 util.go:34] Using dhirajsb@gmail.com credentials for pulling camel-hello-world-sti
I0811 13:42:20.479539 60224 docker.go:199] Image camel-hello-world-sti:latest available locally
panic: runtime error: index out of range

goroutine 1 [running]:
main.func·004(0x208760ea0, 0x20875d5c0, 0x1, 0x3)
	/go/src/github.com/openshift/source-to-image/_output/local/go/src/github.com/openshift/source-to-image/cmd/sti/main.go:217 +0x694
github.com/spf13/cobra.(*Command).execute(0x208760ea0, 0x20875d500, 0x3, 0x3, 0x0, 0x0)
	/go/src/github.com/openshift/source-to-image/Godeps/_workspace/src/github.com/spf13/cobra/command.go:476 +0x65c
github.com/spf13/cobra.(*Command).Execute(0x208760820, 0x0, 0x0)
	/go/src/github.com/openshift/source-to-image/Godeps/_workspace/src/github.com/spf13/cobra/command.go:541 +0x18d
main.main()
	/go/src/github.com/openshift/source-to-image/_output/local/go/src/github.com/openshift/source-to-image/cmd/sti/main.go:366 +0x619

goroutine 5 [chan receive]:
github.com/golang/glog.(*loggingT).flushDaemon(0x68b1e0)
	/go/src/github.com/openshift/source-to-image/Godeps/_workspace/src/github.com/golang/glog/glog.go:839 +0x78
created by github.com/golang/glog.init·1
	/go/src/github.com/openshift/source-to-image/Godeps/_workspace/src/github.com/golang/glog/glog.go:406 +0x2a7

goroutine 7 [syscall]:
os/signal.loop()
	/usr/lib/golang/src/os/signal/signal_unix.go:21 +0x1f
created by os/signal.init·1
	/usr/lib/golang/src/os/signal/signal_unix.go:27 +0x35

goroutine 9 [runnable]:
[...]

PR #268, opened 2015-08-19
Fix arguments length for sti rebuild

nak3:


nak3:
Without this fix, `sti rebuild` outputs this error https://paste.fedoraproject.org/256563/

bparees:
lgtm [merge]

soltysh:
@mfojtik isn't this fix related to #257 ?

mfojtik:
@soltysh most likely yes :-)

bparees:
I don't think so. This was just a failure if you ran with no args. It
shouldn't change behavior for with args.

I mean related in that it's parameter processing for rebuild I suppose, but
probably not beyond that?

Ben Parees | OpenShift

PR #271, opened 2015-08-20
sti --run=true test

gabemontero:
@bparees here's a script for testing sti --run=true

In another git newbie instance, it also includes the cherry picked gofmt related change from test, but the change is the same, so it should be no harm, right?

I'm also working on the vagrant-openshift change to call this script, which of course will be a separate PR.

thanks

### DIFF
--- a/hack/test-stirunimage.sh
+++ b/hack/test-stirunimage.sh
@@ -42,7 +42,7 @@ sti build git://github.com/bparees/openshift-jee-sample openshift/wildfly-8-cent
 export STI_PID=$!
 TIME_SEC=1000
 TIME_MIN=$((60 * $TIME_SEC))
-max_wait=2*TIME_MIN
+max_wait=5*TIME_MIN
 echo "waiting up to ${max_wait}"
 expire=$(($(time_now) + $max_wait))
 
--- a/hack/test-stirunimage.sh
+++ b/hack/test-stirunimage.sh
@@ -1,0 +1,77 @@
+#!/bin/bash
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+STI_ROOT=$(dirname "${BASH_SOURCE}")/..
+
+function time_now()
+{
+  echo $(date +%s000)
+}
+
+function cleanup()
+{
+    set +e
+    rm -f  "${STI_ROOT}/hack/sti-run.log"
+    # use sigint so that sti post processing will remove docker container
+    if [ -z "$KILLDONE" ]; then
+	kill -2 "${STI_PID}"
+    fi
+    echo
+    echo "Complete"
+}
+
+set +e
+img_count=$(docker images | grep -c sti_test/sti-fake)
+set -e
+
+if [ "${img_count}" != "10" ]; then
+    echo "You do not have necessary test images, be sure to run 'hack/build-test-images.sh' beforehand."
+    exit 1
+fi
+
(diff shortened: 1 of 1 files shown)


hack/test-stirunimage.sh
```diff
@@ -0,0 +1,60 @@
...
+while [[ $(time_now) -lt $expire ]]; do
+    grep  "as a result of the --run=true option" "${STI_ROOT}/hack/sti-run.log"
+    if [ $? -eq 0 ]; then
+      set -e
+      echo "[INFO] Success running command sti --run=true"
+      rm -f  "${STI_ROOT}/hack/sti-run.log"
+      exit 0
+    fi
+    sleep 1
+done
```
bparees: would make sense to kill the pid here and then confirm the container also gets killed since that's part of the point of the signal handling.

bparees:
the cherry-picked commit would normally not be a problem except that it appears your cherry-picked commit id is different from the commit id that actually got merged.  not entirely sure why, my only theory is that in the original PR the commit was updated/force-pushed after you cherry-picked and before the PR was merged.

so as it is you'll need to rebase and delete the cherry-picked commit during the rebase.

regarding the PR itself....  i'm wary of adding a standalone shell script for a single test, but i guess we can live with it for now and refactor is we find ourselves adding more tests of this nature (which we probably should, someday... or move to the extended test model we're doing in openshift).

however in order to get this test run on test/merges you'll need to go edit the vagrant plugin so it actually invokes this script when testing PRs.  you can do that after this PR merges.

hack/test-stirunimage.sh
```diff
@@ -0,0 +1,60 @@
...
+
+echo "[INFO] Problem with sti --run=true, see ${STI_ROOT}/hack/sti-run.log"
```
bparees: might be worth just cat'ing the log here, makes life a little easier when it fails.

bparees:
2 comments but generally lgtm.  seems like you've mastered your shell-script fu :)

gabemontero:
re: cherry pick ... ahhhh, I was worried about that ... ok, I'm looking at various git man pages / doc, but I'll probably look for a consult before pulling the trigger on the next git operation ;-)

re: script vs. extended model ... yeah, this crossed my mind, but given how we landed here, and the lack of tests in general, I shied away from heavy investment at this time and went quick and cheap.  If you want to simply punt on this script at this time, let me know.

hack/test-stirunimage.sh
gabemontero: good point ... will include once I sort out the rebase

hack/test-stirunimage.sh
gabemontero: ok ... 1 quick follow up:  remove the log after we cat it ?

bparees: yeah, i'd just remove it in the existing cleanup method so it always gets removed.

gabemontero:
Oh, and as I noted in the description, yep, already started on the vagrant plugin, and will merge after this one does.

bparees:
whoops. glossed over that last sentence..cool.

gabemontero:
ok ... those couple of edits are now pushed (minus the cherry pick commit)

bparees:
lgtm [merge]

gabemontero:
[test]

bparees:
[merge]

PR #272, opened 2015-08-23
fix signal log message formatting

bparees:


bparees:
@gabemontero fyi, @nak3 fyi.
[merge]

gabemontero:
Another thought occurred to me this weekend after the initial review - glog was a potential *participant* in the `sti` hang from a few weeks ago.

Dumping the stack trace data to stdout, stderr and/or a separate file (perhaps with date/time as part of the file name) *could* be a consideration.

bparees:
had to remove the set -e so the grep for the container doesn't cause the test to fail.
[merge]

PR #276, opened 2015-08-24
Added support for go1.5

soltysh:
@bparees @mfojtik ptal

PR #277, opened 2015-08-25
Adding govet

soltysh:
@bparees this is post #276 I'm adding govet, though it'll not fail yet with errors, similarly to what we do with golint. I've also cleaned the `Makefile` a bit to follow origin patterns.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,6 +24,26 @@ all build:
 	hack/build-go.sh
 .PHONY: all build
 
+
+# Verify code is properly organized.
+#
+# Example:
+#   make verify
+verify: build
+	hack/verify-gofmt.sh
+	hack/verify-golint.sh || true
+	hack/verify-govet.sh || true
+.PHONY: verify
+
+
+# Install travis dependencies
+#
+# Example:
+#   make install-travis
+install-travis:
+	hack/install-tools.sh
+.PHONY: install-travis
+
 # Build and run tests.
 #
 # Args:
@@ -36,8 +56,8 @@ all build:
 #   make check
 #   make test
 #   make check WHAT=pkg/build GOFLAGS=-v
-check test:
-	hack/test-go.sh $(WHAT) $(TESTS)
+check test: verify
+	STI_COVER="-cover -covermode=atomic" STI_RACE="-race" hack/test-go.sh $(WHAT) $(TESTS)
 .PHONY: check test
 
 # Remove all build artifacts.
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,14 +5,12 @@ go:
   - 1.5
 
 install:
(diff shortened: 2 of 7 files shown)


Makefile
```diff
@@ -24,6 +24,26 @@ all build:
...
 
+
+# Verify code is properly organized.
+#
+# Example:
+#   make verify
+verify: build
+	hack/verify-gofmt.sh
+	hack/verify-golint.sh || true
+	hack/verify-govet.sh || true
```
bparees: why || true?  what's the point of the verify target if it doesn't fail when things don't verify?

Makefile
```diff
@@ -36,7 +56,9 @@ all build:
 #   make check
 #   make test
 #   make check WHAT=pkg/build GOFLAGS=-v
-check test:
+check-test: export STI_COVER= -cover -covermode=atomic
+check-test: export STI_RACE=  -race
+check-test: verify
```
bparees: why'd you change the name of the target?

bparees: why not just put STI_RACE and STI_COVER inline w/ the hack/test-go invocation?  why 3 targets?

Makefile
soltysh: Once I clean up the errors, I'll remove that.

soltysh: But that will be a separate PR.

Makefile
soltysh: I've copied that solution from origin/k8s, I'm not a makefile expert but a bit of googling gave me [Target-specific Variable Values](https://www.gnu.org/software/make/manual/html_node/Target_002dspecific.html), that's why we have separate for each value. But you're right the target name shouldn't be changed.

Makefile
```diff
@@ -36,7 +56,9 @@ all build:
 #   make check
 #   make test
 #   make check WHAT=pkg/build GOFLAGS=-v
-check test:
+check test: export STI_COVER= -cover -covermode=atomic
+check test: export STI_RACE=  -race
+check test: verify
```
bparees: I still dont' see why you don't just write:  
check test:
  STI_COVER=" - cover -covermode=atomic" STI_RACE=" -race" hack/test-go.sh $(WHAT) $(TESTS)

the example you linked to doesn't "export" the env variable and it's not clear you need to?

bparees: also your example is for Makefile variables, not shell variables... but you want STI_COVER to be a shell variable...right?

soltysh: I guess you're right. Fixed.

soltysh: At the same time, it also means we should fix that in origin, then.

bparees:
lgtm.. merge?

soltysh:
[merge]

PR #278, opened 2015-08-26
clean up waitgroup and attach logic

bparees:


bparees:
@soltysh @mfojtik @gabemontero @csrwng ptal.  i'm still testing this locally but i think it's the source of our hangs.

two issues it addresses:
1) waitgroup objects need to be passed by ref.
2) i didn't see any reason for containerAttachTwo?  containerAttachOne was already attaching stderr so it seemed redundant.  but i'd love a second opinion on that.

bparees:
apparently that second attach gets us the output of the assemble script..  putting it back in and testing if things still hang.  hopefully the real issue here is the wg pointer, which would make sense since it would coincide with the changes to add the "run after build" feature which is when i started noticing these hangs.

csrwng:
The reason we do it in 2 calls was because there was a bug in Docker attaching stdin, stdout and stderr all at once... you may want to remove the if/else in containerAttachOne to see if we can do without the 2 methods. Also call it something other than containerAttachOne if you're going to remove the other one :-)

bparees:
@csrwng yeah i saw the bug comment, but it still looked like attachone was attaching them all at once anyway.  (and yet based on the observed behavior when i removed it, that's apparently not the case)

bparees:
ok this seems to be sufficient to fix the hangs i was getting.

csrwng:
@bparees it's either or:
https://github.com/openshift/source-to-image/blob/master/pkg/docker/docker.go#L422-L428

csrwng:
LGTM

gabemontero:
lgtm

bparees:
[merge]

pkg/docker/docker.go
```diff
@@ -528,22 +506,11 @@ func (d *stiDocker) RunContainer(opts RunContainerOptions) (err error) {
 	defer d.RemoveContainer(container.ID)
 
 	glog.V(2).Infof("Attaching to container")
-	// creating / piping the channels in runContainerAttachOne lead to unintended hangs
+	// creating / piping the channels in runContainerAttach lead to unintended hangs
 	attached := make(chan struct{})
-	wg := runContainerAttachOne(attached, container, opts, d)
+	wg := runContainerAttach(attached, container, opts, d)
 	attached <- <-attached
```
liggitt: why are we feeding the output of the attached channel back into attached channel?

liggitt: nevermind, I read the docs

soltysh:
Nice job @bparees :+1:

PR #285, opened 2015-09-10
Initial support for 'file' SCM

mfojtik:
* Handle file:// protocol in source with fs.Copy()
* GIT is no longer required when using file://
* Initial work on supporting more SCM


mfojtik:
@soltysh @bparees PTAL

mfojtik:
[test]

soltysh:
LGTM [merge]

PR #286, opened 2015-09-10
Update docs for --recursive and file://

mfojtik:


soltysh:
LGTM [merge]

docs/cli.md
```diff
@@ -121,9 +122,14 @@ Build a Node.js application from a local directory, using a local image, the res
...
 
 ```
-$ s2i build --force-pull=false ~/nodejs-app local-nodejs-builder nodejs-app
+$ s2i build --force-pull=false /home/user/nodejs-app local-nodejs-builder nodejs-app
 ```
 
+In case of building from the local directory, the sources will be copied into
+the builder images using plain filesystem copy and the output image will not
+have the Git specific labels. Use this method only for development or local
+testing.
```
bparees: so even if the local dir is a git repo we won't get the labels?  seems like that's a loss of function.  if it's a repo and git is available, we ought to still do the labels...

mfojtik: @bparees we can lookup the git binary when making decisions about what SCM plugin to use. I will fix that.

PR #288, opened 2015-09-10
Prefer git SCM plugin over file when git binary exists

mfojtik:


mfojtik:
@bparees PTAL

bparees:
lgtm [merge]

PR #291, opened 2015-09-24
Expose the CreateTarStream function publicly

csrwng:
Also allows setting the exclusion pattern for tar and allows
specifying whether the source directory should be included in
the path for the tarred files.


csrwng:
@bparees ptal

bparees:
lgtm [merge]

csrwng:
@bparees sorry I had not updated the fake classes with the new interface
[merge] again

csrwng:
Attempting [merge] again with fixed integration tests

PR #293, opened 2015-09-28
Fix integration test by merging labels of committed image

csrwng:
Merges existing image labels with other labels when committing image.

### DIFF
--- a/hack/test-integration.sh
+++ b/hack/test-integration.sh
@@ -27,5 +27,5 @@ trap cleanup EXIT SIGINT
 echo
 echo Integration test cases ...
 echo
-export STI_TIMEOUT="-timeout 240s"
+export STI_TIMEOUT="-timeout 600s"
 "${STI_ROOT}/hack/test-go.sh" test/integration -tags 'integration' "${@:1}"
--- a/hack/test-stirunimage.sh
+++ b/hack/test-stirunimage.sh
@@ -42,7 +42,7 @@ sti build git://github.com/bparees/openshift-jee-sample openshift/wildfly-8-cent
 export STI_PID=$!
 TIME_SEC=1000
 TIME_MIN=$((60 * $TIME_SEC))
-max_wait=5*TIME_MIN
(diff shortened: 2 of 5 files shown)


csrwng:
[test]

csrwng:
@bparees @gabemontero ptal - Integration tests are currently broken. (See #292). Because nothing has changed in the s2i code since the last time something was merged (and succeeded integration tests), it looks like the way Docker handles labels passed in when committing an image has changed. The STI script location wasn't getting set in the committed image, causing save-artifacts to not be called properly. It seems that now  you have to merge labels explicitly when committing the image.

csrwng:
[test]

csrwng:
[test]

csrwng:
I've had to increment timeouts for tests so they can complete. @danmcp did the virtual hardware that these tests run on change?

danmcp:
@csrwng It has not.

bparees:
lgtm, squash?

gabemontero:
@csrwng ital - nothing specific to add to the code changes (and thx for the call out, I got exposure to yet another element of the system I was unaware of).

The only curiosity I have that could be construed as a comment is from a documentation perspective orthogonal to the test failure - would it may sense to document the fact that we add those 7 specific labels in `GenerateLabelsFromSourceInfo`, listing the precise name, and what they mean, etc. (I did not find any with some internet searching)?  

Would a customer conceivably want to leverage the information stored in any of those labels as part of using/possibly extending the image?

csrwng:
@gabemontero if we haven't documented the labels we're setting, we definitely should. I know @mfojtik looked into defining the set of labels used for STI images a while ago. @mfojtik do we have the current ones documented somewhere?

@bparees squashed, thx.

[merge]

csrwng:
jenkins -> aws flake
[merge] again

gabemontero:
thx @csrwng ... based on what @mfojtik says, if need be I can take the to-do of (learning/confirming, then) documenting those labels.

PR #295, opened 2015-09-30
Use unix-style path separators when creating tar files

csrwng:
When tarring windows paths, we store the windows-style path separators in the tar. If the tar is extracted on a unix system, the separators are not interpreted correctly. However, extracting a tar on a windows machine always works correctly with unix-style path separators.


csrwng:
@bparees ptal

bparees:
lgtm [merge]

PR #302, opened 2015-10-06
avoid scripts copy, fix scripts location, when no tar and no --script…

gabemontero:
…s-url, update unit tests accordingly

@bparees @mfojtik gentlemen, here is my PR for the no tar / no scripts specified bugzilla item test found.

@bparees - quick reminder: turns out, even test was going down the layered path, so no updates to the error parsing were needed to make things work;  I left things as is, but of course, if you think we still ought to switch simply looking at the error code, I can change accordingly.

thanks

### DIFF
--- a/pkg/build/strategies/layered/layered.go
+++ b/pkg/build/strategies/layered/layered.go
@@ -5,6 +5,8 @@ import (
 	"bytes"
 	"fmt"
 	"io"
+	"io/ioutil"
+	"os"
 	"path"
 	"path/filepath"
 	"time"
@@ -42,6 +44,7 @@ func New(config *api.Config, scripts build.ScriptsHandler, overrides build.Overr
 	}, nil
 }
 
+//getDestination returns the destination directory from the config
 func getDestination(config *api.Config) string {
 	destination := config.Destination
 	if len(destination) == 0 {
@@ -50,6 +53,17 @@ func getDestination(config *api.Config) string {
 	return destination
 }
 
+//checkValidDirWithContents will return true if the parameter provided is a valid, accessible directory that has contents (i.e. is not empty
+func checkValidDirWithContents(name string) bool {
+	items, err := ioutil.ReadDir(name)
+	if os.IsNotExist(err) {
+		glog.Warningf("Unable to access directory %q: %v", name, err)
+	}
+	return !(err != nil || len(items) == 0)
+}
+
+//CreateDockerfile takes the various inputs and creates the Dockerfile used by the docker cmd
+// to create the image produces by s2i
 func (b *Layered) CreateDockerfile(config *api.Config) error {
 	buffer := bytes.Buffer{}
 
@@ -64,14 +78,27 @@ func (b *Layered) CreateDockerfile(config *api.Config) error {
 	}
 
 	buffer.WriteString(fmt.Sprintf("FROM %s\n", b.config.BuilderImage))
-	buffer.WriteString(fmt.Sprintf("COPY scripts %s\n", locations[0]))
+	// only COPY scripts dir if required scripts are present, i.e. the dir is not empty;
+	// even if the "scripts" dir exists, the COPY would fail if it was empty
+	scriptsIncluded := checkValidDirWithContents(path.Join(config.WorkingDir, api.UploadScripts))
+	if scriptsIncluded {
+		glog.V(2).Infof("The scripts are included in %q directory", path.Join(config.WorkingDir, api.UploadScripts))
+		buffer.WriteString(fmt.Sprintf("COPY scripts %s\n", locations[0]))
+	} else {
+		// if an err on reading or opening dir, can't copy it
+		glog.V(2).Infof("Could not gather scripts from the directory %q", path.Join(config.WorkingDir, api.UploadScripts))
+	}
 	buffer.WriteString(fmt.Sprintf("COPY src %s\n", locations[1]))
 
 	//TODO: We need to account for images that may not have chown. There is a proposal
 	//      to specify the owner for COPY here: https://github.com/docker/docker/pull/9934
 	if len(user) > 0 {
 		buffer.WriteString("USER root\n")
-		buffer.WriteString(fmt.Sprintf("RUN chown -R %s %s %s\n", user, locations[0], locations[1]))
+		if scriptsIncluded {
+			buffer.WriteString(fmt.Sprintf("RUN chown -R %s %s %s\n", user, locations[0], locations[1]))
+		} else {
+			buffer.WriteString(fmt.Sprintf("RUN chown -R %s %s\n", user, locations[1]))
+		}
 		buffer.WriteString(fmt.Sprintf("USER %s\n", user))
 	}
 
@@ -84,6 +111,7 @@ func (b *Layered) CreateDockerfile(config *api.Config) error {
 }
 
 // TODO: this should stop generating a file, and instead stream the tar.
+//SourceTar returns a stream to the source tar file
 func (b *Layered) SourceTar(config *api.Config) (io.ReadCloser, error) {
 	uploadDir := filepath.Join(config.WorkingDir, "upload")
 	tarFileName, err := b.tar.CreateTarFile(b.config.WorkingDir, uploadDir)
@@ -93,6 +121,7 @@ func (b *Layered) SourceTar(config *api.Config) (io.ReadCloser, error) {
 	return b.fs.Open(tarFileName)
 }
 
+//Build handles the `docker build` equivalent execution, returning the success/failure details
 func (b *Layered) Build(config *api.Config) (*api.Result, error) {
 	if err := b.CreateDockerfile(config); err != nil {
 		return nil, err
@@ -141,8 +170,17 @@ func (b *Layered) Build(config *api.Config) (*api.Result, error) {
 	b.config.LayeredBuild = true
 	// new image name
 	b.config.BuilderImage = newBuilderImage
-	// the scripts are inside the image
-	b.config.ScriptsURL = "image://" + path.Join(getDestination(config), "scripts")
+	// see CreateDockerfile, conditional copy, location of scripts
+	scriptsIncluded := checkValidDirWithContents(path.Join(config.WorkingDir, api.UploadScripts))
+	glog.V(2).Infof("Scripts dir has contents %v", scriptsIncluded)
+	if scriptsIncluded {
+		b.config.ScriptsURL = "image://" + path.Join(getDestination(config), "scripts")
+	} else {
+		b.config.ScriptsURL, err = b.docker.GetScriptsURL(newBuilderImage)
+		if err != nil {
+			return nil, err
+		}
+	}
 
 	glog.V(2).Infof("Building %s using sti-enabled image", b.config.Tag)
 	if err := b.scripts.Execute(api.Assemble, config.AssembleUser, b.config); err != nil {
--- a/pkg/build/strategies/layered/layered_test.go
+++ b/pkg/build/strategies/layered/layered_test.go
@@ -2,6 +2,9 @@ package layered
 
 import (
 	"errors"
+	"io/ioutil"
+	"os"
+	"path/filepath"
 	"regexp"
 	"testing"
 
@@ -26,7 +29,50 @@ func newFakeLayered() *Layered {
 	}
 }
 
+func newFakeLayeredWithScripts(assemble, workDir string) *Layered {
+	return &Layered{
+		docker:  &docker.FakeDocker{},
+		config:  &api.Config{WorkingDir: workDir},
+		fs:      &test.FakeFileSystem{},
+		tar:     &test.FakeTar{},
+		scripts: &FakeExecutor{},
+	}
+}
+
 func TestBuildOK(t *testing.T) {
+	workDir, _ := ioutil.TempDir("", "sti")
+	scriptDir := filepath.Join(workDir, api.UploadScripts)
+	err := os.MkdirAll(scriptDir, 0700)
+	assemble := filepath.Join(scriptDir, api.Assemble)
+	file, err := os.Create(assemble)
+	if err != nil {
+		t.Errorf("Unexpected error returned: %v", err)
+	}
+	defer file.Close()
+	defer os.RemoveAll(workDir)
+	l := newFakeLayeredWithScripts(assemble, workDir)
+	l.config.BuilderImage = "test/image"
+	_, err = l.Build(l.config)
+	if err != nil {
+		t.Errorf("Unexpected error returned: %v", err)
+	}
+	if !l.config.LayeredBuild {
+		t.Errorf("Expected LayeredBuild to be true!")
+	}
+	if m, _ := regexp.MatchString(`test/image-\d+`, l.config.BuilderImage); !m {
+		t.Errorf("Expected BuilderImage test/image-withnumbers, but got %s ", l.config.BuilderImage)
+	}
+	// without config.Destination explicitly set, we should get /tmp/scripts for the scripts url
+	// assuming the assemble script we created above is off the working dir
+	if l.config.ScriptsURL != "image:///tmp/scripts" {
+		t.Errorf("Expected ScriptsURL image:///tmp/scripts, but got %s", l.config.ScriptsURL)
+	}
+	if len(l.config.Destination) != 0 {
+		t.Errorf("Unexpected Destination %s", l.config.Destination)
+	}
+}
+
+func TestBuildNoScriptsProvided(t *testing.T) {
 	l := newFakeLayered()
 	l.config.BuilderImage = "test/image"
 	_, err := l.Build(l.config)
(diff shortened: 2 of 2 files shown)


gabemontero:
[test]

gabemontero:
[test]

pkg/build/strategies/layered/layered.go
```diff
@@ -63,14 +64,24 @@ func (b *Layered) CreateDockerfile(config *api.Config) error {
 	}
 
 	buffer.WriteString(fmt.Sprintf("FROM %s\n", b.config.BuilderImage))
-	buffer.WriteString(fmt.Sprintf("COPY scripts %s\n", locations[0]))
+	// if -s / --scripts / --scripts-url are not specified, there will not be a scripts subdir
+	// off of the working dir, so don't copy here, and we'll have to tweak the scripts URL to use the env var
+	// STI_SCRIPTS_PATH
+	scriptsIncluded := len(config.ScriptsURL) > 0
```
bparees: there are two ways to inject scripts other than via the image itself.
1) scriptsurl (covered here)
2) scripts included in the source repo itself. seems like this check would break that flow?

pkg/build/strategies/layered/layered.go
```diff
@@ -139,8 +150,12 @@ func (b *Layered) Build(config *api.Config) (*api.Result, error) {
 	b.config.LayeredBuild = true
 	// new image name
 	b.config.BuilderImage = newBuilderImage
-	// the scripts are inside the image
-	b.config.ScriptsURL = "image://" + filepath.Join(getDestination(config), "scripts")
+	// the scripts are inside the image, but locatnoi may vary based on -s / --scripts / --scripts-url usage
```
bparees: s/locatnoi/location/

pkg/build/strategies/layered/layered.go
bparees: actually there is also a 3) scripts are included in the image at some other path and user specified --scriptsUrl image://some/path/in/the/image

which i suspect is also not handled by this change?

pkg/build/strategies/layered/layered_test.go
```diff
@@ -36,10 +47,32 @@ func TestBuildOK(t *testing.T) {
 		t.Errorf("Expected LayeredBuild to be true!")
 	}
 	if m, _ := regexp.MatchString(`test/image-\d+`, l.config.BuilderImage); !m {
-		t.Errorf("Expected BuilderImage test/image-withnumbers, but got %s", l.config.BuilderImage)
+		t.Errorf("Expected BuilderImage test/image-withnumbers, but got ", l.config.BuilderImage)
```
bparees: needs %s

pkg/build/strategies/layered/layered_test.go
```diff
@@ -36,10 +47,32 @@ func TestBuildOK(t *testing.T) {
...
 	}
 	if l.config.ScriptsURL != "image:///tmp/scripts" {
-		t.Error("Expected ScriptsURL image:///tmp/scripts, but got %s", l.config.ScriptsURL)
+		t.Error("Expected ScriptsURL image:///tmp/scripts, but got ", l.config.ScriptsURL)
```
bparees: needs %s and should be Errorf not Error

pkg/build/strategies/layered/layered.go
gabemontero: Yep, revisited the doc, and the three potential sources are listed [here](https://github.com/openshift/source-to-image/blob/master/docs/builder_image.md#s2i-scripts)

For reference, in order of precedence:
1. A script found at the `--scripts-url` URL
2. A script found in the application source `.sti/bin` directory
3. A script found at the default image URL (`io.openshift.s2i.scripts-url` label)

For 3), the value from the label is stored in the config.ScriptURL variable that is checked in layered.go.  Look [here](https://github.com/openshift/source-to-image/blob/master/pkg/build/config.go#L35).  So I think we are fine there, no?

For 2), I *think* if I read [this module](https://github.com/openshift/source-to-image/blob/master/pkg/scripts/install.go#L88) correctly, since the code is part of the source, it will get copied and referenced as part of the COPY src %s line.

I'll try constructing a .sti/bin example on Wednesday and either confirm that interpretation, or address if it goes elsewhere.

bparees: for (3) i'm not talking about a default value in the label, i'm talking about a case where the user explicitly invokes s2i with a "--scripts-url image://some/path" argument, which means they are specifying "use these external scripts" (case 1) but the scripts are really in the image, not downloaded from some external location, meaning you stlil still not have a "scripts" directory in the tar.

so it's a hybrid of case 1+3.  you can't rely on scriptsurl being len==0 because it will be non-zero but there will still be no scripts directory.  (I think.  i'm really just raising this as a "something to doublecheck", i'm not 100% certain on the behavior)

mfojtik: @bparees @gabemontero wouldn't be better to just check if the "./scripts" directory exists? If not, then do nothing about it and leave it up to image (STI_SCRIPTS_PATH)

pkg/build/strategies/layered/layered.go
```diff
@@ -139,8 +150,12 @@ func (b *Layered) Build(config *api.Config) (*api.Result, error) {
...
+	if len(config.ScriptsURL) > 0 {
+		b.config.ScriptsURL = "image://" + filepath.Join(getDestination(config), "scripts")
+	} else {
+		b.config.ScriptsURL = os.Getenv("STI_SCRIPTS_PATH")
```
mfojtik: does this run inside the container?

mfojtik: if there is no ./scripts directory locally, just skip settings this and leave it to default

pkg/build/strategies/layered/layered.go
gabemontero: @bparees - testing could be warranted, but when I look at https://github.com/openshift/source-to-image/blob/master/pkg/scripts/install.go#L113 and https://github.com/openshift/source-to-image/blob/master/pkg/scripts/download.go#L54-L63, I believe the result is that based on the `image` preface to the url, it will essentially copy from the image destination "scripts" directory that will be in the tar, so we would be ok there.  Or am I misreading this?

@mfojtik - just want to make sure I understand you:  you mean changing the `COPY` statement currently generated in the layered Dockerfile to a `RUN` that has a bash snippet that tests ./scripts existence before doing a cp, correct ?  If so, I can tell you I did the equivalent of that already, and it wasn't sufficient - we needed the conditional setting of ScriptsURL below based on the same conditional, which we would not have if done in the docker file.
If you mean checking literally for the presence of /tmp/..../scripts in the go code, I think the scripts dir is there when it isn't in the builder image (though not as sure there, and can double check if this is the path we want to go down.

gabemontero: all that said, still unclear / may have to adjust for the .sti/bin case - still need to try that one.

gabemontero: OK ... the light bulb finally lit ... better understand what is going on.  The error message from the Dockerfile COPY is misleading.  It is not that the ./scripts dir is missing.  It is that it is empty !!!

So to tweak the suggestion from @mfojtik, I switched from examining ScriptsURL to checking for the existence of the required `assemble` script in the scripts directory.  Using that to avoid the COPY scripts when appropriate, and analogously setting ScriptsURL later on to either the existing image:// url on the STI_SCRIPTS_PATH env var have things working.

Now, I'm curious about whether:
a) does anything ever show up in the scripts dir with the layered path
b) should we (at least when the scripts dir is empty) just star calling InstallRequired / InstallOptional https://github.com/openshift/source-to-image/blob/master/pkg/scripts/install.go#L50-L69 to pull the scripts from wherever (including .sti/bin)

gabemontero: ok the answer to the above a) is "yes", as proved out when I added `.sti/bin` to the source repo, hence the answer to b) is "no";  testing was successful with `.sti/bin`.

New push coming shortly.

gabemontero:
[test]

pkg/build/strategies/layered/layered.go
```diff
@@ -63,14 +64,25 @@ func (b *Layered) CreateDockerfile(config *api.Config) error {
 	}
 
 	buffer.WriteString(fmt.Sprintf("FROM %s\n", b.config.BuilderImage))
-	buffer.WriteString(fmt.Sprintf("COPY scripts %s\n", locations[0]))
+	// only COPY scripts dir if required scripts are present, i.e. the dir is not empty;
+	// even if the "scripts" dir exists, the COPY would fail if it was empty
+	_, err = os.Stat(filepath.Join(config.WorkingDir, api.UploadScripts, api.Assemble))
```
bparees: what if there's a run script only?

bparees: (or a save-artifacts only)

pkg/build/strategies/layered/layered_test.go
bparees: comment bump

pkg/build/strategies/layered/layered_test.go
```diff
@@ -36,10 +83,10 @@ func TestBuildOK(t *testing.T) {
 		t.Errorf("Expected LayeredBuild to be true!")
 	}
 	if m, _ := regexp.MatchString(`test/image-\d+`, l.config.BuilderImage); !m {
-		t.Errorf("Expected BuilderImage test/image-withnumbers, but got %s", l.config.BuilderImage)
+		t.Errorf("Expected BuilderImage test/image-withnumbers, but got ", l.config.BuilderImage)
 	}
-	if l.config.ScriptsURL != "image:///tmp/scripts" {
-		t.Error("Expected ScriptsURL image:///tmp/scripts, but got %s", l.config.ScriptsURL)
+	if l.config.ScriptsURL != "/usr/foo" {
+		t.Errorf("Expected ScriptsURL /usr/foo, but got %s ", l.config.ScriptsURL)
```
bparees: extraneous trailng space

pkg/build/strategies/layered/layered.go
gabemontero: I had interpreted the "InstallRequied" method name on the installer interface and meaning the script had to be provided if you were providing any scripts.

So a customer can provide only 1 of the required scripts through , and the others say provided by the defaults in STI_SCRIPTS_PATH ?

Yeah, looking back at the code, yes, I now see that it consolidates the scripts from the various potential sources.

OK, truly see if the dir is empty.  Found a src snippet on stackoverflow ... will go that route.

gabemontero:
[test]

gabemontero:
switch to seeing if dir has contents, etc. pushed

pkg/build/strategies/layered/layered.go
```diff
@@ -139,8 +167,14 @@ func (b *Layered) Build(config *api.Config) (*api.Result, error) {
...
 	b.config.BuilderImage = newBuilderImage
-	// the scripts are inside the image
-	b.config.ScriptsURL = "image://" + filepath.Join(getDestination(config), "scripts")
+	// see CreateDockerfile, conditional copy, location of scripts
+	scriptsIncluded := scriptsProvided(filepath.Join(config.WorkingDir, api.UploadScripts))
+	glog.V(2).Infof("Scripts dir has contents %v", scriptsIncluded)
+	if scriptsIncluded {
+		b.config.ScriptsURL = "image://" + filepath.Join(getDestination(config), "scripts")
+	} else {
+		b.config.ScriptsURL = os.Getenv("STI_SCRIPTS_PATH")
```
bparees: STI_SCRIPTS_PATH is an image-specific implementation detail/convenience.  the API for defining where the scripts exist within the image is the label:
https://github.com/openshift/sti-base/blob/master/Dockerfile#L8

I think we probably need to sit down for 10-15 minutes and actually define what the flows look like when one or more scripts come from the various potential locations.

In an ideal implementation, "scriptsIncluded" would always be true because no matter where we got them from (a user provided external url, the user's source repo, a user provided image:// url, or a url defined by the image itself via the aforementioned label) we would always go fetch them from there and copy them into the "scripts" directory in the working directory (peer to the source directory) and then run.

if no scripts are found at that point, it's an error because none were provided.

the key point is that:
https://github.com/openshift/sti-base/blob/master/Dockerfile#L8

is really just another url (it would be valid for that to be an http url), it's just a URL with the lowest/worst priority, with the files at that URL being used only if the user didn't explicilty specify a URL or provide scripts in their source dir.

gabemontero: OK, @bparees and I chatted off line.  He wants to switch gears on this one, and I'm amenable.  To simplify things, we'll move to getting rid of the "scripts are in the image" special case logic and always copy the scripts to the scripts working dir.  Then, we are back to always copying the working scripts dir in the layered docker file.

A couple of notes:
a) this parallels a PR @mfojtik has for being able to pull source-code out of an image ... so some synergy there
b) we'll want feedback from @soltysh on all this, since he originally did the "image reader" stuff

Priority wise, this is deemed lower priority for this sprint, but I'll spend a few cycles on this next week while it is all still fresh in my mind.

soltysh:
@gabemontero we need to talk about it, so I have a better understanding before reviewing it.

pkg/build/strategies/layered/layered.go
```diff
@@ -49,6 +50,22 @@ func getDestination(config *api.Config) string {
 	return destination
 }
 
+// if dir is empty, or not accessible, we say scripts are not provided
+func scriptsProvided(name string) bool {
```
mfojtik: use proper godoc format:

`scriptsProvided returns true if ...`

pkg/build/strategies/layered/layered.go
```diff
@@ -49,6 +50,22 @@ func getDestination(config *api.Config) string {
...
+	f, err := os.Open(name)
+	if err != nil {
+		return false
+	}
+	defer f.Close()
+
+	_, err = f.Readdir(1)
+	if err == io.EOF {
+		// empty - means no scripts provided
```
mfojtik: You just re-implemented ioutil.Readdir here :-) See: https://golang.org/src/io/ioutil/ioutil.go?s=3208:3259#L90

This should be utility function, something like:

```go
func checkEmptyDir(name string) {
  items, err := ioutil.ReadDir(name)
  if !isNotFound(err) {
    glog.Warningf("Unable to access directory %q: %v", name, err)
  }
  return !(err != nil || len(items) == 0)
}
```

Also I think we should report the errors here as the script might be provided but for some reason (permissions, corrupted filesystem, etc) we can't get them. So I would also log the error when then error is other than NotFound

pkg/build/strategies/layered/layered.go
```diff
@@ -49,6 +50,22 @@ func getDestination(config *api.Config) string {
...
+		return false
+	}
+	return true
+}
+
 func (b *Layered) CreateDockerfile(config *api.Config) error {
```
mfojtik: can you please add godoc here since you are here :-)

pkg/build/strategies/layered/layered.go
```diff
@@ -63,14 +80,25 @@ func (b *Layered) CreateDockerfile(config *api.Config) error {
 	}
 
 	buffer.WriteString(fmt.Sprintf("FROM %s\n", b.config.BuilderImage))
-	buffer.WriteString(fmt.Sprintf("COPY scripts %s\n", locations[0]))
+	// only COPY scripts dir if required scripts are present, i.e. the dir is not empty;
+	// even if the "scripts" dir exists, the COPY would fail if it was empty
+	scriptsIncluded := scriptsProvided(filepath.Join(config.WorkingDir, api.UploadScripts))
+	glog.V(2).Infof("Scripts dir has contents %v", scriptsIncluded)
```
mfojtik: This will just return boolean to log, we can do better logging :-)

```go
if scriptsIncluded {
  glog.V(2).Infof("The scripts are included in %q directory", filepath.Join(config.WorkingDir, api.UploadScripts))
}
```

Also please use `path.Join()` here as this can run on Windows and we don't want to use Windows filepath separator.

mfojtik:
@gabemontero needs rebase :-)

gabemontero:
thanks for the feedback @mfojtik ; I wonder though, before diving into them and referring to my comment on Oct 9 in this PR, and the change of direction @bparees wanted to do wrt always doing the copy, if this PR is a throwaway.  

 Something for use to perhaps discuss voice to voice when it makes sense to do so.

gabemontero:
I went ahead and rebased and included the comments from @mfojtik , even though the point from the comment form Oct 9 still holds in that @bparees wants to rework this and always do the copy / avoid the script dir stuff (pending a common conversation including @soltysh / @mfojtik minimally, if not as a part of the s2i redesign work discussed in devexp scrums).

pkg/build/strategies/layered/layered.go
```diff
@@ -141,8 +170,14 @@ func (b *Layered) Build(config *api.Config) (*api.Result, error) {
...
 	b.config.BuilderImage = newBuilderImage
-	// the scripts are inside the image
-	b.config.ScriptsURL = "image://" + path.Join(getDestination(config), "scripts")
+	// see CreateDockerfile, conditional copy, location of scripts
+	scriptsIncluded := checkValidDirWithContents(path.Join(config.WorkingDir, api.UploadScripts))
+	glog.V(2).Infof("Scripts dir has contents %v", scriptsIncluded)
+	if scriptsIncluded {
+		b.config.ScriptsURL = "image://" + path.Join(getDestination(config), "scripts")
+	} else {
+		b.config.ScriptsURL = os.Getenv("STI_SCRIPTS_PATH")
```
bparees: hm.  b.config.ScriptsURL should still be derived in the normal flow, meaning if the image has a label use that first, STI_SCRIPTS_PATH is considered deprecated.

bparees: (see docker.go:324 getScriptsURL())

gabemontero: pushed update ... using b.docker.GetScriptURL (which is tested in docker_test.go)

bparees:
[merge]

PR #304, opened 2015-10-08
Get the whole commitID hash

jhadvig:
Get the whole commit ID hash instead of the short version.

Based on conversation at the end of the https://github.com/openshift/origin/issues/615 issue.

@bparees @smarterclayton PTAL


smarterclayton:
LGTM

mfojtik:
[merge]

PR #307, opened 2015-10-13
[#290] s2i build --incremental=true: fix restore of symblic links.

### LINKED ISSUE #290: Symlink artifacts extracted as zero byte files
In reference to Bug #87 , symbolic links are restored only as zero byte files. I have setup a small github project `https://github.com/nitintutlani/s2iSymlinks` to help reproduce this:

First time build:
```
$ s2i build --incremental=false --force-pull=false https://github.com/nitintutlani/s2iSymlinks centos:centos7 test --loglevel=5
```
The assemble script generates two files `src/content.txt` and `src/content.lnk`. Resulting image works well.
```
$ docker run --rm -it test cat /tmp/src/content.lnk
Content shown from symbolic link
$ docker run --rm -it test
Content shown from symbolic link
```
content.txt has valid content:
```
$ docker run --rm -it test cat /tmp/src/content.txt
```

The `save-artifacts` should preserve two files content.txt and content.lnk.

Subsequent build:
```
$ s2i build --incremental=true --force-pull=false https://github.com/nitintutlani/s2iSymlinks centos:centos7 test --loglevel=5 --save-temp-dir=true
```
This time fails.
```
$ docker run --rm -it test cat /tmp/src/content.lnk
$ docker run --rm -it test
```
However content.txt is restored well:
```
$ docker run --rm -it test cat /tmp/src/content.txt
```
Npm install, composer install and many other time taking process produce symbolic links. Please review and guide how to make artifacts work properly.

php-coder:
Here is my (first version of) fix for symbolic links (fixes #290).

PTAL @rhcarvalho 

I want to discuss the following questions:

1) Currently we are supporting only directories, files and now symbolic links. What about other file types? For instance, hard links, devices, etc? Because we don't have any special code that work with it we probably will have bugs when someone will put such files inside an archive.

2) I tried to set atime/mtime of symbolic link when it's restored (like we do for files) but with no luck. We can't use `os.Chtimes()` because it doesn't work with symbolic links and I didn't find a function for that. If we decided to restore mtime/atime for symlinks we should use syscall for that. Probably, this link may be interesting: http://grokbase.com/t/gg/golang-nuts/154wnph4y8/go-nuts-no-way-to-utimes-a-symlink

3) Is it possible to have a security issue there? Or it shouldn't because we're running this code inside a container?

4) I feel that I should add a unit test or even integration test. Should I? May be there is similar tests in our code base?

TIA

### DIFF
--- a/pkg/tar/tar.go
+++ b/pkg/tar/tar.go
@@ -180,7 +180,7 @@ func (t *stiTar) ExtractTarStream(dir string, reader io.Reader) error {
 			if header.FileInfo().IsDir() {
 				dirPath := filepath.Join(dir, header.Name)
 				if err = os.MkdirAll(dirPath, 0700); err != nil {
-					glog.Errorf("Error creating dir %s: %v", dirPath, err)
+					glog.Errorf("Error creating dir %q: %v", dirPath, err)
 					errorChannel <- err
 					break
 				}
@@ -188,13 +188,22 @@ func (t *stiTar) ExtractTarStream(dir string, reader io.Reader) error {
 				fileDir := filepath.Dir(header.Name)
 				dirPath := filepath.Join(dir, fileDir)
 				if err = os.MkdirAll(dirPath, 0700); err != nil {
-					glog.Errorf("Error creating dir %s: %v", dirPath, err)
+					glog.Errorf("Error creating dir %q: %v", dirPath, err)
 					errorChannel <- err
 					break
 				}
+				if header.Mode&tar.TypeSymlink == tar.TypeSymlink {
+					if err := extractLink(dir, header, tarReader); err != nil {
+						glog.Errorf("Error extracting link %q: %v", header.Name, err)
+						errorChannel <- err
+						break
+					}
+					continue
+				}
 				if err := extractFile(dir, header, tarReader); err != nil {
-					glog.Errorf("Error extracting file %s: %v", header.Name, err)
+					glog.Errorf("Error extracting file %q: %v", header.Name, err)
 					errorChannel <- err
+					break
 				}
 			}
 		}
@@ -213,6 +222,16 @@ func (t *stiTar) ExtractTarStream(dir string, reader io.Reader) error {
 			return errors.NewTarTimeoutError()
 		}
 	}
+}
+
+func extractLink(dir string, header *tar.Header, tarReader io.Reader) error {
+	dest := filepath.Join(dir, header.Name)
+	source := header.Linkname
+
+	glog.V(3).Infof("Creating symbolic link from %q to %q", dest, source)
+
+	// TODO: set mtime for symlink (unfortunately we can't use os.Chtimes() and probably should use syscall)
+	return os.Symlink(source, dest)
 }
 
 func extractFile(dir string, header *tar.Header, tarReader io.Reader) error {
@@ -235,10 +254,5 @@ func extractFile(dir string, header *tar.Header, tarReader io.Reader) error {
 	if written != header.Size {
 		return fmt.Errorf("Wrote %d bytes, expected to write %d", written, header.Size)
 	}
-	if err = file.Chmod(header.FileInfo().Mode()); err != nil {
-		return err
-	}
-
-	glog.V(3).Infof("Done with %s", path)
-	return nil
+	return file.Chmod(header.FileInfo().Mode())
 }
--- a/pkg/tar/tar_test.go
+++ b/pkg/tar/tar_test.go
@@ -2,6 +2,7 @@ package tar
 
 import (
 	"archive/tar"
+	"fmt"
 	"io"
 	"io/ioutil"
 	"os"
@@ -19,6 +20,7 @@ type fileDesc struct {
 	mode         os.FileMode
 	content      string
 	shouldSkip   bool
+	target       string
 }
 
 type linkDesc struct {
@@ -129,10 +131,10 @@ func TestCreateTarStreamIncludeParentDir(t *testing.T) {
 	}
 	modificationDate := time.Date(2011, time.March, 5, 23, 30, 1, 0, time.UTC)
 	testFiles := []fileDesc{
-		{"dir01/dir02/test1.txt", modificationDate, 0700, "Test1 file content", false},
-		{"dir01/test2.git", modificationDate, 0660, "Test2 file content", false},
-		{"dir01/dir03/test3.txt", modificationDate, 0444, "Test3 file content", false},
-		{"dir01/.git/hello.txt", modificationDate, 0600, "Ignore file content", true},
+		{"dir01/dir02/test1.txt", modificationDate, 0700, "Test1 file content", false, ""},
+		{"dir01/test2.git", modificationDate, 0660, "Test2 file content", false, ""},
+		{"dir01/dir03/test3.txt", modificationDate, 0444, "Test3 file content", false, ""},
+		{"dir01/.git/hello.txt", modificationDate, 0600, "Ignore file content", true, ""},
 	}
 	if err := createTestFiles(tempDir, testFiles); err != nil {
 		t.Fatalf("Cannot create test files: %v", err)
@@ -164,10 +166,10 @@ func TestCreateTar(t *testing.T) {
 	}
 	modificationDate := time.Date(2011, time.March, 5, 23, 30, 1, 0, time.UTC)
 	testFiles := []fileDesc{
-		{"dir01/dir02/test1.txt", modificationDate, 0700, "Test1 file content", false},
-		{"dir01/test2.git", modificationDate, 0660, "Test2 file content", false},
-		{"dir01/dir03/test3.txt", modificationDate, 0444, "Test3 file content", false},
-		{"dir01/.git/hello.txt", modificationDate, 0600, "Ignore file content", true},
+		{"dir01/dir02/test1.txt", modificationDate, 0700, "Test1 file content", false, ""},
+		{"dir01/test2.git", modificationDate, 0660, "Test2 file content", false, ""},
+		{"dir01/dir03/test3.txt", modificationDate, 0444, "Test3 file content", false, ""},
+		{"dir01/.git/hello.txt", modificationDate, 0600, "Ignore file content", true, ""},
 	}
 	if err := createTestFiles(tempDir, testFiles); err != nil {
 		t.Fatalf("Cannot create test files: %v", err)
@@ -190,23 +192,60 @@ func TestCreateTar(t *testing.T) {
 	verifyTarFile(t, tarFile, testFiles, testLinks)
 }
 
-func createTestTar(files []fileDesc, writer io.Writer) {
+func createTestTar(files []fileDesc, writer io.Writer) error {
 	tw := tar.NewWriter(writer)
+	defer tw.Close()
 	for _, fd := range files {
-		contentBytes := []byte(fd.content)
-		hdr := &tar.Header{
-			Name:       fd.name,
-			Mode:       int64(fd.mode),
-			Size:       int64(len(contentBytes)),
-			Typeflag:   tar.TypeReg,
-			AccessTime: time.Now(),
-			ModTime:    fd.modifiedDate,
-			ChangeTime: fd.modifiedDate,
+		if isSymLink(fd.mode) {
+			if err := addSymLink(tw, &fd); err != nil {
+				msg := "unable to add symbolic link %q (points to %q) to archive: %v"
+				return fmt.Errorf(msg, fd.name, fd.target, err)
+			}
+			continue
 		}
-		tw.WriteHeader(hdr)
-		tw.Write(contentBytes)
+		if err := addRegularFile(tw, &fd); err != nil {
+			return fmt.Errorf("unable to add file %q to archive: %v", fd.name, err)
+		}
 	}
-	tw.Close()
+	return nil
+}
+
+func addRegularFile(tw *tar.Writer, fd *fileDesc) error {
+	contentBytes := []byte(fd.content)
+	hdr := &tar.Header{
+		Name:       fd.name,
+		Mode:       int64(fd.mode),
+		Size:       int64(len(contentBytes)),
+		Typeflag:   tar.TypeReg,
+		AccessTime: time.Now(),
+		ModTime:    fd.modifiedDate,
+		ChangeTime: fd.modifiedDate,
+	}
+	if err := tw.WriteHeader(hdr); err != nil {
+		return err
+	}
+	_, err := tw.Write(contentBytes)
+	return err
+}
+
+func addSymLink(tw *tar.Writer, fd *fileDesc) error {
+	if len(fd.target) == 0 {
+		return fmt.Errorf("link %q must point to somewhere, but target wasn't defined", fd.name)
+	}
+
+	hdr := &tar.Header{
+		Name:     fd.name,
+		Linkname: fd.target,
+		Mode:     int64(fd.mode & os.ModePerm),
+		Typeflag: tar.TypeSymlink,
+		ModTime:  fd.modifiedDate,
+	}
+
+	return tw.WriteHeader(hdr)
+}
+
+func isSymLink(mode os.FileMode) bool {
+	return mode&os.ModeSymlink == os.ModeSymlink
 }
 
 func verifyDirectory(t *testing.T, dir string, files []fileDesc) {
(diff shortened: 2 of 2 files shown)


rhcarvalho:
> 1\. What about other file types? For instance, hard links, devices, etc?

I think we're not trying to handle all possible inputs here.

> 2\. set atime/mtime of symbolic link when it's restored

I believe we want to do that, but I'm not familiar with this code base. Better ask @csrwng or @mfojtik who worked on this file before.

> 4\. I feel that I should add a unit test or even integration test.

Yes, adding a test for this will prevent us from breaking it in the future, we want a test in this PR. I'm not familiar with existing tests to give you a precise link.

php-coder:
I've test fix against more sophisticated cases (broken link, link inside directory that refers to file in top-level directory). Looks like it work:

```
[vagrant@openshiftdev ~]$ docker run --rm -it test2 find /tmp/src -ls
105235    0 drwxr-xr-x   3 root     root          140 Oct 13 14:23 /tmp/src
105243    0 lrwxrwxrwx   1 root     root           11 Oct 13 14:23 /tmp/src/content.lnk -> content.txt
105242    4 -rw-r--r--   1 root     root           33 Oct 13 14:23 /tmp/src/content.txt
105241    4 -rw-rw-r--   1 1000     1000           88 Oct 13 14:23 /tmp/src/README.md
105237    0 drwxr-xr-x   3 root     root           60 Oct 13 14:23 /tmp/src/nested_dir
105238    0 drwxr-xr-x   3 root     root           60 Oct 13 14:23 /tmp/src/nested_dir/dir1
105239    0 drwxr-xr-x   2 root     root           60 Oct 13 14:23 /tmp/src/nested_dir/dir1/dir2
105253    0 lrwxrwxrwx   1 root     root           20 Oct 13 14:23 /tmp/src/nested_dir/dir1/dir2/content.lnk -> ../../../content.txt
105236    0 lrwxrwxrwx   1 root     root           14 Oct 13 14:23 /tmp/src/bad_link -> does-not-exist
```

php-coder:
@mfojtik Could you review this PR and especially I need advice about what to do with modification time of symlinks. Thanks!

pkg/tar/tar_test.go
```diff
@@ -185,23 +187,60 @@ func TestCreateTar(t *testing.T) {
 	verifyTarFile(t, tarFile, testFiles, testLinks)
 }
 
-func createTestTar(files []fileDesc, writer io.Writer) {
+func createTestTar(files []fileDesc, writer io.Writer, t *testing.T) {
```
rhcarvalho: Since `createTestTar` can "fail", the common idiom in Go is to return an error.
So I suggest:

```diff
-func createTestTar(files []fileDesc, writer io.Writer, t *testing.T)
+func createTestTar(files []fileDesc, writer io.Writer) error
```

Then the caller handles the error.

pkg/tar/tar_test.go
```diff
@@ -185,23 +187,60 @@ func TestCreateTar(t *testing.T) {
...
+					fd.name, fd.target, err,
+				)
+			}
+		} else {
+			addRegularFile(tw, &fd)
 		}
-		tw.WriteHeader(hdr)
-		tw.Write(contentBytes)
 	}
-	tw.Close()
```
rhcarvalho: perhaps more like trivia, it's good to know that all three `tw.Close`, `tw.WriteHeader` and `tw.Write` all return error.

pkg/tar/tar_test.go
```diff
@@ -185,23 +187,60 @@ func TestCreateTar(t *testing.T) {
...
+	hdr := &tar.Header{
+		Name:       fd.name,
+		Mode:       int64(fd.mode),
+		Size:       int64(len(contentBytes)),
+		Typeflag:   tar.TypeReg,
+		AccessTime: time.Now(),
+		ModTime:    fd.modifiedDate,
+		ChangeTime: fd.modifiedDate,
+	}
+	tw.WriteHeader(hdr)
```
rhcarvalho: why here we are not checking the error returned by WriteHeader while we do a few lines below?
https://github.com/openshift/source-to-image/pull/307/files#diff-203bceaa948b8f85722a4d86502613f6R235

pkg/tar/tar.go
```diff
@@ -192,9 +192,16 @@ func (t *stiTar) ExtractTarStream(dir string, reader io.Reader) error {
...
 				}
-				if err := extractFile(dir, header, tarReader); err != nil {
-					glog.Errorf("Error extracting file %s: %v", header.Name, err)
-					errorChannel <- err
+				if header.Mode&tar.TypeSymlink == tar.TypeSymlink {
+					if err := extractLink(dir, header, tarReader); err != nil {
+						glog.Errorf("Error extracting link %s: %v", header.Name, err)
+						errorChannel <- err
+					}
+				} else {
```
mfojtik: no need for else here, just return from the if above.

pkg/tar/tar.go
```diff
@@ -215,6 +222,22 @@ func (t *stiTar) ExtractTarStream(dir string, reader io.Reader) error {
 	}
 }
 
+func extractLink(dir string, header *tar.Header, tarReader io.Reader) error {
+	dest := filepath.Join(dir, header.Name)
+	source := header.Linkname
+
+	glog.V(3).Infof("Creating symbolic link from %s to %s", dest, source)
```
mfojtik: use %q here which will quote the string... it is better when the filename has something weird in it (like space ;-) and also better for copy&paste.

pkg/tar/tar.go
```diff
@@ -215,6 +222,22 @@ func (t *stiTar) ExtractTarStream(dir string, reader io.Reader) error {
...
+
+	if err := os.Symlink(source, dest); err != nil {
```
mfojtik: just `return os.Symlink(source, dest)` here, no need for the glog below (or if you really want to do glog, make it as `defer`

pkg/tar/tar_test.go
```diff
@@ -185,23 +187,60 @@ func TestCreateTar(t *testing.T) {
...
-			ModTime:    fd.modifiedDate,
-			ChangeTime: fd.modifiedDate,
+		if isSymLink(fd.mode) {
+			if err := addSymLink(tw, &fd, t); err != nil {
+				t.Fatalf(
+					"Unable to add symbolic link %s (that points to %s) to archive: %v",
+					fd.name, fd.target, err,
+				)
+			}
+		} else {
```
mfojtik: no need for else, Fatalf exit

pkg/tar/tar_test.go
php-coder: It's because I tried to don't touch original code (I only extracted it to the separate method). But, I agree with you and I will add check here too. Thank you!

pkg/tar/tar_test.go
```diff
@@ -185,23 +187,60 @@ func TestCreateTar(t *testing.T) {
...
+		Mode:     int64(fd.mode & os.ModePerm),
+		Typeflag: tar.TypeSymlink,
+		ModTime:  fd.modifiedDate,
+	}
+
+	if err := tw.WriteHeader(hdr); err != nil {
+		return err
+	}
+
+	return nil
```
mfojtik: no need to return nil, just return the `tw.WriteHeader`

mfojtik:
@php-coder couple small coding nits but the rest LGTM. Also please squash the commits into one before merging.

pkg/tar/tar_test.go
mfojtik: @php-coder it is fine to refactor bad code :-)

pkg/tar/tar.go
rhcarvalho: I don't know why we had these glogs here, but just noting that if there's an error, we did not print "Done with...", while if we use defer that way the message would be printed unconditionally.

pkg/tar/tar_test.go
php-coder: As far I understand this method doesn't return `error` and when it fails it do it himself by calling `r.Fatalf()`. I don't understand why I need to return `error`? Could you explain, please?

pkg/tar/tar_test.go
rhcarvalho: this else refers to not being a symlink. It seems correct to me.

pkg/tar/tar.go
mfojtik: i would prefer to not print anything ;-) we don't need to add more verbosity to places where we don't need it.

pkg/tar/tar_test.go
mfojtik:
```
if !isSymlink(fd.mode) { addRegularFile(rw, &fd); continue }
if err := addSymlink(tw, &fd, t); err != nil {
  t.Fatal()
}

Comment (mfojtik) [inline pkg/tar/tar_test.go]: @rhcarvalho else's are ugly :-)

File: pkg/tar/tar_test.go (hunk 2 above)
Comment (php-coder) [inline pkg/tar/tar_test.go]: How to do it for `tw.Close` that is `defer`-ed?

File: pkg/tar/tar_test.go (hunk 8 above)
Comment (php-coder) [inline pkg/tar/tar_test.go]: Is it possible to to the same with the following code?

```go
	if _, err := tw.Write(contentBytes); err != nil {
		return err
	}

	return nil
```

mfojtik:
```go
return tw.Write(contentBytes)
```

Since the function just return the err the if check does not make sense ;-) If there is no err, then the function will return nil, right?

php-coder: Right, but how it works for functions that returns 2 values? Which of them will be used?

pkg/tar/tar.go
php-coder: Didn't know about `%q`, thanks! Is it `glog` specific modifier or it's something standard?

pkg/tar/tar_test.go
rhcarvalho: In this context I'd say it's overkill, and `defer tw.Close()`, ignoring the close error, is fine.

But there are options discussed here: https://groups.google.com/d/topic/golang-nuts/57t7_sZcyDw/discussion

pkg/tar/tar_test.go
mfojtik: @php-coder oh sorry, I missed the `_` there, you are right. You can't do that with function that returns 2 values, you can do this:

```go
_, err := tw.Write()
return err
```

pkg/tar/tar_test.go
rhcarvalho: It would be more idiomatic if, instead of passing `t *testing.T` as an argument to `createTestTar` and have `createTestTar` call `t.Fatal*`, `createTestTar` would return an error and the caller (`Test*` function) handles the error and call `t.Fatal*`/`t.Error*` as appropriate.

pkg/tar/tar_test.go
rhcarvalho: @php-coder this pattern applies when the function signatures are the same. E.g.:

```go
func foo(x, y int) (int, error) {
	if y == 1 {
		return 0, fmt.Errorf("bad value 1")
	}
	return x + y, nil
}

func bar(z int) (int, error) {
	if z == 2 {
		return 0, fmt.Errorf("bad value 2")
	}
	// simply return _foo_, because it has
	// the same signature of _bar_
	return foo(1, z)
}
```

pkg/tar/tar.go
rhcarvalho: Standard: https://golang.org/pkg/fmt/

pkg/tar/tar_test.go
php-coder: Oh, I see what you mean. Ok, I'll change it, thanks!

php-coder:
@rhcarvalho @mfojtik Thanks for a feedback! I've fixed all of your comments (and squash commits to one).

mfojtik:
[test]

pkg/tar/tar_test.go
```diff
@@ -236,6 +278,18 @@ func verifyDirectory(t *testing.T, dir string, files []fileDesc) {
...
+				if isSymLink(fd.mode) {
+					target, err := os.Readlink(path)
+					if err != nil {
+						return err
+					}
+					if target != fd.target {
+						t.Errorf(
+							"Symbolic link %q points to wrong path. Expected: %s, Actual: %s\n",
+							fd.name, fd.target, target,
+						)
```
rhcarvalho: just a small nit, I think it's more idiomatic in Go to have this as a single (long) line than write it in 4 lines. Don't know if we have project-specific preferences in this regard. @mfojtik any stylistic opinion?

php-coder: Do we have limitation for line width?

mfojtik: I have vim window set to 80 characters, long lines are annoying for me :) Uber-idiomatic would be to create special error and pass these three as arguments, but that is too much for this.

pkg/tar/tar_test.go
```diff
@@ -190,23 +192,62 @@ func TestCreateTar(t *testing.T) {
...
+	if err := tw.WriteHeader(hdr); err != nil {
+		return err
+	}
+	_, err := tw.Write(contentBytes)
+	return err
+}
+
+func addSymLink(tw *tar.Writer, fd *fileDesc) error {
+	if len(fd.target) == 0 {
+		return fmt.Errorf("Link %q must points to somewhere, but it points to nothing", fd.name)
```
rhcarvalho: "must point", or even make it more succinct "link %q points to an invalid target"

pkg/tar/tar_test.go
```diff
@@ -190,23 +192,62 @@ func TestCreateTar(t *testing.T) {
...
-			Typeflag:   tar.TypeReg,
-			AccessTime: time.Now(),
-			ModTime:    fd.modifiedDate,
-			ChangeTime: fd.modifiedDate,
+		if isSymLink(fd.mode) {
+			if err := addSymLink(tw, &fd); err != nil {
+				return fmt.Errorf(
+					"Unable to add symbolic link %q (that points to %q) to archive: %v",
+					fd.name, fd.target, err,
+				)
```
rhcarvalho: same stylistic note as below

(btw -- this would be idiomatic in Python :-) )

pkg/tar/tar_test.go
rhcarvalho: https://github.com/golang/go/wiki/CodeReviewComments#line-length

pkg/tar/tar_test.go
php-coder: @rhcarvalho I will rephrase it to "Link %q must point to somewhere, but target wasn't defined" if you don't mind.

rhcarvalho:
LGTM

pkg/tar/tar_test.go
csrwng: @php-coder just a couple of things... the error should start with lowercase. It's also more grammatically correct to say "link %q must point somewhere... ". Sorry if my comment is late :(

pkg/tar/tar_test.go
php-coder: @rhcarvalho @mfojtik I've corrected it my introducing local variable (and removing one word from message).

csrwng:
Please change error messages to start with lowercase

php-coder:
@csrwng Should I change all errors or only some of them (leave as is in tests, for instance)?

csrwng:
@php-coder preferably for all, but it's more important for the regular code errors.

php-coder:
@csrwng Thank, I've fixed it. But all of them were in tests.

pkg/tar/tar.go
```diff
@@ -192,9 +192,18 @@ func (t *stiTar) ExtractTarStream(dir string, reader io.Reader) error {
...
+					if err := extractLink(dir, header, tarReader); err != nil {
+						glog.Errorf("Error extracting link %q: %v", header.Name, err)
+						errorChannel <- err
+						break
+					}
+					continue
+				}
 				if err := extractFile(dir, header, tarReader); err != nil {
-					glog.Errorf("Error extracting file %s: %v", header.Name, err)
+					glog.Errorf("Error extracting file %q: %v", header.Name, err)
```
rhcarvalho: since you're changing this from %s to %q, could you do the same for the surrounding code?
https://github.com/openshift/source-to-image/pull/307/files#diff-6b4b95c3fb2a650ac8b9a0c5eabdb8b3R183
and
https://github.com/openshift/source-to-image/pull/307/files#diff-6b4b95c3fb2a650ac8b9a0c5eabdb8b3R191

pkg/tar/tar_test.go
rhcarvalho: maybe you could do here the same as in [here](https://github.com/openshift/source-to-image/pull/307/files#diff-203bceaa948b8f85722a4d86502613f6R201)?

pkg/tar/tar.go
php-coder: Yes, I've just pushed an update.

pkg/tar/tar_test.go
php-coder: @rhcarvalho Ok, fixed.

mfojtik:
[merge]

bparees:
[merge]

PR #310, opened 2015-10-14
Allow S2I behavior to be overriden by caller

smarterclayton:
Origin needs to be able to download and assemble the source
itself and then pass it to s2i. This commit adds changes to
allow an Overrides object to be passed to the s2i initializer
code to provide a custom Downloader.

This also makes changes to allow better reuse of upstream s2i
code and removes a write to disk for the intermediate tar (which
can be streamed instead).

@mfojtik @bparees review, this is part of binary build

### DIFF
--- a/pkg/build/strategies/onbuild/onbuild.go
+++ b/pkg/build/strategies/onbuild/onbuild.go
@@ -38,7 +38,7 @@ type onBuildSourceHandler struct {
 }
 
 // New returns a new instance of OnBuild builder
-func New(config *api.Config) (*OnBuild, error) {
+func New(config *api.Config, overrides build.Overrides) (*OnBuild, error) {
 	dockerHandler, err := docker.New(config.DockerConfig, config.PullAuthentication)
 	if err != nil {
 		return nil, err
@@ -50,15 +50,25 @@ func New(config *api.Config) (*OnBuild, error) {
 		tar:    tar.New(),
 	}
 	// Use STI Prepare() and download the 'run' script optionally.
-	s, err := sti.New(config)
+	s, err := sti.New(config, overrides)
 	s.SetScripts([]string{}, []string{api.Assemble, api.Run})
-	downloader, sourceUrl, err := scm.DownloaderForSource(config.Source)
-	if err != nil {
-		return nil, err
-	}
-	config.Source = sourceUrl
 
-	b.source = onBuildSourceHandler{downloader, s, &ignore.DockerIgnorer{}}
+	downloader := overrides.Downloader
+	if downloader == nil {
+		d, sourceURL, err := scm.DownloaderForSource(config.Source)
+		if err != nil {
+			return nil, err
+		}
+		downloader = d
+		config.Source = sourceURL
+	}
+
+	b.source = onBuildSourceHandler{
+		Downloader: downloader,
+		Preparer:   s,
+		Ignorer:    &ignore.DockerIgnorer{},
+	}
+
 	b.garbage = &build.DefaultCleaner{b.fs, b.docker}
 	return b, nil
 }
--- a/pkg/build/strategies/sti/sti.go
+++ b/pkg/build/strategies/sti/sti.go
@@ -6,6 +6,7 @@ import (
 	"path/filepath"
 	"regexp"
 	"strings"
+	"sync"
 
 	"github.com/golang/glog"
 	"github.com/openshift/source-to-image/pkg/api"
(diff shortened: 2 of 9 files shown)


pkg/build/strategies/onbuild/onbuild.go
```diff
@@ -50,13 +50,24 @@ func New(config *api.Config) (*OnBuild, error) {
...
+	if downloader == nil {
+		d, sourceURL, err := scm.DownloaderForSource(config.Source)
+		if err != nil {
+			return nil, err
+		}
+		downloader = d
+		config.Source = sourceURL
+	}
+
+	b.source = onBuildSourceHandler{
```
soltysh: this is dupe of line 61

soltysh:
One small nit, but I'd like to see tests for that please.

bparees:
Assuming this hasn't changed much since I reviewed it in your origin PR I'm
ok with it.

Otherwise it's going to need to wait until Monday when I have laptop access
again (I guess android just isn't as good at doing code reviews as iOS...)

But +1 to TCs.

Ben Parees | OpenShift

pkg/build/strategies/onbuild/onbuild.go
smarterclayton: Not sure what you mean - if no override is provided, we now create a default downloader and set it.  b.source is only set once.

smarterclayton:
Why does STI execute take an api.Config that might be different from the api.Config passed to it on creation?

smarterclayton:
Tests added.

smarterclayton:
Ok, race averted.

pkg/build/strategies/onbuild/onbuild.go
```diff
@@ -50,13 +50,24 @@ func New(config *api.Config) (*OnBuild, error) {
...
+		downloader,
+		s,
+		&ignore.DockerIgnorer{},
 	}
-	config.Source = sourceUrl
 
 	b.source = onBuildSourceHandler{downloader, s, &ignore.DockerIgnorer{}}
```
bparees: it is duping this line isn't it?

pkg/build/strategies/strategies.go
```diff
@@ -13,17 +13,24 @@ import (
 )
 
 // GetStrategy decides what build strategy will be used for the STI build.
+// TODO: deprecated, use Strategy() instead
```
bparees: is there a reason you can't just delete this?

smarterclayton: Concern about other callers and trying to get this in before 3.1

pkg/build/strategies/sti/sti.go
```diff
@@ -410,8 +406,28 @@ func (b *STI) Execute(command string, config *api.Config) error {
...
+				w.CloseWithError(err)
+				if r := recover(); r != nil {
+					glog.Errorf("recovered panic: %#v", r)
+				}
+				wg.Done()
+			}()
+			err = b.tar.CreateTarStream(uploadDir, false, w)
+		}()
+
+		opts.Stdin = r
```
bparees: is there a reason you can't move all this logic (setting up the stdin stream) above the creation of the RunContainerOptions struct, and populate the opts.Stdin field as part of creating that struct, instead of doing it here?

pkg/build/strategies/onbuild/onbuild.go
smarterclayton: Ah, you meant 61 in the old file.  Fixed.

pkg/build/strategies/sti/sti.go
bparees: i guess it's a scoping issue.  meh.

smarterclayton: The only reason would be that stdin is not always set, so you'd need an empty `var stdin` and go abhors empty variable declarations.

pkg/build/strategies/onbuild/onbuild.go
bparees: i liked the old line 61 better than your new one. :)

smarterclayton: Fixed it for you - including saving you from a govet error for anonymous
field declarations

bparees: thanks, now yours is better :)

bparees:
squish and lgtm.

smarterclayton:
[merge]

smarterclayton:
[merge]

PR #311, opened 2015-10-15
Introduce NetworkMode option

pedro-r-marques:
Provides openshift origin with the ability to control the network settings of
sti containers.

### DIFF
--- a/pkg/api/types.go
+++ b/pkg/api/types.go
@@ -45,6 +45,10 @@ type Config struct {
 	// previous image from private repositories
 	IncrementalAuthentication docker.AuthConfiguration
 
+	// DockerNetworkMode is used to set the docker network setting to --net=container:<id>
+	// when the builder is invoked from a container.
+	DockerNetworkMode DockerNetworkMode
+
 	// PreserveWorkingDir describes if working directory should be left after processing.
 	PreserveWorkingDir bool
 
@@ -207,4 +211,22 @@ type SourceInfo struct {
 type CloneConfig struct {
 	Recursive bool
 	Quiet     bool
+}
+
+// DockerNetworkMode specifies the network mode setting for the docker container
+type DockerNetworkMode string
+
+const (
+	// DockerNetworkModeHost places the container in the default (host) network namespace.
+	DockerNetworkModeHost DockerNetworkMode = "host"
+	// DockerNetworkModeBridge instructs docker to create a network namespace for this container connected to the docker0 bridge via a veth-pair.
+	DockerNetworkModeBridge DockerNetworkMode = "bridge"
+	// DockerNetworkModeContainerPrefix is the string prefix used by NewDockerNetworkModeContainer.
+	DockerNetworkModeContainerPrefix string = "container:"
+)
+
+// NewDockerNetworkModeContainer creates a DockerNetworkMode value which instructs docker to place the container in the network namespace of an existing container.
+// It can be used, for instance, to place the s2i container in the network namespace of the infrastructure container of a k8s pod.
+func NewDockerNetworkModeContainer(id string) DockerNetworkMode {
+	return DockerNetworkMode(DockerNetworkModeContainerPrefix + id)
 }
--- a/pkg/api/validation/validation.go
+++ b/pkg/api/validation/validation.go
@@ -2,6 +2,7 @@ package validation
 
 import (
 	"fmt"
+	"strings"
 
 	"github.com/openshift/source-to-image/pkg/api"
 )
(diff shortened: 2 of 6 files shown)


pedro-r-marques:
@rajatchopra This knob is used in order to propagate the network namespace from the builder Pod to the s2i docker container; this removes assumption on the networking settings on docker. The s2i docker pod can use the SDN solution used by the pod.

pedro-r-marques:
The following diff shows the proposed usage of this knob:
https://github.com/openshift/origin/compare/master...pedro-r-marques:sti_network
I would like to submit that as a PR once the dependency from source-to-build is merged, assuming that the maintainers consider the proposal acceptable.

rajatchopra:
The idea is helpful. But just asking the reverse question - is it too much hassle to be able to provide external network connectivity through docker's default network mode?

pedro-r-marques:
@rajatchopra The access allowed to the builder can be different than the network access allowed to the node. e.g. builder should have access to the enterprise source control system. The node itself, probably not. I think this would be also useful to the redhat sdn plugin: from a practical standpoint, one would probably not want to have to manage and operate docker networking (docker bridge) plus SDN.
With this proposal the deployment template can set options to the builder that would control its network access: that is set on the netns of origin-pod and inherited by s2i. It seems to me that this is the desirable workflow: control all settings via deployment .json.

rajatchopra:
Agreed, we had to go through hoops to accommodate docker's Linux bridge with our solution.
Do you mind putting this and the usage PR together in this one?

pedro-r-marques:
@rajatchopra Pardon my cluelessness. How do i put the 2 PRs together ? The usage PR is in a separate git repository (origin). Which then copies this repository through its Godeps; i was under the impression that i would have to submit this PR; wait for it to get merged; update the Godeps on origin and then submit the PR on origin with Godeps + change.

pkg/api/types.go
```diff
@@ -45,6 +45,10 @@ type Config struct {
 	// previous image from private repositories
 	IncrementalAuthentication docker.AuthConfiguration
 
+	// DockerNetworkMode is used to set the docker network setting to --net=container:<id>
+	// when the builder is invoked from a container.
+	DockerNetworkMode string
```
bparees: is there a set of valid values that can be defined in the api as constants?

bparees:
@pedro-r-marques @rajatchopra will this also provide the builder-pods DNS settings to the s2i container?  specifically i'm wondering about this issue:
https://trello.com/c/f3azg2K1/685-builder-containers-don-t-have-openshift-cluster-dns
https://github.com/openshift/origin/issues/4303

pedro-r-marques:
@bparees This is required in order to fix the DNS issue, when any network solution other than what is provided by docker is used. However the patch doesn't add the resolv.conf file from the parent Pod (which has the correct DNS settings). Would you like me to add this functionality through this patch or a separate PR ?

pkg/api/types.go
pedro-r-marques: Added a validator to check allowed values.

bparees:
Separate PR would be great. Thanks!

Ben Parees | OpenShift

rajatchopra:
LGTM, @bparees please mark merge if you are okay.

pkg/api/validation/validation.go
```diff
@@ -18,14 +19,34 @@ func ValidateConfig(config *api.Config) []ValidationError {
...
+	}
+	return false
+}
+
 // NewFieldRequired returns a *ValidationError indicating "value required"
 func NewFieldRequired(field string) ValidationError {
 	return ValidationError{ValidationErrorTypeRequired, field}
 }
 
+func NewFieldInvalidValue(field string) ValidationError {
```
mfojtik: need godoc

mfojtik:
LGTM, please add the godoc and it is ready for merge

pedro-r-marques:
@mfojtik Apologies for the newbie question... The PR has a comment in api/types.go type Config that is picked up when i run godoc ./pkg/api

output from godoc:
```
    // DockerNetworkMode is used to set the docker network setting to --net=container:<id>
    // when the builder is invoked from a container.
    DockerNetworkMode string
```

Can you please clarify your comment ? Are you referring to other documentation that is missing ?

pedro-r-marques:
@mfojtik The godoc has been added.

pkg/api/types.go
bparees: i still think you should define constants here for this.

pedro-r-marques:
@bparees Added DockerNetworkMode and corresponding string constants.

bparees:
thanks, lgtm.  [merge]

Issue #303, opened 2015-10-07
s2i help <cmd> prints no extra topics


php-coder:
At the end of output of any `s2i help <cmd>` command I see the empty list of additional help topics:

```
[vagrant@openshiftdev sti-php]$ ./s2i help version
Display version

Usage: 
  s2i version [flags]
Flags:
  -h, --help=false: help for version

Global Flags:
      --ca="": Set the path of the docker TLS ca file
      --cert="": Set the path of the docker TLS certificate file
      --key="": Set the path of the docker TLS key file
      --loglevel=0: Set the level of log output (0-5)
  -U, --url="unix:///var/run/docker.sock": Set the url of the docker socket to use

Additional help topics:


[vagrant@openshiftdev sti-php]$
```

If there is no "additional help topics" why do we print these useless lines?

I'm using source-to-image-v1.0.2-00d1cb3-linux-amd64

bparees:
that's coming out of cobra (the library we we use to handle CLI args).  We use the same thing for openshift but i don't see openshift printing the additional help topics line.  might be worth looking at the difference between the two, perhaps there's a setting we should be using in s2i to avoid it.

it's also possible we just need to update the cobra dep in s2i.

soltysh:
It would be good to see how we do this in origin and apply the same pattern in here as well.

php-coder:
This is bug in cobra (https://github.com/spf13/cobra/issues/113) that was fixed at the end of the August. We don't have this bug in origin because there we are using version from October. In `s2i` we are using version from May. I'm going to update cobra in s2i to the same version as we have in origin.

bparees:
+1

soltysh:
@php-coder please make sure not to use a newer version of spf13/cobra than the one in origin and k8s.

php-coder:
@soltysh Yes, they are the same:
```
$ grep -A1 cobra source-to-image/Godeps/Godeps.json origin/Godeps/Godeps.json 
source-to-image/Godeps/Godeps.json:			"ImportPath": "github.com/spf13/cobra",
source-to-image/Godeps/Godeps.json-			"Rev": "d732ab3a34e6e9e6b5bdac80707c2b6bad852936"
--
origin/Godeps/Godeps.json:			"ImportPath": "github.com/spf13/cobra",
origin/Godeps/Godeps.json-			"Rev": "d732ab3a34e6e9e6b5bdac80707c2b6bad852936"
```

bparees:
[...]

PR #313, opened 2015-10-16
[#303] Fix empty additional topics in s2i help output

php-coder:
I've updated spf13/cobra (with all its dependencies) to the same version as we're using in origin.

Updated dependencies:
- github.com/spf13/pflag
- github.com/spf13/cobra

New dependencies:
- github.com/cpuguy83/go-md2man
- github.com/russross/blackfriday
- github.com/shurcooL/sanitized_anchor_name

Fix for #303

PATAL @rhcarvalho


soltysh:
Godeps bumps needs to conform to the rules described [here](https://github.com/openshift/origin/blob/master/HACKING.md#updating-other-godeps-from-upstream). Each dep should be a separate commit. All the rest looks good.

php-coder:
>Each dep should be a separate commit.

@soltysh But in this case I'll break the build because with new version of `spf13/cobra` doesn't build without `cpuguy83/go-md2man` and it won't build without `russross/blackfriday` that depends on `shurcooL/sanitized_anchor_name`. Do I still need to create separate commits?

soltysh:
Yes please. Separate commits but in one PR. It's for having that info in
history.

php-coder:
@soltysh Done. Thank you for a fast review!

soltysh:
One tiny nit, replace ':' with a space please. I know I'm being picky it would be good to be consistent still. Otherwise LGTM.

rhcarvalho:
@soltysh is it documented wrong in origin? https://github.com/openshift/origin/blob/master/HACKING.md#updating-other-godeps-from-upstream

soltysh:
Hmm... now I am not sure.  Need to confirm that. Because the couple of
recent bumps in origin used version with space instead of colon.
@smarterclayton which form is the preferred?

smarterclayton:
We use `bump(foo):bar` - colon.

mfojtik:
[test]

mfojtik:
@php-coder can you please remove all files from the Godeps that are not Go files and leave just those that we actually import? I know that using godeps is pain, but we should at least keep the Godeps directory clean ;-)

php-coder:
>can you please remove all files from the Godeps that are not Go files and leave just those that we actually import?

@mfojtik Yes, I can. A couple of questions if you don't mind: 1) is it documented somewhere? 2) should I keep `LICENSE`/`README.md` files? 3) May be there is a special way of determining a set of go files that we are importing and their dependencies?

mfojtik:
@php-coder the original way is to do `godep restore` and `godep save` which will only copy the files (packages) you import, nothing else. But because that is pain, we often just copy the files/directories over... You don't need to keep the LICENSE/README.md afaik (I don't think godeps does that).

php-coder:
@mfojtik I've updated my commits. I've just removed non-go files. May be I also should delete `_test.go` files?

mfojtik:
@php-coder yes, they are not imported with the package.

php-coder:
@mfojtik Ok, I've removed also test files (btw, other dependencies still have non-go files and tests).

soltysh:
@mfojtik actually if you have a look at origins' Godeps you'll see them as well, just fyi :wink:

mfojtik:
@soltysh that does not mean it is correct, it means we are lazy in origin to clean them up :-)

mfojtik:
[merge]

PR #316, opened 2015-10-22
tar: avoid calling chmod when OS is windows and clean input dir

csrwng:
Clean the input dir when creating a tar (prevents path truncation when creating tar header) and avoid calling chmod when OS is windows (otherwise an error is thrown).


csrwng:
[test]

csrwng:
@bparees ptal

bparees:
lgtm [merge]

PR #319, opened 2015-10-26
don't require .git suffix; consolidate validation, empty protocol URL processing

gabemontero:
@bparees @csrwng from email exchange this weekend and @csrwng heads up on Friday.  Also added a quick verification for bugzilla 1268478

### DIFF
--- a/pkg/scm/git/git.go
+++ b/pkg/scm/git/git.go
@@ -2,6 +2,8 @@ package git
 
 import (
 	"bufio"
+	"errors"
+	"fmt"
 	"io"
 	"io/ioutil"
 	"net/url"
@@ -18,6 +20,8 @@ import (
 // Git is an interface used by main STI code to extract/checkout git repositories
 type Git interface {
 	ValidCloneSpec(source string) bool
+	ValidCloneSpecRemoteOnly(source string) bool
+	MungeNoProtocolURL(source string, url *url.URL) error
 	Clone(source, target string, opts api.CloneConfig) error
 	Checkout(repo, ref string) error
 	GetInfo(string) *api.SourceInfo
@@ -34,7 +38,34 @@ type stiGit struct {
 	runner util.CommandRunner
 }
 
-var gitSshURLExp = regexp.MustCompile(`\A([\w\d\-_\.+]+@[\w\d\-_\.+]+:[\w\d\-_\.+%/]+\.git)$`)
+// URLMods encapsulates potential changes to similarly named fields in the URL struct defined in golang
+// when a protocol is not explicitly specified in a clone spec (which is possible for certain file:// and ssh:// permutations)
+type URLMods struct {
+	Scheme string
+	User   string
+	Host   string
+	Path   string
+	// Corresponds to Fragment in the URL struct, but we use Ref since URL Fragments are used and git commit refs
+	Ref string
+}
+
+// FileProtoDetails encapsulates certain determinations from examining a given clone spec under the assumption
+// that it is a local file protocol clone spec
+type FileProtoDetails struct {
+	// Using the clone spec as a literal file path, does it actually exis
+	FileExists bool
+	// Use OS level file copy commands instead of the git binary
+	UseCopy bool
+	// Did the clone spec have the prefix of file://
+	ProtoSpecified bool
+	// Was the text for a fragment/ref that follows the last # incorrect
+	BadRef bool
+}
+
+var gitSshURLUser = regexp.MustCompile(`^([\w\-_\.]+)$`)
+var gitSshURLIPv4 = regexp.MustCompile(`^([\w\.\-_]+)$`)         // should cover textual hostnames and x.x.x.x
+var gitSshURLIPv6 = regexp.MustCompile(`^(\[[a-fA-F0-9\:]+\])$`) // should cover [hex:hex ... :hex]
+var gitSshURLPathRef = regexp.MustCompile(`^([\w\.\-_+\/\\]+)$`)
 
 var allowedSchemes = []string{"git", "http", "https", "file", "ssh"}
 
@@ -62,18 +93,272 @@ func stringInSlice(s string, slice []string) bool {
 // ValidCloneSpec determines if the given string reference points to a valid git
 // repository
 func (h *stiGit) ValidCloneSpec(source string) bool {
-	url, err := url.Parse(source)
-	if err != nil {
-		return false
-	}
-
-	if stringInSlice(url.Scheme, allowedSchemes) {
+	details, _ := ParseFile(source)
+	if details.FileExists && !details.BadRef {
 		return true
 	}
 
-	// support 'git@' ssh urls and local protocol without 'file://' scheme
-	return url.Scheme == "" && (strings.HasSuffix(source, ".git") ||
-		(strings.HasPrefix(source, "git@") && gitSshURLExp.MatchString(source)))
+	return h.ValidCloneSpecRemoteOnly(source)
+}
+
+// ValidCloneSpecRemoteOnly determines if the given string reference points to a valid remote git
+// repository; valid local repository specs will result in a return of false
+func (h *stiGit) ValidCloneSpecRemoteOnly(source string) bool {
+	_, err := ParseSSH(source)
+	if err == nil {
+		return true
+	}
+
+	url, err := ParseURL(source)
+	if url != nil && err == nil && url.Scheme != "file" {
+		return true
+	}
+
+	return false
+}
+
+// MungeNoProtocolURL will take a URL returned from net/url.Parse and make
+// corrections to the URL when no protocol is specified in the Scheme, where there
+// are valid protocol-less git url spec formats that result in either file:// or ssh:// protocol usage;
+// if an explicit protocol is already specified, then the
+// URL is left unchaged and the method simply returns with no error reported,
+// since the URL is
+func (h *stiGit) MungeNoProtocolURL(source string, uri *url.URL) error {
+	if uri == nil {
+		return nil
+	}
+
+	// only deal with protocol-less url's
+	if uri.Scheme != "" {
+		return nil
+	}
+
+	details, mods := ParseFile(source)
+
+	if details.BadRef {
+		return errors.New(fmt.Sprintf("bad reference following # in %s", source))
+	}
+	if !details.FileExists {
+		mods2, err := ParseSSH(source)
+		if err != nil {
+			glog.Errorf("ssh git clone spec error: %v", err)
+			return err
+		}
+		mods = mods2
+	}
+
+	// update the either file or ssh url accordingly
+	if mods != nil {
+		if len(mods.User) > 0 {
+			uri.User = url.User(mods.User)
+		}
+		if len(mods.Scheme) > 0 {
+			uri.Scheme = mods.Scheme
+		}
+		if len(mods.Host) > 0 {
+			uri.Host = mods.Host
+		}
+		if len(mods.Path) > 0 {
+			uri.Path = mods.Path
+		}
+		if len(mods.Ref) > 0 {
+			uri.Fragment = mods.Ref
+		}
+	}
+	return nil
+}
+
+// ParseURL sees if the spec is properly handled by golang's URL parser,
+// with a valid, explicit protocol specified,
+// and deals with some of the known issues we have seen with the logic
+func ParseURL(source string) (*url.URL, error) {
+	uri, err := url.Parse(source)
+	if err != nil {
+		return nil, err
+	}
+
+	// to work around the url.Parse bug, deal with the fact it seemed to split on just :
+	if strings.Index(source, "://") == -1 && uri.Scheme != "" {
+		uri.Scheme = ""
+		return nil, errors.New(fmt.Sprintf("url source %s mistakingly interpreted as protocol %s by golang", source, uri.Scheme))
+	}
+
+	// now see if an invalid protocol is specified
+	if !stringInSlice(uri.Scheme, allowedSchemes) {
+		return nil, errors.New(fmt.Sprintf("unsupported protocol specfied:  %s", uri.Scheme))
+	}
+
+	// have a valid protocol, return sucess
+	return uri, nil
+}
+
+// ParseFile will see if the input string is a valid file location, where
+// file names have a great deal of flexibility and can even match
+// expect git clone spec syntax; it also provides details if the file://
+// proto was explicitly specified, if we should use OS copy vs. the git
+// binary, and if a frag/ref has a bad format
+func ParseFile(source string) (details *FileProtoDetails, mods *URLMods) {
+	protoSpecified := false
+	if strings.HasPrefix(source, "file://") && len(source) > 7 {
+		protoSpecified = true
+	}
+
+	// if source, minus potential file:// prefix, exists as is, denote
+	// and return
+	if doesExist(source) {
+		details = &FileProtoDetails{
+			FileExists:     true,
+			UseCopy:        !isLocalGitRepository(source) || !hasGitBinary(),
+			ProtoSpecified: protoSpecified,
+			BadRef:         false,
+		}
+		mods = &URLMods{
+			Scheme: "file",
+			Path:   strings.TrimPrefix(source, "file://"),
+		}
+		return
+	}
+
+	// need to see if this was a file://<valid file>#ref
+	if protoSpecified && strings.LastIndex(source, "#") != -1 {
+		segments := strings.SplitN(source, "#", 2)
+		// given last index check above, segement should
+		// be of len 2
+		path, ref := segments[0], segments[1]
+		// file does not exist, return bad
+		if !doesExist(path) {
+			details = &FileProtoDetails{
+				UseCopy:        false,
+				FileExists:     false,
+				ProtoSpecified: protoSpecified,
+				BadRef:         false,
+			}
+			mods = nil
+			return
+		}
+		// if ref/frag bad, return bad
+		if !gitSshURLPathRef.MatchString(ref) {
+			details = &FileProtoDetails{
+				UseCopy:        false,
+				FileExists:     true,
+				ProtoSpecified: protoSpecified,
+				BadRef:         true,
+			}
+			mods = nil
+			return
+		}
+
+		// return good
+		details = &FileProtoDetails{
+			UseCopy:        !isLocalGitRepository(source) || !hasGitBinary(),
+			FileExists:     true,
+			ProtoSpecified: protoSpecified,
+			BadRef:         false,
+		}
+		mods = &URLMods{
+			Scheme: "file",
+			Path:   path,
+			Ref:    ref,
+		}
+		return
+	}
+
+	// general return bad
+	details = &FileProtoDetails{
+		UseCopy:        false,
+		FileExists:     false,
+		BadRef:         false,
+		ProtoSpecified: protoSpecified,
+	}
+	return
+}
+
+// ParseSSH will see if the input string is a valid git clone spec
+// which follows the rules for using the ssh protocol either with or without
+// the ssh:// prefix
+func ParseSSH(source string) (*URLMods, error) {
+	// if not ssh protcol, return bad
+	if strings.Index(source, "://") != -1 && !strings.HasPrefix(source, "ssh") {
+		return nil, errors.New(fmt.Sprintf("not ssh protocol: %s", source))
+	}
+
+	lastColonIdx := strings.LastIndex(source, ":")
+	atSignPresent := strings.Index(source, "@") != -1
+	if lastColonIdx != -1 {
+		host, path, user, ref := "", "", "", ""
+
+		if atSignPresent {
+			segments := strings.SplitN(source, "@", 2)
+			// with index check above, can assume there are 2 segments
+			user, host = segments[0], segments[1]
+
+			// bad user, return
+			if !gitSshURLUser.MatchString(user) {
+				return nil, errors.New(fmt.Sprintf("invalid user name provided: %s from %s", user, source))
+			}
+
+			// because of ipv6, need to redo last index of :
+			lastColonIdx = strings.LastIndex(host, ":")
+			if lastColonIdx != -1 {
+				path = host[lastColonIdx+1:]
+				host = host[0:lastColonIdx]
+			} else {
+				return nil, errors.New(fmt.Sprintf("invalid git ssh clone spec, the @ precedes the last: %s", source))
+			}
+		} else {
+			host = source[0:lastColonIdx]
+			path = source[lastColonIdx+1:]
+		}
+
+		// bad host, either ipv6 or ipv4
+		if !gitSshURLIPv6.MatchString(host) && !gitSshURLIPv4.MatchString(host) {
+			return nil, errors.New(fmt.Sprintf("invalid host provided: %s from %s", host, source))
+		}
+
+		segments := strings.SplitN(path, "#", 2)
+		if len(segments) == 2 {
+			path, ref = segments[0], segments[1]
+
+			// bad ref/frag
+			if !gitSshURLPathRef.MatchString(ref) {
+				return nil, errors.New(fmt.Sprintf("invalid reference provided: %s from %s", ref, source))
+			}
+		}
+
+		// bad path
+		if !gitSshURLPathRef.MatchString(path) {
+			return nil, errors.New(fmt.Sprintf("invalid path provided: %s from ", path, source))
+		}
+
+		// return good
+		return &URLMods{
+			Scheme: "ssh",
+			User:   user,
+			Host:   host,
+			Path:   path,
+			Ref:    ref,
+		}, nil
+	}
+	return nil, errors.New(fmt.Sprintf("unable to parse ssh git clone specification:  %s", source))
+}
+
+// isLocalGitRepository checks if the specified directory has .git subdirectory (it
+// is a GIT repository)
+func isLocalGitRepository(dir string) bool {
+	_, err := os.Stat(fmt.Sprintf("%s/.git", strings.TrimPrefix(dir, "file://")))
+	return !(err != nil && os.IsNotExist(err))
+}
+
+// doesExist checks if the path exists, removing file:// if needed for OS FS check
+func doesExist(dir string) bool {
+	_, err := os.Stat(strings.TrimPrefix(dir, "file://"))
+	return !(err != nil && os.IsNotExist(err))
+}
+
+// hasGitBinary checks if the 'git' binary is available on the system
+func hasGitBinary() bool {
+	_, err := exec.LookPath("git")
+	return err == nil
 }
 
 // Clone clones a git repository to a specific target directory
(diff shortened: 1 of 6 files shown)


pkg/scm/git/git.go
```diff
@@ -34,7 +34,7 @@ type stiGit struct {
 	runner util.CommandRunner
 }
 
-var gitSshURLExp = regexp.MustCompile(`\A([\w\d\-_\.+]+@[\w\d\-_\.+]+:[\w\d\-_\.+%/]+\.git)$`)
+var gitSshURLExp = regexp.MustCompile(`\A([\w\d\-_\.+]+@[\w\d\-_\.+]+:[\w\d\-_\.+%/])`)
```
csrwng: Is there a reason you got rid of the '$' at the end?

gabemontero: the test cases did not pass with it present with with \.git removed

gabemontero:
[test]

pkg/scm/git/git.go
csrwng: If I add ".git" to "git@192.168.122.1:repositories/authooks" (as in "git@192.168.122.1:repositories/authooks.git"), the test case passes (with the "$" at the end). I think it needs some more investigation. Also, I'm not sure that we're doing the best validation with the current string. For example, we would allow a username that starts with "-" or "+" which I think is invalid.

gabemontero: But the point of that particular test was to allow URLs without .git at the end.  If I add $, then that URL without .git fails.

csrwng: My point is that it's not clear why it fails

csrwng: If you don't include "$" at the end, then you can have any chars at the end of the string and they don't have to match your regex... like gabe@host:/test!@#.\***blah

csrwng: Also, while you investigate, don't forget about IPv6:
http://www.chtaube.eu/news/2014/0918-git210-parser-bug-in-ipv6-ssh-urls/

csrwng: Also specifying a custom port:
http://stackoverflow.com/questions/5767850/git-on-custom-ssh-port

gabemontero: Note on [custom port](http://stackoverflow.com/questions/5767850/git-on-custom-ssh-port), the commentary there explicitly states that you can only specify a port *IF* you specify the protocol (man git-clone also implies this).  As such, golang's url.Parse handles the strings where the protocol is specified, so we should be good there.

gabemontero:
@csrwng @bparees I've pushed the consolidation of git url validation / munging when no protocol is specified; included are a couple of new methods on the interface which are leveraged by both source-to-image and origin

gabemontero:
@csrwng @bparees actually wait ... I thought I saw hack/test-go.sh passing but now it is not

gabemontero:
@csrwng @bparees ok now it is good (all hack/test-*.sh and relevant hack/verify-*.sh passing) - ptal

pkg/scm/git/git.go
```diff
@@ -62,18 +68,107 @@ func stringInSlice(s string, slice []string) bool {
...
+func validCloneSpec(source string, remoteOnly bool, partialURL *url.URL) bool {
+	if partialURL == nil {
+		uri, err := url.Parse(source)
+		if err != nil {
+			return false
+		}
+		partialURL = uri
+	}
+
+	if strings.Index(source, "~") != -1 {
```
csrwng: Is there a reason you explicitly test for the '~'?

gabemontero: To maintain previous behavior for this test:  https://github.com/openshift/origin/blob/master/pkg/generate/app/cmd/newapp_test.go#L35-L97

The ~ stuff with new-app we talked about on Monday.

csrwng: I guess I don't understand why you check for ~ explicitly... wouldn't it be detected by the regex's for the individual pieces?

gabemontero: It wasn't originally, but some of my changes to the regex's may have occurred after I added that check.  I'll try again now.

gabemontero: Yep ... it works fine now with the latest regex's and that check removed.  I'll update the PR.

gabemontero:
[test]

pkg/scm/git/git.go
```diff
@@ -62,18 +68,103 @@ func stringInSlice(s string, slice []string) bool {
...
+				return false
+			}
+		}
+		partialURL.Scheme = "ssh"
+		partialURL.User = url.User(user)
+		partialURL.Host = host
+		partialURL.Path = path
+		partialURL.Fragment = ref
+		return true
+	} else if !remoteOnly {
```
csrwng: you can have a valid remote clone spec without a user:
git clone linux-box:/data/openshift-docs works fine on my Mac

pkg/scm/git/git_test.go
```diff
@@ -10,7 +10,7 @@ import (
 )
```
csrwng: You should add a set of tests for ValidCloneSpecRemoteOnly

pkg/scm/git/git.go
gabemontero: If the user string is empty, we end up with the same empty Username struct that url.Parse(string) started us on.  My take was that a simple re-initialization was cleaner that and if len(user) > 0 check.

Based on that clarification, would you agree?

pkg/scm/git/git_test.go
gabemontero: Yep - missed that.  I'll add some.

gabemontero:
ok, pushed changes resulting from some face to face back and forth between @csrwng and I, as well as some additional fixes in scm.go/scm_test.go based on similar errors on what is a valid git clone spec

gabemontero:
@csrwng @bparees woke up this AM and a couple of minor tweaks came to mind

ptal at your earliest convenience

pkg/scm/git/git.go
```diff
@@ -62,18 +67,136 @@ func stringInSlice(s string, slice []string) bool {
...
+// corrections to the URL when no protocol is specified in the Scheme
+func (h *stiGit) MungeNoProtocolURL(source string, uri *url.URL) bool {
+	if uri == nil {
+		return false
+	}
+	rc, scheme, user, host, path, ref := validCloneSpec(source, false, uri)
+	if !rc {
+		return false
+	}
+	if scheme != "ssh" && scheme != "file" {
```
bparees: this is confusing behavior....  at a minimum i think it needs some comments.

if the uri wasn't a validCloneSpec, the function returns false.  but it also returns false if it was valid, but the protocol was anything other than "file" or "ssh"?  (which is an odd pairing of protocols to be treated the same)

pkg/test/git.go
```diff
@@ -24,6 +25,18 @@ func (f *FakeGit) ValidCloneSpec(source string) bool {
 	return f.ValidCloneSpecResult
 }
 
+//ValidCloneSpecRemoteOmly returns a valid GIT clone specification
```
bparees: s/Omly/Only/

pkg/scm/scm.go
```diff
@@ -16,41 +15,48 @@ import (
...
 
-	// Convert relative path to absolute path.
-	if !strings.HasPrefix(s, "/") {
-		if absolutePath, err := filepath.Abs(s); err == nil {
-			s = absolutePath
+	if fileCopy {
+		if doesExist(s, true) {
+			if !isLocalGitRepository(s) {
+				// since not using git, any resulting URLs need to be explicit with file:// protocol specified
+				s = "file://" + s
```
bparees: what if the user provided the path with a file:// prefix already on it?

pkg/scm/scm.go
```diff
@@ -16,41 +15,48 @@ import (
...
-	// If we have local directory and that directory exists, use file copy
-	if _, err := os.Stat(s); err == nil {
-		return &file.File{util.NewFileSystem()}, s, nil
+	// If the source is valid  GIT protocol (file://, git://, git@, etc..) use GIT
+	// binary to download the sources
+	if g.ValidCloneSpec(s) {
+		// if local file system, without file://, when using git, should not need file://, but we'll be safe;
+		// satisfies previous constructed test case in scm_test.go as well
+		if doesExist(s, false) {
+			s = "file://" + s
```
bparees: again what if they provided the path with file:// on it?

pkg/scm/git/git.go
```diff
@@ -18,6 +18,8 @@ import (
 // Git is an interface used by main STI code to extract/checkout git repositories
 type Git interface {
 	ValidCloneSpec(source string) bool
+	ValidCloneSpecRemoteOnly(source string) bool
+	MungeNoProtocolURL(source string, url *url.URL) bool
```
bparees: it looks like this is never called, what am i missing?

gabemontero: MungeNoProtocolURL will be called by the pending origin changes.  There was duplicate logic in origin modules that have been consolidated in pkg/scm/git/git.go

pkg/scm/scm.go
gabemontero: ah ...late night coding ... the boolean param on doesExist was part of an attempt to deal with file:// either being there or not, without multiple checks for the prefix, but did not complete the job;

i'll rework and update

pkg/scm/scm.go
gabemontero: see comment above

pkg/scm/git/git.go
gabemontero: yeah, the boolean return was in fact a placeholder for future use - its actual use in my corresponding origin change does not leverage it... certainly with lack of use I did not think it through.

I'm inclined to switch gears and have a void return.  Will do that as part of next push.

On comments, yeah, I had some on the origin side but not here.  I'll address.

On the file and ssh pairing, just to be clear, there "association" such as it is stems from the fact that from a git clone spec perspective, there are valid url's where neither file:// or ssh:// are specified, but based on other characters being present (or not present) the file:// or ssh:// can be derived.

gabemontero:
@bparees @csrwng updates in response to @bparees comments pushed

pkg/scm/git/git.go
```diff
@@ -62,18 +69,145 @@ func stringInSlice(s string, slice []string) bool {
...
+		shouldReturn, returnCode := schemeNecessitatesReturn(partialURL)
+		if shouldReturn {
+			return returnCode, scheme, user, host, path, ref
+		}
 	}
 
-	if stringInSlice(url.Scheme, allowedSchemes) {
-		return true
+	lastColonIdx := strings.LastIndex(source, ":")
+	if lastColonIdx != -1 {
```
bparees: can't this ":" be part of "file:///some/local/path" rather than part of an ssh reference?

bparees: oh i guess that would have been caught earlier by the "schemeNecessitatesReturn" check.

bparees: definitely calls for a "// at this point we have elimated all source paths that have an explicit protocol, now we need to distinguish between a local file path that wasn't prefixed with file:// and an ssh path that wasn't prefixed with ssh://" comment.

bparees: also "colon" is  a valid character in a filename, so is this really sufficient?

pkg/scm/git/git.go
```diff
@@ -62,18 +69,145 @@ func stringInSlice(s string, slice []string) bool {
...
+	return false, true
+}
+
+func validCloneSpec(source string, remoteOnly bool, partialURL *url.URL) (rc bool, scheme, user, host, path, ref string) {
+	// note, as part of parsing the clone spec to see if it is valid, various values for the fields in url.URL are saved
+	// and returned for possible URL corrections from MungeNoProtocolURL
+	if partialURL == nil {
+		uri, err := url.Parse(source)
+		if err != nil {
+			return false, scheme, user, host, path, ref
```
csrwng: When you have named return parameters, you can just set the value of each var and just do a "return" with no arguments. It also seems to me that what you are returning could be expressed in a URL struct.

pkg/scm/git/git.go
```diff
@@ -62,18 +69,145 @@ func stringInSlice(s string, slice []string) bool {
...
+		uri.User = url.User(user)
+	}
+	uri.Scheme = scheme
+	uri.Host = host
+	uri.Path = path
+	uri.Fragment = ref
+	return nil
+}
+
+func schemeNecessitatesReturn(uri *url.URL) (shouldReturn, returnCode bool) {
```
csrwng: It's not clear from the name of the function what it does or what the return values mean. At the very least you should have a comment describing what it does.

pkg/scm/git/git.go
```diff
@@ -62,18 +69,145 @@ func stringInSlice(s string, slice []string) bool {
...
+	// note, there also appears to be a bug in url.Parse, where a url like
+	// linux-box:/path/to/code will report the scheme as "linux-box"
+	if uri.Scheme != "" {
+		return true, false
+	}
+
+	return false, true
+}
+
+func validCloneSpec(source string, remoteOnly bool, partialURL *url.URL) (rc bool, scheme, user, host, path, ref string) {
```
csrwng: I think you can do without the partialURL parameter and call this parseSpec()... let the MungeNoProtocol function call it if it needs to.

pkg/scm/git/git.go
csrwng: a colon is not valid in a filename in a git clone spec

pkg/scm/git/git.go
gabemontero: true (return no args) ... I consciously added the arguments on the return because I just like things explicit 

true (a URL struct) .... my thinking was that if this method was public and consumed outside of this go file, then formalize the return with a struct.  But since it is only used locally. it wasn't necessary.

Given those rationales, how strong is your opinion here?  Still change it?

csrwng: hmm ... for return no args... I don't know that explicitly including all vars makes the code clearer. For example, in the line above you really haven't set any values to scheme, user, host,  etc. Setting rc to false and then a return might make it clearer what it is you're itending to do. 

BTW, rc doesn't tell me what it means... valid might be a better name?

I guess as far as the struct goes, I guess it's a matter of preference. A struct would make it clearer for me that your return has 2 parts... one (the bool) indicating whether the input is valid or not and the other one breaking down the parts of the spec.

gabemontero: Ah ... the "2 parts" thing is a good point.  Ok, I've been converted on the struct point.  And sure, I'll come up with something more specific than "rc".

On the return no args, I suppose it comes down to where you draw the line with being explicit/implicit ... i.e. the implicit setting of string vars to "" ;-) ... I'll revisit after the struct conversion and see if I want to continue the discussion on that point then. Thx.

pkg/scm/git/git.go
gabemontero: added additional comments for now ... this function as it currently exists might change (or even go away) based on the outcome of the next comment.

pkg/scm/git/git.go
gabemontero: yep - i like it - thx - part of next push

pkg/scm/git/git.go
bparees: so what's going to happen if i run:  "s2i build some/lo:cal/path"?

gabemontero: just added that to the test cases - it is correctly identified as an invalid git clone spec

bparees: but it should be accepted as a valid source path because it's a local file system path, is it?

gabemontero: Ah - I see now.  Yep, you are right.

And y'all are probably aware, but just in case, today the s2i code does a golang os level copy of files for either file:// or just plain paths.

So to me it would make sense to allow ":" in both those cases.  Agreed?

The trick will be distinguishing between protocol-less file spec and ssh spec.

gabemontero: FYI - this looks like the description of what is valid for git specs:  https://git-scm.com/docs/git-check-ref-format

Definitely more restrictive that what is allowed in URLs:  http://stackoverflow.com/questions/2849756/list-of-valid-characters-for-the-fragment-identifier-in-an-url

gabemontero:
@bparees @csrwng updates from this AM's comments have been pushed

pkg/scm/git/git.go
```diff
@@ -62,18 +78,178 @@ func stringInSlice(s string, slice []string) bool {
...
+	if len(mods.path) > 0 {
+		uri.Path = mods.path
+	}
+	if len(mods.frag) > 0 {
+		uri.Fragment = mods.frag
+	}
+	return nil
+}
+
+// schemeNeccessitatesReturn examines the URL and determines whether further
```
bparees: schemeNecessitatesReturn

pkg/scm/git/git.go
```diff
@@ -62,18 +78,178 @@ func stringInSlice(s string, slice []string) bool {
...
+// url / git spec string processing is needed or not, and if no more processing is needed,
+// whether the input should be deemed valid or not;
+// the combos of interest are:
+// - if a valid protocol/scheme is specifed, just return (leaving the valid URL alone)
+// - if an invalid protocol/scheme is specified
```
bparees: do what?

pkg/scm/git/git.go
```diff
@@ -62,18 +78,178 @@ func stringInSlice(s string, slice []string) bool {
...
+			}
+		}
+
+		if !gitSshURLIPv6.MatchString(host) && !gitSshURLIPv4.MatchString(host) {
+			return nil, errors.New(fmt.Sprintf("invalid host provided: %s from %s", host, source))
+		}
+
+		segments = strings.SplitN(path, "#", 2)
+		if len(segments) == 2 {
+			path, frag = segments[0], segments[1]
```
bparees: why the change from "ref" to "frag" as the name here?

pkg/scm/git/git.go
```diff
@@ -62,18 +78,178 @@ func stringInSlice(s string, slice []string) bool {
...
+	if invalidProtocol {
+		return nil, errors.New(fmt.Sprintf("unsupported protocol specfied:  %s", uri.Scheme))
+	}
+	if justReturn {
+		return nil, nil
+	}
+
+	lastColonIdx := strings.LastIndex(source, ":")
+	atSignPresent := strings.Index(source, "@") != -1
+	if lastColonIdx != -1 && atSignPresent {
```
bparees: not to make you feel like i'm going back and forth on you, but this would imply that:
github.com:openshit/origin.git  would not be a valid ssh clone spec, but it appears that it is:
```
$ git clone user@server:project.git

You can also not specify a user, and Git assumes the user you’re currently logged in as.
```
(from https://git-scm.com/book/en/v2/Git-on-the-Server-The-Protocols)

at the end of the day there may not be any valid way to distinguish between a protocol-less ssh spec, and a protocoless local file spec in which the path includes a colon.

I wonder how git distinguishes, since git does allow local file paths as clone specs too.  Perhaps it just tries the local path first and if it doesn't exist, tries it as an ssh spec.  Presumably we could do the same thing here...  i know you already have some logic that checks if the path provided exists locally, maybe if it does you should just immediately augment the path with file:// (if it doesn't already have that prefix) and then you can avoid a lot of this nastiness.

pkg/scm/git/git.go
gabemontero: cause in the url doc the describe the piece after the # as a fragment

http://stackoverflow.com/questions/2849756/list-of-valid-characters-for-the-fragment-identifier-in-an-url
http://tools.ietf.org/html/rfc3986

pkg/scm/git/git.go
gabemontero: i need new glasses :-)

pkg/scm/git/git.go
gabemontero: comment update in hand

pkg/scm/git/git.go
bparees: yeah but this error message "invalid fragment provided:" is likely going to be confusing to users.  per our syntax it's a ref, because we're treating the value of that fragment as a git ref.

pkg/scm/git/git.go
gabemontero: yeah, @csrwng and I ran into this a day or two ago, and we decided not to support this (presumably) little used form given the ugliness of the fix.

I'll give it a go and see if I can get a solution that is easy enough on the stomach.

gabemontero:
@bparees @csrwng ptal

gabemontero:
missed the comeback on ref/frag ... i'll switch to ref

gabemontero:
@bparees @csrwng ptal

pkg/scm/git/git.go
```diff
@@ -62,18 +78,199 @@ func stringInSlice(s string, slice []string) bool {
...
+	}
+
+	lastColonIdx := strings.LastIndex(source, ":")
+	atSignPresent := strings.Index(source, "@") != -1
+	// deal with the special git case where a user is not specified (i.e. no @);
+	// to do so, see if the entire spec as is exists as a file
+	trySSH := false
+	if lastColonIdx != -1 && !atSignPresent {
+		path := source
+		hashLoc := strings.LastIndex(source, "#")
```
bparees: filenames can contain # too.  i'd just check the source exactly as is to see if it's a valid local path.

pkg/scm/git/git.go
```diff
@@ -62,18 +78,199 @@ func stringInSlice(s string, slice []string) bool {
...
+	if justReturn {
+		return nil, nil
+	}
+
+	lastColonIdx := strings.LastIndex(source, ":")
+	atSignPresent := strings.Index(source, "@") != -1
+	// deal with the special git case where a user is not specified (i.e. no @);
+	// to do so, see if the entire spec as is exists as a file
+	trySSH := false
+	if lastColonIdx != -1 && !atSignPresent {
```
bparees: && !remoteOnly  ?

pkg/scm/git/git.go
gabemontero: ok - change coming soon

pkg/scm/git/git.go
gabemontero: yep - change coming soon

gabemontero:
ok .. with change re: the exchange between @bparees and I last night - @bparees @csrwng ptal

gabemontero:
@bparees @csrwng the latest refactor from this AM's scrum has been pushed to the PR

pkg/scm/git/git.go
```diff
@@ -62,18 +84,272 @@ func stringInSlice(s string, slice []string) bool {
...
+func ParseURL(source string) (*url.URL, error) {
+	uri, err := url.Parse(source)
+	if err != nil {
+		return nil, err
+	}
+
+	// to work around the url.Parse bug, deal with the fact it seemed to split on just :
+	if strings.Index(source, "://") == -1 && uri.Scheme != "" {
+		uri.Scheme = ""
+		return nil, nil
```
bparees: should this not return an error also?  (that it's not a valid url with a protocol)

pkg/scm/git/git.go
```diff
@@ -62,18 +84,272 @@ func stringInSlice(s string, slice []string) bool {
...
+// expect git clone spec syntax; it also provides details if the file://
+// proto was explicitly specified, if we should use OS copy vs. the git
+// binary, and if a frag/ref has a bad format
+func ParseFile(source string) (details *FileProtoDetails, mods *URLMods) {
+	protoSpecified := false
+	if strings.HasPrefix(source, "file://") && len(source) > 7 {
+		protoSpecified = true
+	}
+
+	// if source, minus potential file::// prefix, exists as is, denote
```
bparees: file://

pkg/scm/git/git_test.go
```diff
@@ -9,26 +12,105 @@ import (
...
+		"git@192.168.122.1:repositories/authooks",
+		"mbalazs@build.ulx.hu:/var/git/eap-ulx.git",
+		"ssh://git@[2001:db8::1]/repository.git",
+		"ssh://git@mydomain.com:8080/foo/bar",
+		"git@[2001:db8::1]:repository.git",
+		"git@[2001:db8::1]:/repository.git",
+		"g_m@foo.bar:foo/bar",
+		"g-m@foo.bar:foo/bar",
+		"g.m@foo.bar:foo/bar",
+		"github.com:openshit/origin.git",
```
bparees: openshift.  (don't worry, we've all done it)

pkg/scm/git/git_test.go
```diff
@@ -9,26 +12,105 @@ import (
...
+		t.Errorf("The %q directory is not git repository", d)
+	}
+}
+
 func TestValidCloneSpec(t *testing.T) {
-	scenarios := []string{"git@github.com:user/repo.git",
+	gitLocalDir := createLocalGitDirectory(t)
+	defer os.RemoveAll(gitLocalDir)
+
+	valid := []string{"git@github.com:user/repo.git",
```
bparees: needs some tests for values with # refs.

bparees:
i think the refactor made this much more readable and easier to reason about...  good suggestion @csrwng, nice implementation @gabemontero

gabemontero:
i totally agree on the readability, etc. improvement - thx @csrwng for the suggestion - looking at the remaining comments now

pkg/scm/git/git.go
gabemontero: yep - part of next push

pkg/scm/git/git.go
gabemontero: part of next push

pkg/scm/git/git_test.go
gabemontero: :-)

pkg/scm/git/git_test.go
gabemontero: i've added a couple

gabemontero:
@bparees @csrwng next round - ptal

bparees:
lgtm, will let @csrwng give it a once over monday before we merge.

gabemontero:
Cool deal sounds good

pkg/scm/git/git.go
```diff
@@ -34,7 +38,25 @@ type stiGit struct {
 	runner util.CommandRunner
 }
 
-var gitSshURLExp = regexp.MustCompile(`\A([\w\d\-_\.+]+@[\w\d\-_\.+]+:[\w\d\-_\.+%/]+\.git)$`)
+type URLMods struct {
```
csrwng: godoc (otherwise make private)

pkg/scm/git/git.go
```diff
@@ -34,7 +38,25 @@ type stiGit struct {
...
+	scheme string
+	user   string
+	host   string
+	path   string
+	ref    string
+}
+
+type FileProtoDetails struct {
```
csrwng: godoc (otherwise make private)

csrwng:
really minor comments, otherwise LGTM

gabemontero:
@csrwng godoc comments added (also capitalized fields in the case where struct was public but fields were not)

bparees:
[merge]

PR #320, opened 2015-10-27
Add logging to tar, fix extract when parent dir is .

csrwng:
Adds additional methods to the tar interface to log the files being extracted/archived (similar to -v flag on the tar command). 
Also fixes name truncation when the parent directory is the current directory.

### DIFF
--- a/pkg/tar/tar.go
+++ b/pkg/tar/tar.go
@@ -36,6 +36,12 @@ type Tar interface {
 	// The name of the new tar file is returned if successful
 	CreateTarFile(base, dir string) (string, error)
 
+	// CreateTarStreamWithLogging creates a tar from the given directory
+	// and streams it to the given writer.
+	// An error is returned if an error occurs during streaming.
+	// Archived file names are written to the logger if provided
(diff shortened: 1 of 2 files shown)


csrwng:
@bparees or @gabemontero ptal

csrwng:
[test]

pkg/tar/tar.go
```diff
@@ -45,6 +51,12 @@ type Tar interface {
 	// Times out if reading from the stream for any given file
 	// exceeds the value of timeout
 	ExtractTarStream(dir string, reader io.Reader) error
+
+	// ExtractTarStreamWithLogging extracts files from a given tar stream.
+	// Times out if reading from the stream for any given file
+	// exceeds the value of timeout.
```
bparees: the value of what timeout?

bparees:
lgtm [merge]

pkg/tar/tar.go
csrwng: should be defaultTimeout in this file

PR #323, opened 2015-10-30
Remove UseLogger

mfojtik:
realized that it does not make sense to keep this as nobody will ever use it... also cleaned up the messy 'if' in printer

### DIFF
--- a/pkg/build/strategies/sti/sti.go
+++ b/pkg/build/strategies/sti/sti.go
@@ -446,13 +446,16 @@ func (b *STI) Execute(command string, config *api.Config) error {
 				}
 				break
 			}
-			if glog.V(2) || config.Quiet != true || command == api.Usage {
-				if (config.UseLogger == true) || glog.V(3) {
-					glog.Info(text)
-					continue
-				}
-				fmt.Fprintf(os.Stdout, "%s\n", strings.TrimSpace(text))
+			// Nothing is printed when the quiet option is set
+			if config.Quiet {
+				continue
 			}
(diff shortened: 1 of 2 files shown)


mfojtik:
@soltysh PTAL

pkg/build/strategies/sti/sti.go
```diff
@@ -446,13 +446,16 @@ func (b *STI) Execute(command string, config *api.Config) error {
...
+			// Nothing is printed when the quiet option is set
+			if config.Quiet {
+				continue
+			}
+			// The log level > 3 forces to use glog instead of printing to stdout
+			if glog.V(3) {
+				glog.Info(text)
+				continue
 			}
+			fmt.Fprintf(os.Stdout, "%s\n", strings.TrimSpace(text))
```
soltysh: Remove that and you're good to go.

mfojtik: @soltysh remove what?

soltysh: The fprintf, you're using glog for v>=3, but still fprintf for v<3, what for?

soltysh: Nvmd :wink: I was thinking about something else.

soltysh:
LGTM [merge]

pkg/build/strategies/sti/sti.go
mfojtik: @soltysh if you have v>=3 then we use glog (with ugly timestamp and prefix), in case you are v<3 we just print the output from assemble to stdout (without timestamps)

Issue #273, opened 2015-08-24
Add --use-local-image option


mfojtik:
The `--force-pull=false` seems pretty confusing if somebody wants to do local development and don't want S2I to pull images from external registries. I would suggest to add `--use-local-image` to be more explicit.  We don't need to change anything in `Config`. If user set this option, it will be just an alias for `--force-pull=false`.

mfojtik:
@soltysh @goldmann @bparees FYI

mfojtik:
```console
13:52 < mfojtik> i propose adding new option
13:52 < mfojtik> supplementary to --force-pull
13:52 < mfojtik> 1) it will act just as --force-pull=false
13:52 < mfojtik> 2) it will fail if image not found
13:52 < mfojtik> 3) it won't even attempt to connect to external registry

```

soltysh:
:+1:

bparees:
just setting force-pull to false will not have all the behaviors you describe, specifically it will not prevent pulling images from remote registries if they are not present locally.

bparees:
so you'd have to add other new logic.

mfojtik:
@bparees reason why I think this is important is to support the CI/CD flows that users do with the S2I tool, where they build the image elsewhere (or locally) and then they use `s2i` binary to build the application image they test.

We should have option that turns `s2i` into mode that it won't even attempt to call remote registry.

bparees:
@mfojtik ok that makes sense.  but you can't just have it be an alias for "force-pull=false" then, you need to actually go change a lot of logic where we do things like CheckAndPull.

soltysh:
@bparees maybe that's a good chance to clean the logic around force-pull, we'll deprecate it and educate users about the new flag. Even at the cost of having slightly different behavior.

bparees:
@soltysh force pull has value.  there are three distinct options here:
1) always pull (force pull)
2) pull if not present (what you get today w/ force pull =false)
3) never pull (what @mfojtik is proposing).

we can't get rid of (1).

soltysh:
@bparees in that case instead of adding additional boolean flag, how about refactoring this into `--pull-policy`, witch values `always`, `never`, `ifnotpresent`?

soltysh:
I just realized I've seen those values somewhere :wink: and I did, [here](https://github.com/kubernetes/kubernetes/blob/master/pkg/api/types.go#L869-L871)

mfojtik:
@bparees assign this to me pls :-)

[...]

PR #324, opened 2015-11-03
Replace --force-pull with --pull-policy

mfojtik:
This will replace the --force-pull with --pull-policy option. This option default to "if-not-present" which means that the Docker image will be pull only if it does not exists locally.

Valid options are "always", "never" or "if-not-present".

* `always` will always try to pull the Docker image first and then check if the local image exists
* `if-not-present` will honor the local image and only in case the local image does not exists, it will attempt to pull it
* `never` will never perform pull and always use the local image

Fixes: https://github.com/openshift/source-to-image/issues/273

### DIFF
--- a/cmd/s2i/main.go
+++ b/cmd/s2i/main.go
@@ -122,7 +122,18 @@ $ s2i build . centos/ruby-22-centos7 hello-world-app
 				}
 			}
 
-			if len(validation.ValidateConfig(cfg)) != 0 {
+			if cfg.ForcePull {
+				glog.Warning("DEPRECATED: The '--force-pull' option is deprecated. Use '--pull-policy' instead")
+			}
+
+			if len(cfg.BuilderPullPolicy) == 0 {
+				cfg.BuilderPullPolicy = api.DefaultBuilderPullPolicy
+			}
+
+			if errs := validation.ValidateConfig(cfg); len(errs) > 0 {
+				for _, e := range errs {
+					glog.Errorf("%s", e)
+				}
 				cmd.Help()
 				os.Exit(1)
 			}
@@ -160,7 +171,7 @@ $ s2i build . centos/ruby-22-centos7 hello-world-app
 			}
 
 			if len(oldScriptsFlag) != 0 {
-				glog.Warning("Flag --scripts is deprecated, use --scripts-url instead")
+				glog.Warning("DEPRECATED: Flag --scripts is deprecated, use --scripts-url instead")
 				cfg.ScriptsURL = oldScriptsFlag
 			}
 			if len(oldDestination) != 0 {
@@ -200,10 +211,11 @@ $ s2i build . centos/ruby-22-centos7 hello-world-app
 	buildCmd.Flags().StringVarP(&(cfg.Ref), "ref", "r", "", "Specify a ref to check-out")
 	buildCmd.Flags().StringVar(&(cfg.CallbackURL), "callback-url", "", "Specify a URL to invoke via HTTP POST upon build completion")
 	buildCmd.Flags().StringVarP(&(cfg.ScriptsURL), "scripts-url", "s", "", "Specify a URL for the assemble and run scripts")
-	buildCmd.Flags().StringVar(&(oldScriptsFlag), "scripts", "", "Specify a URL for the assemble and run scripts")
+	buildCmd.Flags().StringVar(&(oldScriptsFlag), "scripts", "", "DEPRECATED: Specify a URL for the assemble and run scripts")
 	buildCmd.Flags().StringVarP(&(oldDestination), "location", "l", "", "Specify a destination location for untar operation")
 	buildCmd.Flags().StringVarP(&(cfg.Destination), "destination", "d", "", "Specify a destination location for untar operation")
-	buildCmd.Flags().BoolVar(&(cfg.ForcePull), "force-pull", true, "Always pull the builder image even if it is present locally")
+	buildCmd.Flags().BoolVar(&(cfg.ForcePull), "force-pull", false, "DEPRECATED: Always pull the builder image even if it is present locally")
+	buildCmd.Flags().VarP(&(cfg.BuilderPullPolicy), "pull-policy", "p", "Specify when to pull the builder image (always, never or if-not-present)")
 	buildCmd.Flags().BoolVar(&(cfg.PreserveWorkingDir), "save-temp-dir", false, "Save the temporary directory used by S2I instead of deleting it")
 	buildCmd.Flags().BoolVar(&(useConfig), "use-config", false, "Store command line options to .stifile")
 	buildCmd.Flags().StringVarP(&(cfg.ContextDir), "context-dir", "", "", "Specify the sub-directory inside the repository with the application sources")
@@ -247,6 +259,10 @@ func newCmdRebuild(cfg *api.Config) *cobra.Command {
 				cfg.PullAuthentication = docker.LoadAndGetImageRegistryAuth(r, cfg.BuilderImage)
 			}
 
+			if len(cfg.BuilderPullPolicy) == 0 {
+				cfg.BuilderPullPolicy = api.DefaultBuilderPullPolicy
+			}
+
 			if glog.V(2) {
 				fmt.Printf("\n%s\n", describe.DescribeConfig(cfg))
 			}
@@ -267,7 +283,8 @@ func newCmdRebuild(cfg *api.Config) *cobra.Command {
 	buildCmd.Flags().BoolVar(&(cfg.Incremental), "incremental", false, "Perform an incremental build")
 	buildCmd.Flags().BoolVar(&(cfg.RemovePreviousImage), "rm", false, "Remove the previous image during incremental builds")
 	buildCmd.Flags().StringVar(&(cfg.CallbackURL), "callback-url", "", "Specify a URL to invoke via HTTP POST upon build completion")
-	buildCmd.Flags().BoolVar(&(cfg.ForcePull), "force-pull", true, "Always pull the builder image even if it is present locally")
+	buildCmd.Flags().BoolVar(&(cfg.ForcePull), "force-pull", false, "DEPRECATED: Always pull the builder image even if it is present locally")
+	buildCmd.Flags().VarP(&(cfg.BuilderPullPolicy), "pull-policy", "p", "Specify when to pull the builder image (always, never or if-not-present)")
 	buildCmd.Flags().BoolVar(&(cfg.PreserveWorkingDir), "save-temp-dir", false, "Save the temporary directory used by S2I instead of deleting it")
 	buildCmd.Flags().StringVarP(&(cfg.DockerCfgPath), "dockercfg-path", "", filepath.Join(os.Getenv("HOME"), ".docker/config.json"), "Specify the path to the Docker configuration file")
 	return buildCmd
@@ -311,8 +328,12 @@ func newCmdUsage(cfg *api.Config) *cobra.Command {
 			cfg.Environment = envs
 
 			if len(oldScriptsFlag) != 0 {
-				glog.Warning("Flag --scripts is deprecated, use --scripts-url instead")
+				glog.Warning("DEPRECATED: Flag --scripts is deprecated, use --scripts-url instead")
 				cfg.ScriptsURL = oldScriptsFlag
+			}
+
+			if len(cfg.BuilderPullPolicy) == 0 {
+				cfg.BuilderPullPolicy = api.DefaultBuilderPullPolicy
 			}
 
 			uh, err := sti.NewUsage(cfg)
@@ -323,8 +344,9 @@ func newCmdUsage(cfg *api.Config) *cobra.Command {
 	}
 	usageCmd.Flags().StringP("env", "e", "", "Specify an environment var NAME=VALUE,NAME2=VALUE2,...")
 	usageCmd.Flags().StringVarP(&(cfg.ScriptsURL), "scripts-url", "s", "", "Specify a URL for the assemble and run scripts")
-	usageCmd.Flags().StringVar(&(oldScriptsFlag), "scripts", "", "Specify a URL for the assemble and run scripts")
-	usageCmd.Flags().BoolVar(&(cfg.ForcePull), "force-pull", true, "Always pull the builder image even if it is present locally")
+	usageCmd.Flags().StringVar(&(oldScriptsFlag), "scripts", "", "DEPRECATED: Specify a URL for the assemble and run scripts")
+	usageCmd.Flags().BoolVar(&(cfg.ForcePull), "force-pull", false, "DEPRECATED: Always pull the builder image even if it is present locally")
+	usageCmd.Flags().VarP(&(cfg.BuilderPullPolicy), "pull-policy", "p", "Specify when to pull the builder image (always, never or if-not-present)")
 	usageCmd.Flags().BoolVar(&(cfg.PreserveWorkingDir), "save-temp-dir", false, "Save the temporary directory used by S2I instead of deleting it")
 	usageCmd.Flags().StringVarP(&(oldDestination), "location", "l", "", "Specify a destination location for untar operation")
 	usageCmd.Flags().StringVarP(&(cfg.Destination), "destination", "d", "", "Specify a destination location for untar operation")
--- a/hack/test-stirunimage.sh
+++ b/hack/test-stirunimage.sh
@@ -5,6 +5,7 @@ set -o nounset
 set -o pipefail
 
 STI_ROOT=$(dirname "${BASH_SOURCE}")/..
+export KILLDONE=""
 
 function time_now()
 {
@@ -17,7 +18,7 @@ function cleanup()
     rm -f  "${STI_ROOT}/hack/sti-run.log"
     # use sigint so that sti post processing will remove docker container
     if [ -z "$KILLDONE" ]; then
-	kill -2 "${STI_PID}"
+	    kill -2 "${STI_PID}"
     fi
     echo
     echo "Complete"
@@ -34,16 +35,13 @@ fi
 
 trap cleanup EXIT SIGINT
 
-echo
-echo 
-echo
-
-sti build git://github.com/bparees/openshift-jee-sample openshift/wildfly-8-centos test-jee-app --run=true &> "${STI_ROOT}/hack/sti-run.log" &
+echo "Running 'sti build --run=true ...'"
+s2i build git://github.com/bparees/openshift-jee-sample openshift/wildfly-8-centos test-jee-app --run=true &> "${STI_ROOT}/hack/sti-run.log" &
 export STI_PID=$!
 TIME_SEC=1000
 TIME_MIN=$((60 * $TIME_SEC))
 max_wait=10*TIME_MIN
-echo "waiting up to ${max_wait}"
+echo "Waiting up to ${max_wait} for the build to finish ..."
 expire=$(($(time_now) + $max_wait))
 
 set +e
--- a/pkg/api/describe/describer.go
+++ b/pkg/api/describe/describer.go
@@ -39,7 +39,7 @@ func DescribeConfig(config *api.Config) string {
 			fmt.Fprintf(out, "Incremental Image Pull User:\t%s\n", config.IncrementalAuthentication.Username)
 		}
 		fmt.Fprintf(out, "Remove Old Build:\t%s\n", printBool(config.RemovePreviousImage))
-		fmt.Fprintf(out, "Force Pull:\t%s\n", printBool(config.ForcePull))
+		fmt.Fprintf(out, "Builder Pull Policy:\t%s\n", config.BuilderPullPolicy)
 		fmt.Fprintf(out, "Quiet:\t%s\n", printBool(config.Quiet))
 		fmt.Fprintf(out, "Layered Build:\t%s\n", printBool(config.LayeredBuild))
 		if len(config.Destination) > 0 {
@@ -76,7 +76,7 @@ func describeBuilderImage(config *api.Config, image string, out io.Writer) {
 		DockerConfig:       config.DockerConfig,
 		PullAuthentication: config.PullAuthentication,
 		BuilderImage:       config.BuilderImage,
-		ForcePull:          config.ForcePull,
+		BuilderPullPolicy:  config.BuilderPullPolicy,
 		Tag:                config.Tag,
 		IncrementalAuthentication: config.IncrementalAuthentication,
 	}
--- a/pkg/api/types.go
+++ b/pkg/api/types.go
@@ -1,6 +1,8 @@
 package api
 
 import (
+	"fmt"
+
 	docker "github.com/fsouza/go-dockerclient"
 
 	"github.com/openshift/source-to-image/pkg/util/user"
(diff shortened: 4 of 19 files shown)


mfojtik:
@bparees @soltysh PTAL

mfojtik:
[test]

cmd/s2i/main.go
```diff
@@ -203,7 +207,7 @@ $ s2i build . centos/ruby-22-centos7 hello-world-app
 	buildCmd.Flags().StringVar(&(oldScriptsFlag), "scripts", "", "Specify a URL for the assemble and run scripts")
 	buildCmd.Flags().StringVarP(&(oldDestination), "location", "l", "", "Specify a destination location for untar operation")
 	buildCmd.Flags().StringVarP(&(cfg.Destination), "destination", "d", "", "Specify a destination location for untar operation")
-	buildCmd.Flags().BoolVar(&(cfg.ForcePull), "force-pull", true, "Always pull the builder image even if it is present locally")
+	buildCmd.Flags().VarP(&(cfg.BuilderPullPolicy), "pull-policy", "p", "Specify how to pull the builder image (always, never or if-not-present)")
```
soltysh: just PullPolicy, please

mfojtik: PullPolicy is the type ;-)

mfojtik: maybe would be good to separate Builder and the Incremental image pulls in future

pkg/api/types.go
```diff
@@ -230,3 +241,37 @@ const (
...
 }
+
+// PullPolicy specifies a type for the method used to retrieve the Docker image
+type PullPolicy string
+
+// String implements the String() function of pflags.Value so this can be used as
+// command line parameter.
+// This will also default to "if-not-present" if this parameter is not set.
+func (p *PullPolicy) String() string {
+	if len(string(*p)) == 0 {
```
soltysh: Hmmm... not sure, I'd rather see when it's actually empty and not default it in some hidden place.

mfojtik: @soltysh this is really just a hack to make s2i build --help show "if-not-present" as a default option, it is not used anywhere else in code

pkg/build/config.go
```diff
@@ -17,10 +17,16 @@ func GenerateConfigFromLabels(image string, config *api.Config) error {
...
+
+	switch config.BuilderPullPolicy {
+	case api.PullAlways:
 		source, err = d.PullImage(image)
-	} else {
+	case api.PullIfNotPresent:
 		source, err = d.CheckAndPullImage(image)
+	case api.PullNever:
+		source, err = d.CheckImage(image)
+	default:
```
soltysh: That default should not happen. IIRC, we have a validation method (see `pkg/api/validation`) which is being called before running a build. You should create a validation there.

cmd/s2i/main.go
soltysh: For the latter, I can agree, for the former, I don't care, we can have both names being identical. One is type and other is a field in a struct, that doesn't conflict, imho.

mfojtik: @soltysh I hate having "PullPolicy PullPolicy", so will keep it this way. It is more expressive now as it describes more that this policy is for fetching the Builder image

soltysh:
You're getting my LGTM, but I still dislike the string hack for printing the default value for PullPolicy :stuck_out_tongue_winking_eye:

mfojtik:
@soltysh it is not hack, just not ideal interface that pflags gives us :-)

mfojtik:
amazon hates me, [test]

soltysh:
Did we got banned by gh :wink: for testing too much 
[test]

cmd/s2i/main.go
```diff
@@ -203,7 +207,7 @@ $ s2i build . centos/ruby-22-centos7 hello-world-app
 	buildCmd.Flags().StringVar(&(oldScriptsFlag), "scripts", "", "Specify a URL for the assemble and run scripts")
 	buildCmd.Flags().StringVarP(&(oldDestination), "location", "l", "", "Specify a destination location for untar operation")
 	buildCmd.Flags().StringVarP(&(cfg.Destination), "destination", "d", "", "Specify a destination location for untar operation")
-	buildCmd.Flags().BoolVar(&(cfg.ForcePull), "force-pull", true, "Always pull the builder image even if it is present locally")
```
bparees: deprecate, don't remove.

cmd/s2i/main.go
bparees: +1 to keeping this as is for both reasons @mfojtik stated.

cmd/s2i/main.go
```diff
@@ -267,7 +275,7 @@ func newCmdRebuild(cfg *api.Config) *cobra.Command {
 	buildCmd.Flags().BoolVar(&(cfg.Incremental), "incremental", false, "Perform an incremental build")
 	buildCmd.Flags().BoolVar(&(cfg.RemovePreviousImage), "rm", false, "Remove the previous image during incremental builds")
 	buildCmd.Flags().StringVar(&(cfg.CallbackURL), "callback-url", "", "Specify a URL to invoke via HTTP POST upon build completion")
-	buildCmd.Flags().BoolVar(&(cfg.ForcePull), "force-pull", true, "Always pull the builder image even if it is present locally")
```
bparees: again deprecate

cmd/s2i/main.go
```diff
@@ -324,7 +336,7 @@ func newCmdUsage(cfg *api.Config) *cobra.Command {
 	usageCmd.Flags().StringP("env", "e", "", "Specify an environment var NAME=VALUE,NAME2=VALUE2,...")
 	usageCmd.Flags().StringVarP(&(cfg.ScriptsURL), "scripts-url", "s", "", "Specify a URL for the assemble and run scripts")
 	usageCmd.Flags().StringVar(&(oldScriptsFlag), "scripts", "", "Specify a URL for the assemble and run scripts")
-	usageCmd.Flags().BoolVar(&(cfg.ForcePull), "force-pull", true, "Always pull the builder image even if it is present locally")
```
bparees: deprecate.

cmd/s2i/main.go
```diff
@@ -324,7 +336,7 @@ func newCmdUsage(cfg *api.Config) *cobra.Command {
...
+	usageCmd.Flags().VarP(&(cfg.BuilderPullPolicy), "pull-policy", "p", "Specify how to pull the builder image (always, never or if-not-present)")
```
bparees: s/how/when/

(same everywhere else)

pkg/api/types.go
```diff
@@ -12,6 +14,15 @@ const (
 	KubernetesNamespace = "io.k8s."
 )
 
+const (
+	// PullAlways means that we always attempts to pull the latest image.
```
bparees: attempt

pkg/api/types.go
```diff
@@ -12,6 +14,15 @@ const (
...
+	PullAlways PullPolicy = "always"
+	// PullNever means that we never pulls an image, but only uses a local image.
```
bparees: pull
use

pkg/build/strategies/sti/sti.go
```diff
@@ -323,15 +324,19 @@ func (b *STI) Exists(config *api.Config) bool {
 		return false
 	}
 
-	// can only do incremental build if runtime image exists, so always pull image
-	previousImageExists, _ := b.docker.IsImageInLocalRegistry(config.Tag)
```
bparees: you've broken this behavior which was by design to pull the previous version of the application image if it didn't exist, regardless of the "force pull" flag.  

whether or not we pull the previous application image should be entirely independent of the BuilderPullPolicy.

The right way to fix this is to go ahead and introduce a PreviousImagePullPolicy.

cmd/s2i/main.go
bparees: ForcePull=true should set pull policy always, and ForcePull=false should set PullIfNotPresent.  

and we'll have to change the default, unfortunately, to ForcePull=false, otherwise it would conflict with the default of PINP for BuilderPullPolicy.

mfojtik: done

pkg/build/strategies/sti/sti.go
mfojtik: fixed

pkg/docker/util.go
```diff
@@ -155,3 +159,86 @@ func parseRepositoryTag(repos string) (string, string, string) {
 	}
 	return repos, "", ""
 }
+
+// PullImage pulls the Docker image specifies by name taking the pull policy
+// into the account.
+// TODO: The 'force' option will be removed
+func PullImage(name string, d Docker, policy api.PullPolicy, force bool) (*PullResult, error) {
```
mfojtik: fyi this is here just to avoid writing the same switch everytime we want to pull the image (usage, incremental, build, etc)

mfojtik:
@soltysh @bparees ready for the second round, summary of changes:

* Keep `--force-pull` around but default it to 'false' now (if you set it to true, you get nice deprecation warning)
* Refactored the pkg/test thing (moved docker to pkg/dockerfake_docker.go) to avoid import cycles
* Refactored image pulling to avoid code duplication

mfojtik:
As a follow up we should:

* Refactor the pkg/test entirely (ideally get rid of that package as it causes import cycles)
* Reorganize the packages a little bit (currently there are things mixed together like CheckAllowedUsers in pkg/docker where it should really be in util/user, etc... there are much examples like this and that require more general refactoring)

cmd/s2i/main.go
```diff
@@ -122,6 +122,10 @@ $ s2i build . centos/ruby-22-centos7 hello-world-app
 				}
 			}
 
+			if cfg.ForcePull == true {
+				glog.Warning("DEPRECATED: The '--force-pull' option is deprecated. Use '--pull-policy' instead")
```
soltysh: Make all the deprecation info identical, see [here](https://github.com/openshift/source-to-image/blob/master/cmd/s2i/main.go#L163) and [here](https://github.com/openshift/source-to-image/blob/master/cmd/s2i/main.go#L167). Doing the latter two your way is also a good option, or making a function for that would be the best :wink:

cmd/s2i/main.go
```diff
@@ -122,6 +122,10 @@ $ s2i build . centos/ruby-22-centos7 hello-world-app
 				}
 			}
 
+			if cfg.ForcePull == true {
```
soltysh: `if cfg.ForcePull {` should be enough.

cmd/s2i/main.go
```diff
@@ -267,7 +276,8 @@ func newCmdRebuild(cfg *api.Config) *cobra.Command {
 	buildCmd.Flags().BoolVar(&(cfg.Incremental), "incremental", false, "Perform an incremental build")
 	buildCmd.Flags().BoolVar(&(cfg.RemovePreviousImage), "rm", false, "Remove the previous image during incremental builds")
 	buildCmd.Flags().StringVar(&(cfg.CallbackURL), "callback-url", "", "Specify a URL to invoke via HTTP POST upon build completion")
-	buildCmd.Flags().BoolVar(&(cfg.ForcePull), "force-pull", true, "Always pull the builder image even if it is present locally")
+	buildCmd.Flags().BoolVar(&(cfg.ForcePull), "force-pull", false, "DEPRECATED: Always pull the builder image even if it is present locally")
```
soltysh: Again, it would be good to make all the detracted flags look identical, preferring your version.

pkg/api/types.go
```diff
@@ -12,6 +14,24 @@ const (
...
+
+	// PullIfNotPresent means that we pull if the image isn't present on disk.
+	PullIfNotPresent PullPolicy = "if-not-present"
+
+	// DefaultBuilderPullPolicy specifies the default pull policy to use
+	DefaultBuilderPullPolicy = PullIfNotPresent
+
+	// DefaultPreviousImagePullPolicy specifies policy for pulling the previously
+	// build Docker image when doing incremental build
+	DefaultPreviousImagePullPolicy = PullAlways
```
soltysh: If you're providing that option, how about making it accessible for users to set?

pkg/api/types.go
```diff
@@ -65,6 +85,19 @@ type Config struct {
...
+
+	// PreviousImagePullPolicy specifies when to pull the previously build image
+	// when doing incremental build
+	PreviousImagePullPolicy PullPolicy
+
+	// ForcePull defines if the builder image should be always pulled or not.
+	// This is now deprecated by BuilderPullPolicy and will be removed soon.
+	// Setting this to 'true' equals setting BuilderPullPolicy to 'PullAlways'.
+	// Setting this to 'false' equals setting BuilderPullPolicy to 'PullIfNotPresent'
+	ForcePull bool
```
soltysh: Add a deprecation comment.

pkg/api/types.go
mfojtik: 1) how would you call it? --previous-pull-policy? ;-)
2) none if of the names I have in my mind are good
3) you don't want to confuse people (at least for now)

mfojtik: @soltysh so if you

1) come up with a use-case where setting this is actually useful
2) and give me reasonable option name

I can expose it to users :-)

pkg/api/types.go
mfojtik: This is now deprecated by BuilderPullPolicy and will be removed soon. 

this is not enough?

pkg/api/types.go
soltysh: Yeah, the usual naming problem :wink:

pkg/api/types.go
soltysh: I'm blind, sorry.

pkg/build/strategies/sti/sti.go
```diff
@@ -323,15 +323,18 @@ func (b *STI) Exists(config *api.Config) bool {
...
-	if !previousImageExists || config.ForcePull {
-		if image, _ := b.incrementalDocker.PullImage(config.Tag); image != nil {
-			previousImageExists = true
-		}
+	policy := config.PreviousImagePullPolicy
+	if len(policy) == 0 {
+		policy = api.DefaultPreviousImagePullPolicy
+	}
```
soltysh: Shouldn't that defaulting happen in `cmd/s2i/main.go` like all the others? I told you to create defaults :stuck_out_tongue_winking_eye:

pkg/docker/util.go
soltysh: :+1:

soltysh:
Nits mostly, other than that LGTM.

pkg/build/strategies/sti/sti.go
mfojtik: what about origin? i think we want to set default there as well, although in origin it will be PullAlways

soltysh: It can always have own defaulting function, remember origin is using slightly different flow, and having such defaulting function gives us confidence when and what happens.

mfojtik: @soltysh I think defaulter will be also on my list for the future refactoring :-) but currently we have just PullPolicy we have to default and don't want to spend a lot of time refactoring code just to default one option ;-)

soltysh: Create issues, so we don't forget.

pkg/api/describe/describer.go
```diff
@@ -39,7 +39,7 @@ func DescribeConfig(config *api.Config) string {
 			fmt.Fprintf(out, "Incremental Image Pull User:\t%s\n", config.IncrementalAuthentication.Username)
 		}
 		fmt.Fprintf(out, "Remove Old Build:\t%s\n", printBool(config.RemovePreviousImage))
-		fmt.Fprintf(out, "Force Pull:\t%s\n", printBool(config.ForcePull))
+		fmt.Fprintf(out, "Pull Policy:\t%s\n", config.BuilderPullPolicy)
```
bparees: Builder Pull Policy

bparees:
needs squash

cmd/s2i/main.go
mfojtik: done

mfojtik:
@bparees i want to remove the last commit before push, it is there to help debug the integration failure which is not clear to me...

pkg/docker/util.go
```diff
@@ -155,3 +159,85 @@ func parseRepositoryTag(repos string) (string, string, string) {
...
+	var (
+		image *client.Image
+		err   error
+	)
+	switch policy {
+	case api.PullIfNotPresent:
+		image, err = d.CheckAndPullImage(name)
+	case api.PullAlways:
+		image, err = d.PullImage(name)
+		fallthrough
```
soltysh: :8ball:

cmd/s2i/main.go
```diff
@@ -200,10 +211,11 @@ $ s2i build . centos/ruby-22-centos7 hello-world-app
 	buildCmd.Flags().StringVarP(&(cfg.Ref), "ref", "r", "", "Specify a ref to check-out")
 	buildCmd.Flags().StringVar(&(cfg.CallbackURL), "callback-url", "", "Specify a URL to invoke via HTTP POST upon build completion")
 	buildCmd.Flags().StringVarP(&(cfg.ScriptsURL), "scripts-url", "s", "", "Specify a URL for the assemble and run scripts")
-	buildCmd.Flags().StringVar(&(oldScriptsFlag), "scripts", "", "Specify a URL for the assemble and run scripts")
+	buildCmd.Flags().StringVar(&(oldScriptsFlag), "scripts", "", "DEPRECATED: Specify a URL for the assemble and run scripts")
 	buildCmd.Flags().StringVarP(&(oldDestination), "location", "l", "", "Specify a destination location for untar operation")
```
soltysh: Location is also deprecated, pretty please add DEPRECATED in front of it.

mfojtik: i'm doing that in follow up with main.go refactoring.

soltysh: k

hack/test-stirunimage.sh
```diff
@@ -34,16 +34,13 @@ fi
 
 trap cleanup EXIT SIGINT
 
-echo
-echo 
-echo
-
-sti build git://github.com/bparees/openshift-jee-sample openshift/wildfly-8-centos test-jee-app --run=true &> "${STI_ROOT}/hack/sti-run.log" &
+echo "Running 'sti build --run=true ...'"
+s2i build git://github.com/bparees/openshift-jee-sample openshift/wildfly-8-centos test-jee-app --run=true &> "${STI_ROOT}/hack/sti-run.log" &
```
soltysh: This should be also fixed to use the `openshift/wildfly-81-centos7` instead, can be part of that refactor of yours.

soltysh:
LGTM, @bparees merge it :smile:

hack/test-stirunimage.sh
mfojtik: @soltysh this is a tribute to @bparees :-)

soltysh: I know, I'm just saying to use newer image he still maintains :wink:

bparees: yeah please change it :)

bparees: i'd even consider changing it to the php app or something because that'll build faster than the JEE app does, so tests will be faster.

pkg/build/strategies/sti/sti.go
```diff
@@ -401,7 +404,7 @@ func (b *STI) Execute(command string, config *api.Config) error {
 		Image:           config.BuilderImage,
 		Stdout:          outWriter,
 		Stderr:          errWriter,
-		PullImage:       config.ForcePull,
+		PullImage:       config.BuilderPullPolicy == api.PullAlways || config.BuilderPullPolicy == api.PullIfNotPresent,
```
bparees: doesn't this always result in a Pull for PullIfNotPresent?  In the case of PullIfNotPresent, don't you need to manually check if the image exists locally first, and then set this to true or false accordingly?

pkg/docker/util.go
```diff
@@ -155,3 +159,85 @@ func parseRepositoryTag(repos string) (string, string, string) {
...
+// TODO: The 'force' option will be removed
+func PullImage(name string, d Docker, policy api.PullPolicy, force bool) (*PullResult, error) {
+	// TODO: Remove this after we deprecate --force-pull
+	if force {
+		policy = api.PullAlways
+	}
+
+	// When pull policy not specified, default to DefaultPullPolicy
+	if len(policy) == 0 {
+		policy = api.DefaultBuilderPullPolicy
```
bparees: you're defaulting to the Builder policy, but this method might be invoked for Builder images or Previous images...  that's a problem.  You should just require a policy be passed in and error out otherwise.

pkg/build/strategies/sti/sti.go
bparees: alternatively if you assume by this point in the code you've already called PullImage (which will have pulled the image if appropriate), then you should always be passing "false" here because the right image (freshly pulled or not) will already exist locally.

pkg/docker/util.go
mfojtik: @bparees the other place this is used is to pull previous image that has the policy explicitly set.
I think we should not default to any policy and just return error when the policy is not set.

pkg/build/strategies/sti/sti.go
mfojtik: right, false should be safe because the image should be pulled in all cases

hack/test-stirunimage.sh
mfojtik: i'll do that as follow up, don't want to make this PR fatter :-)

pkg/docker/util.go
soltysh: I'd again suggest creating defaults, but I remember you've it already in the followup PR :wink:

mfojtik: It is not needed as we run PullImage() before so the image should always be
available or this should error out.

On Fri, Nov 6, 2015 at 2:14 PM, Maciej Szulik <notifications@github.com>
wrote:

> In pkg/docker/util.go
> <https://github.com/openshift/source-to-image/pull/324#discussion_r44134559>
> :
>
> > @@ -155,3 +159,85 @@ func parseRepositoryTag(repos string) (string, string, string) {
> >  	}
> >  	return repos, "", ""
> >  }
> > +
> > +// PullImage pulls the Docker image specifies by name taking the pull policy
> > +// into the account.
> > +// TODO: The 'force' option will be removed
> > +func PullImage(name string, d Docker, policy api.PullPolicy, force bool) (*PullResult, error) {
> > +	// TODO: Remove this after we deprecate --force-pull
> > +	if force {
> > +		policy = api.PullAlways
> > +	}
> > +
> > +	// When pull policy not specified, default to DefaultPullPolicy
> > +	if len(policy) == 0 {
> > +		policy = api.DefaultBuilderPullPolicy
>
> I'd again suggest creating defaults, but I remember you've it already in
> the followup PR [image: :wink:]
>
> —
> <https://github.com/openshift/source-to-image/pull/324/files#r44134559>.
>



-- 

bparees: > @bparees the other place this is used is to pull previous image that has the policy explicitly set.
I think we should not default to any policy and just return error when the policy is not set.

isn't that what i suggested? :)

mfojtik: Yes, it is :-)

On Fri, Nov 6, 2015 at 4:00 PM, Ben Parees <notifications@github.com> wrote:

> In pkg/docker/util.go
> <https://github.com/openshift/source-to-image/pull/324#discussion_r44144586>
> :
>
> > @@ -155,3 +159,85 @@ func parseRepositoryTag(repos string) (string, string, string) {
> >  	}
> >  	return repos, "", ""
> >  }
> > +
> > +// PullImage pulls the Docker image specifies by name taking the pull policy
> > +// into the account.
> > +// TODO: The 'force' option will be removed
> > +func PullImage(name string, d Docker, policy api.PullPolicy, force bool) (*PullResult, error) {
> > +	// TODO: Remove this after we deprecate --force-pull
> > +	if force {
> > +		policy = api.PullAlways
> > +	}
> > +
> > +	// When pull policy not specified, default to DefaultPullPolicy
> > +	if len(policy) == 0 {
> > +		policy = api.DefaultBuilderPullPolicy
>
> @bparees <https://github.com/bparees> the other place this is used is to
> pull previous image that has the policy explicitly set.
> I think we should not default to any policy and just return error when the
> policy is not set.
>
> isn't that what i suggested? :)
>
> —
> <https://github.com/openshift/source-to-image/pull/324/files#r44144586>.
>



-- 

bparees:
lgtm.

@gabemontero heads up, this may cause some conflicts in your branch.  hopefully only around the test scripts.

[merge]

soltysh:
:tada:

PR #327, opened 2015-11-04
always use copy when the source is local

bparees:


bparees:
@mfojtik @gabemontero this is needed because with the changes @gabemontero did to clean up the local path handling, s2i now fails on urls like "file:///some/path" where /some/path is a git repo.  Because we ultimately go down the git clone path which ends up checking if "file:///some/path" exists as a file, which of course it doesn't (due to the protocol prefix).  

basically the problems start here:
https://github.com/openshift/source-to-image/blob/master/pkg/scm/git/clone.go#L25

that logic should probably be just checking if it's a valid clone spec period, not a valid remote clone spec.  However "file://relative/path" is then considered a valid clone spec, but git doesn't actually successfully clone a path like that.

this logic also isn't working for relative paths (git clone against a relative path fails with this code), so that's another reason to do away with it.

There are other fixes, but this is expedient and takes us in the direction of ripping out git support from s2i, which at this point i'm strongly in favor of.

bparees:
[test]

bparees:
it's actually worse than i thought, the changes also broke s2i usage, i'm just going to revert it for now so we can get the push job working again.

mfojtik:
@bparees we should get rid of GIT integration in S2I and just make a wrapper script that will clone and stream the content of GIT repo as tar to s2i builder (long term)

bparees:
>  @bparees we should get rid of GIT integration in S2I and just make a wrapper script that will clone and stream the content of GIT repo as tar to s2i builder (long term)

yes.  that's what @smarterclayton suggested a week or so ago and i agreed then, i agree even more now.

mfojtik:
@bparees trello coverage: https://trello.com/c/4bxcGz5w/750-remove-git-support-from-s2i

PR #329, opened 2015-11-04
Revert "don't require .git suffix; consolidate URL validation and no protocol construction"

bparees:
This reverts commit 73dd2c2d0946e098a233f6ac56a4e2d91e22204d.


bparees:
@gabemontero this commit appears to be breaking at least 2 use cases that we test with our s2i image testing.

if you run a "make test" on one of the s2i image repos (with a bad version of s2i on your path) you'll see the issues.

bparees:
[merge]

bparees:
@mfojtik also fyi.

bparees:
(clearly we also need some new test cases in the s2i repo)

PR #330, opened 2015-11-05
Dotgit2

gabemontero:
@bparees fixes for bugs that lead to the recent revert of the git clone spec validation / modification changes, as well as the addition of tests for the failing scenarios.

WRT the git mechanics behind this PR, I did a `git reset HEAD~1` to undo the revert, then applied my fixes which were mods to the original changes reverted.

Please let me know if those git mechanics were wrong (and perhaps I just nuke this branch and try again with whatever git calls are proscribed.

Otherwise, PTAL

### DIFF
--- a/hack/test-stirunimage.sh
+++ b/hack/test-stirunimage.sh
@@ -5,22 +5,44 @@ set -o nounset
 set -o pipefail
 
 STI_ROOT=$(dirname "${BASH_SOURCE}")/..
+KILLDONE=""
+STI_PID=""
 
 function time_now()
 {
   echo $(date +%s000)
 }
 
+tmpdir=$(time_now) 
+WORK_DIR="/tmp/${tmpdir}"
 function cleanup()
 {
     set +e
-    rm -f  "${STI_ROOT}/hack/sti-run.log"
+    rm -rf ${WORK_DIR}
     # use sigint so that sti post processing will remove docker container
-    if [ -z "$KILLDONE" ]; then
-	kill -2 "${STI_PID}"
+    if [ -R "$KILLDONE" ]; then
+	if [ -R "$STI_PID" ]; then
+	    kill -2 "${STI_PID}"
+	fi
     fi
     echo
     echo "Complete"
+}
+
+function check_result() {
+  local result="$1"
+  if [[ "$result" != "0" ]]; then
+      info "TEST FAILED (${result})"
+      cat $2
+      cleanup
+      exit $result
+  fi
+}
+
+function test_debug() {
+    echo
+    echo $1
+    echo
 }
 
 set +e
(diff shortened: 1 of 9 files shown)


bparees:
needs a rebase

bparees:
the content in commit 892ee875418b7994315f0375b526049cc899879d does not look like a proper reversion of my revert.  Did you use "git revert 0cae25aa59f9b99a213a4cb9b63c4ab9974505fd" to undo my revert?

It seems like the un-revert is mixed between your two commits.

gabemontero:
Weird - the PR comments were not embedded in line like they usually are - perhaps cause it needs a rebase.  I'll answer things here, then work on the rebase.
- as I mentioned in the opening text, I was worried I did the revert wrong ... was unaware of the `git revert` and did a `git rebase` ... the results were not optimal;  I'll see if I can recover, but I may just blow away this PR/branch and retry with `git revert`
- for base -R, see http://www.gnu.org/software/bash/manual/bashref.html#Bash-Conditional-Expressions:  `True if the shell variable varname is set and is a name reference.`
- usage/downloader:  the path was https://github.com/openshift/source-to-image/blob/master/pkg/build/strategies/sti/usage.go#L23-L24 and then https://github.com/openshift/source-to-image/blob/master/pkg/build/strategies/sti/sti.go#L105-L110 ... with the new validation, an empty string  (since there is no source url with a `s2i usage`) was passed into DownloaderForSource and with the more strict validation logic, an error was getting returned
- I think I can map the other suggestions to their intended locations (again, the commenting on this PR is showing up in an atypical fashion for me), but I'll sort out the rebase/revert stuff first.

bparees:
> Weird - the PR comments were not embedded in line like they usually are - perhaps cause it needs a rebase

that's because i commented directly on your commits, not on the PR files.  I was looking at the commits individually to try to differentiate the un-reversion from your new fixes.

gabemontero:
I see - as i am sure you saw - i nuked the pr/branch and am redoing it with
git revert - looking cleaner

bparees:
and yeah, sorry, I should have read your intro more carefully instead of jumping into the review.

gabemontero:
No worries

bparees:
> usage/downloader: the path was https://github.com/openshift/source-to-image/blob/master/pkg/build/strategies/sti/usage.go#L23-L24 and then https://github.com/openshift/source-to-image/blob/master/pkg/build/strategies/sti/sti.go#L105-L110 ... with the new validation, an empty string (since there is no source url with a s2i usage) was passed into DownloaderForSource and with the more strict validation logic, an error was getting returned

I'm thinking it would be better for New() to be more intelligent and know when it's being invoked for Usage and that way it could skip the downloader initialization entirely (and maybe other stuff).  Getting that complexity out of the usage path would make it less likely to get broken in the future.

> for base -R, see http://www.gnu.org/software/bash/manual/bashref.html#Bash-Conditional-Expressions: True if the shell variable varname is set and is a name reference. 

i'm not entirely clear on what a "name reference" is in bash (my best guess is it's checking if the value of the variable is the name of another variable) but i don't think it's what you want here"-n" should be right, or possibly -v.  At least in my quick test i couldn't get -R to return true.

bparees:
(ok I got -R to return true for a variable of value "" and a completely undefined variable name, neither of which sounds like what the doc is describing the behavior to be, but either way -n is the more typical approach to check for a non-zero length value)

gabemontero:
Ok (new) - i was originally on that path but saw the override and it looked
like an existing customizer intended for cases like this (and i would be
told toleverage it ;-) ).  I can switch gears.

On the bash options, i could not get -v or - n to do what i needed.  Will
circle back and revisit after the other comments. Thx

gabemontero:
Sanity check:  looking into Usage, with its calls to Cleanup, Prepare and Execute, while it does not need the downloader, it seems it would need the other artifacts set up in New (including the Ignorer and Layered, since those are called by Prepare/Execute).

Thoughts/corrections?

bparees:
I think Usage is way overly complicated right now.  It looks like it has the ability to pull down remote scripts using the "-s" argument, then layer that usage script into the image via a layered build (in case that image does not have tar), then execute the resulting image.

We should just strip this down to exec'ing the image they specified with a usage command.  It seems ridiculous to run "s2i usage someimage -s remotescripts" just to get the usage information... you might as well go look at the usage script at "remotescripts" at that point.

@soltysh @mfojtik agree?

soltysh:
:+1:

PR #331, opened 2015-11-06
Updating build process to match origin.

soltysh:
1. Added make release
2. Builds are placed in _output/local/bin/$(platform) directory
3. Added i386 binaries in release

@mfojtik @bparees ptal


bparees:
I think this has the same bug i have a PR open for here:
https://github.com/openshift/origin/pull/5567

soltysh:
Will check, thx for the heads up.

bparees:
[merge] to see if this resolves the issue @danmcp is seeing in travis, or at least ensure we're debugging the scripts we want and not the scripts we had.

bparees:
@soltysh looks like you need to fix up some scripts since you changed the output path: 
```
[INFO] Problem with sti --run=true, dumping hack/../hack/sti-run.log
hack/test-stirunimage.sh: line 39: s2i: command not found
```

soltysh:
@bparees this relies on https://github.com/openshift/vagrant-openshift/pull/360 being merged and deployed first. Otherwise we won't be able to run s2i tests nor image tests.

bparees:
vagrant has been updated.  [merge]

Issue #314, opened 2015-10-20
s2i on Windows fails with E1020 


myfear:
D:\temp\test>s2i build git://github.com/bparees/openshift-jee-sample openshift/wildfly-8-centos test-jee-app
W1020 08:43:10.926903 07004 docker.go:330] The 'io.s2i.destination' label is deprecated. Use "io.openshift.s2i.destination" instead.
E1020 08:43:11.095424 07004 util.go:85] /bin/sh: usrlocalstiassemble: command not found
E1020 08:43:11.233442 07004 fs.go:103] Error removing directory 'C:\Users\meisele\AppData\Local\Temp\sti292704367': remove C:\Users\meisele\AppDa
ta\Local\Temp\sti292704367\upload\src\.git\objects\pack\pack-efabe02dcd36d820e628c92fadd90f6d773b3fee.idx: Access is denied.

D:\temp\test>s2i version
s2i v1.0.2

bparees:
can you rerun with --loglevel=5 ?

bparees:
also please use the "openshift/wildfly-81-centos7" image instead, though i don't expect it to resolve the access denied issue.

bparees:
i've also just created a new release of s2i (v1.0.3), it might be worth using that release instead, though again i don't think it will resolve this particular issue.

myfear:
d:\temp\test>s2i build git://github.com/bparees/openshift-jee-sample openshift/wildfly-8-centos test-jee-app --loglevel=5
I1020 20:34:49.531757 09736 docker.go:228] Pulling image openshift/wildfly-8-centos:latest

Builder Image:          openshift/wildfly-8-centos
Source:                 git://github.com/bparees/openshift-jee-sample
Output Image Tag:       test-jee-app
Incremental Build:      disabled
Remove Old Build:       disabled
Force Pull:             ←[1menabled←[0m
Quiet:                  disabled
Layered Build:          disabled
Docker Endpoint:        tcp://vagrant.f8:2375

I1020 20:34:54.917940 09736 docker.go:228] Pulling image openshift/wildfly-8-centos:latest
I1020 20:34:57.989331 09736 sti.go:123] Preparing to build test-jee-app
I1020 20:34:57.998832 09736 clone.go:28] Cloning into C:\Users\meisele\AppData\Local\Temp\sti128951035\upload\src
I1020 20:35:05.151740 09736 docker.go:211] Image openshift/wildfly-8-centos:latest available locally
[...]

PR #337, opened 2015-11-10
WIP: Stop using filepath.Join on windows

mfojtik:
The problem is that we should never using `filepath.Join` when constructing the path we later use in container (like assemble path). If this code is called on Windows, the final path will not be UNIX but Windows path, which will lead to stripping all path separators.

Fixes: https://github.com/openshift/source-to-image/issues/314

### DIFF
--- a/pkg/docker/docker.go
+++ b/pkg/docker/docker.go
@@ -3,7 +3,7 @@ package docker
 import (
 	"fmt"
 	"io"
-	"path/filepath"
+	"path"
 	"strings"
 	"sync"
 
@@ -371,7 +371,9 @@ func runContainerTar(opts RunContainerOptions, config docker.Config, imageMetada
 	if opts.ExternalScripts {
 		// for external scripts we must always append 'scripts' because this is
 		// the default subdirectory inside tar for them
-		commandBaseDir = filepath.Join(tarDestination, "scripts")
+		// NOTE: We use path.Join instead of filepath.Join to avoid converting the
+		// path to UNC (Windows) format as we always run this inside container.
+		commandBaseDir = path.Join(tarDestination, "scripts")
 		glog.V(2).Infof("Both scripts and untarred source will be placed in '%s'", tarDestination)
 	} else {
 		// for internal scripts we can have separate path for scripts and untar operation destination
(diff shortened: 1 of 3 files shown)


mfojtik:
This is WIP because I need to test this on Windows.

pkg/util/fs.go
```diff
@@ -40,6 +42,15 @@ type fs struct {
 	runner CommandRunner
 }
 
+// JoinPath joins the path elements always using the UNIX way. This function
+// should be used when constructing path that is valid for running in container.
+func JoinPath(elem ...string) string {
+	if runtime.GOOS == "windows" {
+		return strings.Join(elem, "/")
```
soltysh: Isn't windows using the other one, iow `\` ?

mfojtik: read the comment :-)

pkg/util/fs.go
```diff
@@ -40,6 +42,15 @@ type fs struct {
 	runner CommandRunner
 }
 
+// JoinPath joins the path elements always using the UNIX way. This function
+// should be used when constructing path that is valid for running in container.
+func JoinPath(elem ...string) string {
+	if runtime.GOOS == "windows" {
+		return path.Join(elem...)
+	}
+	return filepath.Join(elem...)
```
soltysh: What's the point in using `filepath.Join` if you always want to have the unix path, just use `path.Join` in that case. 
Currently I'm reading this as: if I'm on windows use `/`, otherwise use `os.Separator` which ends up being `/` as well. Am I missing something here?

mfojtik: @soltysh yeah this function is pointless, I can just use path.Join on places where it is required... I'm reworking this PR anyway :-)

mfojtik: fixed

pkg/docker/docker.go
```diff
@@ -371,7 +371,9 @@ func runContainerTar(opts RunContainerOptions, config docker.Config, imageMetada
 	if opts.ExternalScripts {
 		// for external scripts we must always append 'scripts' because this is
 		// the default subdirectory inside tar for them
-		commandBaseDir = filepath.Join(tarDestination, "scripts")
+		// NOTE: We use path.Join instead of filepath.Join to avoid converting the
+		// path to UNC (Windows) format as we always run this inside container.
+		commandBaseDir = path.Join(tarDestination, "scripts")
```
bparees: need a comment like this everywhere you're doing this.

bparees:
would like a comment on all the places where the path is being used inside the container, otherwise lgtm.

soltysh:
LGTM

pkg/docker/docker.go
mfojtik: comments added.

mfojtik:
[merge]

PR #339, opened 2015-11-11
log errors when docker client fails

bparees:


bparees:
these errors often get wrapped/eaten by callers, so it'd be better to log them when they occur in case we need the details.

@mfojtik @gabemontero ptal.

bparees:
[test]

mfojtik:
LGTM

gabemontero:
+1

bparees:
[merge]

PR #341, opened 2015-11-12
readme path and sample are wrong

bparees:


bparees:
@mfojtik @soltysh fyi
[merge]

Issue #306, opened 2015-10-13
s2i build should print a warning when s2i scripts are in a wrong dir


jhadvig:
I've seen several times (mostly in QA TCs) that they put s2i scripts in a wrong path, usually in `.sti/` and not `.sti/bin/`. We should detect that and print a warning in that case, to make debugging easier for users.
@soltysh @mfojtik @bparees thoughts ?

mfojtik:
+1

soltysh:
I wonder how complicated the code would be if we'd pick both `.sti/` and `.sti/bin` like you proposed on IRC. But the warning has big fat :+1: from my side.

mfojtik:
don't that will cause inconsistency :-)

soltysh:
Yeah, that's the other thing I'm a bit worried as well.

jhadvig:
Agreed, I think it would be better to have it on one place. What if user would have different run script in both of those directions. Which one would have priority ?
Warning is enough I think :)

mfojtik:
@jhadvig I assume this is only the case when you have s2i scripts in the source repository

PR #343, opened 2015-11-12
Increate the timeout when tar closes the stream

mfojtik:
Fixes: https://github.com/openshift/source-to-image/issues/306

@bparees @csrwng I think increasing the timeout here will help with extracting large files.


bparees:
[merge]

Issue #81, opened 2014-11-10
Build and run as different users


accursoft:
I would like the built image to run with a more restricted user than the one which performed the build. This could be supported by running `assemble` as a non-default user, or by changing the image's default user after assembly.

It can be done now with `docker run -u`, but it would be nice to support this automatically.

accursoft:
`--assemble-user` changes the image's default user, so `docker run` will still use the same user as assemble.

wanghaoran1988:
if we specify the --assemble-user option when do sti build with the language images openshift provided, the build will failed, since the images are using default user 1001, and the src path directory inside of the images like :/opt/app-root/src is owned by user 1001 . so , there is a problem here.

mfojtik:
@accursoft good catch, I have to take the User from the builder image and specify that user when committing the container after build. The `Config.User` will then be set to 1001 (which is the openshift s2i image default user), however the `Config.ContainerConfig.User` will be set to `--assemble-user` (there is no way in docker to change that other than 're-running' the container and committing it. 

With https://github.com/openshift/source-to-image/pull/354 you should docker run the built image and it will run using the original builder id.

mfojtik:
@wanghaoran1988 that is correct, however you might have builder image that will allow assemble to run only as user X and the run the final image as user Y. The OpenShift image are not build that way. However, you can try to build using `--assemble-user=root` (which is how I'm testing this).

accursoft:
Should `save-artifacts` also run as `assemble-user`? This tripped me up when `cd ~` went to the wrong directory, but it could also cause issues if the artifacts can't be read by the run-time user.

mfojtik:
@accursoft probably yes, because the save-artifacts is reading stuff created by assemble. Currently the assemble script has to make it readable for the default image user.

PR #345, opened 2015-11-12
Add --assemble-user option

mfojtik:
Fixes: https://github.com/openshift/source-to-image/issues/81

This allows to execute the assemble script under specified user.


mfojtik:
@soltysh @bparees PTAL

soltysh:
I'd like to see an integration test for that use case, otherwise looks good.

mfojtik:
@soltysh I can promise you I will add one, in future.

soltysh:
Then merge will have to wait for the future :stuck_out_tongue_winking_eye:  But seriously this kind of functionality is easily breakable and adding the int test case right now is highly required, imho. But @bparees has the power to overrule my statement.

mfojtik:
@soltysh this PR is just about passing the User down to RunContainerOpt and we have to verify that the User field is set properly which I do in the unit test. The integration test can run with different user, but it will not do too much and it would be hard to check if it really ran as that user ;-) (I can craft assemble that will store the output of `id -u`).

soltysh:
> I can craft assemble that will store the output of id -u

That's all I'm asking for.

mfojtik:
@soltysh we should simplify the integration tests to make it possible to do this easily. I can write that test, but it will make test integration more big and complex. We should brainstorm about how to make the integration simple.

soltysh:
With #347 in mind, LGTM.

bparees:
[merge]

Issue #66, opened 2014-10-23
Add more information to STI callback payload


csrwng:
Include builder image, source repo, date, time, image tag

bparees:
@mfojtik given that you just added a bunch of this information as labels, you want to add it to the callback too?

mfojtik:
@rhcarvalho what is the state of this? I can take care if this if you want.

rhcarvalho:
@mfojtik haven't started on it. Feel free to take it back, I still have other issues assigned to me ;)

bparees:
this is super low priority, please find something else to do... even docs work would be more important than this.  I'm not aware of anyone using the s2i callback feature, and if anything we're trying to strip down the amount of stuff s2i can do, not add to it.

PR #346, opened 2015-11-12
Add source info into callback

mfojtik:
Fixes: https://github.com/openshift/source-to-image/issues/66

### DIFF
--- a/pkg/util/callback.go
+++ b/pkg/util/callback.go
@@ -11,7 +11,7 @@ import (
 
 // CallbackInvoker posts results to a callback URL when a STI build is done.
 type CallbackInvoker interface {
-	ExecuteCallback(callbackURL string, success bool, messages []string) []string
+	ExecuteCallback(callbackURL string, success bool, labels map[string]string, messages []string) []string
 }
 
 // NewCallbackInvoker creates an instance of the default CallbackInvoker implementation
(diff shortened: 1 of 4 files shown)


mfojtik:
@csrwng @soltysh PTAL

soltysh:
LGTM

pkg/util/callback.go
```diff
@@ -39,6 +39,10 @@ func (c *callbackInvoker) ExecuteCallback(callbackURL string, success bool, mess
 		"success": success,
 	}
 
+	for name, value := range labels {
+		d[name] = value
+	}
```
csrwng: Is it necessary to keep the json that we send flat? Why not add a labels map under the main map?

mfojtik: @csrwng flat json is more easy to parse ;-) but yeah, I can add this into a labels field

mfojtik: @csrwng fixed

csrwng: My concern was that at some point the labels would conflict with the other entries we add to the map

csrwng:
LGTM

pkg/util/callback.go
mfojtik: @csrwng right, especially if we allow users to set arbitrary label

mfojtik:
[merge]

PR #348, opened 2015-11-12
Fix copying of local file directory

csrwng:
When specifying a local directory (that is not a git repo), we copy the files from it to a temporary upload/src directory using ```cp -a``` For the contents of the directory to be copied and not the directory itself to be copied, the directory spec must end in "/".

### DIFF
--- a/pkg/util/fs.go
+++ b/pkg/util/fs.go
@@ -1,10 +1,12 @@
 package util
 
 import (
+	"fmt"
 	"io"
 	"io/ioutil"
 	"os"
 	"path/filepath"
+	"strings"
 
 	"github.com/golang/glog"
 
@@ -20,6 +22,7 @@ type FileSystem interface {
 	Mkdir(dirname string) error
 	Exists(file string) bool
 	Copy(sourcePath, targetPath string) error
+	CopyContents(sourcePath, targetPath string) error
 	RemoveDirectory(dir string) error
 	CreateWorkingDirectory() (string, error)
 	Open(file string) (io.ReadCloser, error)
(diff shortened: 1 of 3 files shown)


bparees:
lgtm
[merge]

csrwng:
@soltysh @GrahamDumpleton fyi

GrahamDumpleton:
Confirmed now working for me okay.

pkg/util/fs.go
```diff
@@ -90,10 +93,25 @@ func (h *fs) Copy(sourcePath string, targetPath string) error {
 
 		targetPath = filepath.Join(targetPath, filepath.Base(sourcePath))
 	}
+	// TODO: Use the appropriate command for Windows
 	glog.V(5).Infof("cp -a %s %s", sourcePath, targetPath)
 	return h.runner.Run("cp", "-a", sourcePath, targetPath)
 }
 
+func (h *fs) CopyContents(sourcePath string, targetPath string) error {
```
mfojtik: need godoc :-)

mfojtik:
@csrwng thanks for fixing this! good catch

soltysh:
@csrwng I owe with this one :+1:

GrahamDumpleton:
I will mention it here before I do some more testing and creating a separate issue if need be, but should the copy carry across empty directories into the image?

I don't know what it was doing before and whether this change has caused a regression, but if I have an empty directory in a local directory, it never makes it to the image.

It may well have been an existing problem which was never noticed because as you can't really have an empty directory in a git repo and would need to have a dummy file such as ``.gitkeep`` in it. So is particular to working against a local directory.

soltysh:
I guess we never tested that solution, so this will be ever existing problem I think we won't be willing to fix ;) Still for these kind of topics just open an issue and we can discuss options in there.

PR #349, opened 2015-11-12
Add additional fields for author and committer to git.SourceInfo

csrwng:
Separates the author info and adds committer info so it can be used by Origin code.


csrwng:
@bparees ptal

bparees:
lgtm.
[merge]

PR #350, opened 2015-11-13
don't prepend file:// to url path with MungeNoProto

gabemontero:
@bparees the one small piece of clean up from the origin testing of the clone spec / url validation pull ; there is a short term work around to the caller of MungeNoProtocolURL on the origin side (pkg/generate/git/git.go) (no callers on the source-to-image side).  The origin side change can tolerate MungeNoProtocolURL either prefixing or not prefixing the path with "file://".

We can godeps merge this change and remove the workaround in origin at our convenience.

PTAL when you get the chance - thx

### DIFF
--- a/pkg/scm/git/git.go
+++ b/pkg/scm/git/git.go
@@ -236,7 +236,7 @@ func ParseFile(source string) (details *FileProtoDetails, mods *URLMods) {
 		}
 		mods = &URLMods{
 			Scheme: "file",
-			Path:   "file://" + makePathAbsolute(strings.TrimPrefix(source, "file://")),
+			Path:   makePathAbsolute(strings.TrimPrefix(source, "file://")),
 		}
 		return
 	}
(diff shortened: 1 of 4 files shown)


pkg/scm/git/git.go
```diff
@@ -48,6 +48,8 @@ type URLMods struct {
 	Path   string
 	// Corresponds to Fragment in the URL struct, but we use Ref since URL Fragments are used and git commit refs
 	Ref string
+	// for use by scm.go DownloaderForSource for updating source input parameter into the s2i command
+	NewSource string
```
bparees: since this appears to only be set/used when the original path was a local filesystem path, i think the name needs work.  (as does the comment).

alternatively change the instantiation of URLMods such that this is always populated.

pkg/scm/git/git_test.go
```diff
@@ -115,6 +116,127 @@ func TestValidCloneSpecRemoteOnly(t *testing.T) {
 	}
 }
 
+func TestMungeNoProtocolURL(t *testing.T) {
```
bparees: should test the value of the new field you're adding to URLMods.

gabemontero: yep - testing of the new URLMods field - good catch.  As far as where to test, the TestMungeNoProtocolURL already confirms that with the no protocol file URL, the path is not prepended with "file://"

However, I think I should add a test for ParseFile in the git dir, and tweak the TestDownloaderForSource test to ensure the new field (and its resulting mods) are correct.

gabemontero:
updates pushed

pkg/scm/git/git_test.go
bparees: even better.

bparees:
lgtm [merge]

PR #353, opened 2015-11-16
Add submodule commands

paralin:
Adds submodule update and init commands. See issue in origin:

https://github.com/openshift/origin/issues/5910

### DIFF
--- a/pkg/scm/git/clone.go
+++ b/pkg/scm/git/clone.go
@@ -20,27 +20,41 @@ func (c *Clone) Download(config *api.Config) (*api.SourceInfo, error) {
 	targetSourceDir := filepath.Join(config.WorkingDir, api.Source)
 	config.WorkingSourceDir = targetSourceDir
 	var info *api.SourceInfo
-	cloneConfig := api.CloneConfig{Quiet: true, Recursive: !config.DisableRecursive}
+	hasRef := len(config.Ref) > 0
+	hasSubmodules := !config.DisableRecursive
+	cloneConfig := api.CloneConfig{Quiet: true, Recursive: hasSubmodules && !hasRef}
 
 	if c.ValidCloneSpec(config.Source) {
 		if len(config.ContextDir) > 0 {
 			targetSourceDir = filepath.Join(config.WorkingDir, api.ContextTmp)
 		}
-		if cloneConfig.Recursive {
+
+		// If we have a specific checkout ref, use submodule update instead of recursive
+		// Otherwise the versions will be incorrect.
+		if hasRef && hasSubmodules {
+			glog.V(2).Infof("Cloning sources (deferring submodule init) into %q", targetSourceDir)
+		} else if cloneConfig.Recursive {
 			glog.V(2).Infof("Cloning sources and all GIT submodules into %q", targetSourceDir)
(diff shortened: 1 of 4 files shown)


bparees:
can you want to update the s2i clone path to use this new functionality so it works here too?

paralin:
@bparees I've done the change, have a look please. I'm not sure if it's the best way to do it. But it should work.

paralin:
@bparees Test is failing because the last command is now the submodule update instead of the checkout. I can fix this a few different ways. I think I'll add another test for the submodule update.

paralin:
@bparees I've updated the tests to not use Recursive for now. It shouldn't change anything in the context of the test but it might be wise to add more submodule-related tests for coverage later on.

pkg/scm/git/clone.go
```diff
@@ -20,27 +20,41 @@ func (c *Clone) Download(config *api.Config) (*api.SourceInfo, error) {
 	targetSourceDir := filepath.Join(config.WorkingDir, api.Source)
 	config.WorkingSourceDir = targetSourceDir
 	var info *api.SourceInfo
-	cloneConfig := api.CloneConfig{Quiet: true, Recursive: !config.DisableRecursive}
+	hasRef := len(config.Ref) > 0
+	hasSubmodules := !config.DisableRecursive
+	cloneConfig := api.CloneConfig{Quiet: true, Recursive: !config.DisableRecursive && !hasRef}
```
liggitt: `Recursive: hasSubmodules && !hasRef` might be easier to read

pkg/scm/git/clone_test.go
```diff
@@ -16,9 +16,10 @@ func TestCloneWithContext(t *testing.T) {
 	c := &Clone{gh, fs}
 
 	fakeConfig := &api.Config{
-		Source:     "https://foo/bar.git",
-		ContextDir: "subdir",
-		Ref:        "ref1",
+		Source:           "https://foo/bar.git",
+		ContextDir:       "subdir",
+		Ref:              "ref1",
+		DisableRecursive: true,
```
liggitt: why did this test need to change?

paralin: @liggitt Please read what I write on the pull request :)

I'll quote again...

```
@bparees Test is failing because the last command is now the submodule update instead of the checkout. I can fix this a few different ways. I think I'll add another test for the submodule update.
```

As with recursive the last command (the only command actually checked in the test) is `submodule update --init` etc, the test fails. I disable recursive so that the last command is the checkout, which is what the test is checking for. It's checking the same thing. Just not the submodule checkout, which wasn't tested before anyway.

pkg/scm/git/clone.go
paralin: Good point

bparees:
lgtm, would like @mfojtik to take a look also.

paralin:
made the 1 minor syntastic change suggested by @liggitt, otherwise commit is the same

mfojtik:
[test]

lgtm as soon as the test pass

mfojtik:
Thanks @paralin !

mfojtik:
[merge]

PR #354, opened 2015-11-18
Commit container with original builder user

mfojtik:
Fixes: https://github.com/openshift/source-to-image/issues/81

### DIFF
--- a/hack/test-stirunimage.sh
+++ b/hack/test-stirunimage.sh
@@ -12,7 +12,8 @@ function time_now()
   echo $(date +%s000)
 }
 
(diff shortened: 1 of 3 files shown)


bparees:
lgtm [merge]
(should probably get a test case at some point...)

bparees:
[merge]

bparees:
[merge]

bparees:
@mfojtik looks like you're hitting a timeout that needs to be bumped.

mfojtik:
[merge]

it seems like the run test is flaky, I fixed the path where it stores logs so we can inspect them in Jenkins.

PR #356, opened 2015-11-18
use --tags for git-describe

### LINKED ISSUE #355: Wrong version in CLI
When I build the `v1.0.3` tag from source and I run `s2i version` I get: `s2i v1.0.1-99-g08e6631` which is kinda confusing :)

jmtd:
Fixes #355

### DIFF
--- a/hack/common.sh
+++ b/hack/common.sh
@@ -391,7 +391,7 @@ sti::build::sti_version_vars() {
     fi
 
     # Use git describe to find the version based on annotated tags.
-    if [[ -n ${STI_GIT_VERSION-} ]] || STI_GIT_VERSION=$("${git[@]}" describe "${STI_GIT_COMMIT}^{commit}" 2>/dev/null); then
+    if [[ -n ${STI_GIT_VERSION-} ]] || STI_GIT_VERSION=$("${git[@]}" describe --tags "${STI_GIT_COMMIT}^{commit}" 2>/dev/null); then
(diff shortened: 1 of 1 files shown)


mfojtik:
LGTM

@bparees we should fix the release binaries to have correct version

[merge]

bparees:
@mfojtik i'm planning to do a new s2i release soon anyway, hoping to include @gabemontero's upcoming git clone fix first though.

gabemontero:
@mfojtik @bparees I believe that fix is going to be origin only, if time is of the essence

mfojtik:
@bparees I think we should still fix the binaries for 1.0.3 to report correct version :-)

bparees:
@gabemontero it's not particularly urgent.  why is the fix origin only?  is it not broken for s2i?

gabemontero:
@bparees roger that (urgency) - the problematic code path is unique to secret processing in the build config, which does not apply to s2i

mfojtik:
@gabemontero @bparees how is that related to this issue? ;-)

bparees:
@mfojtik it relates to when we're going to do a new release w/ proper versioning and whether it's worth re-doing the existing release or not.

mfojtik:
@bparees I think it worth redoing the previous release. I think the only thing that is broken currently is the binary that we published (it is reporting wrong version). The tag is is properly, so we just need to upload a new binary built by ./hack/build-release.sh with this fix.

Issue #358, opened 2015-11-18
s2i incremental build failures: tries to unpack symlinks which aren't there


jmtd:
I'm hitting a problem with s2i version v1.0.3 (I will attempt to reproduce with HEAD shortly; having trouble building it) and incremental builds using recent JBoss OpenShift EAP images.

The issue is with the second build, upon receiving the tar stream. The logs seem to suggest that it is attempting to unpack a symlink and failing. However, there is no symlink in the tar that is being streamed via `save-artifacts`:

```
...
2015-11-18 14:25:20,549 - root - ERROR - I1118 14:25:20.549707 13108 sti.go:348] Saving build artifacts from image integ-os-eap to path /tmp/sti177660341/upload/artifacts
2015-11-18 14:25:20,611 - root - ERROR - I1118 14:25:20.611884 13108 docker.go:329] Image contains io.openshift.s2i.scripts-url set to 'image:///usr/local/s2i'
2015-11-18 14:25:20,612 - root - ERROR - I1118 14:25:20.611905 13108 docker.go:382] Base directory for STI scripts is '/usr/local/s2i'. Untarring destination is '/tmp'.
2015-11-18 14:25:20,612 - root - ERROR - I1118 14:25:20.611912 13108 docker.go:509] Creating container using config: {Hostname: Domainname: User: Memory:0 MemorySwap:0 CPUShares:0 CPUSet: AttachStdin:false AttachStdout:true AttachStderr:false PortSpecs:[] ExposedPorts:map[] Tty:false OpenStdin:false StdinOnce:false Env:[] Cmd:[/usr/local/s2i/save-artifacts] DNS:[] Image:integ-os-eap:latest Volumes:map[] VolumesFrom: WorkingDir: MacAddress: Entrypoint:[] NetworkDisabled:false SecurityOpts:[] OnBuild:[] Mounts:[] Labels:map[]}
2015-11-18 14:25:21,039 - root - ERROR - I1118 14:25:21.039487 13108 docker.go:526] Starting container
2015-11-18 14:25:21,323 - root - ERROR - I1118 14:25:21.323634 13108 tar.go:233] Creating symbolic link from "/tmp/sti177660341/upload/artifacts/.m2/settings.xml" to ""
2015-11-18 14:25:21,538 - root - ERROR - W1118 14:25:21.538480 13108 sti.go:148] Clean build will be performed because of error saving previous build artifacts
[...]

PR #360, opened 2015-11-19
Fix how we detect symlinks

mfojtik:
Fixes: https://github.com/openshift/source-to-image/issues/358


mfojtik:
@gabemontero @bparees sorry for stealing this from Gabe, but I have @goldmann available in my TZ so it is fast to communicate what needs to be fixed.

mfojtik:
[merge]

goldmann:
LGTM, fixes my issue

PR #370, opened 2015-11-27
changed /usr/local to /usr/libexec/

PI-Victor:
i noticed that the generator scripts still use the old /usr/local for assemble, run, etc. so i've changed it to the new localtion /usr/libexec/s2i as mentioned in this [sti-base image PR](https://github.com/openshift/sti-base/pull/60) also modified it in the tests.

my question is, should it be done this way, hard coded or should we set it according to the env variable $STI_SCRIPTS_PATH set in sti-base in case this path changes again?

Later Edit: nvm, found my answer to the question: https://github.com/openshift/source-to-image/pull/302/files#r41652962


bparees:
lgtm [merge]

PR #371, opened 2015-11-30
Minimal s2i builder image tutorial

PI-Victor:
Minimal reference for s2i builder image.
this is related to https://trello.com/c/d6WQuBS4/537-3-add-minimal-reference-sti-image

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,6 @@ Interested in learning more? Read on!
 
 Want to just get started now? Check out the [instructions](#getting-started).
 
-
 # Philosophy
 
 1. Simplify the process of application source + builder image -> usable image for most use cases (the
@@ -28,6 +27,8 @@ Want to just get started now? Check out the [instructions](#getting-started).
 
 
 # Anatomy of a builder image
+
+See a practical tutorial on how to create a builder image [here](examples/TUTORIAL.md)
 
 Creating builder images is easy. `s2i` looks for you to supply the following scripts to use with an
 image:
--- a/examples/README.md
+++ b/examples/README.md
@@ -1,0 +1,50 @@
+Creating a basic S2I builder image  
+---
+
+### Getting started  
+
+Directory skeleton:  
+
+* `Dockerfile` – standard Dockerfile where we’ll define the base builder image.  
+* `assemble` - responsible for building the application.  
+* `run` - responsible for running the application.  
+* `save-artifacts` - optional script for incremental builds that save built artifacts.  
+* `usage` - optional script responsible for printing the usage of the builder.  
+
+The first step is to create a `Dockerfile` that installs all the necessary tools and libraries that are needed to build and run our application.  
+You can find an example of the Dockerfile [`here`](nginx-app/Dockerfile).   
+
+The next step is to create an `assemble` script that will, for example, build python modules, bundle install our required components or setup application specific configuration based on the logic we define in it. We can specify a way to restore any saved artifacts from the previous image. In [`this`](nginx-app/assemble) example it will only copy our index.html over the default one.  
+
+Now we can create our `run` script that will start the application. You can see an example [`here`](nginx-app/run).
+
+Optionally we can also specify a `save-artifacts` script, which allows a new build to reuses content from a previous version of the application image.
+An example can be found [`here`](nginx-app/save-artifacts).  
+
+We can provide some help to the user on how to use it as a base for an application image via the `usage` script. An example can be found [`here`](nginx-app/usage).
+
+The next step is to create the builder image. In the nginx-app directory issue `docker build -t nginx-centos7 .`  
+This will create a builder image from the current Dockerfile.
+
+Once the builder image is done, the user can issue `s2i usage nginx-centos7` which will print out the help info that was defined in our `usage` script.
+
+The next step is to create `the application image`. We will create this with the content from the source directory `test` from this repo. In this source directory we have only one file for this example, `index.html`.
+
+```
+s2i build test/ nginx-centos7 nginx-app
+---> Building and installing application from source...
+```
+All the logic defined previously in the `assemble` script will now be executed thus compiling your assets or setting up application specific configuration.
+
+Running the application image is as simple as invoking the docker run command:
+`docker run -d -p 8080:8080 nginx-app`
+
+Now you should be able to access a static web page served by our newly created application image on [http://localhost:8080](http://localhost:8080).
+
+If we want to rebuild the application with the saved artifacts, then we can do:
+```
+s2i build --incremental=true test/ nginx-centos7 nginx-app
+---> Restoring build artifacts...
+---> Building and installing application from source...
+```
+This will run the `save-artifacts` script that has the code which will save your artifacts from the previously built application image, and then inject those artifacts into the new image according to the logic you specified in the `assemble` script. 
--- a/examples/nginx-app/Dockerfile
+++ b/examples/nginx-app/Dockerfile
@@ -1,0 +1,36 @@
+# nginx-centos7
+# Here you can use whatever image base is relevant for your application.
+FROM centos:centos7
+
+# Here you can specify the maintainer for the image that you're building
+MAINTAINER Victor Palade <ipalade@redhat.com>
+
+# Export the environment variable that provides information about the application.
+# Replace this with the according version for your application.
+ENV NGINX_VERSION=1.6.3
+
+# Set the labels that are used for Openshift to describe the builder image.
+LABEL io.k8s.description="Nginx Webserver" \
+    io.k8s.display-name="Nginx 1.6.3" \
+    io.openshift.expose-services="8080:http" \
+    io.openshift.tags="builder,webserver,html,nginx" \
+    # this label tells s2i where to find its mandatory scripts
+    # (run, assemble, save-artifacts)
+    io.openshift.s2i.scripts-url="image:///usr/libexec/s2i"
+
+# Install our Nginx package and clean the yum cache so that we don't have any
+# cached files waste space.
+RUN yum install -y epel-release  && \
+    yum install -y --setopt=tsflags=nodocs nginx && \
+    yum clean all
+
+# We will change the default port for nginx (It's required if you plan on
+# running images as non-root user).
+RUN sed -i 's/80/8080/' /etc/nginx/nginx.conf
+
+# Copy the S2I scripts to /usr/libexec/s2i since we set the label that way
+COPY  ["run", "assemble", "save-artifacts", "usage", "/usr/libexec/s2i/"]
+
+# Modify the usage script in your application dir to inform the user how to run
+# this image.
+CMD ["/usr/libexec/s2i/usage"]
--- a/examples/nginx-app/assemble
+++ b/examples/nginx-app/assemble
@@ -1,0 +1,28 @@
+#!/bin/bash -e
+#
+# S2I assemble script for the 'nginx-centos7' image.
+# The 'assemble' script builds your application source so that it is ready to run.
+#
+# For more information refer to the documentation:
+#	https://github.com/openshift/source-to-image/blob/master/docs/builder_image.md
+#
+
+if [[ "$1" == "-h" ]]; then
+	# If the 'nginx-centos7' assemble script is executed with '-h' flag,
+	# print the usage.
+	exec /usr/libexec/s2i/usage
+fi
+
+# Restore artifacts from the previous build (if they exist).
+# We set them here just for show, but you will need to set this up with your logic
+# according to the application directory that you chose.
+if [ "$(ls /tmp/artifacts/ 2>/dev/null)" ]; then
+  echo "---> Restoring build artifacts..."
+  mv /tmp/artifacts/* /usr/share/nginx/html/
+fi
+# Override our default nginx index.html file.
+# This is what we consider in this example `installing the application`
+# here you can go ahead an replace this with the actual building of python modules,
+# bundle install, and anything else you need.
+echo "---> Building and installing application from source..."
+cp -Rf /tmp/src/test/index.html /usr/share/nginx/html/index.html
--- a/examples/nginx-app/run
+++ b/examples/nginx-app/run
@@ -1,0 +1,11 @@
+#!/bin/bash -e
+# The run script executes the server that runs your application.
+#
+# For more information see the documentation:
+#	https://github.com/openshift/source-to-image/blob/master/docs/builder_image.md
+#
+
+# We will turn off daemonizing for the nginx process so that the container
+# doesn't exit after the process runs.
+# We will also specify the default port on which it should listen.
+exec /usr/sbin/nginx -g "daemon off;"
--- a/examples/nginx-app/save-artifacts
+++ b/examples/nginx-app/save-artifacts
@@ -1,0 +1,16 @@
+#!/bin/sh -e
+#
+# S2I save-artifacts script for the 'nginx-centos7' image.
+# The save-artifacts script streams a tar archive to standard output.
+# The archive contains the files and folders you want to re-use in the next build.
+#
+# For more information see the documentation:
+#	https://github.com/openshift/source-to-image/blob/master/docs/builder_image.md
+#
+# Replace this with any logic that you need in order to save all of your built artifacts from
+# your application so they can be reused in a future build to save time.
+touch /tmp/artifact
+cd /tmp
+# the final step of the assemble script is to stream a tar of the artifacts to be saved, to stdout.
+# This tar stream will be received by s2i and used as an input to the build
+tar cf - artifact
--- a/examples/nginx-app/test/index.html
+++ b/examples/nginx-app/test/index.html
@@ -1,0 +1,13 @@
+<!DOCTYPE html>
+<html lang="en">
+<html xmlns="http://www.w3.org/1999/xhtml">
+<head>
+  <meta charset="utf-8">
+  <title>openshift/source-to-image tutorial</title>
+  <meta http-equiv="X-UA-Compatible" content="IE=edge">
+  <meta name="description" content="source-to-image tutorial example page">
+</head>
+  <body>
+    <h1>Hello World</h1>
+  </body>
+</html>
--- a/examples/nginx-app/usage
+++ b/examples/nginx-app/usage
@@ -1,0 +1,13 @@
+#!/bin/bash -e
+cat <<EOF
+This is the nginx-centos7 S2I image:
+To use it, install S2I: https://github.com/openshift/source-to-image
+
+Sample invocation:
+
+s2i build test/ nginx-centos7 nginx-app
+
+You can then run the resulting image via:
+docker run -d -p 8080:8080 nginx-app
+and see the test via http://localhost:8080
+EOF


README.md
```diff
@@ -14,6 +14,7 @@ Interested in learning more? Read on!
 
 Want to just get started now? Check out the [instructions](#getting-started).
 
+See a practical tutorial on how to create a builder image [here](exmaples/TUTORIAL.md)
```
bparees: examples

examples/TUTORIAL.md
```diff
@@ -0,0 +1,253 @@
...
+Directory skeleton:
+
+* `s2i-nginx`
+  * `Dockerfile` – standard Dockerfile where we’ll define the builder image
+  * `Makefile` – helper script for building and testing the builder image
+  * `test`
+    * `run` – test script, testing if the builder image works correctly
+    * `test-app/ ` – directory for your test application
+
+* `.s2i.bin/ `
```
bparees: .s2i/bin

examples/TUTORIAL.md
```diff
@@ -0,0 +1,253 @@
...
+This will create a directory structure for our image and generate the artifacts, injecting the name of the desired image inside them, in our case `nginx-centos7`. This helps streamline the process of building and running the image.
+
+Directory skeleton:
+
+* `s2i-nginx`
+  * `Dockerfile` – standard Dockerfile where we’ll define the builder image
+  * `Makefile` – helper script for building and testing the builder image
+  * `test`
+    * `run` – test script, testing if the builder image works correctly
+    * `test-app/ ` – directory for your test application
```
bparees: get rid of the Makefile and test scripts.  They're just more stuff that someone has to dig through to understand what they're doing.  For this we want barebones "here's the minimal stuff you need to make an s2i image".  They can develop their own build+test processes, but let's not muddy the waters here.

examples/TUTORIAL.md
```diff
@@ -0,0 +1,253 @@
+Creating a basic S2I builder image
+---
+
+### Getting started
+
+The builder image that you specify must contain all required libraries and tools to be able to build and run the intended application.
+This tutorial uses the [openshift/base-centos7](https://github.com/openshift/sti-base) image that has, among other pre-installed libraries and tools, the epel repository added which allows us to install Nginx, since it does come with the standard centos 7 repositories.
```
bparees: i'd rather see an example that starts from just centos, rather than our sti-base.  Again we want barebones and don't want them to have to go dig into that image to see what magic we're doing there.  Everything they need to know should be provided in the files that are in this tutorial, not inherited from elsewhere.

bparees:
including @smarterclayton because this was originally his ask and I want to make sure it fulfills his goals, but i think what's here kinda misses the mark (the good news is i think you've done more than we want, not less)

What I think we really want is just a very basic directory containing:
 - a Dockerfile, as simple as possible.  installing nginx is fine, httpd would be fine too.  Minimal set of recommended Docker Labels that we care about for s2i/openshift.
 - assemble script - looks for saved-artifacts, maybe prints the filenames to show it recognized them, copies static html to appropriate directory to be served by nginx/httpd
  - save-artifacts script - copies some dummy file from the image to illustrate to how save an artifact
 - run script - starts nginx/httpd
 - usage script
 - README

And then lots of comments in those scripts of "replace this section with your logic to......."

Things not to worry about:
- running as an arbitrary user+permission related issues.  that's not s2i's problem, it's openshift's.
- Makefile/build process - just tell them how to run docker build to build the image (in the README)
- testing the image - again just tell them how to use s2i to build a new application image and run it, don't provide a test script.

The thing to keep in mind is this is intended to serve more as a skeleton template for someone to start from, rather than a fully fleshed out example that illustrates a real world working example.  It should be showing them where to plugin and modify things.  

Not sure how much it will help since you don't have an openshift v2 background, but this is an example of the philosophy we're going for here:
https://github.com/openshift/origin-server/tree/master/cartridges/openshift-origin-cartridge-diy
https://docs.openshift.org/origin-m4/oo_cartridge_guide.html#diy

PI-Victor:
@bparees thanks for taking the time to review. i will update the example asap. i took the blogpost more as a reference point and i was thinking that this was less than was needed, but i understand  better now what the goal is.

README.md
soltysh: I'd move this sentence to the end of [Anatomy of a builder image](https://github.com/openshift/source-to-image/#anatomy-of-a-builder-image) section. Not everyone who's interested in s2i will be interested in creating builder images. Besides that section provides good intro to your tutorial.

examples/TUTORIAL.md
soltysh: :+1:

examples/TUTORIAL.md
```diff
@@ -0,0 +1,253 @@
...
+The tool comes with a bootstrap command that creates the directory structure and needed files to get started.
```
soltysh: The s2i comes with...

examples/TUTORIAL.md
```diff
@@ -0,0 +1,253 @@
...
+
+`s2i create <image name> <destination directory>` - [command help](https://github.com/openshift/source-to-image/blob/master/docs/cli.md#s2i-create)
```
soltysh: Don't repeat commands, provide example and explain each parameter from there.

examples/TUTORIAL.md
```diff
@@ -0,0 +1,253 @@
...
+
+`s2i create nginx-centos7 s2i-nginx`
+
+This will create a directory structure for our image and generate the artifacts, injecting the name of the desired image inside them, in our case `nginx-centos7`. This helps streamline the process of building and running the image.
```
soltysh: s/injecting/using

examples/TUTORIAL.md
soltysh: @bparees are you suggesting we should get rid of it from `s2i create` or just from this tutorial? If the latter - :+1: You can mention they are created but ignore then in this tutorial.

examples/TUTORIAL.md
```diff
@@ -0,0 +1,253 @@
...
+  * `Makefile` – helper script for building and testing the builder image
+  * `test`
+    * `run` – test script, testing if the builder image works correctly
+    * `test-app/ ` – directory for your test application
+
+* `.s2i/bin/ `
+  * `assemble` - responsible for building the application.
+  * `run` - responsible for running the application.
+  * `save-artifacts` - optional script responsible for incremental builds.
+  * `usage` - optional script responsible for printing the usage of the builder.
```
soltysh: Provide links where user can find more info. This document https://github.com/openshift/source-to-image/blob/master/docs/builder_image.md contains section per each script, point to it.

examples/TUTORIAL.md
```diff
@@ -0,0 +1,253 @@
...
+
+We are going to start by modifying the `Dockerfile`.
+
+```dockerfile
+# nginx-centos7
+# We are specifying the base for our builder image. Keep in mind that openshift's sti base image has pre-installed tools and packages to streamline the building of the application.
```
soltysh: s/sti/2i

examples/TUTORIAL.md
```diff
@@ -0,0 +1,253 @@
...
+
+        error_page 404 /404.html;
+            location = /40x.html {
+        }
+
+        error_page 500 502 503 504 /50x.html;
+            location = /50x.html {
+        }
+    }
+}
```
soltysh: That's very lengthy. I'd rather seeing short listings, easily graspable, have you seen my article in the topic https://blog.openshift.com/create-s2i-builder-image/ ? Why not using lighttpd here instead of nginx, which is simpler to setup?

examples/TUTORIAL.md
soltysh: I'd also omit incremental builds, which means leaving out `save-artifacts` as well. It's not needed in the simplest use-case, we're targeting.

examples/TUTORIAL.md
```diff
@@ -0,0 +1,253 @@
...
+ ---> ceb6f4c86b5b
+Removing intermediate container 8422927dfd36
+Step 9 : EXPOSE 8080
+ ---> Running in f10b4055dbcd
+ ---> 2b234ffa6558
+Removing intermediate container f10b4055dbcd
+Step 10 : CMD usage
+ ---> Running in dd8a407c9441
+ ---> f81e42101d9c
+Removing intermediate container dd8a407c9441
```
soltysh: Shorten the listing as much as possible, it's not necessary to have it here. You can do:
```
[user@source-to-image s2i-nginx] $ make
docker build -t nginx-centos7 .
...
Successfully built f81e42101d9c
```

examples/TUTORIAL.md
```diff
@@ -0,0 +1,253 @@
...
+Successfully built f81e42101d9c
+```
+
+`s2i build <source> <image> [<tag>] [flags]`
+
+Now we build our image, specifying our source directory, which is `test/app-test`, the base source image which is `nginx-centos7` resulted from the Dockerfile build and then we can specify the name of the resulted image `s2i-nginx`.
```
soltysh: You've just built the image, now you're testing it. You need clear distinction between builder image and image the user will build with s2i.

examples/TUTORIAL.md
```diff
@@ -0,0 +1,253 @@
...
+#
+
+# We will turn off daemonizing for the nginx process so that the container
+# doesn't exit after the process runs.
+# We will also specify our configuration that we copied in the builder image.
+exec /usr/sbin/nginx -c /opt/app-root/etc/nginx.conf -g "daemon off;"
+```
+Next, we continue by creating our base builder image with make.
+
+In the current project directory (`source-to-image/examples/s2i-nginx`) just issue the `make` command. Make creates our base builder image from the instructions in our `Dockerfile`. After it finishes we can use our builder image to test out the webserver configuration.
```
soltysh: Assume a user is following you step by step, which means you should not link to examples directory, but rather provide them with commands he can easily follow on their own. Again have a look at my article. I'd point a user to it at the beginning or the end of the article, there's a accompanying repo one might be interested in.

README.md
bparees: this example is specifically for people who want to create builder images.

examples/TUTORIAL.md
bparees: > are you suggesting we should get rid of it from s2i create or just from this tutorial? If the latter 

just getting rid of it from this tutorial.

> I'd also omit incremental builds, which means leaving out save-artifacts

i debated that, but i think providing one still makes sense.  again the idea here is not "tutorial" it's "skeleton image that they can fill in their own logic" so we should give them all the skeleton bits they need.  If they choose not to add any logic to the save-artifacts script, it won't hurt anything, but it'll be there if they want to.

soltysh: sgtm.

PI-Victor:
@bparees ping! i've updated the tutorial.

examples/README.md
```diff
@@ -0,0 +1,159 @@
...
+#!/bin/bash -e
+#
+# S2I assemble script for the 'nginx-centos7' image.
+# The 'assemble' script builds your application source ready to run.
+#
+# For more information refer to the documentation:
+#	https://github.com/openshift/source-to-image/blob/master/docs/builder_image.md
+#
+
+if [ "$1" = "-h" ]; then
```
mfojtik: ==

examples/README.md
```diff
@@ -0,0 +1,159 @@
...
+	# If the 'nginx-centos7' assemble script is executed with '-h' flag,
+	# print the usage.
+	exec /usr/libexec/s2i/usage
+fi
+
+# Restore artifacts from the previous build (if they exist).
+# We set them here just for show, but you will need to set this up with your logic
+# according to the application directory that you chose.
+if [ "$(ls /tmp/artifacts/ 2>/dev/null)" ]; then
+  echo "---> Restoring build artifacts"
```
mfojtik: append ` ...` at the end (every time you reporting progress, append these)

examples/s2i-nginx/Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+    io.openshift.expose-services="8080:http" \
+    io.openshift.tags="builder,webserver,html,nginx" \
+    io.openshift.s2i.scripts-url="image:///usr/libexec/s2i"
+
+# Install our Nginx package and clean the yum cache so that we don't have any cached files waste space.
+RUN yum install -y epel-release  && \
+    yum install -y --setopt=tsflags=nodocs nginx && \
+    yum clean all
+
+# We will change the default port for nginx (This step is not required, but it's good to run services in containers on non-standard ports)
```
mfojtik: maybe worth explaining that when you set this to "80" you won't be able to run this image other than "root"

mfojtik: also does this image run as `root` by default? it should set random user :-)

examples/s2i-nginx/assemble
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash -e
+#
+# S2I assemble script for the 'nginx-centos7' image.
+# The 'assemble' script builds your application source ready to run.
+#
+# For more information refer to the documentation:
+#	https://github.com/openshift/source-to-image/blob/master/docs/builder_image.md
+#
+
+if [ "$1" = "-h" ]; then
```
mfojtik:
```
if [[ "$1" == "-h" ]]
```

examples/s2i-nginx/Dockerfile
PI-Victor: see @bparees's comment regarding that.

examples/s2i-nginx/assemble
PI-Victor: i understand, and i will change them. however i kept the templates that are created by `s2i create`. so those need to be changed. i will probably submit a pull request with your suggestion a bit later on :)

examples/README.md
```diff
@@ -0,0 +1,159 @@
...
+# We will change the default port for nginx (It's required if you plan on running images as non-root user).
+RUN sed -i 's/80/8080/' /etc/nginx/nginx.conf
+
+# Copy the S2I scripts to /usr/libexec/s2i since we set the label that way
+COPY  ["run", "assemble", "save-artifacts", "usage", "/usr/libexec/s2i/"]
+
+# Modify the usage script in your application dir to inform the user how to run this image.
+CMD ["/usr/libexec/s2i/usage"]
+```
+In our assemble script we define other actions that need to be ran when we want to install our application, this is where we build python modules or bundle install our required components, however in this example we leave it as it is. It will only override the default index.html file. It will also copy over any previously saved artifacts.
```
bparees: s/ran/run/

examples/README.md
```diff
@@ -0,0 +1,159 @@
...
+# Modify the usage script in your application dir to inform the user how to run this image.
+CMD ["/usr/libexec/s2i/usage"]
+```
+In our assemble script we define other actions that need to be ran when we want to install our application, this is where we build python modules or bundle install our required components, however in this example we leave it as it is. It will only override the default index.html file. It will also copy over any previously saved artifacts.  
+
+```
+#!/bin/bash -e
+#
+# S2I assemble script for the 'nginx-centos7' image.
+# The 'assemble' script builds your application source ready to run.
```
bparees: "The assemble script builds your application source so it is ready to run"

examples/README.md
```diff
@@ -0,0 +1,159 @@
...
+# We set them here just for show, but you will need to set this up with your logic
+# according to the application directory that you chose.
+if [ "$(ls /tmp/artifacts/ 2>/dev/null)" ]; then
+  echo "---> Restoring build artifacts..."
+  mv /tmp/artifacts/* /usr/share/nginx/html/
+fi
+
+# Override our default nginx index.html file.
+# This is what we consider in this example `installing the application`
+# here you can go ahead an replace this with the actual building of python modules,
```
bparees: s/an/and/

examples/README.md
```diff
@@ -0,0 +1,159 @@
...
+# bundle install, and anything else you need.
+echo "---> Installing application source..."
+cp -Rf /tmp/src/test/index.html /usr/share/nginx/html/index.html
+```
+What we need to do is tell the run script what process to run on start.
```
bparees: "We also need to create a run script that will start the application"

examples/README.md
```diff
@@ -0,0 +1,159 @@
...
+#!/bin/sh -e
+#
+# S2I save-artifacts script for the 'nginx-centos7' image.
+# The save-artifacts script streams a tar archive to standard output.
+# The archive contains the files and folders you want to re-use in the next build.
+#
+# For more information see the documentation:
+#	https://github.com/openshift/source-to-image/blob/master/docs/builder_image.md
+#
+# Replace this with any logic that you need in order to save all of your built artifacts from your application.
```
bparees: "Replace this with any logic that you need in order to save all of your built artifacts from your application so they can be reused in a future build to save time."

examples/README.md
```diff
@@ -0,0 +1,159 @@
...
+touch /tmp/artifact
+cd /tmp
+tar cf - artifact
```
bparees: put a comment above this.
"# the final step of the assemble script is to stream a tar of the artifacts to be saved, to stdout.  This tar stream will be received by s2i and used as an input to the build"

examples/README.md
```diff
@@ -0,0 +1,159 @@
...
+```
+
+We are now ready to create our base image that contains nginx and we will name it `nginx-centos7`.
```
bparees: this is creating the builder image.

examples/README.md
```diff
@@ -0,0 +1,159 @@
...
+
+`docker build -t nginx-centos7 .`
+
+With the base image created we can now create the builder image.
```
bparees: this is creating the application image (using the builder image we constructed plus a sample source repository, to produce an application image)

PI-Victor:
updated, i think i've addressed all issues.

examples/README.md
```diff
@@ -0,0 +1,159 @@
...
+More [here](https://github.com/openshift/source-to-image/blob/master/docs/builder_image.md).
+
+```
+s2i build test/ nginx-centos7 s2i-nginx
+---> Installing application source
+```
+Printing out the usage format of the image
+
+```
+s2i usage s2i-nginx
```
bparees: this should be s2i usage nginx-centos7, because nginx-centos7 is the s2i builder image.

s2i-nginx is a (badly named) application image.  should call it nginx-app or something instead.

examples/README.md
```diff
@@ -0,0 +1,159 @@
...
+
+Directory skeleton:  
+
+* `Dockerfile` – standard Dockerfile where we’ll define the base builder image.  
+* `assemble` - responsible for building the application.  
+* `run` - responsible for running the application.  
+* `usage` - optional script responsible for printing the usage of the builder.  
+* `save-artifacts` - optional script for incremental builds that save built artifacts.  
+
+We are going to start by adding our logic to the `Dockerfile`.
```
bparees: maintaining this content in two places (here and in the actual file) is going to be a problem.  just link to it.

same for the other scripts.

examples/s2i-nginx/assemble
```diff
@@ -0,0 +1,28 @@
...
+# according to the application directory that you chose.
+if [ "$(ls /tmp/artifacts/ 2>/dev/null)" ]; then
+  echo "---> Restoring build artifacts..."
+  mv /tmp/artifacts/* /usr/share/nginx/html/
+fi
+# Override our default nginx index.html file.
+# This is what we consider in this example `installing the application`
+# here you can go ahead an replace this with the actuall building of python modules,
+# bundle install, and anything else you need.
+echo "---> Installing application source..."
```
bparees: change to this "building and installing application from source..." as that more generically covers what should be done here.

examples/s2i-nginx/save-artifacts
```diff
@@ -0,0 +1,14 @@
...
+#
+# S2I save-artifacts script for the 'nginx-centos7' image.
+# The save-artifacts script streams a tar archive to standard output.
+# The archive contains the files and folders you want to re-use in the next build.
+#
+# For more information see the documentation:
+#	https://github.com/openshift/source-to-image/blob/master/docs/builder_image.md
+#
+# Replace this with any logic that you need in order to save all of your built artifacts
+# from your application.
```
bparees: "from your application that you want to reuse in a future build."

examples/s2i-nginx/test/index.html
```diff
@@ -0,0 +1,177 @@
...
+		<meta charset="utf-8">
+		<title>openshift/source-to-image tutorial</title>
+		<meta http-equiv="X-UA-Compatible" content="IE=edge">
+		<!-- ... -->
+		<meta name="description" content="source-to-image tutorial example page">
+		<meta name="author" content="Palade Victor">
+	</head>
+	<body>
+  <xmp theme="readable" style="display:none;">
+Creating a basic S2I builder image
```
bparees: now you've got this content in multiple places.  just make this page do a basic hello world instead.

examples/s2i-nginx/assemble
PI-Victor: you'd like this changed for the script s2i-template as well?

PI-Victor:
@bparees updated. PTAL.

examples/s2i-nginx/assemble
bparees: it's less of a concern, but sure.

examples/nginx-app/Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+# Export the environment variable that provides information about the application.
+# Replace this with the according version for your application.
+ENV NGINX_VERSION=1.6.3
+
+# Set the labels that are used for Openshift to describe the builder image.
+LABEL io.k8s.description="Nginx Webserver" \
+    io.k8s.display-name="Nginx 1.6.3" \
+    io.openshift.expose-services="8080:http" \
+    io.openshift.tags="builder,webserver,html,nginx" \
+    io.openshift.s2i.scripts-url="image:///usr/libexec/s2i"
```
bparees: comment on this particular label as being the one that tells s2i where to find the assemble/run/save-artifacts scripts.  it's the only one of these labels that is really critical.

examples/nginx-app/assemble
```diff
@@ -0,0 +1,28 @@
...
+# Restore artifacts from the previous build (if they exist).
+# We set them here just for show, but you will need to set this up with your logic
+# according to the application directory that you chose.
+if [ "$(ls /tmp/artifacts/ 2>/dev/null)" ]; then
+  echo "---> Restoring build artifacts..."
+  mv /tmp/artifacts/* /usr/share/nginx/html/
+fi
+# Override our default nginx index.html file.
+# This is what we consider in this example `installing the application`
+# here you can go ahead an replace this with the actuall building of python modules,
```
bparees: s/actuall/actual/

examples/README.md
```diff
@@ -0,0 +1,64 @@
...
+* `usage` - optional script responsible for printing the usage of the builder.  
+* `save-artifacts` - optional script for incremental builds that save built artifacts.  
+
+We are going to start by adding our logic to the [`Dockerfile`](nginx-app/Dockerfile).  
+
+In our [`assemble`](nginx-app/assemble) script we define other actions that need to be run when we want to install our application, this is where we build python modules or bundle install our required components, however in this example we leave it as it is. It will only override the default index.html file. It will also copy over any previously saved artifacts.  
+
+Let's create our [`run`](nginx-app/run) script that will start the application.  
+
+We can also specify a logic for an incremental build in [`save-artifacts`](nginx-app/save-artifacts).
```
bparees: Since you're now linking, this section needs some rewording, like:

"The first piece of the builder image is a Dockerfile that installs anything your image is going to need to build or run application code, you can see an example here[link]"

"Next we need an assemble script which....... the assemble script for this example is found here[link].  This assemble script just overrides the default index.html, but normally you would do ........"

"The final required piece is a run script which starts the application.  The example run script can be seen here[link]"

"You can also optionally define a save-artifacts script which allows a build to reuse content from the last version of the application image that was built.  The sample save-artifacts script can be seen here[link]"

(please don't copy the above suggestions verbatim, they probably need more polish than what i spent time on, but hopefully you get the idea)

examples/README.md
```diff
@@ -0,0 +1,64 @@
...
+
+We are going to start by adding our logic to the [`Dockerfile`](nginx-app/Dockerfile).  
+
+In our [`assemble`](nginx-app/assemble) script we define other actions that need to be run when we want to install our application, this is where we build python modules or bundle install our required components, however in this example we leave it as it is. It will only override the default index.html file. It will also copy over any previously saved artifacts.  
+
+Let's create our [`run`](nginx-app/run) script that will start the application.  
+
+We can also specify a logic for an incremental build in [`save-artifacts`](nginx-app/save-artifacts).    
+
+We are now ready to create our builder image that contains nginx and we will name it `nginx-centos7`.
```
bparees: Need to tell them to run this from within the directory containing the sample pieces referenced above.

examples/README.md
```diff
@@ -0,0 +1,64 @@
...
+`docker build -t nginx-centos7 .`
+
+With the builder image created we can now create the application image (the builder image plus a sample source repository).
+More [here](https://github.com/openshift/source-to-image/blob/master/docs/builder_image.md).
+
+```
+s2i build test/ nginx-centos7 nginx-app
+---> Building and installing application from source...
+```
+We can also print out the [`usage`](nginx-app/usage) format of the image.
```
bparees: you should have them run this against the nginx-centos7 image, not the nginx-app image.

put this step before the "s2i build" invocation.  that is, the flow should be:

1) docker build the builder image
2) illustrate the usage being printed, which also tells you how to invoke s2i build with it
3) have them invoke s2i build with it (which, coincidentally, is what the usage just told them to do)
4) have them run the resulting application

examples/README.md
```diff
@@ -0,0 +1,64 @@
...
+
+s2i build test/ nginx-centos7 nginx-app
+
+You can then run the resulting image via:
+docker run -d -p 8080:8080 nginx-app
+and see the test via http://localhost:8080
+EOF
+```
+
+We can now run the resulted image:
```
bparees: s/resulted/resulting/

bparees:
@PI-Victor this is very close. :)

PI-Victor:
:+1: thanks this must have taken some time. i'll update it asap

PI-Victor:
@bparees Ping! i've updated this yesterday and forgot to tag you.

examples/README.md
```diff
@@ -0,0 +1,61 @@
...
+To use it, install S2I: https://github.com/openshift/source-to-image
+
+Sample invocation:
+
+s2i build test/ nginx-centos7 nginx-app
+
+You can then run the resulting image via:
+docker run -d -p 8080:8080 nginx-app
+and see the test via http://localhost:8080
+```
```
bparees: you can get rid of this usage script content since you linked it above.

examples/README.md
```diff
@@ -0,0 +1,61 @@
...
+
+The next step is to create the builder image. In the nginx-app directory issue `docker build -t nginx-centos7 .`  
+This will create a builder image from the current Dockerfile.
+
```
bparees: you left out the illustrating of running "s2i usage" with the builder image we just created.

PI-Victor: right, you wanted the usage against the builder image since that is the base for our application image.

examples/README.md
```diff
@@ -0,0 +1,49 @@
...
+The next step is to create an `assemble` script that will install our application. This is where we build python modules or bundle install our required components. We can also restore any saved artifacts from the previous image. In our example that we have [`here`](nginx-app/assemble) it will only copy our index.html over the default one.  
+
+Now we can create our `run` script that will start the application. You can see an example  [`here`](nginx-app/run).
+
+
+Optionally we can also specify a `save-artifacts` script, which allows a new build to reuses content from a previous version of the application image.
+An example can be found [`here`](nginx-app/save-artifacts).  
+
+We can also inform the user on how to build an application image based on our builder image via the `usage` script. An example can be found [`here`](nginx-app/usage).
+You can print this output with `s2i usage nginx-centos7`.
```
bparees: this line (26) needs to come after you've built the builder image.  you can't invoke s2i usage until you've built the image.

PI-Victor:
@bparees my eight ball said today is the day it will get merge. :D i've updated it and remade some sentences that i wasn't comfortable with

PI-Victor:
@bparees i was thinking of adding this as the last sentence, right below the --incremental=true output but i'm not very sure about how it sounds.
```
This will run the `save-artifacts` script that has the code which will save your artifacts and then restore it to the new image according to the logic you specified in the `assemble` script. 
```

bparees:
reworded:

This will run the `save-artifacts` script that has the code which will save your artifacts from the previously built application image, and then inject those artifacts into the new image according to the logic you specified in the `assemble` script.

PI-Victor:
done, thanks. i think now it has more info about what --incremental=true does in the background.

bparees:
[merge]

@smarterclayton fyi.

PI-Victor:
Thanks! If anyone wants updates, just tag me and i will do them.

PI-Victor:
hmm, wonder if i can re[test]

bparees:
[merge]

bparees:
[merge]

Issue #363, opened 2015-11-19
filesystem build adds an extra directory


accursoft:
`s2i build {git repository}` adds the files from the repository to `/tmp/src`. `s2i build some/path` adds the files to `tmp/src/path`.

This makes it tricky to write an `assemble` script which works for sources from both git repos and the local file system.

GrahamDumpleton:
Is this related to #348. That is, fixed by it, or caused by it.

mfojtik:
@accursoft can you please confirm that this issue was fixed?

accursoft:
It's not fixed, and I'm not clear which changeset was supposed to have fixed it.

PR #376, opened 2015-12-15
Fix copying the local directory

mfojtik:
Fixes: https://github.com/openshift/source-to-image/issues/363

As a bonus, this should also work on Windows is it does not rely on "cp" binary.

@csrwng @bparees PTAL

### DIFF
--- a/pkg/util/fs.go
+++ b/pkg/util/fs.go
@@ -5,8 +5,8 @@ import (
 	"io"
 	"io/ioutil"
 	"os"
-	"path/filepath"
-	"strings"
+	"path"
+	"runtime"
 
 	"github.com/golang/glog"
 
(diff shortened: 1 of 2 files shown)


mfojtik:
[test]

pkg/util/fs.go
```diff
@@ -77,39 +75,66 @@ func (h *fs) Exists(file string) bool {
...
+		return err
+	}
+	sourceinfo, err := os.Stat(source)
+	if err != nil {
+		return err
 	}
-	// TODO: Use the appropriate command for Windows
-	glog.V(5).Infof("cp -a %s %s", sourcePath, targetPath)
-	return h.runner.Run("cp", "-a", sourcePath, targetPath)
+	return os.Chmod(dest, sourceinfo.Mode())
```
csrwng: You can make the chmod conditional on os != "windows" for it to truly work on windows

pkg/util/fs.go
```diff
@@ -77,39 +75,66 @@ func (h *fs) Exists(file string) bool {
...
-	info, err := os.Stat(targetPath)
-
-	if err != nil || (info != nil && !info.IsDir()) {
-		err = os.Mkdir(targetPath, 0700)
-		if err != nil {
-			return err
-		}
+	defer sourcefile.Close()
+	glog.V(5).Infof("%q -> %q", source, dest)
+	destfile, err := os.Create(dest)
```
csrwng: If we never expect the destination to be a directory, we should change the comment on line 79.

pkg/util/fs.go
mfojtik: good catch, will do

pkg/util/fs.go
mfojtik: i reworked this a little bit, now the CopyContents use Copy function to avoid duplication. Also Copy acts more less as "cp -a" now.

pkg/util/fs.go
```diff
@@ -77,39 +76,75 @@ func (h *fs) Exists(file string) bool {
...
 
-// Copy copies a set of files from sourcePath to targetPath
-func (h *fs) Copy(sourcePath string, targetPath string) error {
-	if _, err := os.Stat(sourcePath); err != nil {
+// Copy copies the source file to a destination directory.
+// If the destination is a filename, the file will be replaced.
+// If the source is directory, the directory itself and also the content of the
+// source directory will be copied into a target directory.
+func (h *fs) Copy(source string, dest string) (err error) {
+	sourcefile, err := os.Open(source)
```
csrwng: wouldn't this fail if it's a directory?

mfojtik: according to my test it will not fail

csrwng: why would we want to open it though (in case of directory)?

pkg/util/fs.go
```diff
@@ -77,39 +76,75 @@ func (h *fs) Exists(file string) bool {
...
 	}
-	// TODO: Use the appropriate command for Windows
-	glog.V(5).Infof("cp -a %s %s", sourcePath, targetPath)
-	return h.runner.Run("cp", "-a", sourcePath, targetPath)
+	return os.Chmod(dest, sourceinfo.Mode())
 }
 
-func (h *fs) CopyContents(sourcePath string, targetPath string) error {
-	info, err := os.Stat(sourcePath)
+// CopyContents copies the files recursivelt from the source directory to
```
csrwng: s/recursivelt/recursively/

pkg/util/fs.go
mfojtik: a check if the directory exists? (it will throw an error when it does not)

pkg/util/fs.go
```diff
@@ -77,39 +76,75 @@ func (h *fs) Exists(file string) bool {
...
+	if sourceinfo.IsDir() {
+		glog.V(5).Infof("D %q -> %q", source, dest)
+		err = os.MkdirAll(dest, sourceinfo.Mode())
 		if err != nil {
 			return err
 		}
+		return h.CopyContents(source, dest)
+	}
+
+	destfile, err := os.Create(dest)
```
csrwng: what happens if src is a file and dest is a directory?

pkg/util/fs.go
```diff
@@ -77,39 +76,75 @@ func (h *fs) Exists(file string) bool {
...
+	}
+	defer sourcefile.Close()
 
-	info, err := os.Stat(targetPath)
-
-	if err != nil || (info != nil && !info.IsDir()) {
-		err = os.Mkdir(targetPath, 0700)
+	if sourceinfo.IsDir() {
+		glog.V(5).Infof("D %q -> %q", source, dest)
+		err = os.MkdirAll(dest, sourceinfo.Mode())
```
csrwng: this seems redundant, since you create the dirs in CopyContents

pkg/util/fs.go
mfojtik: yeah I just hit that problem with integration test... it will not work :-/ you have to specify the target file as well. The CopyContents is doing that, the Copy() does not. However, we use that only in onbuild right now and I already fixed that (I specify the dest as full path to a file)

pkg/util/fs.go
```diff
@@ -77,39 +76,75 @@ func (h *fs) Exists(file string) bool {
...
 	}
-	if !strings.HasSuffix(sourcePath, string(filepath.Separator)) {
-		sourcePath += string(filepath.Separator)
+	objects, err := directory.Readdir(-1)
+	if err != nil {
+		return err
+	}
+	for _, obj := range objects {
+		source := path.Join(src, "/", obj.Name())
+		destination := path.Join(dest, "/", obj.Name())
```
PI-Victor: what about using os.PathSeparator ?

mfojtik: @PI-Victor path.Join always expand "/" as os.PathSeparator

csrwng: filepath.Join uses os.PathSeparator ... and actually the "/" in the second arg is not needed

mfojtik: @PI-Victor err filepath.Join does that, here we always want to have "/" as this run in docker container which is always unix

pkg/util/fs.go
```diff
@@ -77,39 +77,77 @@ func (h *fs) Exists(file string) bool {
...
 
-// Copy copies a set of files from sourcePath to targetPath
-func (h *fs) Copy(sourcePath string, targetPath string) error {
-	if _, err := os.Stat(sourcePath); err != nil {
+// Copy copies the source file to a destination directory.
+// If the destination is a filename, the file will be replaced.
+// If the source is directory, the directory itself and also the content of the
+// source directory will be copied into a target directory.
+// If the destination is a directory an error will be returned as the
+// destination has to be full path to a file.
```
bparees: the previous two statements are contradictory.  either the content will be copied into the target directory, or the target cannot be a directory, which is it?

and what is the behavior when the target does not exist?

bparees: and if the source is a directory, the directory itself does not appear to be copied (the code calls CopyContents, which explicitly does NOT copy the source directory, just the content)

mfojtik: @bparees 

`Copy()` copies file or a directory into a destination. If the source is file, then the destination has to be file as well. If the source is a directory, then we use CopyContents().

mfojtik: @bparees ok, I rewrote the godoc for both functions to be more clear on what is going on :-)

pkg/util/fs.go
```diff
@@ -77,39 +77,78 @@ func (h *fs) Exists(file string) bool {
...
+	return os.Chmod(dest, sourceinfo.Mode())
 }
 
-func (h *fs) CopyContents(sourcePath string, targetPath string) error {
-	info, err := os.Stat(sourcePath)
+// CopyContents copies the content of the source directory to a destination
+// directory.
+// If the destination directory does not exists, it will be created.
+// The source directory itself will not be copied, only it's content. If you
+// want this behavior, the destrination must include the source directory name.
```
bparees: destination.

bparees:
one nit and then lgtm.

mfojtik:
fixed, thanks!

[merge]

PR #377, opened 2015-12-17
use updated wildfly image in tests

bparees:


bparees:
attempt to fix flake being seen in https://github.com/openshift/source-to-image/pull/371

bparees:
[merge]

PR #378, opened 2015-12-18
Update docs: local git changes have to be commited

jhadvig:
I have been approach several times with question: "why the s2i isn't building the latest changes in my repository?" and the answer usually was that "you need to commit all your changes", so I figured out it would be wise to have this info in our docs, i case anyone would read them :P


bparees:
thanks. [merge]

PR #380, opened 2016-01-05
Support injecting files into a container that runs assemble script

mfojtik:
Trello: https://trello.com/c/m5bPxwh4/643-8-allow-to-consume-secrets-in-builds

How this works?

```console
$ echo supersecretstuff > /tmp/foo/secret
$ s2i build -p never --loglevel=5 --inject /tmp/foo:/tmp/bar \
  file://${GOPATH}/src/github.com/openshift/sti-ruby/2.2/test/rack-test-app \
  centos/ruby-22-centos7 testapp
```

What will happen is that the content of `/tmp/foo` folder will get injected (copied) into the container where the `assemble` script runs into the `/tmp/bar` directory. Assemble script can then read the "secret" file from there. The `UploadToContainer` approach allows is to copy any file into any arbitrary location in the container, which is great, because you can replace any file in container (or write to any directory)...

The drawback of this solution is that the file that is copied is owned by `root` and thus we can't erase it entirely, but we can instead `truncate` it ;-) I think, personally, that it is enough (we can add shred if we want to be pedant)

### DIFF
--- a/cmd/s2i/main.go
+++ b/cmd/s2i/main.go
@@ -175,6 +175,7 @@ $ s2i build . centos/ruby-22-centos7 hello-world-app
 	buildCmd.Flags().StringVarP(&(cfg.DisplayName), "application-name", "n", "", "Specify the display name for the application (default: output image name)")
 	buildCmd.Flags().StringVarP(&(cfg.Description), "description", "", "", "Specify the description of the application")
 	buildCmd.Flags().VarP(&(cfg.AllowedUIDs), "allowed-uids", "u", "Specify a range of allowed user ids for the builder image")
+	buildCmd.Flags().VarP(&(cfg.Injections), "inject", "i", "Specify a list of directories to inject into the assemble container")
 	buildCmd.Flags().StringVarP(&(oldDestination), "location", "l", "",
 		"DEPRECATED: Specify a destination location for untar operation")
 
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -86,12 +86,36 @@ that image and add them to the tar streamed to the container into `/artifacts`.
 | `-s (--scripts-url)`       | URL of S2I scripts (see [S2I Scripts](https://github.com/openshift/source-to-image/blob/master/docs/builder_image.md#s2i-scripts)) |
 | `--recursive`              | Perform recursive git clone when getting sources using git|
 | `-q (--quiet)`             | Operate quietly, suppressing all non-error output |
+| `--inject`                 | Inject the content of the specified directory into the path in the container that runs the assemble script |
 
 #### Context directory
 
 In the case where your application resides in a directory other than your repository root
 folder, you can specify that directory using the `--context-dir` parameter. The
 specified directory will be used as your application root folder.
+
+#### Injecting directories to build
+
+If you want to inject files that should only be available during the build (ie
+when the assemble script is invoked), you can specify the directories from which
+the files will be copied into the container that runs the assemble script. To do
+so you can invoke S2I as follows:
+
+```console
+$ s2i build --inject /mydir:/container/dir file://source builder-image output-image
+```
+
+In this case the content of the `/mydir` directory will get copied into
+`/container/dir` inside the container which runs the assemble script.
+After the `assemble` script finishes, all files copied will be truncated and thus
+not available in the output image. The files are truncated instead of deleted
+because the user under which we run the container with the assemble script might not
+have permissions to delete files in the destination directory (eg. `/etc/ssl`).
+
+You can also specify multiple directories, for example: `--inject /dir1:/container/dir1 --inject /dir2:container/dir2`.
+
+You can use this feature to provide SSL certificates, private configuration
+files which contains credentials, etc. 
 
 #### Callback URL
 
--- a/pkg/api/types.go
+++ b/pkg/api/types.go
@@ -2,6 +2,8 @@ package api
 
 import (
 	"fmt"
+	"path/filepath"
+	"strings"
 
 	docker "github.com/fsouza/go-dockerclient"
 
@@ -13,6 +15,10 @@ const (
 	DefaultNamespace    = "io.openshift.s2i."
 	KubernetesNamespace = "io.k8s."
 )
+
+// invalidFilenameCharacters contains a list of character we consider malicious
+// when injecting the directories into containers.
+const invalidFilenameCharacters = `\:;*?"<>|%#$!+{}&[],"'` + "`"
 
 const (
 	// PullAlways means that we always attempt to pull the latest image.
@@ -156,9 +162,23 @@ type Config struct {
 
 	// Usage allows for properly shortcircuiting s2i logic when `s2i usage` is invoked
 	Usage bool
+
+	// Injections specifies a list source/destination folders that are injected to
+	// the container that runs assemble.
+	// All files we inject will be truncated after the assemble script finishes.
+	Injections InjectionList
 }
 
-// DockerConfig contains the configuration for a Docker connection
+// InjectPath contains definition of source directory and the injection path.
+type InjectPath struct {
+	SourcePath     string
+	DestinationDir string
+}
+
+// InjectionList contains list of InjectPath.
+type InjectionList []InjectPath
+
+// DockerConfig contains the configuration for a Docker connection.
 type DockerConfig struct {
 	// Endpoint is the docker network endpoint or socket
 	Endpoint string
@@ -310,4 +330,50 @@ func (p *PullPolicy) Set(v string) error {
 		return fmt.Errorf("invalid value %q, valid values are: always, never or if-not-present")
 	}
 	return nil
+}
+
+// IsInvalidFilename verifies if the provided filename contains malicious
+// characters.
+func IsInvalidFilename(name string) bool {
+	return strings.ContainsAny(name, invalidFilenameCharacters)
+}
+
+// Set implements the Set() function of pflags.Value interface.
+// This function parses the string that contains source:destination pair.
+// When the destination is not specified, the source get copied into current
+// working directory in container.
+func (il *InjectionList) Set(value string) error {
+	mount := strings.Split(value, ":")
+	switch len(mount) {
+	case 0:
+		return fmt.Errorf("invalid format, must be source:destination")
+	case 1:
+		mount = append(mount, "")
+		fallthrough
+	case 2:
+		mount[0] = strings.Trim(mount[0], `"'`)
+		mount[1] = strings.Trim(mount[1], `"'`)
+	default:
+		return fmt.Errorf("invalid source:path definition")
+	}
+	s := InjectPath{SourcePath: filepath.Clean(mount[0]), DestinationDir: filepath.Clean(mount[1])}
+	if IsInvalidFilename(s.SourcePath) || IsInvalidFilename(s.DestinationDir) {
+		return fmt.Errorf("invalid characters in filename: %q", value)
+	}
+	*il = append(*il, s)
+	return nil
+}
+
+// String implements the String() function of pflags.Value interface.
+func (il *InjectionList) String() string {
+	result := []string{}
+	for _, i := range *il {
+		result = append(result, strings.Join([]string{i.SourcePath, i.DestinationDir}, ":"))
+	}
+	return strings.Join(result, ",")
+}
+
+// Type implements the Type() function of pflags.Value interface.
+func (il *InjectionList) Type() string {
+	return "string"
 }
--- a/pkg/build/strategies/sti/sti.go
+++ b/pkg/build/strategies/sti/sti.go
@@ -368,7 +368,7 @@ func (b *STI) Save(config *api.Config) (err error) {
 	defer errReader.Close()
 	defer errWriter.Close()
 	glog.V(1).Infof("Saving build artifacts from image %s to path %s", image, artifactTmpDir)
-	extractFunc := func() error {
+	extractFunc := func(string) error {
 		defer outReader.Close()
 		return b.tar.ExtractTarStream(artifactTmpDir, outReader)
 	}
@@ -446,6 +446,44 @@ func (b *STI) Execute(command string, user string, config *api.Config) error {
 		User:            user,
 		PostExec:        b.postExecutor,
 		NetworkMode:     string(config.DockerNetworkMode),
+	}
+
+	// If there are injections specified, override the original assemble script
+	// and wait till all injections are uploaded into the container that runs the
+	// assemble script.
+	if len(config.Injections) > 0 && command == api.Assemble {
+		injectedFiles, err := util.ExpandInjectedFiles(config.Injections)
+		if err != nil {
+			return err
+		}
+		rmScript, err := util.CreateInjectedFilesRemovalScript(injectedFiles)
+		if err != nil {
+			return err
+		}
+		defer os.Remove(rmScript)
+		glog.V(5).Infof("Waiting for injected files to be copied into assemble container...")
+		opts.CommandOverrides = func(cmd string) string {
+			return fmt.Sprintf("while [ ! -f %q ]; do sleep 0.5; done; %s; result=$?; source %[1]s; exit $result",
+				"/tmp/rm-injections", cmd)
+		}
+		originalOnStart := opts.OnStart
+		opts.OnStart = func(containerID string) error {
+			if err != nil {
+				return err
+			}
+			for _, s := range config.Injections {
+				if err := b.docker.UploadToContainer(s.SourcePath, s.DestinationDir, containerID); err != nil {
+					return err
+				}
+			}
+			if err := b.docker.UploadToContainer(rmScript, "/tmp/rm-injections", containerID); err != nil {
+				return err
+			}
+			if originalOnStart != nil {
+				return originalOnStart(containerID)
+			}
+			return nil
+		}
 	}
 
 	if !config.LayeredBuild {
--- a/pkg/docker/docker.go
+++ b/pkg/docker/docker.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"io"
 	"path"
+	"path/filepath"
 	"strings"
 	"sync"
 
@@ -15,6 +16,7 @@ import (
 
 	"github.com/openshift/source-to-image/pkg/api"
 	"github.com/openshift/source-to-image/pkg/errors"
+	"github.com/openshift/source-to-image/pkg/tar"
 )
 
 const (
@@ -63,6 +65,7 @@ type Docker interface {
 	BuildImage(opts BuildImageOptions) error
 	GetImageUser(name string) (string, error)
 	GetLabels(name string) (map[string]string, error)
+	UploadToContainer(srcPath, destPath, name string) error
 	Ping() error
 }
 
@@ -76,6 +79,7 @@ type Client interface {
 	AttachToContainer(opts docker.AttachToContainerOptions) error
 	StartContainer(id string, hostConfig *docker.HostConfig) error
 	WaitContainer(id string) (int, error)
+	UploadToContainer(id string, opts docker.UploadToContainerOptions) error
 	RemoveContainer(opts docker.RemoveContainerOptions) error
 	CommitContainer(opts docker.CommitContainerOptions) (*docker.Image, error)
 	CopyFromContainer(opts docker.CopyFromContainerOptions) error
@@ -100,22 +104,23 @@ type PullResult struct {
 
 // RunContainerOptions are options passed in to the RunContainer method
 type RunContainerOptions struct {
-	Image           string
-	PullImage       bool
-	PullAuth        docker.AuthConfiguration
-	ExternalScripts bool
-	ScriptsURL      string
-	Destination     string
-	Command         string
-	Env             []string
-	Stdin           io.Reader
-	Stdout          io.Writer
-	Stderr          io.Writer
-	OnStart         func() error
-	PostExec        PostExecutor
-	TargetImage     bool
-	NetworkMode     string
-	User            string
+	Image            string
+	PullImage        bool
+	PullAuth         docker.AuthConfiguration
+	ExternalScripts  bool
+	ScriptsURL       string
+	Destination      string
+	Command          string
+	CommandOverrides func(originalCmd string) string
+	Env              []string
+	Stdin            io.Reader
+	Stdout           io.Writer
+	Stderr           io.Writer
+	OnStart          func(containerID string) error
+	PostExec         PostExecutor
+	TargetImage      bool
+	NetworkMode      string
+	User             string
 }
 
 // CommitContainerOptions are options passed in to the CommitContainer method
@@ -155,6 +160,42 @@ func New(config *api.DockerConfig, auth docker.AuthConfiguration) (Docker, error
 		client:   client,
 		pullAuth: auth,
 	}, nil
+}
+
+// UploadToContainer uploads artifacts to the container.
+// If the source is a directory, then all files and sub-folders are copied into
+// the destination (which has to be directory as well).
+// If the source is a single file, then the file copied into destination (which
+// has to be full path to a file inside the container).
+func (d *stiDocker) UploadToContainer(src, dest, name string) error {
+	path := filepath.Dir(dest)
+	f, err := os.Open(src)
+	if err != nil {
+		return err
+	}
+	info, _ := f.Stat()
+	defer f.Close()
+	t := tar.New()
+	r, w := io.Pipe()
+	if info.IsDir() {
+		path = dest
+		go func() {
+			defer w.Close()
+			if err := t.StreamDirAsTar(src, dest, w); err != nil {
+				glog.Errorf("Uploading directory to container failed: %v", err)
+			}
+		}()
+	} else {
+		go func() {
+			defer w.Close()
+			if err := t.StreamFileAsTar(src, filepath.Base(dest), w); err != nil {
+				glog.Errorf("Uploading files to container failed: %v", err)
+			}
+		}()
+	}
+	glog.V(3).Infof("Uploading %q to %q ...", src, path)
+	opts := docker.UploadToContainerOptions{Path: path, InputStream: r}
+	return d.client.UploadToContainer(name, opts)
 }
 
 // IsImageInLocalRegistry determines whether the supplied image is in the local registry.
@@ -405,9 +446,12 @@ func runContainerTar(opts RunContainerOptions, config docker.Config, imageMetada
 	// when calling assemble script with Stdin parameter set (the tar file)
 	// we need to first untar the whole archive and only then call the assemble script
 	if opts.Stdin != nil && (opts.Command == api.Assemble || opts.Command == api.Usage) {
-		cmd = []string{"/bin/sh", "-c", fmt.Sprintf("tar -C %s -xf - && %s",
-			tarDestination, path.Join(commandBaseDir, string(opts.Command)))}
+		cmd = []string{"/bin/sh", "-c", fmt.Sprintf("tar -C %s -xf - && %s", tarDestination, cmd[0])}
+		if opts.CommandOverrides != nil {
+			cmd = []string{"/bin/sh", "-c", opts.CommandOverrides(strings.Join(cmd[2:], " "))}
+		}
 	}
+	glog.V(5).Infof("Running %q command in container ...", strings.Join(cmd, " "))
 	config.Cmd = cmd
 	return config, tarDestination
 }
@@ -551,7 +595,7 @@ func (d *stiDocker) RunContainer(opts RunContainerOptions) (err error) {
 		return err
 	}
 	if opts.OnStart != nil {
-		if err = opts.OnStart(); err != nil {
+		if err = opts.OnStart(container.ID); err != nil {
 			return err
 		}
 	}
--- a/pkg/api/types_test.go
+++ b/pkg/api/types_test.go
@@ -1,0 +1,45 @@
+package api
+
+import "testing"
+
+func TestInjectionListSet(t *testing.T) {
+	table := map[string][]InjectPath{
+		"/test:":             {{SourcePath: "/test", DestinationDir: "."}},
+		"/test:/test":        {{SourcePath: "/test", DestinationDir: "/test"}},
+		"/test/foo:/etc/ssl": {{SourcePath: "/test/foo", DestinationDir: "/etc/ssl"}},
+		":/foo":              {{SourcePath: ".", DestinationDir: "/foo"}},
+		"/foo":               {{SourcePath: "/foo", DestinationDir: "."}},
+		":":                  {{SourcePath: ".", DestinationDir: "."}},
+		"/t est/foo:":        {{SourcePath: "/t est/foo", DestinationDir: "."}},
+		`"/test":"/foo"`:     {{SourcePath: "/test", DestinationDir: "/foo"}},
+		`'/test':"/foo"`:     {{SourcePath: "/test", DestinationDir: "/foo"}},
+		`"/te"st":"/foo"`:    {},
+		"/test/foo:/ss;ss":   {},
+		"/test;foo:/ssss":    {},
+	}
+	for v, expected := range table {
+		got := InjectionList{}
+		err := got.Set(v)
+		if len(expected) == 0 {
+			if err == nil {
+				t.Errorf("Expected error for %q, got %#v", v, got)
+			} else {
+				continue
+			}
+		}
+		if len(got) != len(expected) {
+			t.Errorf("Expected %d injection in the list for %q, got %d", len(expected), v, len(got))
+		}
+		for _, exp := range expected {
+			found := false
+			for _, g := range got {
+				if g.SourcePath == exp.SourcePath && g.DestinationDir == exp.DestinationDir {
+					found = true
+				}
+			}
+			if !found {
+				t.Errorf("Expected %+v injection found in %#v list", exp, got)
+			}
+		}
+	}
+}
--- a/pkg/docker/fake_docker.go
+++ b/pkg/docker/fake_docker.go
@@ -89,7 +89,7 @@ func (f *FakeDocker) RunContainer(opts RunContainerOptions) error {
 		return f.RunContainerError
 	}
 	if opts.OnStart != nil {
-		if err := opts.OnStart(); err != nil {
+		if err := opts.OnStart(""); err != nil {
 			return err
 		}
 	}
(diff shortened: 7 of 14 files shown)


mfojtik:
@csrwng @php-coder @bparees PTAL

mfojtik:
@bparees one architectural problem I found in S2I regarding to this is that the way we run `assemble` is not really flexible... Currently everything has to happen in the Command you pass to container, ideally we should start the assemble container with something like `/bin/sleep infinity`, then upload secrets, docker exec assemble, remove secrets and then commit the container...

mfojtik:
@php-coder k, I think I fixed all nits you found and I hope I answered all other comments... replying to commit comments is pretty hard ;-)

bparees:
@mfojtik i'd suggest you start from the openshift side and determine what the api is going to look like for users wanting to inject secrets, and use that to drive the s2i api, rather than vice versa where you might paint yourself into a corner.

mfojtik:
@bparees I the API will look something like:

```json
"strategy": {
  "type": "Source",
  "sourceStrategy": {
    "from": {
      "kind": "ImageStreamTag",
      "name": "ruby-22-centos7:latest"
    },
    "secrets": [
      { "name": "secret1", "destination": "/etc/ssl/" },
      { "name": "secret2" },
    ],
  }
}
```

The content of "secret1" gets copied into `/etc/ssl`. The content of "secret2" gets copied into application source directory (or contextDir)).
For S2I think will become `SecretMounts` where you specify the Source and Destination (the source will be original mount that is in the builder container).
For Docker, we should mount all secrets into the "context" directory, where your Dockerfile exists, so you can then add them in Dockerfile. That is not secure at all, but we can warn you. Another approach, more complex for docker strategy would be that we provide a web server, where you can "curl" the secrets from in Dockerfile and remove them in one step. But I think for this you want to use something like "vault" or other service you consume from the build, so I don't want to go there.

What do you think? @smarterclayton ?

mfojtik:
@bparees @php-coder I updated this PR to also handle copying the directories (which is the case in OpenShift secret volumes), the syntax is now:

```console
s2i build -p never --loglevel=5 --secrets /tmp/test:/etc/ssl,/tmp/foos:/foo file:///data/src/github.com/openshift/sti-ruby/2.2/test/rack-test-app centos/ruby-22-centos7 testapp
```

In this example `/tmp/test` is a directory in my local machine where I have 3 files. They all get copied into `/etc/ssl` directory inside the container. The `/tmp/foos` is a regular file, which will be copied into `/foo` file inside the container.
All files from `/etc/ssl` which we copy are truncated when the `assemble` script finishes (also the `/foo` file). 

The sample run produces this output, so you have idea what is going on:

```console
<a lot of irrelevant noise here>
I0106 11:55:20.839525 05551 docker.go:455] Running "/bin/sh -c while [ ! -f \"/etc/ssl/3\" ]; do sleep 0.5; done; tar -C /tmp -xf - && /usr/libexec/s2i/assemble; result=$?; truncate -s0 \"/etc/ssl/1\" \"/etc/ssl/2\" \"/etc/ssl/3\" ; exit $result" command in container ...
I0106 11:55:20.839543 05551 docker.go:574] Creating container using config: {Hostname: Domainname: User: Memory:0 MemorySwap:0 MemoryReservation:0 KernelMemory:0 CPUShares:0 CPUSet: AttachStdin:false AttachStdout:true AttachStderr:false PortSpecs:[] ExposedPorts:map[] StopSignal: Tty:false OpenStdin:true StdinOnce:true Env:[] Cmd:[/bin/sh -c while [ ! -f "/etc/ssl/3" ]; do sleep 0.5; done; tar -C /tmp -xf - && /usr/libexec/s2i/assemble; result=$?; truncate -s0 "/etc/ssl/1" "/etc/ssl/2" "/etc/ssl/3" ; exit $result] DNS:[] Image:centos/ruby-22-centos7:latest Volumes:map[] VolumeDriver: VolumesFrom: WorkingDir: MacAddress: Entrypoint:[] NetworkDisabled:false SecurityOpts:[] OnBuild:[] Mounts:[] Labels:map[]}
I0106 11:55:21.121277 05551 docker.go:588] Attaching to container
I0106 11:55:21.124248 05551 docker.go:594] Starting container
I0106 11:55:21.124640 05551 tar.go:255] Adding to tar: /tmp/sti942014619/upload/src/Gemfile.lock as src/Gemfile.lock
I0106 11:55:21.124776 05551 tar.go:255] Adding to tar: /tmp/sti942014619/upload/src/app.rb as src/app.rb
I0106 11:55:21.124887 05551 tar.go:255] Adding to tar: /tmp/sti942014619/upload/src/config.ru as src/config.ru
I0106 11:55:21.381865 05551 docker.go:171] Uploading "/tmp/test" to "/etc/ssl" ...
I0106 11:55:21.382365 05551 docker.go:197] Uploading "/tmp/test" to "/etc/ssl" ...
I0106 11:55:21.383302 05551 fs.go:98] D "/tmp/test" -> "/tmp/s2i-071672126"
I0106 11:55:21.384373 05551 fs.go:111] F "/tmp/test/1" -> "/tmp/s2i-071672126/1"
I0106 11:55:21.384611 05551 fs.go:111] F "/tmp/test/2" -> "/tmp/s2i-071672126/2"
I0106 11:55:21.384835 05551 fs.go:111] F "/tmp/test/3" -> "/tmp/s2i-071672126/3"
I0106 11:55:21.385043 05551 tar.go:188] Adding "/tmp/s2i-071672126" to tar ...
I0106 11:55:21.385214 05551 tar.go:255] Adding to tar: /tmp/s2i-071672126/1 as 1
I0106 11:55:21.385380 05551 tar.go:255] Adding to tar: /tmp/s2i-071672126/2 as 2
I0106 11:55:21.385434 05551 tar.go:255] Adding to tar: /tmp/s2i-071672126/3 as 3
I0106 11:55:21.415338 05551 docker.go:524] Waiting for container
I0106 11:55:21.913698 05551 sti.go:529] ---> Installing application source ...
I0106 11:55:21.916180 05551 sti.go:529] ---> Building your Ruby application from source ...
I0106 11:55:21.916202 05551 sti.go:529] ---> Running 'bundle install --deployment' ...
```

bparees:
> The /tmp/foos is a regular file, which will be copied into /foo file inside the container.

what happens if the target /foo is a directory that exists?  does it copy "foos" into "/foo" or does it fail?  

what if /foo exists as a regular file, is it overwritten or fails?

> All files from /etc/ssl which we copy are truncated when the assemble script finishes (also the /foo file).

are nested directory structures handled?  (not sure secrets can have nested dirs, but...)

mfojtik:
@bparees if the "/foo" already exists in container and it is a dir then docker uploads that file into "/foo/foos" which is not what you want... I'm thinking about dropping the support for copying "single" file and keeping just directory to make it closer to OpenShift. 

The nested directories are handled and copied as well.

php-coder:
```json
    "secrets": [
      { "name": "secret1", "destination": "/etc/ssl/" },
      { "name": "secret2" },
    ],
```
@mfojtik if we rename `destination` to `mountPath` then we could reuse [`SecretSpec` object](https://docs.openshift.org/latest/rest_api/openshift_v1.html#v1-secretspec). Or maybe we could name in the same style, like `copyPath`. (Just one more idea.)

mfojtik:
@php-coder it is not really a mount path but a copy to path... depends on how much we want to make that visible to users. Copy behaves differently than mounts, we adding files to directories that already exists instead of over-mounting them, which might be unexpected when you use `mountPath`.

php-coder:
@mfojtik I agree. "maybe we could name in the same style, like `copyPath`?"

mfojtik:
@php-coder `copyToPath` is more explicit but I'm ok just with copyPath

bparees:
> I'm thinking about dropping the support for copying "single" file and keeping just directory to make it closer to OpenShift. 

so if i specify [ "secret1", "/etc/ssl"]

and "secret1" contains "mysecret" and "yoursecret"

it gets mounted as /var/run/secrets/secret1 (which contains 2 files) right?

what's the final running container going to look like?
/etc/ssl/secret1 (which contains mysecret and yoursecret)?

or
/etc/ssl (which contains mysecret and yoursecret)?

mfojtik:
@bparees the second case, it all files/dirs from /var/run/secrets/secret1 gets copied into /etc/ssl/

bparees:
@mfojtik ok good. :)

mfojtik:
@php-coder the removal is now done by a script we upload to container (it is the last file, so we can use it as a marker). That fixes the issue with long command line.

mfojtik:
[test]

mfojtik:
@php-coder yeah

mfojtik:
@bparees @php-coder I think this is ready for regular review. I added one integration test to prove that the secrets are delivered to assemble container AND truncated afterwards. Also I added bunch unit tests to check the helper methods and parsing of the command line.

One last thing that remains here is the naming, current I use:

* `SecretMounts` in the api/types.go which has a type of `SecretList`. Are you OK with this name? 
* There is no "short" option to specify this on command line, you have to use `--secrets` as the `-s` is already taken. If we want to provide short option, I'm OK with `-k`.
* The single secret definition has a  `Secret`  type which has two fields `Source` and `Destination`. Does it make sense to sync up this with ImageSourcePath here, or you are OK leaving it here and make the API fields compatible instead?

As discussed, only directories are allowed now, so copying a single file will fail (the assemble will fail, so the container will never get comited). I'm fine with that since it makes the logic simpler and prevent us from errors like when destination exists and it is a directory (we can't truncate the secret then).

mfojtik:
and yeah, Travis go 1.4 is broken, but it is broken also for origin, there is an issue for that there, so not related to this PR.

bparees:
So instead of Secrets, how about Inject or Injections?
InjectionMounts, InjectionList, --inject, -i 
?

bparees:
@mfojtik comments made directly on the commits.

mfojtik:
@bparees updated the PR to rename "secret" to "inject" (also all types/tests/docs). Fixed the integration test to verify the content of the file can be picked up by assemble script itself. Also testing the nested directory now.

mfojtik:
@php-coder I did separate PR to bump the go-dockerclient so this PR will be more clean for sake of review :) will rebase as soon as it gets merged.

mfojtik:
@rhcarvalho now this is single commit :)

docs/cli.md
```diff
@@ -93,6 +94,28 @@ In the case where your application resides in a directory other than your reposi
 folder, you can specify that directory using the `--context-dir` parameter. The
 specified directory will be used as your application root folder.
 
+#### Injecting directories to build
+
+In case you want to inject a files that should be available only during the
```
rhcarvalho: s/a files/files/

docs/cli.md
```diff
@@ -93,6 +94,28 @@ In the case where your application resides in a directory other than your reposi
...
+this script. To do so you can invoke S2I as following:
+
+```console
+$ s2i build --inject /mydir:/container/dir file://source builder-image output-image
+```
+
+In this case the content of the `/mydir` directory will get copied into
+`/container/dir` inside the container which runs the assemble script.
+After the `assemble` script finish, all files copied will be truncated and thus
+not available in the output image.
```
rhcarvalho: I think it's well deserved to explain why files are truncated and not removed.

docs/cli.md
```diff
@@ -93,6 +94,28 @@ In the case where your application resides in a directory other than your reposi
...
+
+You can also specify multiple directories which in that case must be delimited
+by `,`. For example: `--inject /dir1:/container/dir1,/dir2/container/dir2,...`.
```
rhcarvalho: should be `--inject /dir1:/container/dir1,/dir2:container/dir2,...` (colon after /dir2).

The problem with this syntax is that your dirs cannot have commas...
I'd prefer being able to pass --inject multiple times, just like `docker run -v` works.

mfojtik: The comma is not allowed character in the directory name (we validate that in `IsSafeFilename()` so you will get an error if you will try to do that. I think the syntax you propose might work already as we use `pflags.Value` here (but I haven't tried that). However, I'm not that much concerned about the edge cases (like dirs with commas here).

pkg/api/types.go
```diff
@@ -156,8 +158,22 @@ type Config struct {
 
 	// Usage allows for properly shortcircuiting s2i logic when `s2i usage` is invoked
 	Usage bool
+
+	// Injections specifies a list source/destination folders that are injected to
+	// the container that runs assemble. All files are truncated after the
+	// assemble script runs, so they wont be present in the output image.
```
rhcarvalho: the files **will** be present, but with size 0, right? That's an important distinction as the file names will be still available and might disclose information to an attacker.

BTW, do you know if after truncation it is possible to "recover" the truncated files by bypassing the file system? That would also be a security concern, as the file contents would still be there, but "hidden".

docs/cli.md
mfojtik: OK it is working that way already :-) (you can specify multiple `--inject`)

pkg/api/types.go
```diff
@@ -156,8 +158,22 @@ type Config struct {
...
+	Injections InjectionList
+}
+
+// InjectPath contains definition of source directory and the injection path
```
rhcarvalho: godoc should end with ".", here and few other places ;)

pkg/api/types.go
mfojtik: @rhcarvalho it is not possible to recover them when the image is comitted. I can add "shred" there before I truncate them, but I think that is an overkill.

pkg/api/types.go
```diff
@@ -311,3 +327,53 @@ func (p *PullPolicy) Set(v string) error {
 	}
 	return nil
 }
+
+// IsSafeFilename checks if the filename provided does not contain any invalid
+// characters
+func IsSafeFilename(name string) bool {
```
rhcarvalho: what's the intent of this?
why blacklisting instead of whitelisting?
why is `•` there, but not any other funny unicode symbol?

mfojtik: It is there to prevent injection of malicious directory names. There are just few characters that can do damage so for that I think blacklisting them makes more sense.

rhcarvalho:
@mfojtik is it still to be considered a "working prototype" as the description says?
If so, what's our take on adding new prototype-like features? Do we add and remove them (if need be) and avoid releases? Would we hide it under something like "ex(perimental)" prefix or some other way?
Or shall we just keep it in a parallel, non-master, branch for the time being?

mfojtik:
@rhcarvalho not a prototype anymore

docs/cli.md
rhcarvalho: :) good, then we can remove a bunch of code for parsing that flag.

pkg/api/types.go
rhcarvalho: How is • one of them?!

docs/cli.md
mfojtik: @rhcarvalho I still like to be able to pass multiple directories if I want instead of duplicating that flag on command line.

pkg/api/types.go
mfojtik: removed.

On Fri, Jan 8, 2016 at 3:08 PM, Rodolfo Carvalho <notifications@github.com>
wrote:

> In pkg/api/types.go
> <https://github.com/openshift/source-to-image/pull/380#discussion_r49190812>
> :
>
> > @@ -311,3 +327,53 @@ func (p *PullPolicy) Set(v string) error {
> >  	}
> >  	return nil
> >  }
> > +
> > +// IsSafeFilename checks if the filename provided does not contain any invalid
> > +// characters
> > +func IsSafeFilename(name string) bool {
>
> How is • one of them?!
>
> —
> <https://github.com/openshift/source-to-image/pull/380/files#r49190812>.
>



-- 

docs/cli.md
rhcarvalho: I think because I watched Dope last night I got this thing in my head: slippery slope...
Adding custom format for this one flag opens an unnecessary precedent to have inconsistent CLI.
I'd much rather leave handling of flags to cobra.

docs/cli.md
```diff
@@ -93,6 +94,30 @@ In the case where your application resides in a directory other than your reposi
...
+
+In this case the content of the `/mydir` directory will get copied into
+`/container/dir` inside the container which runs the assemble script.
+After the `assemble` script finish, all files copied will be truncated and thus
+not available in the output image. The files are truncated instead of deleted
+because the user under which we run the container with assemble script might not
+have permissions to delete files in the destination directory (eg. `/etc/ssl`).
+
+You can also specify multiple directories which in that case must be delimited
+by `,`. For example: `--inject /dir1:/container/dir1,/dir2/container/dir2,...`.
```
rhcarvalho: you missed the typo of `/dir2/container/dir2` -> `/dir2:container/dir2`

docs/cli.md
mfojtik: we have similar approach for defining multiple env vars, but i don't have strong opinion here, so I removed ',' syntax and updated the docs.

pkg/api/types.go
```diff
@@ -311,3 +327,49 @@ func (p *PullPolicy) Set(v string) error {
 	}
 	return nil
 }
+
+// IsSafeFilename checks if the filename provided does not contain any invalid
+// characters.
+func IsSafeFilename(name string) bool {
```
rhcarvalho: I see we negate twice here, once in the return statement, and later in all usages of `IsSafeFilename`.
Perhaps what we want is a `IsInvalidFilename`?
```go
const invalidChars = `\:;*?"<>|%#$!+{}&[],"'` + "`"

func IsInvalidFilename(name string) bool {
	return strings.ContainsAny(name, invalidChars)
}

// ...

	if IsInvalidFilename(s.SourcePath) || IsInvalidFilename(s.DestinationDir) {
 		return fmt.Errorf("invalid characters in file name: %q", value)
 	}
```

docs/cli.md
```diff
@@ -93,6 +94,30 @@ In the case where your application resides in a directory other than your reposi
...
+
+In this case the content of the `/mydir` directory will get copied into
+`/container/dir` inside the container which runs the assemble script.
+After the `assemble` script finish, all files copied will be truncated and thus
+not available in the output image. The files are truncated instead of deleted
+because the user under which we run the container with assemble script might not
+have permissions to delete files in the destination directory (eg. `/etc/ssl`).
+
+You can also specify multiple directories which in that case must be delimited
+by `,`. For example: `--inject /dir1:/container/dir1 --inject /dir2:container/dir2`.
```
rhcarvalho: no more "must be delimited by `,`", right?

docs/cli.md
```diff
@@ -93,6 +94,30 @@ In the case where your application resides in a directory other than your reposi
...
+```console
+$ s2i build --inject /mydir:/container/dir file://source builder-image output-image
+```
+
+In this case the content of the `/mydir` directory will get copied into
+`/container/dir` inside the container which runs the assemble script.
+After the `assemble` script finish, all files copied will be truncated and thus
+not available in the output image. The files are truncated instead of deleted
+because the user under which we run the container with assemble script might not
+have permissions to delete files in the destination directory (eg. `/etc/ssl`).
```
rhcarvalho: In which case you can truncate a file but not remove it? Is it dependent to the permissions of the directory that contains the file?

pkg/api/types.go
mfojtik: @rhcarvalho fixed.

docs/cli.md
mfojtik: @rhcarvalho in order to be able to remove the file from a directory you need to be an owner of that directory or have write access to that directory. However, the file can allow write access to anyone (0666) so we are able to truncate it.

rhcarvalho: we need to test that this will satisfy the use cases we have in mind, e.g. /etc/ssh/ssh_config, ~/.ssh/config cannot be 0666.

pkg/api/types.go
```diff
@@ -14,6 +16,10 @@ const (
 	KubernetesNamespace = "io.k8s."
 )
 
+// InvalidFilenameCharacters contains a list of character we consifed malicious
```
rhcarvalho: s/we consifed/considered

docs/cli.md
mfojtik: @rhcarvalho you can modify the permissions in assemble script if you have that use case.

pkg/api/types.go
```diff
@@ -14,6 +16,10 @@ const (
...
+// when injecting the directories into containers.
+const InvalidFilenameCharacters = `\:;*?"<>|%#$!+{}&[],"'` + "`"
```
rhcarvalho: I think we don't need to export it unless we have a reason to do so

pkg/api/types.go
```diff
@@ -311,3 +331,48 @@ func (p *PullPolicy) Set(v string) error {
...
+// This function parses the string that contains source:destination pair.
+// When the destination is not specified, the source get copied into current
+// working directory in container.
+func (il *InjectionList) Set(value string) error {
+	mount := strings.Split(value, ":")
+	switch len(mount) {
+	case 0:
+		return fmt.Errorf("invalid format, must be source:destination")
+	case 1:
+		mount = append(mount, "")
```
rhcarvalho: why here we don't trim quotes like below?

docs/cli.md
mfojtik: @rhcarvalho also I have zero use cases for that file (.ssh/ssh_config). For things like certs or npmrc or .gemrc, etc this will work just fine.

pkg/api/types.go
mfojtik: if second mount is empty, we add it as empty mount which then expand to ".". No need to strip quotes on empty string.

rhcarvalho: perhaps a `fallthrough` here would do the job

mfojtik: @rhcarvalho not really as mount[1] cannot be set without appending to list.

rhcarvalho: I mean:

```go
case 1:
    mount = append(mount, "")
    fallthrough
case 2:
...
```

You are not stripping quotes from the first argument in the case 1 now.

rhcarvalho: `fallthrough` is a DRY solution, otherwise:

```go
case 1:
    mount[0] = strings.Trim(mount[0], `"'`)
    mount = append(mount, "")
case 2:
...
```

mfojtik: i got what you want now, fixed.

mfojtik:
[test] flake in integration

cmd/s2i/main.go
```diff
@@ -175,6 +175,7 @@ $ s2i build . centos/ruby-22-centos7 hello-world-app
 	buildCmd.Flags().StringVarP(&(cfg.DisplayName), "application-name", "n", "", "Specify the display name for the application (default: output image name)")
 	buildCmd.Flags().StringVarP(&(cfg.Description), "description", "", "", "Specify the description of the application")
 	buildCmd.Flags().VarP(&(cfg.AllowedUIDs), "allowed-uids", "u", "Specify a range of allowed user ids for the builder image")
+	buildCmd.Flags().VarP(&(cfg.Injections), "inject", "i", "Specify a list of directories to inject into assemble container")
```
rhcarvalho: assemble container? Is that how it is called?

docs/cli.md
```diff
@@ -93,6 +94,29 @@ In the case where your application resides in a directory other than your reposi
...
+
+You can also specify multiple directories, for example: `--inject /dir1:/container/dir1 --inject /dir2:container/dir2`.
```
rhcarvalho: 1. Can the directory in the container be relative, or must it be an absolute path?
2. The example uses `/container/dir1` and `container/dir2` (note missing slash).

pkg/build/strategies/sti/sti.go
```diff
@@ -448,6 +448,44 @@ func (b *STI) Execute(command string, user string, config *api.Config) error {
...
+		}
+		rmScript, err := util.CreateInjectedFilesRemovalScript(injectedFiles)
+		if err != nil {
+			return err
+		}
+		defer os.Remove(rmScript)
+		glog.V(5).Infof("Waiting for injected files to be copied into assemble container...")
+		opts.CommandOverrides = func(cmd string) string {
+			return fmt.Sprintf("while [ ! -f %q ]; do sleep 0.5; done; %s; result=$?; source %s; exit $result",
+				"/tmp/rm-injections", cmd, "/tmp/rm-injections")
```
rhcarvalho: Why do we pass "/tmp/rm-injections" twice? can we make this easier to read/understand?

pkg/docker/docker.go
```diff
@@ -100,22 +104,23 @@ type PullResult struct {
...
+	ExternalScripts  bool
+	ScriptsURL       string
+	Destination      string
+	Command          string
+	CommandOverrides func(originalCmd string) string
+	Env              []string
+	Stdin            io.Reader
+	Stdout           io.Writer
+	Stderr           io.Writer
+	OnStart          func(string) error
```
rhcarvalho: Did this change because of the bump on the docker client lib?!
What does it do?

mfojtik: goftm

docs/cli.md
mfojtik: it can be relative or ommited.

pkg/docker/docker.go
rhcarvalho: OnStart took no arguments before, now it takes a string.

php-coder:
>@bparees I the API will look something like:
>```
>"strategy": {
>  "type": "Source",
>  "sourceStrategy": {
>    "from": {
>      "kind": "ImageStreamTag",
>      "name": "ruby-22-centos7:latest"
>    },
>    "secrets": [
>      { "name": "secret1", "destination": "/etc/ssl/" },
>      { "name": "secret2" },
>    ],
>  }
>}
>```

I see that `customStrategy` alredy have possibility to specify secrets: https://docs.openshift.org/latest/dev_guide/builds.html#custom-secrets

@mfojtik Could we re-use it for other strategies?

pkg/docker/docker.go
php-coder: >OnStart took no arguments before, now it takes a string.

... and also `CommandOverrides func(originalCmd string) string` as added.

mfojtik:
@php-coder probably not, we have to compatible with ImageSourcePath

pkg/docker/docker.go
mfojtik: @rhcarvalho that is valid change, I have to pass container id to know where to upload files...

pkg/build/strategies/sti/sti.go
mfojtik: @rhcarvalho because we wait for that file to exists and then we execute it... i can make it a variable, but I'm not big fun of hiding stuff to variables.

docs/cli.md
mfojtik: and yes, it is like that on purpose.

cmd/s2i/main.go
mfojtik: "a container where the assemble runs" is long and breaks the feng-shui of my terminal.

pkg/build/strategies/sti/sti.go
rhcarvalho: the bash script could be in multiple lines, and if you need to reuse the same args in Sprintf you can use %[i]s syntax.

rhcarvalho: What happens if the execution of `/tmp/rm-injections` returns non-zero exit code? We're ignoring it here.

mfojtik: @rhcarvalho i will set `-e` in the script so it will exit the parent with error immediately.

mfojtik: @rhcarvalho fixed.

pkg/docker/docker.go
rhcarvalho: are those functions and the arguments they take documented anywhere?

mfojtik: i don't think so, but that is out of scope of this PR, we should add godoc to all field, but not in this PR

pkg/build/strategies/sti/sti.go
rhcarvalho: does that mean that if `cmd` returns non-zero exit code it will also exit immediately without sourcing/running `/tmp/rm-injections`?

mfojtik: no, if assemble returns non-zero then we still run injection cleanup... if
that returns non-zero we exit immediately. i updated the PR (set +e in the
script)

On Fri, Jan 8, 2016 at 5:54 PM, Rodolfo Carvalho <notifications@github.com>
wrote:

> In pkg/build/strategies/sti/sti.go
> <https://github.com/openshift/source-to-image/pull/380#discussion_r49210030>
> :
>
> > +	// and wait till all injections are uploaded into the container that runs the
> > +	// assemble script.
> > +	if len(config.Injections) > 0 && command == api.Assemble {
> > +		injectedFiles, err := util.ExpandInjectedFiles(config.Injections)
> > +		if err != nil {
> > +			return err
> > +		}
> > +		rmScript, err := util.CreateInjectedFilesRemovalScript(injectedFiles)
> > +		if err != nil {
> > +			return err
> > +		}
> > +		defer os.Remove(rmScript)
> > +		glog.V(5).Infof("Waiting for injected files to be copied into assemble container...")
> > +		opts.CommandOverrides = func(cmd string) string {
> > +			return fmt.Sprintf("while [ ! -f %q ]; do sleep 0.5; done; %s; result=$?; source %s; exit $result",
> > +				"/tmp/rm-injections", cmd, "/tmp/rm-injections")
>
> does that mean that if cmd returns non-zero exit code it will also exit
> immediately without sourcing/running /tmp/rm-injections?
>
> —
> <https://github.com/openshift/source-to-image/pull/380/files#r49210030>.
>



-- 

pkg/docker/docker.go
rhcarvalho: you are adding a new undocument "string" argument in this PR

pkg/build/strategies/sti/sti.go
mfojtik: actually we don't care about cleanup when assemble returns non-zero because
that means the assemble fails and we don't commit container in that case

On Fri, Jan 8, 2016 at 5:56 PM, Michal Fojtik <mi@mifo.sk> wrote:

> no, if assemble returns non-zero then we still run injection cleanup... if
> that returns non-zero we exit immediately. i updated the PR (set +e in the
> script)
>
> On Fri, Jan 8, 2016 at 5:54 PM, Rodolfo Carvalho <notifications@github.com
> > wrote:
>
>> In pkg/build/strategies/sti/sti.go
>> <https://github.com/openshift/source-to-image/pull/380#discussion_r49210030>
>> :
>>
>> > +	// and wait till all injections are uploaded into the container that runs the
>> > +	// assemble script.
>> > +	if len(config.Injections) > 0 && command == api.Assemble {
>> > +		injectedFiles, err := util.ExpandInjectedFiles(config.Injections)
>> > +		if err != nil {
>> > +			return err
>> > +		}
>> > +		rmScript, err := util.CreateInjectedFilesRemovalScript(injectedFiles)
>> > +		if err != nil {
>> > +			return err
>> > +		}
>> > +		defer os.Remove(rmScript)
>> > +		glog.V(5).Infof("Waiting for injected files to be copied into assemble container...")
>> > +		opts.CommandOverrides = func(cmd string) string {
>> > +			return fmt.Sprintf("while [ ! -f %q ]; do sleep 0.5; done; %s; result=$?; source %s; exit $result",
>> > +				"/tmp/rm-injections", cmd, "/tmp/rm-injections")
>>
>> does that mean that if cmd returns non-zero exit code it will also exit
>> immediately without sourcing/running /tmp/rm-injections?
>>
>> —
>> Reply to this email directly or view it on GitHub
>> <https://github.com/openshift/source-to-image/pull/380/files#r49210030>.
>>
>
>
>
> --
> -----------------------------------------------
> Michal Fojtik <mi@mifo.sk <mfojtik@mifo.sk>>
> Red Hat OpenShift, Engineering
>



-- 

pkg/docker/docker.go
rhcarvalho: at least give it a name then, like `CommandOverrides func(**originalCmd** string) string`

mfojtik: I can add name to that parameter if that makes you feel better :)

rhcarvalho: not me, but let's call it "the future you"

cmd/s2i/main.go
bparees: "the assemble container"

docs/cli.md
```diff
@@ -86,6 +86,7 @@ that image and add them to the tar streamed to the container into `/artifacts`.
 | `-s (--scripts-url)`       | URL of S2I scripts (see [S2I Scripts](https://github.com/openshift/source-to-image/blob/master/docs/builder_image.md#s2i-scripts)) |
 | `--recursive`              | Perform recursive git clone when getting sources using git|
 | `-q (--quiet)`             | Operate quietly, suppressing all non-error output |
+| `--inject`                 | Inject the content of specified directory into the path in container that runs assemble script |
```
bparees: in the container

bparees: runs the assemble script

bparees: of the specified directory

docs/cli.md
```diff
@@ -93,6 +94,29 @@ In the case where your application resides in a directory other than your reposi
 folder, you can specify that directory using the `--context-dir` parameter. The
 specified directory will be used as your application root folder.
 
+#### Injecting directories to build
+
+In case you want to inject files that should be available only during the
+build time (iow. when the `assemble` script is invoked), you can specify the
+directories from where the files will be copied into the container that runs
+this script. To do so you can invoke S2I as following:
```
bparees: "If you want to inject files that should only be available during the build (ie when the `assemble` script is invoked), you can specify the directories from which the files will be copied into the container that runs the assemble script.  To do so you can invoke S2I as follows:"

docs/cli.md
```diff
@@ -93,6 +94,29 @@ In the case where your application resides in a directory other than your reposi
...
+
+```console
+$ s2i build --inject /mydir:/container/dir file://source builder-image output-image
+```
+
+In this case the content of the `/mydir` directory will get copied into
+`/container/dir` inside the container which runs the assemble script.
+After the `assemble` script finish, all files copied will be truncated and thus
```
bparees: s/finish/finishes/

docs/cli.md
```diff
@@ -93,6 +94,29 @@ In the case where your application resides in a directory other than your reposi
...
+not available in the output image. The files are truncated instead of deleted
+because the user under which we run the container with assemble script might not
```
bparees: the assemble script

bparees:
grammatical nits and lgtm

mfojtik:
@bparees grammatical nits fixed, thanks!

[merge]

mfojtik:
@php-coder after this merges, we can include this commit as a bump to origin.

PR #382, opened 2016-01-11
Add rebase-to-origin.sh script

mfojtik:
Tired of manual rebase to origin? Try this handy script :-)

This script assumes that the "origin" is cloned into $GOPATH/src/github.com/openshift/origin.
This script will create a new branch, so make sure the "origin" HEAD is the latest master.
After you invoke this script, the branch and the commit will be created with proper commit message, ready to push and create a PR from it.

### DIFF
--- a/hack/rebase-to-origin.sh
+++ b/hack/rebase-to-origin.sh
@@ -1,0 +1,46 @@
+#!/bin/bash -e
+#
+# This script will create a rebase commit in the OpenShift Origin GIT repository
+# based on the current HEAD.
+#
+# NOTE: Make sure all your changes are committed and there are no junk files
+#       present in the pkg/ folder.
+#
+
+# Exclude these packages from source-to-image explicitely
+exclude_pkgs=(
+  pkg/cmd
+  pkg/config
+  pkg/create
+  pkg/run
+  pkg/test
+  pkg/version
+)
+
+s2i_dir=$GOPATH/src/github.com/openshift/source-to-image
+origin_dir=$GOPATH/src/github.com/openshift/origin
+origin_s2i_godep_dir=${origin_dir}/Godeps/_workspace/src/github.com/openshift/source-to-image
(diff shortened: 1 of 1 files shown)


mfojtik:
@soltysh @bparees FYI

hack/rebase-to-origin.sh
```diff
@@ -0,0 +1,46 @@
...
+  # Remove all explicitely excluded packages
+  for pkg in "${exclude_pkgs[@]}"; do
+    rm -rf "${origin_s2i_godep_dir}/${pkg}"
+  done
+
+  # Bump the origin Godeps.json file
+  sed -i '' "s/${s2i_godeps_ref}/${s2i_ref}/g" ${origin_dir}/Godeps/Godeps.json
+
+  # Make a commit with proper message
+  git add Godeps && git commit -m "bump(github.com/openshift/source-to-image) ${s2i_ref}"
```
soltysh: The form is: bump(path): commit id  <- you're missing the colon after closing bracket.
See https://github.com/openshift/origin/blob/master/HACKING.md#updating-other-godeps-from-upstream

soltysh:
One nit, otherwise LGTM.

mfojtik:
@soltysh corrected, thanks! [merge]

bparees:
so we've given up on using the "godeps" tool entirely eh?  maybe that should tell us something instead of writing bash scripts to work around it :(

mfojtik:
@bparees well godeps save/restore on origin does not work anyway as we ship the extra patches to other projects...

soltysh:
There's a WIP for replacing godeps with something different, to be chosen what exactly.

rhcarvalho:
if this is to exist, shouldn't it live in origin instead of here?

mfojtik:
@rhcarvalho yes it could and I will remove this file as soon as we have sane way to do rebase in origin for S2I :-) Without that I'm keeping this file there so whoever needs a rebase to origin don't need to type the commands in script.

PR #383, opened 2016-01-12
Fix go-dockerclient godeps

mfojtik:
Removes files that should be removed and they were not for some reason (I think it was my mistake with assuming `git add Godeps` is the right thing ;-)


mfojtik:
[test]

mfojtik:
[merge]

PR #385, opened 2016-01-13
More things that we missed in the injections PR

mfojtik:
* Truncate the rm-injections script itself
* Add describer

### DIFF
--- a/pkg/build/strategies/sti/sti.go
+++ b/pkg/build/strategies/sti/sti.go
@@ -456,7 +456,7 @@ func (b *STI) Execute(command string, user string, config *api.Config) error {
 		if err != nil {
 			return err
 		}
-		rmScript, err := util.CreateInjectedFilesRemovalScript(injectedFiles)
+		rmScript, err := util.CreateInjectedFilesRemovalScript(injectedFiles, "/tmp/rm-injections")
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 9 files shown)


mfojtik:
@php-coder FYI, I'll keep adding more here if we find something that we are missing.

mfojtik:
[test]

pkg/build/strategies/sti/sti.go
```diff
@@ -452,11 +452,18 @@ func (b *STI) Execute(command string, user string, config *api.Config) error {
 	// and wait till all injections are uploaded into the container that runs the
 	// assemble script.
 	if len(config.Injections) > 0 && command == api.Assemble {
+		workdir, err := b.docker.GetImageWorkdir(config.BuilderImage)
+		if err != nil {
+			return err
+		}
+		if err := util.FixInjectionsWithRelativePath(workdir, &config.Injections); err != nil {
```
mfojtik: @bparees just FYI, I was terribly wrong thinking that the WORKPATH is used automatically, it is not. The default for uploading files into container is set by Docker and it defaults to "/", which is not what we want. This should get the WORKDIR from the builder image and use it as a prefix for relative path...

mfojtik:
[test] (flake)

mfojtik:
@php-coder fixed the `return nil`.

php-coder:
LGTM

mfojtik:
[merge] thanks @php-coder

PR #389, opened 2016-01-17
return error if pull fails

bparees:


bparees:
[test]

bparees:
@mfojtik ptal.  i don't understand why this was doing a fallthrough before.  if PullAlways is set, and we can't pull the image, that should be a failure even if a (possibly older) version of the image exists locally.

this is going to be even more important for openshift where we're going to enforce pullalways to ensure you can't use images that exist on the node unless you also have permission to pull them.

bparees:
@gabemontero ptal also since this might have been your work on forcepull, though git blame points the finger at @mfojtik

gabemontero:
Going back and looking at the commit where @mfojtik switched s2i force pull from the binary forcePull semantic to be similar to what deploy does, that entire method you modified was simply added.  It was not a modification of a method added with forcePull.

However, I then saw in pkg/build/strategies/strategies.go's GetBuilderImage from that commit this logic:

if config.ForcePull {
		image, err = d.PullImage(config.BuilderImage)
		if err != nil {
			glog.Warningf("An error occurred when pulling %s: %v. Attempting to use local image.", config.BuilderImage, err)
			image, err = d.CheckImage(config.BuilderImage)
		}
	} else {
		image, err = d.CheckAndPullImage(config.BuilderImage)
	}

It is in fact trying to use the local image if the force pull failed.  So we can surmise I believe that @mfojtik 's change attempted to preserve that rationale.

Looking at the history of pkg/build/strategies/strategies.go, it was not touched by my changes for forcePull.  the behavior of using the local image was added with this commit by @soltysh :  https://github.com/openshift/source-to-image/commit/b6d798a614ed914cd8a4f6d07eb43d24b2f9b1b6#diff-08d24a3f6e101eecb563cec14c432646

bparees:
Thanks for the analysis @gabemontero, definitely helpful.  I think it was a mistake for this to ever have been a warning (but i can see why we would have gone that route in s2i at the time) and not an immediate failure.  If the user requests forcePull, we should pull or fail.

bparees:
@ncdc we'll need this updated in origin after it goes into s2i.

gabemontero:
Cool deal - glad i could help @bparees

mfojtik:
@bparees my guess is that this code was intended to work similar to what forcePull does before (forcePull pulls the image but if the image does not exists and exists locally, it used that image instead). I think this change is correct, but now there is no way to have the same behavior as with `forcePull`. Or maybe we need a new policy (`pull-or-local`).

bparees:
@mfojtik "try to pull but if the image can't be pulled and a local version exists, use it" is:

1) s2i build pullIfNotPresent <fails>
2) s2i build pullNever

:)

bparees:
[merge]

PR #390, opened 2016-01-20
Improve injection removal from the image

mfojtik:


pkg/build/strategies/sti/sti.go
```diff
@@ -461,15 +482,13 @@ func (b *STI) Execute(command string, user string, config *api.Config) error {
...
-			return err
-		}
-		defer os.Remove(rmScript)
 		glog.V(5).Infof("Waiting for injected files to be copied into assemble container...")
+		lastFileInjected := injectedFiles[len(injectedFiles)-1]
 		opts.CommandOverrides = func(cmd string) string {
-			return fmt.Sprintf("while [ ! -f %q ]; do sleep 0.5; done; %s; result=$?; source %[1]s; exit $result",
-				"/tmp/rm-injections", cmd)
+			// First we have to wait till the last injected file is present.
+			// When assemble is done, we have to want when the last injected file is
```
php-coder: we have to want -> we have to wait?

mfojtik: should not be there at all :) just leftover from my previous attempt

mfojtik:
[test]

mfojtik:
@csrwng @bparees ready for review :)

mfojtik:
@bparees do you know if this will work on OpenShift (we need to docker exec as root).

pkg/docker/docker.go
```diff
@@ -163,6 +171,60 @@ func New(config *api.DockerConfig, auth docker.AuthConfiguration) (Docker, error
...
+	return err
+}
+
+// ChangeContainerFileOwner changes an owner of a single file inside the
+// container.
+func (d *stiDocker) ChangeContainerFilesOwner(containerName, username string, files []string) error {
+	return d.dockerExec(containerName, append([]string{"chown", "-R", "-c", username}, files...))
+}
+
+// ChangeContainerFileOwner changes an owner of a single file inside the
```
csrwng: update go-doc

csrwng:
@mfojtik it should work because we are running as privileged and have full access to the docker socket. Not sure it's a good thing though :)

mfojtik:
@csrwng we already use UploadToContainer which runs as root :)

bparees:
> @bparees do you know if this will work on OpenShift (we need to docker exec as root).

yeah it'll presumably work, but it is slightly scary.

mfojtik:
@bparees there is no other way to add files into any location... so we sacrificed a little bit of security to get better user experience, which I'm not sure is a right thing :-) But I don't think this exposes some vulneability since we control the S2I exec invocations.

csrwng:
@mfojtik just thinking out loud.... what if the provided user image includes its own "rm" binary?

mfojtik:
@csrwng then we run it as root inside the assemble container...

csrwng:
isn't that a security exposure?

pkg/build/strategies/sti/sti.go
```diff
@@ -481,8 +500,10 @@ func (b *STI) Execute(command string, user string, config *api.Config) error {
 					return err
 				}
 			}
-			if err := b.docker.UploadToContainer(rmScript, "/tmp/rm-injections", containerID); err != nil {
-				return err
+			if len(user) > 0 {
+				if err := b.docker.ChangeContainerFilesOwner(containerID, user, injectedFiles); err != nil {
```
bparees: does this have any implication for ssl certs?  not sure if there are any ownership requirements on those.

certainly there are for ssh keys...

pkg/docker/docker.go
```diff
@@ -163,6 +171,60 @@ func New(config *api.DockerConfig, auth docker.AuthConfiguration) (Docker, error
...
+	}()
+	<-success
+	glog.V(5).Infof("Executed %#v command in %q", command, containerName)
+	return err
+}
+
+// ChangeContainerFileOwner changes an owner of a single file inside the
+// container.
+func (d *stiDocker) ChangeContainerFilesOwner(containerName, username string, files []string) error {
+	return d.dockerExec(containerName, append([]string{"chown", "-R", "-c", username}, files...))
```
bparees: same concern @csrwng raised about rm... this chown command is being run as root.  if the user overrode what "chown" is in their image, we just gave them the ability to run arbitrary root operations in the container.

pkg/docker/docker.go
```diff
@@ -163,6 +171,60 @@ func New(config *api.DockerConfig, auth docker.AuthConfiguration) (Docker, error
...
+}
+
+// KillContainerOptions kills the container using SIGKILL.
+func (d *stiDocker) KillContainer(containerID string) error {
+	opts := docker.KillContainerOptions{ID: containerID}
+	return d.client.KillContainer(opts)
+}
+
+// dockerExec executes the command in specified container
+func (d *stiDocker) dockerExec(containerName string, command []string) error {
```
bparees: name this dockerExecAsRoot or make it take a user name as a parameter.  we want to be very clear this is doing things as root.

bparees:
> isn't that a security exposure?

yes, yes it is.

mfojtik:
@csrwng @bparees after some thinking I realized that we allow you to run any command as root in Docker strategy, and I think this is no different, is it?

csrwng:
@mfojtik correct. But that is why we consider the Docker strategy less secure. We wouldn't allow it online for example.

mfojtik:
In that case I am more convinced to leave the truncation in place. That
runs as user. @bparees?

Also I think thos was a good exercise, at least for me to se what are the
boundnaries ww dont want to cross in s2i :)

bparees:
> In that case I am more convinced to leave the truncation in place. That runs as user. @bparees?

for now at least that seems best.

php-coder:
>In that case I am more convinced to leave the truncation in place. That runs as user.

I suggest to add a comment to the sources about why we're using current approach and what is alternative to it.

bparees:
@mfojtik close this?

PR #391, opened 2016-01-22
apply cgroup limits to docker operations

bparees:


bparees:
@mfojtik ptal

pkg/api/types.go
```diff
@@ -167,6 +167,19 @@ type Config struct {
 	// the container that runs assemble.
 	// All files we inject will be truncated after the assemble script finishes.
 	Injections InjectionList
+
+	// CGroupLimits describes the cgroups limits that will be applied to any containers
```
rhcarvalho: The correct is "cgroup" or "cgroups"?

pkg/api/types.go
```diff
@@ -167,6 +167,19 @@ type Config struct {
...
+	// run by s2i.
+	CGroupLimits *CGroupLimits
+}
+
+type CGroupLimits struct {
+	MemoryLimitBytes int64
+	CPUShares        int64
+	CPUPeriod        int64
+	CPUQuota         int64
+	// This value should always be -1
```
rhcarvalho: Maybe add a note briefly explaining why?

pkg/docker/docker.go
```diff
@@ -592,14 +594,23 @@ func (d *stiDocker) RunContainer(opts RunContainerOptions) (err error) {
 		config.AttachStdout = true
 	}
 
-	glog.V(2).Infof("Creating container using config: %+v", config)
```
rhcarvalho: Why are we dropping the this glog call?

pkg/docker/docker.go
```diff
@@ -592,14 +594,23 @@ func (d *stiDocker) RunContainer(opts RunContainerOptions) (err error) {
...
+		ccopts.HostConfig.PublishAllPorts = true
+		ccopts.HostConfig.NetworkMode = opts.NetworkMode
 	} else if opts.NetworkMode != "" {
-		ccopts.HostConfig = &docker.HostConfig{NetworkMode: opts.NetworkMode}
+		ccopts.HostConfig.NetworkMode = opts.NetworkMode
 	}
 
+	if opts.CGroupLimits != nil {
+		ccopts.HostConfig.Memory = opts.CGroupLimits.MemoryLimitBytes
+		ccopts.HostConfig.MemorySwap = opts.CGroupLimits.MemorySwap
```
rhcarvalho: Shouldn't this be "always -1"? Are we going to rely on all callers to set this to -1?

pkg/api/types.go
bparees: we're only talking about the limits for one cgroup, so it's correct as is.

pkg/docker/docker.go
bparees: it's not dropped, look down to line 613

pkg/docker/docker.go
bparees: yes we're going to rely on callers to implement the correct behavior for their use.  in the case of openshift it's -1, but there's no reason to hardcode that into the s2i api.

pkg/docker/docker.go
```diff
@@ -592,14 +594,23 @@ func (d *stiDocker) RunContainer(opts RunContainerOptions) (err error) {
...
+		ccopts.HostConfig.CPUShares = opts.CGroupLimits.CPUShares
+		ccopts.HostConfig.CPUQuota = opts.CGroupLimits.CPUQuota
+		ccopts.HostConfig.CPUPeriod = opts.CGroupLimits.CPUPeriod
+	}
+	glog.V(2).Infof("Creating container using config: %s, %#v, %#v", ccopts.Name, ccopts.Config, ccopts.HostConfig)
```
rhcarvalho: Isn't `%+v` enough?

I've used this in https://github.com/openshift/origin/pull/6715 and it is pretty verbose:

```go
Infof("Creating container with options {Name:%q Config:%+v HostConfig:%+v} ...", createOpts.Name, createOpts.Config, createOpts.HostConfig)
```

Example log output:

```
I0125 16:30:36.987207    1039 dockerutil.go:144] Creating container with options
{Name:"manual_testing_post_commit_hook" Config:&{Hostname: Domain name: User:
Memory:0 MemorySwap:0 MemoryReservation:0 KernelMemory:0 CPUShares:0 CPUSet:
AttachStdin:false AttachStdout:false AttachStderr:false P ortSpecs:[]
ExposedPorts:map[] StopSignal: Tty:false OpenStdin:false StdinOnce:false Env:[]
Cmd:[echo "Hello world!"] DNS:[] Image:c91ab321f208 Volumes:map[] VolumeDriver:
VolumesFrom: WorkingDir: MacAddress: Entrypoint:[sh -c] NetworkDisabled:false
SecurityOpts :[] OnBuild:[] Mounts:[] Labels:map[]} HostConfig:&{Binds:[]
CapAdd:[] CapDrop:[] GroupAdd:[] ContainerIDFile: LxcConf:[] Privileged:false
PortBin dings:map[] Links:[] PublishAllPorts:false DNS:[] DNSOptions:[]
DNSSearch:[] ExtraHosts:[] VolumesFrom:[] NetworkMode: IpcMode: PidMode:
UTSMode:  RestartPolicy:{Name: MaximumRetryCount:0} Devices:[] LogConfig:{Type:
Config:map[]} ReadonlyRootfs:false SecurityOpt:[] CgroupParent: Memory:0 Mem
orySwap:0 MemorySwappiness:0 OOMKillDisable:false CPUShares:1024 CPUSet:
CPUSetCPUs: CPUSetMEMs: CPUQuota:-1 CPUPeriod:100000 BlkioWeight:0 Ulimit s:[]
VolumeDriver:}} ...
```

pkg/docker/docker.go
```diff
@@ -669,7 +680,7 @@ func (d *stiDocker) CommitContainer(opts CommitContainerOptions) (string, error)
 			User:   opts.User,
 		}
 		dockerOpts.Run = &config
-		glog.V(2).Infof("Committing container with config: %+v", config)
+		glog.V(2).Infof("Committing container with config: %#v", dockerOpts)
```
rhcarvalho: Since `config` is a pointer, this will just log the memory address of the pointer, not very useful.
This change actually hides the content of `config` from the log output...

pkg/docker/docker.go
```diff
@@ -695,5 +706,13 @@ func (d *stiDocker) BuildImage(opts BuildImageOptions) error {
 		InputStream:         opts.Stdin,
 		OutputStream:        opts.Stdout,
 	}
+	if opts.CGroupLimits != nil {
+		dockerOpts.Memory = opts.CGroupLimits.MemoryLimitBytes
+		dockerOpts.Memswap = opts.CGroupLimits.MemorySwap
```
rhcarvalho: Ugly inconsistency in https://github.com/fsouza/go-dockerclient :cry:

pkg/docker/docker.go
bparees: it's verbose logging, i don't see the problem.

pkg/api/types.go
rhcarvalho: Shouldn't it be then "describes the *cgroup* limits"?

pkg/docker/docker.go
rhcarvalho: I'm not complaining that it is verbose, but suggesting to use the `%+v` format, since `%#v` is not adding any useful information AFAICT?

pkg/docker/docker.go
rhcarvalho: my bad

pkg/api/types.go
bparees: cgroups is the name of the technology.  so it's cgroups limits that are being enforced.

the struct is the limits for a particular cgroup, so it's cgrouplimits (the limits for this cgroup)

pkg/docker/docker.go
rhcarvalho: So then could you please either remove the comment from the type definition or make it more explicit that -1 is what makes sense in OpenShift's context today?

I remember similar comments we had like "never change this line", that caused doubts when we had to change the line.

bparees: sure

pkg/api/types.go
rhcarvalho: Ok, got it. Thanks for the patience to clarify it.

bparees:
@rhcarvalho fixed various logging formats and removed the comment about the -1 for swap, i'll just make it explicit where openshift is instantiating the struct.

rhcarvalho:
@bparees do you recall using the log output in some real case?
I don't recall using it myself, my only concern is that we'd need to read the source code (of a certain version) to understand what is being printed.
If I had to debug log messages, I'd prefer to see the output of something like "{Name:%s Config:%+v HostConfig:%+v}" more than  "%s, %+v, %+v".

Other than that LGTM.

bparees:
@rhcarvalho fair point, i've made the logging more explicit.

bparees:
[merge]

Issue #394, opened 2016-01-25
When using "scripts" in origin, the build will fail


mfojtik:
The build will fail when the URL does not have both "assemble" and "run". In case you just want to have custom "run" script, but you want to leave the "assemble" as it is, this will break for you and you will have to fork the "assemble" script as well.

Source-To-Image should download **only** scripts that are available and default to image scripts in case the download failed.

mfojtik:
@bparees FYI

PR #396, opened 2016-01-25
Fix scripts installation

mfojtik:
* Deprecate `.sti/bin` is the source with `.s2i/bin` (Fixes: https://github.com/openshift/source-to-image/issues/394)
* Always display when the image is being pulled to give users clue about what is going on (Fixes: https://github.com/openshift/origin/issues/6809)
* Fix contextDir when using FileDownloader
* Refactor `pkg/scripts/install.go` and by that fix the case when scriptURL is provided but one script fails to download which in that case use the other script handlers. (Fixes: https://github.com/openshift/source-to-image/issues/394)


soltysh:
It would be good to have unit tests covering both .sti/bin and .s2i/bin, otherwise lgtm.

pkg/scripts/install.go
```diff
@@ -91,10 +92,18 @@ func (i *installer) run(scripts []string, dstDir string) []api.InstallResult {
 		if !i.fs.Exists(file) {
 			sourceResults[script].err = errors.NewDownloadError(file, -1)
 		}
+		file = filepath.Join(dstDir, api.DeprecatedSourceScripts, script)
+		if i.fs.Exists(file) {
+			glog.Infof("DEPRECATED: The .sti/bin directory will be deprecated, use .s2i/bin instead")
+			sourceResults[script] = &downloadResult{location: api.DeprecatedSourceScripts}
+		}
```
bparees: you should put this first so if they have files in both .s2i and .sti, they get a warning but the files in .s2i get used.

mfojtik: need to refactor this anyway, so consider this pr as wip :)

mfojtik:
@bparees @soltysh https://github.com/mfojtik/source-to-image/blob/75ea61c01d546c3ccc398374c2bffcae962510b6/pkg/scripts/install.go

I refactored how we're downloading S2I scripts in more 'go-way'. At least it works for all cases now and adding "deprecated" `.sti/bin` in not tragedy anymore...

PTAL, still WIP, I will have to fix some integration tests and unit tests.

pkg/scm/file/download.go
```diff
@@ -21,7 +21,8 @@ func (f *File) Download(config *api.Config) (*api.SourceInfo, error) {
 	config.WorkingSourceDir = targetSourceDir
 
 	if len(config.ContextDir) > 0 {
-		targetSourceDir = filepath.Join(targetSourceDir, config.ContextDir, ".")
+		//targetSourceDir = filepath.Join(targetSourceDir, config.ContextDir, ".")
+		targetSourceDir = filepath.Join(config.WorkingDir, api.ContextTmp)
```
bparees: hmm...thought we needed this .

bparees: ah,nm

pkg/scripts/install.go
```diff
@@ -13,177 +15,230 @@ import (
...
+// Get parses the provided URL and the script name.
+func (s *UrlScriptHandler) Get(script string) *api.InstallResult {
+	if len(s.URL) == 0 {
+		return nil
 	}
-	if len(failedScripts) > 0 {
-		err = errors.NewInstallRequiredError(failedScripts, docker.ScriptsURLLabel)
+	scriptURL, err := url.Parse(s.URL + "/" + script)
+	if err != nil {
+		glog.Infof("invalid script url %q: %v", s.URL, err)
```
bparees: Errorf?

pkg/scripts/install.go
```diff
@@ -13,177 +15,230 @@ import (
...
-// with error if installing some of them failed
-func (i *installer) InstallRequired(scripts []string, dstDir string) (results []api.InstallResult, err error) {
-	results = i.run(scripts, dstDir)
-	failedScripts := []string{}
-	for _, r := range results {
-		if !r.Installed && r.Error != nil {
-			failedScripts = append(failedScripts, r.Script)
-		}
+// Get parses the provided URL and the script name.
+func (s *UrlScriptHandler) Get(script string) *api.InstallResult {
```
bparees: why does Get take an InstallResult?  and why does Install take an InstallResult?  i'd expect it to return one.

mfojtik: @bparees Get() verifies if the script is available for given source (or the ScriptSource provides the script). It returns the InstallResult with informations about how to install the script. The Install() then install it and update that information with setting the Installed/Downloaded to true.

mfojtik:
@bparees @soltysh revamped the install_test.go so now it has ~90% coverage and covers all cases for script installation we have.

pkg/scripts/install.go
mfojtik: @bparees better name suggestions are welcome :-)

mfojtik:
[test]

bparees:
squash and lgtm.

pkg/scripts/install.go
soltysh: Installer operates on InstallResult, I'm leaning towards having something along ScriptResult/DownloadResult/GetResult. I agree with @bparees that InstallResult is misleading here a bit.

pkg/scripts/install.go
```diff
@@ -13,177 +15,234 @@ import (
...
-		sourceResults[script] = &downloadResult{location: api.SourceScripts}
-		file := filepath.Join(dstDir, api.SourceScripts, script)
-		if !i.fs.Exists(file) {
-			sourceResults[script].err = errors.NewDownloadError(file, -1)
-		}
-		checkLocation := filepath.Join(dstDir, strings.TrimSuffix(api.SourceScripts, "/bin"), script)
-		if i.fs.Exists(checkLocation) {
-			glog.Warningf("The %q script found in .sti/%s, should be in .sti/bin/%s.", script, script, script)
-		}
+	location = strings.Replace(location, "s2i", "sti", 1)
```
soltysh: Add here a comment explaining why we're doing this, but frankly I'd prefer we have api.DeprecatedSourceScripts var used here instead of renaming :-/

pkg/scripts/install.go
```diff
@@ -13,177 +15,234 @@ import (
...
+}
 
-	// iterate over scripts
-	for idx, script := range scripts {
-		result := api.InstallResult{Script: script}
+// InstallRequired Downloads and installs required scripts into dstDir, the result is a
+// map of scripts with detailed information about each of the scripts install process
+// with error if installing some of them failed
+func (i *DefaultScriptSourceManager) InstallRequired(scripts []string, dstDir string) ([]api.InstallResult, error) {
+	result := i.InstallOptional(scripts, dstDir)
```
soltysh: I prefer the previous construct where both InstallRequired and InstallOptional called private run method. Seeing InstallRequired calling InstallOptional seems odd to me.

pkg/scripts/install.go
```diff
@@ -13,177 +15,234 @@ import (
...
-	}
+// DefaultScriptSourceManager manages the default script lookup and installation
+// for source-to-image.
+type DefaultScriptSourceManager struct {
+	Image      string
+	ScriptsURL string
+	download   Downloader
+	docker     docker.Docker
+	dockerAuth dockerClient.AuthConfiguration
+	sources    *list.List
```
soltysh: What's the advantage of using lists here? Isn't plain old array of ScriptHandler enough for it? I found the places using this confusing. First I found the for loop using Front etc and started wondering why?

pkg/scripts/install.go
```diff
@@ -13,177 +15,234 @@ import (
...
 
-type installer struct {
-	image      string
-	scriptsURL string
-	docker     docker.Docker
-	pullAuth   dockerClient.AuthConfiguration
-	downloader Downloader
-	fs         util.FileSystem
+// UrlScriptHandler handles script download using URL.
+type UrlScriptHandler struct {
```
soltysh: This struct is not used outside this package, if so make it private.

pkg/scripts/install.go
```diff
@@ -13,177 +15,234 @@ import (
...
+				Error:  fmt.Errorf("script %q not installed", script),
+			})
+		}
 	}
+	return result
+}
 
-	return resultList
+// handlerName returns the human friendly name of the script handler.
+func handlerName(s ScriptHandler) string {
```
soltysh: Instead of type switch here, which will have to be updated whenever we add a handler I'd rather see this as a handler interface method, then everything you call against that interface is simpler to read.

soltysh:
The general idea is perfect, but I'd like you to make a few adjustments. Are you planning on exposing Handler outside, if not make it private and the initialization of handlers should be done inside NewInstaller. Most importantly make it an array, there's no need to use lists for 2 elements here. Operate on handlers through interface: no casting, no type switches, plain old interface will be enough.

pkg/scripts/install.go
soltysh: It'll be also easier to find when actually removing those. See [docker.go](https://github.com/openshift/source-to-image/blob/master/pkg/docker/docker.go#L25) as an example.

pkg/scripts/install.go
```diff
@@ -13,177 +15,234 @@ import (
...
+	if err := s.fs.Rename(r.URL, dst); err != nil {
+		return err
+	}
+	if err := s.fs.Chmod(dst, 0755); err != nil {
+		return err
+	}
+	// Make the path to scripts nicer in logs
+	parts := strings.Split(r.URL, "/")
+	if len(parts) > 3 {
+		r.URL = "<source-dir>/" + strings.Join(parts[len(parts)-3:], "/")
```
soltysh: Make <source-dir> a private constant, it'll be clearer in tests as well.

pkg/scripts/install.go
mfojtik: @soltysh will add comment and TODO here, but I don't want to introduce `api. DeprecatedSourceScripts` because 1) the name is super-ugly, 2) it is too much engineering for 1 character change.

pkg/scripts/install.go
mfojtik: I don't like private methods because they are there to hide something you don't want to expose to other programmers. We don't want to hide something here, the InstallRequired is the same as InstallOptional except it is more restrictive in error checking than InstallOptional. I don't think that is a good reason to create third (hidden) function.

pkg/scripts/install.go
mfojtik: @soltysh the array will be probably good here as well, I prefer List because it guarantees the order and it is not that much performance hit. But yeah, I will switch it to use array.

pkg/scripts/install.go
mfojtik: @soltysh agree, I will add String() method to interface.

mfojtik:
@soltysh adding `String()` to ScriptHandler, so no more handlerName() and the `sources` is now `[]ScriptHandler` instead of `list`. PTAL

pkg/scripts/install.go
mfojtik: We discussed this with @soltysh on IRC. We keep this exported as it implements a public interface and allow upstream to use this package and create their own script handlers (and register them).

pkg/scripts/install.go
mfojtik: done

On Wed, Jan 27, 2016 at 9:20 AM, Maciej Szulik <notifications@github.com>
wrote:

> In pkg/scripts/install.go
> <https://github.com/openshift/source-to-image/pull/396#discussion_r50953095>
> :
>
> > -	defaultURL, err := i.docker.GetScriptsURL(i.image)
> > -	if err == nil && defaultURL != "" {
> > -		defaultResults = i.download(defaultURL, scripts, filepath.Join(dstDir, api.DefaultScripts))
> > +// Install copies the script into upload directory and fix its permissions.
> > +func (s *SourceScriptHandler) Install(r *api.InstallResult) error {
> > +	dst := filepath.Join(s.DestinationDir, api.UploadScripts, r.Script)
> > +	if err := s.fs.Rename(r.URL, dst); err != nil {
> > +		return err
> > +	}
> > +	if err := s.fs.Chmod(dst, 0755); err != nil {
> > +		return err
> > +	}
> > +	// Make the path to scripts nicer in logs
> > +	parts := strings.Split(r.URL, "/")
> > +	if len(parts) > 3 {
> > +		r.URL = "<source-dir>/" + strings.Join(parts[len(parts)-3:], "/")
>
> Make a private constant, it'll be clearer in tests as well.
>
> —
> <https://github.com/openshift/source-to-image/pull/396/files#r50953095>.
>



-- 

mfojtik:
@soltysh fixed all your comments, jenkins is happy, so [merge]

Issue #397, opened 2016-01-25
tar file is closed before further usage


rassie:
Apparently, when using source upload with `--inject` the `tar` file is created from two different directories under `/tmp`. However, the file is being closed just after the injection step, so that the actual source and the scripts cannot be written. 

`s2i --loglevel 5 --inject "/home/user/.m2/repository/:/root/.m2/repository/" build . app-builder app:latest` produces the following output:

```
...
I0125 17:12:51.152699 08557 tar.go:255] Adding to tar: /tmp/s2i-323596458/xmlpull/xmlpull/1.1.3.1/xmlpull-1.1.3.1.pom.sha1 as xmlpull/xmlpull/1.1.3.1/xmlpull-1.1.3.1.pom.sha1
I0125 17:12:51.152778 08557 tar.go:255] Adding to tar: /tmp/s2i-323596458/xpp3/xpp3_min/1.1.4c/_remote.repositories as xpp3/xpp3_min/1.1.4c/_remote.repositories
I0125 17:12:51.152845 08557 tar.go:255] Adding to tar: /tmp/s2i-323596458/xpp3/xpp3_min/1.1.4c/xpp3_min-1.1.4c.jar as xpp3/xpp3_min/1.1.4c/xpp3_min-1.1.4c.jar
I0125 17:12:51.152910 08557 tar.go:255] Adding to tar: /tmp/s2i-323596458/xpp3/xpp3_min/1.1.4c/xpp3_min-1.1.4c.jar.sha1 as xpp3/xpp3_min/1.1.4c/xpp3_min-1.1.4c.jar.sha1
I0125 17:12:51.152961 08557 tar.go:255] Adding to tar: /tmp/s2i-323596458/xpp3/xpp3_min/1.1.4c/xpp3_min-1.1.4c.pom as xpp3/xpp3_min/1.1.4c/xpp3_min-1.1.4c.pom
I0125 17:12:51.153013 08557 tar.go:255] Adding to tar: /tmp/s2i-323596458/xpp3/xpp3_min/1.1.4c/xpp3_min-1.1.4c.pom.sha1 as xpp3/xpp3_min/1.1.4c/xpp3_min-1.1.4c.pom.sha1
E0125 17:12:51.455001 08557 tar.go:210] Error writing header for Chart.Doughnut.js: io: read/write on closed pipe
[...]

PR #400, opened 2016-01-28
Fail the build when destinationDir does not exists during injection

mfojtik:
Fixes: https://github.com/openshift/source-to-image/issues/397

Thanks to @php-coder for debugging this problem! Can you please review this?

This will basically stop the build when the injection fails. It also provides 'better' error to user:

```console
E0128 09:52:23.853183 09625 injection.go:99] The destination directory for "/data/src/github.com/openshift/source-to-image/docs" injection must exists in container ("/root/foo")
I0128 09:52:24.776407 09625 main.go:326] An error occurred: API error (404): lstat /var/lib/docker/devicemapper/mnt/664dc416e71434fab5fae2324fdcc6125c141d3643b691442a250788b064e243/rootfs/root/foo: no such file or directory
```

### DIFF
--- a/pkg/build/strategies/sti/sti.go
+++ b/pkg/build/strategies/sti/sti.go
@@ -452,6 +452,8 @@ func (b *STI) Execute(command string, user string, config *api.Config) error {
 	// If there are injections specified, override the original assemble script
 	// and wait till all injections are uploaded into the container that runs the
 	// assemble script.
+	injectionComplete := make(chan struct{})
+	var injectionError error
 	if len(config.Injections) > 0 && command == api.Assemble {
 		workdir, err := b.docker.GetImageWorkdir(config.BuilderImage)
 		if err != nil {
@@ -467,22 +469,26 @@ func (b *STI) Execute(command string, user string, config *api.Config) error {
 			return err
 		}
 		defer os.Remove(rmScript)
-		glog.V(5).Infof("Waiting for injected files to be copied into assemble container...")
 		opts.CommandOverrides = func(cmd string) string {
 			return fmt.Sprintf("while [ ! -f %q ]; do sleep 0.5; done; %s; result=$?; source %[1]s; exit $result",
 				"/tmp/rm-injections", cmd)
 		}
 		originalOnStart := opts.OnStart
 		opts.OnStart = func(containerID string) error {
+			defer close(injectionComplete)
 			if err != nil {
+				injectionError = err
 				return err
 			}
+			glog.V(2).Info("starting the injections uploading ...")
 			for _, s := range config.Injections {
 				if err := b.docker.UploadToContainer(s.SourcePath, s.DestinationDir, containerID); err != nil {
+					injectionError = util.HandleInjectionError(s, err)
 					return err
 				}
 			}
 			if err := b.docker.UploadToContainer(rmScript, "/tmp/rm-injections", containerID); err != nil {
+				injectionError = util.HandleInjectionError(api.InjectPath{SourcePath: rmScript, DestinationDir: "/tmp/rm-injections"}, err)
 				return err
 			}
 			if originalOnStart != nil {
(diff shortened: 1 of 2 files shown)


pkg/build/strategies/sti/sti.go
```diff
@@ -490,16 +496,25 @@ func (b *STI) Execute(command string, user string, config *api.Config) error {
...
 		r, w := io.Pipe()
 		go func() {
+			// Wait for the injections to complete and check the error. Do not start
+			// streaming the sources when the injection failed.
+			<-injectionComplete
+			if injectionError != nil {
+				wg.Done()
+				return
+			}
+			glog.V(2).Infof("starting the source uploading ...")
```
mfojtik: fyi, I want to leave this here for the future debugging... the timing for streaming into container is important as when we fail in OnStart the container is removed.

pkg/util/injection.go
```diff
@@ -88,3 +88,17 @@ func CreateInjectedFilesRemovalScript(files []string, scriptName string) (string
...
 }
+
+// HandleInjectionError handles the error caused by injection and provide
+// reasonable suggestion to users.
+func HandleInjectionError(p api.InjectPath, err error) error {
+	if err == nil {
+		return nil
+	}
+	if strings.Contains(err.Error(), "no such file or directory") {
+		glog.Errorf("The destination directory for %q injection must exists in container (%q)", p.SourcePath, p.DestinationDir)
```
php-coder: I'm not sure but IMHO it's more correct to say "must exist" (without `s`). Correct me if I'm wrong.

pkg/build/strategies/sti/sti.go
php-coder: Is `Info()` should be enough here?

pkg/build/strategies/sti/sti.go
```diff
@@ -467,22 +469,26 @@ func (b *STI) Execute(command string, user string, config *api.Config) error {
...
 				"/tmp/rm-injections", cmd)
 		}
 		originalOnStart := opts.OnStart
 		opts.OnStart = func(containerID string) error {
+			defer close(injectionComplete)
 			if err != nil {
+				injectionError = err
 				return err
 			}
+			glog.V(2).Infof("starting the injections uploading ...")
```
php-coder: Is `Info()` should be enough here?

mfojtik: yeah, bad habit...

pkg/util/injection.go
```diff
@@ -88,3 +88,17 @@ func CreateInjectedFilesRemovalScript(files []string, scriptName string) (string
...
+		return err
+	}
+	glog.Errorf("Error occured during injecting %q: %v", p.SourcePath, err)
```
php-coder: WDYT about showing `destinationDir` here too? Because current error (and maybe some of hidden) is more related to the destination dir.

php-coder:
LGTM

mfojtik:
@php-coder hopefully fixed everything, thanks!

[merge]

pkg/build/strategies/sti/sti.go
```diff
@@ -467,22 +469,26 @@ func (b *STI) Execute(command string, user string, config *api.Config) error {
...
 			}
+			glog.V(2).Info("starting the injections uploading ...")
 			for _, s := range config.Injections {
 				if err := b.docker.UploadToContainer(s.SourcePath, s.DestinationDir, containerID); err != nil {
+					injectionError = util.HandleInjectionError(s, err)
 					return err
 				}
 			}
 			if err := b.docker.UploadToContainer(rmScript, "/tmp/rm-injections", containerID); err != nil {
+				injectionError = util.HandleInjectionError(api.InjectPath{SourcePath: rmScript, DestinationDir: "/tmp/rm-injections"}, err)
```
php-coder: I'm not sure but should we use `/tmp` for `DestinationDir` here? (As name implies?)

mfojtik: right, but this is just an informative (we show this only in error).

PR #404, opened 2016-01-29
handle builder image names ending in hex image ids

gabemontero:
@bparees @csrwng PTAL

this PR is for addressing the new flavor of bugzilla 1261117 (builder img with no tar), where we enter the layered path via `oc new-app` vs. `s2i`.

The build image name is getting set with the docker image ref / ending with @<hex img id>


bparees:
aside from my rapidly growing regret over the existence of the "layered build" feature in s2i, lgtm.
[merge]

PR #405, opened 2016-02-01
Deprecate .sti/environment

mfojtik:
@bparees FYI

This is a leftover from deprecating the `.sti` in `api.ScriptSource`. Also I'm adding glog.Infof() here because I think seeing that the environment of the build is going to change based on provided file is important to user to confirm the file is actually picked up and the env vars are used.

### DIFF
--- a/pkg/scripts/environment.go
+++ b/pkg/scripts/environment.go
@@ -18,30 +18,22 @@ type Environment struct {
 	Value string
 }
 
-// getEnvPath returns prefix/environment path.
-func getEnvPath(config *api.Config, prefix string) (string, error) {
-	envPath := filepath.Join(config.WorkingDir, api.Source, prefix, api.Environment)
-	if _, err := os.Stat(envPath); os.IsNotExist(err) {
-		return "", errors.New("no environment file found in application sources")
-	}
-	return envPath, nil
-}
-
 // GetEnvironment gets the .s2i/environment file located in the sources and
 // parse it into []environment
 func GetEnvironment(config *api.Config) ([]Environment, error) {
-	envPath, err := getEnvPath(config, ".s2i")
-	if err != nil {
-		envPath, err = getEnvPath(config, ".sti")
-		if err != nil {
-			return nil, err
+	envPath := filepath.Join(config.WorkingDir, api.Source, ".s2i", api.Environment)
+	if _, err := os.Stat(envPath); os.IsNotExist(err) {
+		// TODO: Remove this when the '.sti/environment' is deprecated.
+		envPath = filepath.Join(config.WorkingDir, api.Source, ".sti", api.Environment)
+		if _, err := os.Stat(envPath); os.IsNotExist(err) {
+			return nil, errors.New("no environment file found in application sources")
 		}
 		glog.Infof("DEPRECATED: Use .s2i/environment instead of .sti/environment")
 	}
 
 	f, err := os.Open(envPath)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


mfojtik:
@php-coder can you please review?

pkg/scripts/environment.go
```diff
@@ -21,14 +21,19 @@ type Environment struct {
 // GetEnvironment gets the .sti/environment file located in the sources and
```
php-coder: Should we update the comment as well?

php-coder:
I see that we still mention `.sti/environment` in `README.md`, probably, this place also should be updated.

php-coder:
One more place that probably need to be revised:
```
pkg/build/strategies/sti/sti.go:269:            glog.V(1).Infof("No .sti/environment provided (%v)", err)
pkg/build/strategies/sti/sti.go:416:            glog.V(1).Infof("No .sti/environment provided (%v)", err)
```

php-coder:
See my questions below and LGTM overall.

pkg/scripts/environment.go
mfojtik: yes we should, good catch

mfojtik:
@php-coder thanks will fix all that

pkg/scripts/environment.go
```diff
@@ -21,14 +21,19 @@ type Environment struct {
...
+		if _, err := os.Stat(envPath); os.IsNotExist(err) {
+			return nil, errors.New("no environment file found in application sources")
+		}
+		glog.Infof("DEPRECATED: Use .s2i/environment instead of .sti/environment")
 	}
 
 	f, err := os.Open(envPath)
 	if err != nil {
-		return nil, errors.New("unable to read .sti/environment file")
+		return nil, errors.New("unable to read custom environment file")
```
bparees: fmt.Errorf("unable to read custom environment file %q",envPath)

pkg/scripts/environment.go
```diff
@@ -49,10 +54,10 @@ func GetEnvironment(config *api.Config) ([]Environment, error) {
 			Name:  strings.TrimSpace(parts[0]),
 			Value: strings.TrimSpace(parts[1]),
 		}
-		glog.V(1).Infof("Setting '%s' to '%s'", e.Name, e.Value)
 		result = append(result, e)
 	}
 
+	glog.Infof("Setting %d environment variables provided by '.s2i/environment'", len(result))
```
bparees: slightly bothers me this message might be wrong if they used ".sti/environment" but i'm going to live with it :)

pkg/scripts/environment.go
mfojtik: @bparees problem with that is that envPath resolves to full path (including the workdir and it looks **really** ugly in console then ;-)

pkg/scripts/environment.go
mfojtik: yes, same concert... will change that to "provided by environment file in sources"

mfojtik:
@php-coder @bparees thanks for comments, updated.

mfojtik:
[test]

pkg/scripts/environment.go
bparees: Yeah i'm aware but if there's an error reading it, the full path is useful.
This should be very rare anyway.

Ben Parees | OpenShift

bparees:
lgtm following a rebase.

mfojtik:
rebased.

[merge]

Issue #275, opened 2015-08-24
Rename sti leftovers


soltysh:
This is a followup task to #244 and #250. I'm copying here the notes from my PR regarding what still needs to be changed (in that particular order):

* [x] change `.sti/` directory to `.s2i/` directory
* [x] change `/tmp/sti/` to `/tmp/s2i` or other not using `/tmp` directory
* [ ] change internals to use s2i everywhere

mfojtik:
@soltysh 

1) We will have to support both .sti and .s2i at least for some time

soltysh:
Yes, I'm aware of that.

rhcarvalho:
Related issue (/tmp): #211

php-coder:
I see that https://github.com/openshift/source-to-image/blob/master/docs/builder_image.md#s2i-builder-image-requirements uses `sti`: "... should be aware of the two basic requirements for the best possible **sti** performance"

bparees:
@mfojtik I think the work you did recently should be sufficient to close this?

soltysh:
From quick glimpse:
- [ ] .sticonfig 
- [ ] some docs leftovers
- [ ] .sti/environment

All addressed in #408 . The remaining stuff are only internals, but that will be done over time, I guess.

mfojtik:
@bparees I hope so + what @soltysh said. will close this as soon as #408 gets merged.

bparees:
#408 merged.

PR #408, opened 2016-02-03
S2i leftovers

soltysh:
Fixes leftovers from #275.

@mfojtik @bparees

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -11,8 +11,8 @@ import (
 	"github.com/spf13/pflag"
 )
 
-// DefaultConfigPath specifies the default location of the STI config file
-const DefaultConfigPath = ".stifile"
+// DefaultConfigPath specifies the default location of the S2I config file
+const DefaultConfigPath = ".s2ifile"
 
 // Config represents a basic serialization for the STI build options
 type Config struct {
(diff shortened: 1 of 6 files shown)


pkg/config/config.go
```diff
@@ -11,8 +11,9 @@ import (
 	"github.com/spf13/pflag"
 )
 
-// DefaultConfigPath specifies the default location of the STI config file
-const DefaultConfigPath = ".stifile"
+// DefaultConfigPath specifies the default location of the S2I config file
+const DeprecatedConfigPath = ".stifile"
```
mfojtik: i'm not a fan of having these 'Deprecated*' constants... we are not doing that of other deprecated stuff and we will expose them to users as API (where there is no point exposing them).

I would have just one constant and handle deprecation below, what do you think?

mfojtik:
@soltysh one nit, otherwise :+1: 

also please squash you cheater.

soltysh:
Comment addressed.

> also please squash you cheater.

It's to distinguish which change is where :smile:

mfojtik:
[merge]

On Wed, Feb 3, 2016 at 9:43 AM, Maciej Szulik <notifications@github.com>
wrote:

> Comment addressed.
>
> also please squash you cheater.
>
> It's to distinguish which change is where [image: :smile:]
>
> —
> <https://github.com/openshift/source-to-image/pull/408#issuecomment-179094582>
> .
>



-- 

PR #410, opened 2016-02-09
Remove custom SIGQUIT handler

rhcarvalho:
Go prints a stack trace of all goroutines by default on SIGQUIT, except those related to the runtime.
To include all goroutines, set GOTRACEBACK=2 or GOTRACEBACK=crash.

This custom handler suppresses Go's default behavior.

http://stackoverflow.com/a/35290196/4804690


rhcarvalho:
@bparees PTAL

bparees:
lgtm [merge]

PR #411, opened 2016-02-10
Always display source downloading message

mfojtik:


mfojtik:
@bparees FYI (will rebase to origin + do the same for docker strategy)

bparees:
[merge]

bparees:
@mfojtik pretty sure we handle cloning source differently in origin now, right?  I don't think this method gets invoked.

mfojtik:
@bparees hmm I think you're right. going to check.

Issue #412, opened 2016-02-18
no way to set incremental image pull policy explicitly


bparees:
pull policy used to pull the previous image is tied to the pull policy for the builder image.

Mangirdaz:
For now forcepull (in OSEv3.2) does not work with incremental build. From code and description i assume this will enable these two to work together?

bparees:
What do you mean incremental does not work with forcepull today?

this feature simply enables you to set a cluster wide default such that all builds will, by default, use incremental builds.

Mangirdaz:
Still debugging on this, but one of the clients started complaining that incremental builds stopped working when they added this to the template. 

if incremental is set build maven builds copy paste .m2 data into new container. And if they added forcepull: true it apparently stopped doing incremental ones.

But still need to confirm this one before raising new ticket...

bparees:
ok, keep us posted, but it *should* work.

Mangirdaz:
Will do. I suspect that previous build was failed so it defaulted to build from scratch, despite the fact we had one n-2 build running. But at this point its my assumption before debugging.

PR #413, opened 2016-02-18
expose previous image pull policy as a flag

bparees:
fixes https://github.com/openshift/source-to-image/issues/412

### DIFF
--- a/cmd/s2i/main.go
+++ b/cmd/s2i/main.go
@@ -81,6 +81,9 @@ $ s2i build . centos/ruby-22-centos7 hello-world-app
 			if len(cfg.BuilderPullPolicy) == 0 {
 				cfg.BuilderPullPolicy = api.DefaultBuilderPullPolicy
 			}
+			if len(cfg.PreviousImagePullPolicy) == 0 {
+				cfg.PreviousImagePullPolicy = api.DefaultPreviousImagePullPolicy
+			}
 
 			if errs := validation.ValidateConfig(cfg); len(errs) > 0 {
 				for _, e := range errs {
@@ -215,6 +218,9 @@ func newCmdRebuild(cfg *api.Config) *cobra.Command {
 			if len(cfg.BuilderPullPolicy) == 0 {
 				cfg.BuilderPullPolicy = api.DefaultBuilderPullPolicy
 			}
+			if len(cfg.PreviousImagePullPolicy) == 0 {
+				cfg.PreviousImagePullPolicy = api.DefaultPreviousImagePullPolicy
+			}
 
 			if glog.V(2) {
 				fmt.Printf("\n%s\n", describe.DescribeConfig(cfg))
@@ -281,6 +287,9 @@ func newCmdUsage(cfg *api.Config) *cobra.Command {
 
 			if len(cfg.BuilderPullPolicy) == 0 {
 				cfg.BuilderPullPolicy = api.DefaultBuilderPullPolicy
+			}
+			if len(cfg.PreviousImagePullPolicy) == 0 {
+				cfg.PreviousImagePullPolicy = api.DefaultPreviousImagePullPolicy
 			}
 
 			uh, err := sti.NewUsage(cfg)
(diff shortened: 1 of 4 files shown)


bparees:
@soltysh @mfojtik ptal, going to cut a release after this merges.
[test]

pkg/cmd/util.go
```diff
@@ -24,6 +24,8 @@ func AddCommonFlags(c *cobra.Command, cfg *api.Config) {
 		"DEPRECATED: Always pull the builder image even if it is present locally")
 	c.Flags().VarP(&(cfg.BuilderPullPolicy), "pull-policy", "p",
 		"Specify when to pull the builder image (always, never or if-not-present)")
+	c.Flags().Var(&(cfg.PreviousImagePullPolicy), "incremental-pull-policy",
+		"Specify when to pull the previous image for incremental builds (always, never or if-not-present)")
```
soltysh: Unfortunately there's a slight problem with this, and main reason I haven't submitted my changes yet. Since @mfojtik created [this String function](https://github.com/openshift/source-to-image/blob/master/pkg/api/types.go#L319-L324) which *always* shows `if-not-present` which is not true for previous image policy. His reasoning at the time was to show that default, but we can't do it correctly. We need to remove that string function and place the default information inside the description. I'm also looking at other option of using `StringVarP` method and transform the string into the PullPolicy type.

mfojtik: I would not say that this is not true for the previous pull policy. Having this default to "if-not-present" means that we will attempt to pull from registry when the image is not available locally. Which seems to be a good default for this as well.

soltysh: That's other solution, but we set it to always for a reason and I'd rather keep it that way.

mfojtik: @soltysh what was the reason? Does this mean that when I building the image locally, s2i will always attempt to fetch the image that I have pushed to remote registry, ignoring the local version? That has big potential to break me :-)

soltysh: If running inside of openshift we wanted to make sure you actually use the latest build image, which not always is the same for local build :/ In the end I can be convinced to using pull-if-not-present if that will be applied by default, which is not atm.

mfojtik: @soltysh in that case, OpenShift can set this to whatever it wants and it is not affecting the setting here which is intended to be default for the CLI ;-)

soltysh: As discussed on IRC: we'll leave previous image pull policy `always` in the main part of s2i, this way always will be the default for OpenShift usecase, which is embedding the core of s2i. Whereas in `main.go` we'll force previous image pull policy to be `pull-if-not-present`, this will address the CLI usecase. @bparees can you apply those changes?

soltysh: A little explanation why the two different pull policy are needed:
1. for OpenShift we want to make sure we always use the latest built image, the actual build may land on different nodes, which may or may not have the most recent image, thus the `always` policy assures we do have the latest, which will be in the output image stream
2. for CLI usecases, `pull-if-not-present` is the optimal, because when I'm developing locally I'd like to re-use already downloaded deps. @mfojtik gave here a really good example of a conference wifi, when I'm trying to rebuild the image (eg. I fixed a slight type just before the presentation) but without downloading half of the internet (yeah I'm looking at you java/mvn guys) but rather reusing what I have in my previous image.

bparees: none of this is necessary.

i'm setting cfg.PreviousImagePullPolicy in main.go, which ensures that when usage is printed, it prints the correct "default" value.

cmd/s2i/main.go
```diff
@@ -328,7 +337,7 @@ func checkErr(err error) {
 }
 
 func main() {
-	cfg := &api.Config{}
+	cfg := &api.Config{PreviousImagePullPolicy: api.PullAlways}
```
bparees: @soltysh @mfojtik this is the magic right here.

soltysh:
The problem was already solved, I must have been blind earlier when looking at it.
LGTM [merge]

mfojtik:
LGTM, [merge]

bparees:
closing, @soltysh is going to pick this up and resolve the TC issue.

PR #422, opened 2016-03-03
Add support for custom tag for incremental builds

rhcarvalho:
This allows to take artifacts from an image with a tag different than the output tag.

The default behavior is still to use the same tags.

This is needed to fix a regression in incremental builds OpenShift: https://github.com/openshift/origin/issues/7708

### DIFF
--- a/pkg/build/strategies/sti/sti_test.go
+++ b/pkg/build/strategies/sti/sti_test.go
@@ -447,6 +447,32 @@ func TestSaveArtifacts(t *testing.T) {
 	}
 }
 
+func TestSaveArtifactsCustomTag(t *testing.T) {
+	bh := testBuildHandler()
+	bh.config.WorkingDir = "/working-dir"
+	bh.config.IncrementalFromTag = "custom/tag"
+	bh.config.Tag = "image/tag"
+	fs := bh.fs.(*test.FakeFileSystem)
+	fd := bh.docker.(*docker.FakeDocker)
+	th := bh.tar.(*test.FakeTar)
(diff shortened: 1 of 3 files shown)


rhcarvalho:
@bparees @soltysh PTAL

rhcarvalho:
[test]

bparees:
lgtm, thanks @rhcarvalho 
[merge]

soltysh:
@bparees I didn't even get a chance to complain about missing tests and you merged it :disappointed:

bparees:
@soltysh we have upstream tests for incremental builds.

rhcarvalho:
sorry I missed the tests in the commit, my bad.

@soltysh @bparees ptal again

I'm adding a test for the new config option.

pkg/build/strategies/sti/sti_test.go
```diff
@@ -447,6 +447,32 @@ func TestSaveArtifacts(t *testing.T) {
 	}
 }
 
+func TestSaveArtifactsCustomTag(t *testing.T) {
+	bh := testBuildHandler()
+	bh.config.WorkingDir = "/working-dir"
+	bh.config.IncrementalFromTag = "custom/tag"
```
rhcarvalho: When both tags are set, `IncrementalFromTag` takes precedence and gets used.

bparees:
lbtm (looks better to me).
[merge]

rhcarvalho:
> lbtm

:smile: that's a new one to my catalogue, thanks @bparees

soltysh:
Thank you!

PR #423, opened 2016-03-03
Defer closing all ends of pipe

rhcarvalho:
- Let the function that created the pipe arrange freeing resources
  (we shall not rely on extractFunc being executed).
- Remove unnecessary if clause.

### DIFF
--- a/pkg/build/strategies/sti/sti.go
+++ b/pkg/build/strategies/sti/sti.go
@@ -152,9 +152,7 @@ func (b *STI) Build(config *api.Config) (*api.Result, error) {
 	if b.incremental {
 		if err := b.artifacts.Save(config); err != nil {
 			glog.Warningf("Clean build will be performed because of error saving previous build artifacts")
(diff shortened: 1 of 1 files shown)


pkg/build/strategies/sti/sti.go
```diff
@@ -371,12 +369,13 @@ func (b *STI) Save(config *api.Config) (err error) {
 		image = config.Tag
 	}
 	outReader, outWriter := io.Pipe()
+	defer outReader.Close()
+	defer outWriter.Close()
```
rhcarvalho: This matches the pattern of neighboring code:
https://github.com/openshift/source-to-image/pull/423/files#diff-77c4df2e21c6a783c0928af25a29ba8dR431

rhcarvalho:
Little fixes while looking at the `sti.go` file, may I bother you once again? @bparees @soltysh ptal

bparees:
lgtm [merge]

PR #425, opened 2016-03-07
Allowing dropping capabilities when running a container

csrwng:
Adds a flag to the s2i build command to allow specifying a set of capabilities to drop when executing a container.

Example:
```
s2i build --cap-drop=SETGID,SETUID https://github.com/myname/myrepo.git myco/builder:latest my-app
```


csrwng:
This will be used to fix BZ 1315187
@mfojtik or @gabemontero ptal

bparees:
lgtm
[merge]

csrwng:
thx!

Issue #367, opened 2015-11-25
Empty directories from local file system not preserved.


GrahamDumpleton:
As mentioned in #348, when S2I is used against a git repo it doesn't have to worry about empty directories because to have an empty directory in git you need to add a dummy file such as ``.gitkeep`` to it also git will never add the empty directory to the report.

When running S2I against the file system though, an empty directory could well exist as no requirement to have a dummy file in it. Unfortunately the directory when empty isn't preserved by S2I and it never created in the image. This means that if any code expected the empty directory to exist when run in the image, it will fail.

gabemontero:
So after reading this and https://github.com/openshift/source-to-image/pull/348 a couple of times, I suppose we could, after finding the empty directories off of the `targetPath` (after the `cp -a sourcePath targetPath`) via a `find targetPath -type d -empty`
1) add a `.gitkeep` file to any empty directories on the `targetPath` slide found so `git` respects them
2) explicitly document we don't support empty dirs, and remove them from the `targetPath` after the `cp -a`

I suppose 1) is better, but wonder if I'm missing (unfavorable) implications down the line.

@bparees @csrwng @mfojtik @GrahamDumpleton - any opinions / suggestions?

mfojtik:
@gabemontero another option would be to add special flag to force "file" SCM, so the directory with source will be copied instead of 'git cloned'.

gabemontero:
ah, that's right @mfojtik , thanks for the reminder ... s2i today has a generic "scm" concept, where the 2 impls or plugins at the moment are git and file.

Before moving to next steps, a quick recap of what the code does in determining which to use:
1) it takes the source spec and sees if the file exists
2) it then looks to see if either a) the git binary is not in the path, or b) it looks to see if a .git file does not exists off of the root of the source spec provided

If either a) or b) are true, then it uses local copy.

So in @GrahamDumpleton 's scenario, if it ended up doing a git clone where a local dir path was provided, either the path did not in fact exist, or the top dir had a .git file ... i.e. the source dir tree was originally created by a git clone, and then empty directories were created after the fact.

Certainly seems like a reasonable dev path, and expecting the user to remove .git files and force the copy behavior based on how our impl is currently done is a bit much.

I'll start down the add the config option path, unless the typical concerns about adding another config option are too high in this case vs. the value of satisfying this scenario.

bparees:
I think i'm ok w/ the "forceCopy" (vs clone) flag to address this.

[...]

PR #428, opened 2016-03-14
add option to only attempt copy command on source url

gabemontero:
Fixes https://github.com/openshift/source-to-image/issues/367

@bparees @mfojtik PTAL

### DIFF
--- a/cmd/s2i/main.go
+++ b/cmd/s2i/main.go
@@ -181,6 +181,7 @@ $ s2i build . centos/ruby-22-centos7 hello-world-app
 	buildCmd.Flags().StringSliceVar(&(cfg.DropCapabilities), "cap-drop", []string{}, "Specify a comma-separated list of capabilities to drop when running Docker containers")
 	buildCmd.Flags().StringVarP(&(oldDestination), "location", "l", "",
 		"DEPRECATED: Specify a destination location for untar operation")
+	buildCmd.Flags().BoolVarP(&(cfg.ForceCopy), "copy", "c", false, "Use local file system copy instead of git cloning the source url")
 
 	return buildCmd
 }
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -72,6 +72,7 @@ that image and add them to the tar streamed to the container into `/artifacts`.
 | Name                       | Description                                             |
 |:-------------------------- |:--------------------------------------------------------|
 | `--callback-url`           | URL to be invoked after a successful build (see [Callback URL](#callback-url)) |
+| `-c (--copy)`             | Use local file system copy instead of git cloning the source url (allows for inclusion of empty directories) |
 | `-d (--destination)`       | Location where the scripts and sources will be placed prior doing build (see [S2I Scripts](https://github.com/openshift/source-to-image/blob/master/docs/builder_image.md#s2i-scripts)) |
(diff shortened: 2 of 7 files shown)


cmd/s2i/main.go
```diff
@@ -181,6 +181,7 @@ $ s2i build . centos/ruby-22-centos7 hello-world-app
 	buildCmd.Flags().StringSliceVar(&(cfg.DropCapabilities), "cap-drop", []string{}, "Specify a comma-separated list of capabilities to drop when running Docker containers")
 	buildCmd.Flags().StringVarP(&(oldDestination), "location", "l", "",
 		"DEPRECATED: Specify a destination location for untar operation")
+	buildCmd.Flags().BoolVarP(&(cfg.ForceCopy), "copy", "c", false, "Only attempt a local file system copy against the source URL")
```
bparees: "Use local file system copy instead of git cloning the source url"

docs/cli.md
```diff
@@ -72,6 +72,7 @@ that image and add them to the tar streamed to the container into `/artifacts`.
 | Name                       | Description                                             |
 |:-------------------------- |:--------------------------------------------------------|
 | `--callback-url`           | URL to be invoked after a successful build (see [Callback URL](#callback-url)) |
+| `-c (--copy)`             | Only attempt a local file system copy against the source URL (allows for inclusion of empty directories) |
```
bparees: "Use local file system copy instead of git cloning the source URL (allows...."

bparees:
one wording change suggestion and lgtm

gabemontero:
word changes pushed

bparees:
lgtm [merge]

rhcarvalho:
@gabemontero where does the "copy" actually happen?

gabemontero:
@rhcarvalho it happens here:  https://github.com/openshift/source-to-image/blob/master/pkg/scm/file/download.go#L28

rhcarvalho:
Hmm I was surprised that doesn't appear in this patch. Looking at the changes seems that the meat was at https://github.com/openshift/source-to-image/pull/428/files#diff-e9f1cf6224310cfa91dbb395a6db1f85R38 but then I could not understand how this affects the code path. Some comments advertise that `cp` would be used instead of `git clone`, so I'm guessing the two mechanisms were already in place and we're just exposing it here, did I get that right?

bparees:
>  so I'm guessing the two mechanisms were already in place and we're just exposing it here, did I get that right?

right.  we're giving you a way to override the magic that picks the mechanism.

rhcarvalho:
> we're giving you a way to override the magic that picks the mechanism.

I couldn't find where we map `forceCopy = true` to "make sure we use `cp` and not `git clone`"?

The condition reads `details.FileExists && (details.UseCopy || forceCopy)`, so I'm not clear on how `details.UseCopy` gets set or what it means, but I read that as "if the file exists and either UseCopy or forceCopy is true, then return a File instance" -- but no guarantee what will be the mechanism?

Perhaps what happens is that, for empty dirs, `git clone` is not possible so `cp` is the only option left.
But if I use `--force-copy` with a directory containing `.git`, won't it just do a `git clone`?!

rhcarvalho:
The option name is actually `--copy`, my bad.

PR #429, opened 2016-03-17
Bug 1318395: Add timeout check for Docker API calls

mfojtik:


mfojtik:
[test]

@bparees || @rhcarvalho can you review this?

mfojtik:
@bparees there are more calls we can wrap with timeout, but I don't want to go nuts about them unless we get BZ when people hitting timeouts...

bparees:
lgtm [merge]

pkg/util/timeout.go
```diff
@@ -0,0 +1,37 @@
...
+func (t *TimeoutError) Error() string {
+	return fmt.Sprintf("calling the function timeout after %v", t.after)
+}
+
+// TimeoutAfter executes the provide function and return the TimeoutError in
+// case when the execution time of the provided function is bigger than provided
+// time duration.
+func TimeoutAfter(t time.Duration, fn func() error) error {
+	c := make(chan error, 1)
+	defer close(c)
```
rhcarvalho: The buffer in the channel is the right pattern to avoid leaking a goroutine, well done :)

There's still a bug here. When we time out, the goroutine started in the next line will panic sending on a closed channel.

It should be:

```go
c := make(chan error, 1)
go func() {
  defer close(c)
  c <- fn()
}()
select {
```

That ensures the channel is only closed after we send.

mfojtik: i dont think it will panic... that will show up in unot test, right?

mfojtik: can you provide an example where this will panic?

rhcarvalho: http://play.golang.org/p/N5ANWaIdt6

PR #430, opened 2016-03-17
Rework RunContainer implementation

rhcarvalho:
The objective is to fix concurrency issues with the current implementation.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1315607.

### DIFF
--- a/pkg/docker/docker.go
+++ b/pkg/docker/docker.go
@@ -6,7 +6,6 @@ import (
 	"path"
 	"path/filepath"
 	"strings"
-	"sync"
 	"time"
 
 	docker "github.com/fsouza/go-dockerclient"
@@ -83,7 +82,7 @@ type Client interface {
 	InspectImage(name string) (*docker.Image, error)
 	PullImage(opts docker.PullImageOptions, auth docker.AuthConfiguration) error
 	CreateContainer(opts docker.CreateContainerOptions) (*docker.Container, error)
-	AttachToContainer(opts docker.AttachToContainerOptions) error
+	AttachToContainerNonBlocking(opts docker.AttachToContainerOptions) (docker.CloseWaiter, error)
 	StartContainer(id string, hostConfig *docker.HostConfig) error
 	WaitContainer(id string) (int, error)
 	UploadToContainer(id string, opts docker.UploadToContainerOptions) error
@@ -130,6 +129,66 @@ type RunContainerOptions struct {
 	User             string
 	CGroupLimits     *api.CGroupLimits
 	CapDrop          []string
+}
+
+// asDockerConfig converts a RunContainerOptions into a Config understood by the
+// go-dockerclient library.
+func (rco RunContainerOptions) asDockerConfig() docker.Config {
+	return docker.Config{
+		Image:        getImageName(rco.Image),
+		User:         rco.User,
+		Env:          rco.Env,
+		OpenStdin:    rco.Stdin != nil,
+		StdinOnce:    rco.Stdin != nil,
+		AttachStdout: rco.Stdout != nil,
+	}
+}
+
+// asDockerHostConfig converts a RunContainerOptions into a HostConfig
+// understood by the go-dockerclient library.
+func (rco RunContainerOptions) asDockerHostConfig() docker.HostConfig {
+	hostConfig := docker.HostConfig{
+		CapDrop:         rco.CapDrop,
+		PublishAllPorts: rco.TargetImage,
+		NetworkMode:     rco.NetworkMode,
+	}
+	if rco.CGroupLimits != nil {
+		hostConfig.Memory = rco.CGroupLimits.MemoryLimitBytes
+		hostConfig.MemorySwap = rco.CGroupLimits.MemorySwap
+		hostConfig.CPUShares = rco.CGroupLimits.CPUShares
+		hostConfig.CPUQuota = rco.CGroupLimits.CPUQuota
+		hostConfig.CPUPeriod = rco.CGroupLimits.CPUPeriod
+	}
+	return hostConfig
+}
+
+// asDockerCreateContainerOptions converts a RunContainerOptions into a
+// CreateContainerOptions understood by the go-dockerclient library.
+func (rco RunContainerOptions) asDockerCreateContainerOptions() docker.CreateContainerOptions {
+	config := rco.asDockerConfig()
+	hostConfig := rco.asDockerHostConfig()
+	return docker.CreateContainerOptions{
+		Name:       "",
+		Config:     &config,
+		HostConfig: &hostConfig,
+	}
+}
+
+// asDockerAttachToContainerOptions converts a RunContainerOptions into a
+// AttachToContainerOptions understood by the go-dockerclient library.
+func (rco RunContainerOptions) asDockerAttachToContainerOptions() docker.AttachToContainerOptions {
+	return docker.AttachToContainerOptions{
+		InputStream:  rco.Stdin,
+		OutputStream: rco.Stdout,
+		ErrorStream:  rco.Stderr,
+
+		Stdin:  rco.Stdin != nil,
+		Stdout: rco.Stdout != nil,
+		Stderr: rco.Stderr != nil,
+
+		Logs:   rco.Stdout != nil,
+		Stream: rco.Stdout != nil,
+	}
 }
 
 // CommitContainerOptions are options passed in to the CommitContainer method
@@ -441,10 +500,9 @@ func getDestination(image *docker.Image) string {
 }
 
 // this funtion simply abstracts out the tar related processing that was originally inline in RunContainer()
-func runContainerTar(opts RunContainerOptions, config docker.Config, imageMetadata *docker.Image) (docker.Config, string) {
-	tarDestination := ""
+func runContainerTar(opts RunContainerOptions, imageMetadata *docker.Image) (cmd []string, tarDestination string) {
 	if opts.TargetImage {
-		return config, tarDestination
+		return
 	}
 
 	// base directory for all STI commands
@@ -474,7 +532,7 @@ func runContainerTar(opts RunContainerOptions, config docker.Config, imageMetada
 
 	// NOTE: We use path.Join instead of filepath.Join to avoid converting the
 	// path to UNC (Windows) format as we always run this inside container.
-	cmd := []string{path.Join(commandBaseDir, string(opts.Command))}
+	cmd = []string{path.Join(commandBaseDir, string(opts.Command))}
 	// when calling assemble script with Stdin parameter set (the tar file)
 	// we need to first untar the whole archive and only then call the assemble script
 	if opts.Stdin != nil && (opts.Command == api.Assemble || opts.Command == api.Usage) {
@@ -483,9 +541,8 @@ func runContainerTar(opts RunContainerOptions, config docker.Config, imageMetada
 			cmd = []string{"/bin/sh", "-c", opts.CommandOverrides(strings.Join(cmd[2:], " "))}
 		}
 	}
-	glog.V(5).Infof("Running %q command in container ...", strings.Join(cmd, " "))
-	config.Cmd = cmd
-	return config, tarDestination
+	glog.V(5).Infof("Setting %q command for container ...", strings.Join(cmd, " "))
+	return cmd, tarDestination
 }
 
 // this funtion simply abstracts out the running of the newly produced image when the --run=true option is provided
@@ -515,61 +572,17 @@ func runContainerDockerRun(container *docker.Container, d *stiDocker, image stri
 	glog.V(2).Infof("\nReceived signal '%s', stopping services...\n", signal)
 }
 
-// this funtion simply abstracts out the first phase of attaching to the container that was originally in line with the RunContainer() method
-func runContainerAttach(attached chan struct{}, container *docker.Container, opts RunContainerOptions, d *stiDocker) *sync.WaitGroup {
-	attachOpts := docker.AttachToContainerOptions{
-		Container: container.ID,
-		Success:   attached,
-		Stream:    true,
-	}
-	if opts.Stdin != nil {
-		attachOpts.InputStream = opts.Stdin
-		attachOpts.Stdin = true
-	}
-	if opts.Stdout != nil {
-		attachOpts.OutputStream = opts.Stdout
-		attachOpts.Stdout = true
-	}
-	if opts.Stderr != nil {
-		attachOpts.ErrorStream = opts.Stderr
-		attachOpts.Stderr = true
-	}
+// RunContainer creates and starts a container using the image specified in opts
+// with the ability to stream input and/or output.
+func (d *stiDocker) RunContainer(opts RunContainerOptions) error {
+	createOpts := opts.asDockerCreateContainerOptions()
 
-	wg := sync.WaitGroup{}
-	go func() {
-		wg.Add(1)
-		defer wg.Done()
-		if err := d.client.AttachToContainer(attachOpts); err != nil {
-			glog.Errorf("Unable to attach container with %v", attachOpts)
-		}
-	}()
-	return &wg
-}
-
-// this funtion simply abstracts out the waiting on the container hosting the builder function that was originally in line with the RunContainer() method
-func runContainerWait(wg *sync.WaitGroup, d *stiDocker, container *docker.Container) error {
-	glog.V(2).Infof("Waiting for container")
-	exitCode, err := d.client.WaitContainer(container.ID)
-	glog.V(2).Infof("Container wait returns with %d and %v\n", exitCode, err)
-
-	wg.Wait()
-	if err != nil {
-		return err
-	}
-
-	glog.V(2).Infof("Container exited")
-	if exitCode != 0 {
-		return errors.NewContainerError(container.Name, exitCode, "")
-	}
-	return nil
-}
-
-// RunContainer creates and starts a container using the image specified in the options with the ability
-// to stream input or output
-func (d *stiDocker) RunContainer(opts RunContainerOptions) (err error) {
 	// get info about the specified image
-	image := getImageName(opts.Image)
-	var imageMetadata *docker.Image
+	image := createOpts.Config.Image
+	var (
+		imageMetadata *docker.Image
+		err           error
+	)
 	if opts.PullImage {
 		imageMetadata, err = d.CheckAndPullImage(image)
 	} else {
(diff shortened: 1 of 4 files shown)


rhcarvalho:
@bparees I'm not done yet, but there're some questions here that we can start looking into the answers.

If all we need is to send stdin and stream logs, I'm confident we don't need any calls to the "Attach" parts of the Docker API.

PTAL at the "TODO" lines with questions.

pkg/docker/docker.go
```diff
@@ -126,6 +126,62 @@ type RunContainerOptions struct {
...
 }
 
+// asDockerConfig converts a RunContainerOptions into a Config understood by the
+// docker client library.
+func (rco RunContainerOptions) asDockerConfig() docker.Config {
+	return docker.Config{
+		Image: getImageName(rco.Image),
+		User:  rco.User,
+		// TODO why do we call runContainerTar with a "partial"
+		// docker.Config?
```
rhcarvalho: I think I know the answer for this one.

The only use of `config` in `runContainerTar` is to set the `config.Cmd` field. We never read the config, so it doesn't matter what values it has.

pkg/docker/docker.go
```diff
@@ -126,6 +126,62 @@ type RunContainerOptions struct {
...
+}
+
+// asDockerHostConfig converts a RunContainerOptions into a HostConfig
+// understood by the docker client library.
+func (rco RunContainerOptions) asDockerHostConfig() docker.HostConfig {
+	hostConfig := docker.HostConfig{
+		CapDrop: rco.CapDrop,
+
+		PublishAllPorts: rco.TargetImage,
+		// TODO why didn't we always set NetworkMode?!
```
rhcarvalho: What does the `config.TargetImage` boolean mean?

Prior code would only set `NetworkMode` if `TargetImage` was true, or if `NetworkMode` is not the empty string (redundant check). If `TargetImage` was false and `NetworkMode` non-empty, it would be set as well... so seems that "we always set it" anyway, but hidden under conditionals...

Still, what does `TargetImage` means?

bparees: the defaults were probably acceptable so we only needed to set it when overriding it.

pkg/docker/docker.go
bparees: runContainerTar doesn't do anything except flesh out the config and return it...

bparees: (probably not a great function name)

pkg/docker/docker.go
```diff
@@ -126,6 +126,62 @@ type RunContainerOptions struct {
...
+}
+
+// asDockerCreateContainerOptions converts a RunContainerOptions into a
+// CreateContainerOptions understood by the docker client library.
+func (rco RunContainerOptions) asDockerCreateContainerOptions() docker.CreateContainerOptions {
+	config := rco.asDockerConfig()
+	hostConfig := rco.asDockerHostConfig()
+	return docker.CreateContainerOptions{
+		// TODO we should probably be setting some S2I-related name
+		// instead.
```
rhcarvalho: @bparees agree we should give the image an s2i related name? That was the decision we made in origin. Empty string means Docker will generate names like "fluffy_panda" :panda_face:

pkg/docker/docker.go
```diff
@@ -126,6 +126,62 @@ type RunContainerOptions struct {
...
+		Image: getImageName(rco.Image),
+		User:  rco.User,
+		// TODO why do we call runContainerTar with a "partial"
+		// docker.Config?
+
+		Env: rco.Env,
+
+		// TODO review how callers use these options to pass stdin.
+		// Probably they are not necessary, as they're related to
+		// attaching to the container (that is often unnecessary).
```
bparees: yeah so basically we need to attach stdin for `assemble` (to stream in the content).  

we need to attach stdout/stderr for `run` and for when we run the resulting image after creating it (so we can see the logs)

presumably we'd want to attach stderr when running `save-artifacts` in case the save-artifacts script uses it.

I don't know what other conditions we might need to worry about (ie the on-build flows and layered build flows)

pkg/docker/docker.go
mfojtik: we will have to deal with name conflicts in that case

pkg/docker/docker.go
rhcarvalho: The defaults are the zero values of the field types, for strings it's the empty string. We can always override when we're building the struct from scratch. So we're fine here.

Do you know what TargetImage actually means, semantically?

It's used to set `PublishAllPorts`, to terminate early `runContainerTar`, and to call `runContainerDockerRun`...

bparees: I think TargetImage is the case where we're trying to run the result image(the target image) after building it.

bparees: that is, `s2i build blah blah blah --run`

pkg/docker/docker.go
bparees: i like pandas.....

yeah, sure, an s2i name would be good, just make sure it's guaranteed unique.

pkg/docker/docker.go
rhcarvalho: > (so we can see the logs)

This part is what I have shared previously when I implemented `runDocker` in origin, we don't need to attach anything to stream logs. It's much simpler to have a single goroutine calling the log api:
https://github.com/openshift/origin/blob/000489d894c9b54b749fbf9246d4d218197a7280/pkg/build/builder/dockerutil.go#L146-L193

> presumably we'd want to attach stderr when running save-artifacts in case the save-artifacts script uses it.

Again, we can stream stderr using the logs api.



> yeah so basically we need to attach stdin for assemble (to stream in the content).

I believe we don't need to "attach" stdin to send content, it should be enough to set the Stdin field when creating the container. Needs confirmation.

pkg/docker/docker.go
rhcarvalho: I'll leave that as a follow up -> https://github.com/openshift/source-to-image/issues/433

pkg/docker/docker.go
```diff
@@ -132,6 +132,55 @@ type RunContainerOptions struct {
 	CapDrop          []string
 }
 
+// asDockerConfig converts a RunContainerOptions into a Config understood by the
+// docker client library.
```
mfojtik: Docker client (it is trademark actually...)

pkg/docker/docker.go
```diff
@@ -132,6 +132,55 @@ type RunContainerOptions struct {
...
+}
+
+// asDockerCreateContainerOptions converts a RunContainerOptions into a
+// CreateContainerOptions understood by the docker client library.
+func (rco RunContainerOptions) asDockerCreateContainerOptions() docker.CreateContainerOptions {
+	config := rco.asDockerConfig()
+	hostConfig := rco.asDockerHostConfig()
+	return docker.CreateContainerOptions{
+		Name:       "",
+		Config:     &config,
```
mfojtik: why not return the pointer from the function? then you don't need these extra vars?

mfojtik:
lgtm so far :-)

pkg/docker/docker.go
rhcarvalho: because it's invalid Go code to do what you're saying :P

pkg/docker/docker.go
rhcarvalho: that's right, thanks. I had in mind "go-dockerclient", therefore wrote in lowercase. Should be either `go-dockerclient` or `Docker client`. I prefer the former as it's more specific to what I mean. Thanks for pointing it out!

pkg/docker/docker.go
```diff
@@ -483,9 +547,10 @@ func runContainerTar(opts RunContainerOptions, config docker.Config, imageMetada
 			cmd = []string{"/bin/sh", "-c", opts.CommandOverrides(strings.Join(cmd[2:], " "))}
 		}
 	}
+	// TODO this message is misleading, as we're not running any container
+	// yet, we have just determined a value for cmd...
```
rhcarvalho: @bparees @mfojtik would you like us to update this message below to something more precise? If so, suggestions on what should it be are welcome.

pkg/docker/docker.go
```diff
@@ -132,6 +131,72 @@ type RunContainerOptions struct {
...
+// go-dockerclient library.
+func (rco RunContainerOptions) asDockerConfig() docker.Config {
+	return docker.Config{
+		Image: getImageName(rco.Image),
+		User:  rco.User,
+		Env:   rco.Env,
+
+		// TODO review how callers use these options to pass stdin.
+		// Probably they are not necessary, as they're related to
+		// attaching to the container (that is often unnecessary).
```
rhcarvalho: These options seem to work for us, I'm tempted to don't do anything about them (and of course drop this comment from the final patch).

rhcarvalho:
@bparees @mfojtik @gabemontero PTAL

Guys, could you help me battle test this?

```console
$ git fetch upstream pull/430/head:pr430
$ git checkout pr430
$ make build
$ s2i build ...
```

I'll drop some commits (drop: TODO) & squash once we're in good shape.

pkg/docker/docker.go
mfojtik: what is invalid about?

```go
func (rco RunContainerOptions) asDockerConfig() *docker.Config {}
```

Then you can just do:

```go
Config:  rco.asDockerConfig(),
```

rhcarvalho: Aha, I thought you meant

```go
{
     Config:     &rco.asDockerConfig(),
     // ...
}
```

:smiley:

rhcarvalho: @mfojtik we could do what you propose for `asDockerConfig` and `asDockerHostConfig`, but then for `asDockerCreateContainerOptions` and `asDockerAttachToContainerOptions` we don't want them to return pointers (for similar reasons in terms of "convenience of use").

That would lead to similar methods some returning pointers, some not.

WDYT of that situation?

pkg/docker/docker.go
mfojtik: "Setting %q command for container" ? It is V(5) so this is debugging only.

pkg/docker/docker.go
bparees: so we don't need to do anything w/ stderr?

pkg/docker/docker.go
bparees: i think we can live w/ the extra local variables :)

bparees: (and add it to the list of "poor decisions golang has made")

pkg/docker/docker.go
bparees: +1 @mfojtik suggestion

pkg/docker/docker.go
```diff
@@ -584,92 +605,97 @@ func (d *stiDocker) RunContainer(opts RunContainerOptions) (err error) {
...
-	glog.V(2).Infof("Attaching to container")
-	// creating / piping the channels in runContainerAttach lead to unintended hangs
-	attached := make(chan struct{})
-	wg := runContainerAttach(attached, container, opts, d)
-	attached <- <-attached
+	// Attach to the container.
+	glog.V(2).Infof("Attaching to container %q ...", containerName)
+	attachOpts := opts.asDockerAttachToContainerOptions()
+	attachOpts.Container = container.ID
+	if _, err = d.client.AttachToContainerNonBlocking(attachOpts); err != nil {
```
bparees: so is this the meat of the change, that we're calling the non-blocking attach now (and not doing it in a goroutine)?  or are the arguments to attach actually different now from what they were?

pkg/docker/docker.go
rhcarvalho: These options here should be the same as they were [before](https://github.com/openshift/source-to-image/pull/430/files#diff-bf07e8f23e3aaca43157f262317e094cL593).

Stderr streaming is setup in [line 189](https://github.com/openshift/source-to-image/pull/430/files#diff-bf07e8f23e3aaca43157f262317e094cR189) and [193](https://github.com/openshift/source-to-image/pull/430/files#diff-bf07e8f23e3aaca43157f262317e094cR193).

pkg/docker/docker.go
rhcarvalho: > (and add it to the list of "poor decisions golang has made")

Just for my own curiosity (and for compiling the list :smile: )...
What's the particular poor decision here? How would it be in an "ideal" language?

pkg/docker/docker.go
rhcarvalho: +1, updating

pkg/docker/docker.go
bparees: func foo() obj {
  return obj{}
}

var a *obj
a=&foo()

Basically all i'm asking for (in this case) is syntactic sugar that keeps me from having to declare the temporary variable.

rhcarvalho: I see... on the curiosity side, do you guys know if is it possible to do that in C++ or Java?

That would be the `&` operator taking the address of the return value of a function call...
Possibly it would be written as `&(foo())` so that there's no ambiguity with `(&foo)()`?

pkg/docker/docker.go
rhcarvalho: There should not be changes to arguments (`attachOpts`).
Indeed, using the non-blocking variant.

I expected to be able to use the first return value (now ignored, a `CloseWaiter` interface type), and call `cw.Wait()` to wait for it to "detach" when the container terminates.
However that's not the effect it had. Documentation in `go-dockerclient` is really really lacking... :(

It turns out that just waiting for the container to terminate is fine (line 661 on).

There might be goroutine leaking from go-dockerclient.

rhcarvalho: > There might be goroutine leaking from go-dockerclient.

Or not. Since we always force remove the container, all the resources related to the attach call should be released.
I haven't checked the implementation of `client.hijack` enough to tell by heart if they have leaks.

rhcarvalho:
I've manually tested:

- [x] regular builds
- [x] incremental builds
- [x] `--run`
- [x] `--inject`

@bparees @mfojtik something else to trigger a different flow?

bparees:
@rhcarvalho on-build images?

pkg/docker/docker.go
rhcarvalho: > There should not be changes to arguments (attachOpts).

Sorry, that was an incorrect statement. I just remembered that there is a difference, now we're not using the [`Success`](https://godoc.org/github.com/fsouza/go-dockerclient#AttachToContainerOptions) field (`chan struct{}`).

The docs say:
```go
    // If set, after a successful connect, a sentinel will be sent and then the
    // client will block on receive before continuing.
    //
    // It must be an unbuffered channel. Using a buffered channel can lead
    // to unexpected behavior.
    Success chan struct{}
```

We don't really need to do anything special right after the connection is established. My understanding is that we were just going through unnecessary hops.

We were [receiving and sending](https://github.com/openshift/source-to-image/pull/430/files#diff-bf07e8f23e3aaca43157f262317e094cL635) on that channel without any intermediary steps.

rhcarvalho:
> @rhcarvalho on-build images?

Do we have any handy?

bparees:
@rhcarvalho I believe the dockerhub golang image includes ONBUILD commands?

rhcarvalho:
@bparees tested:

```console
$ s2i build https://github.com/openshift/golang-ex golang:1.6-onbuild onbuild-test
I0317 23:21:50.214175 28636 docker.go:352] Image "golang:1.6-onbuild" not available locally, pulling ...
W0317 23:23:10.089476 28636 docker.go:474] Image does not contain a value for the io.openshift.s2i.scripts-url label
I0317 23:23:10.091991 28636 clone.go:32] Downloading "https://github.com/openshift/golang-ex" ...
W0317 23:23:16.142024 28636 sti.go:227] Error getting assemble from : script "assemble" not installed
W0317 23:23:16.142501 28636 sti.go:227] Error getting run from : script "run" not installed
Step 1 : FROM golang:1.6-onbuild
# Executing 3 build triggers...
Step 1 : COPY . /go/src/app
Step 1 : RUN go-wrapper download
 ---> Running in effb72e9ddb2
+ exec go get -v -d
github.com/openshift/golang-ex (download)
Step 1 : RUN go-wrapper install
 ---> Running in 1b1e94703da7
+ exec go install -v
<snip>
# app
./main.go:40: cannot use controllers.AppController literal (type *controllers.AppController) as type "app/vendor/github.com/astaxie/beego".ControllerInterface in argument to
 "app/vendor/github.com/astaxie/beego".Router:
        *controllers.AppController does not implement "app/vendor/github.com/astaxie/beego".ControllerInterface (wrong type for Init method)
<snip>
Removing intermediate container 42553ebfdc10
Removing intermediate container effb72e9ddb2
Removing intermediate container 1b1e94703da7
E0317 23:23:46.345915 28636 main.go:336] An error occurred: The command '/bin/sh -c go-wrapper install' returned a non-zero code: 2
```


That's a good test on the failure case. It "works" as I expected -- the way the ONBUILD image works it ends up trying to use upstream beego instead of the vendored code.

rhcarvalho:
@bparees I tried also your repo with official ruby:2.2-onbuild:

```console
$ s2i build https://github.com/bparees/incremental-app ruby:2.2-onbuild incr-test
W0317 23:29:22.116623 29801 docker.go:474] Image does not contain a value for the io.openshift.s2i.scripts-url label
I0317 23:29:22.120667 29801 clone.go:32] Downloading "https://github.com/bparees/incremental-app" ...
I0317 23:29:27.805512 29801 install.go:111] DEPRECATED: Use .s2i/bin instead of .sti/bin
I0317 23:29:27.805659 29801 install.go:236] Using "assemble" installed from "<source-dir>/.sti/bin/assemble"
W0317 23:29:27.805741 29801 sti.go:227] Error getting run from : script "run" not installed
E0317 23:29:27.806534 29801 main.go:336] An error occurred: No valid entrypoint specified
```

I'm not sure that could possibly work...

bparees:
hm, it would be nice if we could have a successful test of on-build :)  But yeah, you've probably proven it's basically working (launched the build, streamed the output, recognized the failure)

rhcarvalho:
@bparees there you go... using the `ruby-hello-world` repo:

```console
$ s2i build https://github.com/openshift/ruby-hello-world ruby:2.2-onbuild incr-test
W0317 23:31:29.260507 29851 docker.go:474] Image does not contain a value for the io.openshift.s2i.scripts-url label
I0317 23:31:29.263253 29851 clone.go:32] Downloading "https://github.com/openshift/ruby-hello-world" ...
W0317 23:31:35.110075 29851 sti.go:227] Error getting assemble from : script "assemble" not installed
W0317 23:31:35.110864 29851 sti.go:227] Error getting run from : script "run" not installed
I0317 23:31:35.111918 29851 environment.go:31] DEPRECATED: Use .s2i/environment instead of .sti/environment
I0317 23:31:35.112724 29851 environment.go:60] Setting 1 environment variables provided by environment file in sources
Step 1 : FROM ruby:2.2-onbuild
# Executing 4 build triggers...
Step 1 : COPY Gemfile /usr/src/app/
Step 1 : COPY Gemfile.lock /usr/src/app/
Step 1 : RUN bundle install
 ---> Running in dd5f232f0991
Fetching gem metadata from https://rubygems.org/..........
Fetching version metadata from https://rubygems.org/..
Installing rake 10.3.2
<snip>
Bundle complete! 5 Gemfile dependencies, 18 gems now installed.
Bundled gems are installed into /usr/local/bundle.
Step 1 : COPY . /usr/src/app
 ---> 1a65d7d5ce92
Removing intermediate container ded5e255622a
Removing intermediate container dd5f232f0991
Removing intermediate container 0a2fa28fd14d
Removing intermediate container e2d322d85b5e
Step 2 : ENV RACK_ENV "production"
 ---> Running in bf58b2806a5c
 ---> 2741747f9088
Removing intermediate container bf58b2806a5c
Step 3 : ENTRYPOINT ./run.sh
 ---> Running in cd98539171e2
 ---> 5e115267eb5d
Removing intermediate container cd98539171e2
Successfully built 5e115267eb5d
$ docker images | head -2
REPOSITORY                                          TAG                 IMAGE ID            CREATED             VIRTUAL SIZE
incr-test                                           latest              5e115267eb5d        21 seconds ago      735 MB
```

bparees:
@rhcarvalho :boom:

bparees:
lgtm, squash it tomorrow and we'll do a final review.

rhcarvalho:
There's no hang, but still a problem (not introduced here, as I've observed it a couple of weeks ago when working on incremental builds), that is read/write on closed pipe error.
I had hoped to improve/fix this problem in https://github.com/openshift/source-to-image/pull/423, but it is still present and I can reproduce it with this:


```console
$ s2i build https://github.com/GrahamDumpleton/wheelhouse-demo grahamdumpleton/warp0-debian8-python27 wheelhouse-demo --incremental
I0318 09:21:13.798999 09109 clone.go:32] Downloading "https://github.com/GrahamDumpleton/wheelhouse-demo" ...
I0318 09:21:15.031510 09109 install.go:236] Using "assemble" installed from "image:///usr/local/s2i/bin/assemble"
I0318 09:21:15.031776 09109 install.go:236] Using "run" installed from "image:///usr/local/s2i/bin/run"
I0318 09:21:15.031816 09109 install.go:236] Using "save-artifacts" installed from "<source-dir>/.s2i/bin/save-artifacts"
E0318 09:21:16.837349 09109 util.go:91]  -----> Running save-artifacts
W0318 09:21:46.181452 09109 sti.go:158] Clean build will be performed because of error saving previous build artifacts
E0318 09:21:46.181704 09109 tar.go:290] Error reading next tar header: io: read/write on closed pipe
---> Installing application source
---> Building application from source
```

@mfojtik @bparees could you help me validate that the [save-artifacts script](https://github.com/GrahamDumpleton/wheelhouse-demo/blob/25a3b48ac67006880338504dda37c66ad0cf6667/.s2i/bin/save-artifacts) is correct/valid?

I don't think this problem should block this PR (we can merge this and follow up), but anyway I want to investigate why Graham's example is not working.

pkg/docker/docker.go
mfojtik: yeah, the goals for me is to keep the amount of temporary vars low (with exception of making the code more readable) and avoid `if` as much as possible :-)

in this case that function servers as generator for the `Config` field which require `*docker.Config`, so having that function return the pointer makes more sense to me ;-)

however, I will still be ok with what you have here today, it just seems odd to always create temp variable to use it :)

mfojtik:
@rhcarvalho that script seems to be correct (i will try to remove the `echo` just to be sure). Otherwise the `tar` command looks OK.

bparees:
@mfojtik @rhcarvalho looks ok to me too, but i'd try removing the echos just in case somehow the output to stderr is either not going to stderr, or is causing us problems elsewhere.

rhcarvalho:
@bparees squashed. PTAL

bparees:
@rhcarvalho did it work w/ wildfly incremental?

rhcarvalho:
> did it work w/ wildfly incremental?

Yes, just finished a successful flow initial-build -> incremental-build.

```
s2i build git://github.com/bparees/openshift-jee-sample openshift/wildfly-100-centos7 wildflytest --incremental
```

bparees:
lgtm [merge]

PR #434, opened 2016-03-17
Bump dockerclient to match version in origin

rhcarvalho:
I'm gonna need code that is in the version in origin and not in the old version we had here (`AttachToContainerNonBlocking`)

Sending as separate PR to not pollute the diff of #430 with all these changes.

@bparees ptal

### DIFF
--- a/Godeps/Godeps.json
+++ b/Godeps/Godeps.json
@@ -67,8 +67,7 @@
 		},
 		{
 			"ImportPath": "github.com/fsouza/go-dockerclient",
-			"Comment": "0.2.1-860-g25bc220",
-			"Rev": "25bc220b299845ae5489fd19bf89c5278864b050"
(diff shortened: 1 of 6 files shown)


bparees:
lgtm [merge]

rhcarvalho:
@bparees I've updated the commit making sure the patch is the same as what's here:
https://github.com/fsouza/go-dockerclient/compare/25bc220b299845ae5489fd19bf89c5278864b050...0099401a7342ad77e71ca9f9a57c5e72fb80f6b2.

Actually not a must-have for my other PR... I thought we were dealing with a relatively older version (coming from the bad assumption of what I've seen last time).

bparees:
312 files down to 6.... do i want to know what happened the first time around? :)

[merge]

rhcarvalho:
> do i want to know what happened the first time around? :)

I want to know too :)

No way to trust `godep`...

PR #436, opened 2016-03-22
Check error before deferring calls that use file

rhcarvalho:
Error checking should be done earlier, and if there's an error creating
a file we should not try to use the file.

Error creating a file is an exceptional condition, can happen for lack
of permissions, full disk, etc. Calling file.Close() on a nil File
pointer would panic.

### DIFF
--- a/pkg/tar/tar.go
+++ b/pkg/tar/tar.go
@@ -356,13 +356,13 @@ func extractFile(dir string, header *tar.Header, tarReader io.Reader) error {
 	glog.V(3).Infof("Creating %s", path)
 
 	file, err := os.Create(path)
+	if err != nil {
+		return err
+	}
 	// The file times need to be modified after it's been closed thus this function
 	// is deferred after the file close (LIFO order for defer)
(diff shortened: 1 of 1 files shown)


rhcarvalho:
@bparees @mfojtik PTAL

rhcarvalho:
or @soltysh?

soltysh:
LGTM [merge]

soltysh:
LGTM [merge]

pkg/tar/tar.go
```diff
@@ -356,13 +356,13 @@ func extractFile(dir string, header *tar.Header, tarReader io.Reader) error {
 	glog.V(3).Infof("Creating %s", path)
 
 	file, err := os.Create(path)
+	if err != nil {
+		return err
+	}
```
bparees: is the file returned from os.Create considered opened?  If so, is there any condition under which os.Create returns valid open file, and an error? (in which case you are not going to close the file with this logic)

rhcarvalho: You either get an open file or an error, never both.

bparees: cool

PR #437, opened 2016-03-22
allow for no source URL (defer download to assemble script)

gabemontero:
@bparees PTAL

These are the source-to-image changes needed for build config with no inputs.

They stem from [this call in origin](https://github.com/openshift/origin/blob/master/pkg/build/builder/sti.go#L208) failing a source strategy build.

Also, I made this change *origin* only, and did not add any options to `s2i`.  The `s2i` user minimally still should point that command to a local directory with scripts.

### DIFF
--- a/pkg/scm/scm.go
+++ b/pkg/scm/scm.go
@@ -6,6 +6,7 @@ import (
 	"github.com/golang/glog"
 
 	"github.com/openshift/source-to-image/pkg/build"
+	"github.com/openshift/source-to-image/pkg/scm/empty"
 	"github.com/openshift/source-to-image/pkg/scm/file"
 	"github.com/openshift/source-to-image/pkg/scm/git"
 	"github.com/openshift/source-to-image/pkg/util"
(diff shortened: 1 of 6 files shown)


pkg/scm/scm.go
```diff
@@ -46,5 +47,9 @@ func DownloaderForSource(s string, forceCopy bool) (build.Downloader, string, er
 		return &git.Clone{g, util.NewFileSystem()}, s, nil
 	}
 
+	if len(s) == 0 {
+		return &empty.Noop{}, s, nil
+	}
```
bparees: should this be at the top of this function instead of the bottom?  not sure what git.ParseFile() does with an empty string, but probably best not to even call it.

gabemontero: Yep you're correct.

gabemontero:
BTW - the one travis failure was a hiccup building travis via `make install-travis` with golang 1.4.

bparees:
i'll fix travis here: https://github.com/openshift/source-to-image/pull/438

pkg/scm/scm.go
gabemontero: Update pushed.

gabemontero: Dammit - forgot to delete the old location .. just a sec.

gabemontero:
OK ... update pushed.

bparees:
lgtm.  note that someone still can't actually use the s2i command line w/o providing a source arg, but this will allow us to address the origin requirement anyway.

[merge]

PR #440, opened 2016-03-23
Fix test messages

rhcarvalho:
- The majority of the changes in this commit is about removing redundant
  trailing '\n', since t.Logf & related methods automatically handle
  trailing new lines, and essentially ignore that extra line we're
  adding.
- Replace several usages of '%s' where '%q' would be more useful.
- Fix cases where the arguments to format the string were incorrectly
  part of the string itself, not arguments.

### DIFF
--- a/pkg/build/strategies/onbuild/onbuild_test.go
+++ b/pkg/build/strategies/onbuild/onbuild_test.go
@@ -2,7 +2,6 @@ package onbuild
 
 import (
 	"bytes"
-	"fmt"
 	"os"
 	"strings"
 	"testing"
@@ -48,15 +47,15 @@ func checkDockerfile(fs *test.FakeFileSystem, t *testing.T) {
 		t.Errorf("%v", fs.WriteFileError)
 	}
 	if fs.WriteFileName != "upload/src/Dockerfile" {
-		t.Errorf("Expected Dockerfile in 'upload/src/Dockerfile', got %v", fs.WriteFileName)
+		t.Errorf("Expected Dockerfile in 'upload/src/Dockerfile', got %q", fs.WriteFileName)
 	}
 	if !strings.Contains(fs.WriteFileContent, `ENTRYPOINT ["./run"]`) {
-		t.Errorf("The Dockerfile does not set correct entrypoint:\n %s\n", fs.WriteFileContent)
+		t.Errorf("The Dockerfile does not set correct entrypoint, file content:\n%s", fs.WriteFileContent)
 	}
 
 	buf := bytes.NewBuffer([]byte(fs.WriteFileContent))
 	if _, err := parser.Parse(buf); err != nil {
-		t.Errorf("cannot parse new Dockerfile: " + err.Error())
+		t.Errorf("cannot parse new Dockerfile: %v", err)
 	}
 
 }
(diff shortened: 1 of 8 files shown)


rhcarvalho:
I was fixing the "\n" issue in one file I'm working on, thought a global search would do better good.

@mfojtik @soltysh PTAL

pkg/build/strategies/onbuild/onbuild_test.go
```diff
@@ -48,15 +48,15 @@ func checkDockerfile(fs *test.FakeFileSystem, t *testing.T) {
 		t.Errorf("%v", fs.WriteFileError)
 	}
 	if fs.WriteFileName != "upload/src/Dockerfile" {
-		t.Errorf("Expected Dockerfile in 'upload/src/Dockerfile', got %v", fs.WriteFileName)
+		t.Errorf("Expected Dockerfile in 'upload/src/Dockerfile', got %q", fs.WriteFileName)
 	}
 	if !strings.Contains(fs.WriteFileContent, `ENTRYPOINT ["./run"]`) {
-		t.Errorf("The Dockerfile does not set correct entrypoint:\n %s\n", fs.WriteFileContent)
```
soltysh: I think newlines here were intentional, for readability, please leave as is.

pkg/build/strategies/onbuild/onbuild_test.go
```diff
@@ -102,7 +102,7 @@ func TestCreateDockerfileWithAssemble(t *testing.T) {
 	}
 	checkDockerfile(fakeFs, t)
 	if !strings.Contains(fakeFs.WriteFileContent, `RUN sh assemble`) {
-		t.Errorf("The Dockerfile does not run assemble:\n%s\n", fakeFs.WriteFileContent)
```
soltysh: Same here.

pkg/scm/git/git_test.go
```diff
@@ -230,9 +230,22 @@ func TestMungeNoProtocolURL(t *testing.T) {
...
+- Fragment: got %v, ok? %v
+- User: got %v`,
+				scenario,
+				uri.Scheme, uri.Scheme == test.Scheme,
+				uri.Opaque, uri.Opaque == test.Opaque,
+				uri.Host, uri.Host == test.Host,
+				uri.Path, uri.Path == test.Path,
+				uri.RawQuery, uri.RawQuery == test.RawQuery,
+				uri.Fragment, uri.Fragment == test.Fragment,
+				uri.User)
```
soltysh: This is bad ass ;) Please split that into multiple lines one for each element, it'll be simpler to read and modify.

soltysh:
3 nits, but otherwise looks good.

pkg/scm/git/git_test.go
mfojtik: maybe would be better to create some helper function to describe this

soltysh: I don't think so, this is just for tests.

rhcarvalho: I don't think anybody was reading the previous string...

The way to go is to replace that with the simpler `reflect.DeepEqual` code once we drop support for Go 1.4. Been there in origin and reverted because of language differences: https://github.com/openshift/origin/pull/6402, https://github.com/openshift/origin/pull/6578.

pkg/build/strategies/onbuild/onbuild_test.go
rhcarvalho: No as is, but updated to `:\n%s`, removing unnecessary spaces and trailing new line.

rhcarvalho:
@soltysh updated, PTAL

soltysh:
LGTM [merge]

PR #441, opened 2016-03-23
Refactor pkg/run

rhcarvalho:
- Update doc comments to match Go's convention
  https://golang.org/doc/effective_go.html
- Reorganize imports to match convention
- Call errors 'err', no need for prefixes

### DIFF
--- a/pkg/run/run.go
+++ b/pkg/run/run.go
@@ -1,49 +1,44 @@
-/*
-Code that supports the running of the image produced by s2i.  Triggered by the --run=true option specified on the command line.
-*/
+// Package run supports running images produced by S2I. It is used by the
+// --run=true command line option.
 package run
 
 import (
+	"io"
+
 	"github.com/golang/glog"
 	"github.com/openshift/source-to-image/pkg/api"
 	"github.com/openshift/source-to-image/pkg/docker"
 	"github.com/openshift/source-to-image/pkg/errors"
-	"io"
 )
 
-// the struct capturing the list of items needed to support the --run=true options;
-// currently, only running docker containers and only need the docker client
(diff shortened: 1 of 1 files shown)


rhcarvalho:
I'll be touching this file for fixing concurrency issues, didn't want to get stylistic changes mixed up with that.

@soltysh @gabemontero PTAL

pkg/run/run.go
```diff
@@ -61,10 +56,10 @@ func (b *DockerRunner) Run(config *api.Config) error {
...
 	go docker.StreamContainerIO(errReader, nil, glog.Error)
 	go docker.StreamContainerIO(outReader, nil, glog.Info)
-	rerr := b.ContainerClient.RunContainer(opts)
-	if e, ok := rerr.(errors.ContainerError); ok {
-		return errors.NewContainerError(config.Tag, e.ErrorCode, errOutput)
-	}
 
+	err := b.ContainerClient.RunContainer(opts)
+	if e, ok := err.(errors.ContainerError); ok {
+		return errors.NewContainerError(config.Tag, e.ErrorCode, "")
```
rhcarvalho: @gabemontero this is a question to you... why do we replace a `ContainerError` with another `ContainerError`?

gabemontero: @rhcarvalho same answer to the other place you asked ... the update the contents of the message

rhcarvalho: (not a required answer to get this merged, but if it turns out to be not needed we can remove it here or in a follow up)

gabemontero: @rhcarvalho now, I see via `git blame` that docker.go has changed more recently than run.go ... it is certainly not inconceivable that this can be revisited

bparees:
lgtm [merge]

PR #442, opened 2016-03-23
Do not ignore errors

rhcarvalho:
We focused on translating one specific error case, and missed returning
any other error, returning nil instead.

Also document why we recreate a ContainerError, and preserve all
original fields except the one we intend to change.


rhcarvalho:
@bparees PTAL, @gabemontero FYI

bparees:
lgtm [merge]

PR #448, opened 2016-03-28
Allow passing a proxy configuration for script download

csrwng:
Allows specifying a proxy configuration for script download in the api.Config


csrwng:
@bparees ptal

bparees:
lgtm [merge]

PR #449, opened 2016-03-28
Throw an error if a scripts URL is specified but no scripts can be downloaded from it

csrwng:


csrwng:
@bparees ptal

pkg/scripts/install.go
```diff
@@ -29,29 +29,41 @@ type ScriptHandler interface {
...
 }
 
-const sourcesRootAbbrev = "<source-dir>/"
+const (
+	sourcesRootAbbrev = "<source-dir>/"
+
+	// ScriptURLHandler is the name of the script URL handler
+	ScriptURLHandler = "script URL handler"
+
+	// ScriptURLHandler is the name of the image URL handler
```
bparees: ImageURLHandler

bparees:
one nit and squash, then lgtm

csrwng:
thx, [merge]

csrwng:
fixed gofmt error

Issue #415, opened 2016-02-23
Generate Bash completion scripts


elyscape:
It would be nice to have Bash completion scripts for the s2i binary. This shouldn't be too difficult, given that [Cobra](https://github.com/spf13/cobra) is used for the CLI interface and [Cobra offers a Bash completion generator](https://github.com/spf13/cobra/blob/master/bash_completions.md).

mfojtik:
@elyscape sorry for this to take so long, I will have a look tomorrow and provide bash completetions (we already have that for oc and openshift commands, s2i should be no different).

elyscape:
No worries. Thanks for your help on this!

mfojtik:
@elyscape it is being merged now... several options:

1) Use `s2i genbashcompletion` will give you up-to-date completion for the current `s2i` binary
2) You can download generated completion (see contrib/bash) folder.

elyscape:
Fantastic. Thanks for your help!

PR #456, opened 2016-04-05
Add genbashcompletion sub-command

mfojtik:
Fixes: https://github.com/openshift/source-to-image/issues/415

@bparees PTAL

The usage here is to run:

`$ s2i genbashcompletion > completion_file.sh`

It is also possible to add `eval $(s2i genbashcompletion)` into `.bashrc` IMHO.
`EDIT:` Added generated completion per @soltysh request :-)

### DIFF
--- a/cmd/s2i/main.go
+++ b/cmd/s2i/main.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"bytes"
 	"flag"
 	"fmt"
 	"os"
@@ -262,6 +263,20 @@ func newCmdCreate() *cobra.Command {
 	}
 }
 
+func newCmdGenBashCompletion(root *cobra.Command) *cobra.Command {
+	return &cobra.Command{
+		Use:   "genbashcompletion",
+		Short: "Generate Bash completetion for the s2i command",
+		Long:  "Generate Bash completetion for the s2i command into standard output",
+		Run: func(cmd *cobra.Command, args []string) {
+			// TODO: The latest version of cobra require to pass the io.Writer here.
+			var out bytes.Buffer
+			root.GenBashCompletion(&out)
+			fmt.Printf(out.String())
+		},
+	}
+}
+
 func newCmdUsage(cfg *api.Config) *cobra.Command {
 	oldScriptsFlag := ""
 	oldDestination := ""
@@ -372,6 +387,7 @@ func main() {
 		glog.Warning("sti binary is deprecated, use s2i instead")
 	}
 
+	stiCmd.AddCommand(newCmdGenBashCompletion(stiCmd))
 	err := stiCmd.Execute()
 	if err != nil {
 		os.Exit(1)
--- a/hack/update-generated-completions.sh
+++ b/hack/update-generated-completions.sh
@@ -1,0 +1,14 @@
+#!/bin/bash
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+STI_ROOT=$(dirname "${BASH_SOURCE}")/..
+
+source "${STI_ROOT}/hack/common.sh"
+
+sti::build::build_binaries "$@"
+
+echo "+++ Updating Bash completion (${STI_ROOT}/contrib/bash/s2i)"
+${STI_LOCAL_BINPATH}/s2i genbashcompletion > ${STI_ROOT}/contrib/bash/s2i
--- a/contrib/bash/s2i
+++ b/contrib/bash/s2i
@@ -1,0 +1,382 @@
+#!/bin/bash
+
+__debug()
+{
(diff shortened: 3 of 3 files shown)


soltysh:
LGTM

mfojtik:
[merge]

cmd/s2i/main.go
```diff
@@ -262,6 +263,20 @@ func newCmdCreate() *cobra.Command {
 	}
 }
 
+func newCmdGenBashCompletion(root *cobra.Command) *cobra.Command {
+	return &cobra.Command{
+		Use:   "genbashcompletion",
+		Short: "Generate Bash completetion for the s2i command",
+		Long:  "Generate Bash completetion for the s2i command into standard output",
```
rhcarvalho: Typo in the two lines above: s/completetion/completion

cmd/s2i/main.go
```diff
@@ -262,6 +263,20 @@ func newCmdCreate() *cobra.Command {
...
+		Run: func(cmd *cobra.Command, args []string) {
+			// TODO: The latest version of cobra require to pass the io.Writer here.
+			var out bytes.Buffer
+			root.GenBashCompletion(&out)
+			fmt.Printf(out.String())
```
rhcarvalho: Couldn't the 3 lines above be simply:

```go
root.GenBashCompletion(os.Stdout)
```

?

mfojtik: @rhcarvalho no, os.Stdout is file pointer, the version we have in godeps require pointer to a bytes buffer.

mfojtik: which is the reason for the todo ;-)

rhcarvalho:
Missing developer docs for the `update-generated-completions.sh` script?
We should mention somewhere that changes to the CLI commands might require updating the generate completions.

Additionally, we could check in Jenkins that the completions match the current state, so that we don't forget to run the script before merging a PR.

mfojtik:
@rhcarvalho will send a follow up.

cmd/s2i/main.go
rhcarvalho: The comment didn't help me understand that `root.GenBashCompletion` has the signature `GenBashCompletion(out *bytes.Buffer)`, I had to go read the source...

Why not bump cobra?

rhcarvalho: Why use `fmt.Printf` if there is no formatting going on? Using `fmt.Print` (or `fmt.Println`) looks more appropriate.

hack/update-generated-completions.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+STI_ROOT=$(dirname "${BASH_SOURCE}")/..
```
rhcarvalho: readonly

hack/update-generated-completions.sh
```diff
@@ -0,0 +1,14 @@
...
+
+source "${STI_ROOT}/hack/common.sh"
+
+sti::build::build_binaries "$@"
+
+echo "+++ Updating Bash completion (${STI_ROOT}/contrib/bash/s2i)"
```
rhcarvalho: The `${STI_ROOT}` part in the message will look weird, something like `/path/to/s2i/hack/../contrib/bash/s2i`.
Not very useful, we could just drop that part and keep only the relative path `contrib/bash/s2i`, or even drop it altogether. WDYT?

mfojtik: i don't really have any strong preference :-) i can remove it, I thought it would be nice to show where the completion was generated for debugging.

hack/update-generated-completions.sh
mfojtik: we don't have this `readonly` in any other script and I'm not going to change that here and not everywhere else.

cmd/s2i/main.go
mfojtik: @rhcarvalho bumping cobra might cause additional issues, we might do it to sync with origin, out of scope for this PR.

hack/update-generated-completions.sh
rhcarvalho: We do:

- https://github.com/openshift/source-to-image/blob/master/hack/test-integration.sh#L7
- https://github.com/openshift/source-to-image/blob/master/hack/build-test-images.sh#L9

rhcarvalho: And we have that in `origin` as well.

mfojtik: @rhcarvalho that is 2 :-) I can do a follow up fixing all of them, but that was not the reason I started this PR :-)

rhcarvalho: Since you are adding new Bash code, I just wanted to share the "good practices".

PR #458, opened 2016-04-05
Follow up for bash completion

mfojtik:


mfojtik:
[test]

mfojtik:
@rhcarvalho fyi.

hack/verify-bash-completion.sh
```diff
@@ -0,0 +1,27 @@
+#!/bin/bash
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+STI_ROOT=$(dirname "${BASH_SOURCE}")/..
```
rhcarvalho: readonly

hack/verify-bash-completion.sh
```diff
@@ -0,0 +1,27 @@
...
+source "${STI_ROOT}/hack/common.sh"
+
+pushd "${STI_ROOT}" >/dev/null
+
+mv contrib/bash/s2i contrib/bash/s2i-proposed
+hack/update-generated-completions.sh
+
+ret=0
+diff -Naupr contrib/bash/s2i contrib/bash/s2i-proposed || ret=$?
```
rhcarvalho: If you put the `diff ...` as the `if` condition, we don't need to manipulate the return code.

```bash
if diff ...; then
   ...
else
   ...
```

hack/verify-bash-completion.sh
```diff
@@ -0,0 +1,27 @@
...
+
+mv contrib/bash/s2i-proposed contrib/bash/s2i
+if [[ $ret -eq 0 ]]
+then
+  echo "SUCCESS: Generated completions up to date."
+else
+  echo "FAILURE: Generated completions out of date. Please run hack/update-generated-completions.sh"
+  exit 1
```
rhcarvalho: If we `exit 1` here, the `popd` command below will not run.

rhcarvalho:
@mfojtik pending other comments I made in the original PR:

- [ ] https://github.com/openshift/source-to-image/pull/456#discussion_r58526370
- [ ] https://github.com/openshift/source-to-image/pull/456#discussion_r58526773
- [ ] https://github.com/openshift/source-to-image/pull/456#discussion_r58525984
- [ ] https://github.com/openshift/source-to-image/pull/456#discussion_r58525859

hack/verify-bash-completion.sh
mfojtik: this is copied from origin ;-)

hack/verify-bash-completion.sh
mfojtik: switched to `cd`.

contrib/bash/s2i
```diff
@@ -103,7 +103,7 @@ __handle_flag()
     local flagname=${words[c]}
     # if the word contained an =
     if [[ ${words[c]} == *"="* ]]; then
-        flagname=${flagname%!=(MISSING)*} # strip everything after the =
+        flagname=${flagname%=*} # strip everything after the =
```
rhcarvalho: Looks like the `Printf` had further consequences interpreting Bash code :)
I had not think that far!

hack/update-generated-completions.sh
```diff
@@ -10,5 +10,5 @@ source "${STI_ROOT}/hack/common.sh"
 
 sti::build::build_binaries "$@"
 
-echo "+++ Updating Bash completion (${STI_ROOT}/contrib/bash/s2i)"
+echo "+++ Updating Bash completion in contrib/bash/s2i)"
```
rhcarvalho: We have a leftover closing paren there ;)

contrib/bash/s2i
mfojtik: yeah, was a good catch.

hack/verify-bash-completion.sh
rhcarvalho: Can we do it in a subshell? IMHO it's bad practice to change the current working directory when you run some script.
Apart from `cd` itself, a developer don't expect that running a command/script will change his working directory.

rhcarvalho:
LGTM thanks @mfojtik 
Good to have this integrated with `make check`.

hack/verify-bash-completion.sh
mfojtik: well we do this in origin (I think the original code I took from origin used `cd`, but I changed it to pushd). I think `cd` is fine here, don't want to overcomplicated this helper script.

mfojtik:
[merge]

PR #460, opened 2016-04-05
make tar stream exclusion regexp configurable

### LINKED ISSUE #457: No .git directory in source dir
My assemble script expects git repo (because my gradle build calls some git commands), not only the source code (without a .git directory).

 I've tried command build with param "--save-temp-dir=true"

```
s2i build --incremental=true --save-temp-dir=true https://github.com/MartinRumanek/kramerius.git kramerius-builder kramerius 
```

and the path /tmp/s2i269878825/upload/src is ok (with .git directory), but inside the builder image directory .git doesn't exist.

```
$ s2i version
s2i v1.0.5-55-gfb6df26
```

gabemontero:
Fixes #457 

@bparees @mfojtik PTAL


pkg/tar/tar_test.go
```diff
@@ -192,6 +193,42 @@ func TestCreateTar(t *testing.T) {
...
+	defer os.RemoveAll(tempDir)
+	if err != nil {
+		t.Fatalf("Cannot create temp directory for test: %v", err)
+	}
+	modificationDate := time.Date(2011, time.March, 5, 23, 30, 1, 0, time.UTC)
+	testFiles := []fileDesc{
+		{"dir01/dir02/test1.txt", modificationDate, 0700, "Test1 file content", false, ""},
+		{"dir01/test2.git", modificationDate, 0660, "Test2 file content", false, ""},
+		{"dir01/dir03/test3.txt", modificationDate, 0444, "Test3 file content", false, ""},
+		{"dir01/.git/hello.txt", modificationDate, 0600, "Allow .git content", false, ""},
```
bparees: maybe throw one file in here that should be skipped w/ this exclusion pattern?

cmd/s2i/main.go
```diff
@@ -171,6 +171,7 @@ $ s2i build . centos/ruby-22-centos7 hello-world-app
 	buildCmd.Flags().StringVarP(&(cfg.Ref), "ref", "r", "", "Specify a ref to check-out")
 	buildCmd.Flags().StringVarP(&(cfg.AssembleUser), "assemble-user", "", "", "Specify the user to run assemble with")
 	buildCmd.Flags().StringVarP(&(cfg.ContextDir), "context-dir", "", "", "Specify the sub-directory inside the repository with the application sources")
+	buildCmd.Flags().StringVarP(&(cfg.ExcludeRegExp), "exclude-regex", "", "", "Regular expression string for selecting which files from the application source tree to exclude from the tar stream")
```
bparees: i think we can just call this "exclude".  How are multiple patterns handled?  comma separated list?  if so the help should indicate that.

gabemontero:
travis failure stemmed from need to run `hack/update-generated-completions.sh` ... have pushed preliminary update, though looks like will need to run again based on comments.

cmd/s2i/main.go
gabemontero: have renamed queued up;

As to the help update, would it be sufficient to point them to the [doc for the go regexp package](https://golang.org/pkg/regexp/)?

As to how to specify multiple patterns, minimally from looking at [defaul exclusion pattern](https://github.com/openshift/source-to-image/blob/master/pkg/tar/tar.go#L27), the separation parameter (or at least one of the valid ones), is `|`

pkg/tar/tar_test.go
gabemontero: ok ... tweaked the regexp to select one of the files to be excluded

gabemontero: ok ... tweaked the regexp to select one of the files to be excluded

gabemontero: ok ... tweaked the regexp to select one of the files to be excluded

cmd/s2i/main.go
```diff
@@ -171,6 +171,7 @@ $ s2i build . centos/ruby-22-centos7 hello-world-app
 	buildCmd.Flags().StringVarP(&(cfg.Ref), "ref", "r", "", "Specify a ref to check-out")
 	buildCmd.Flags().StringVarP(&(cfg.AssembleUser), "assemble-user", "", "", "Specify the user to run assemble with")
 	buildCmd.Flags().StringVarP(&(cfg.ContextDir), "context-dir", "", "", "Specify the sub-directory inside the repository with the application sources")
+	buildCmd.Flags().StringVarP(&(cfg.ExcludeRegExp), "exclude", "", "", "Regular expression for selecting files from the source tree to exclude from the tar stream; see https://golang.org/pkg/regexp for syntax")
```
bparees: users won't know what the tar stream is, how about "Regular expression for selecting files from the source tree to exclude from the build (blah blah syntax)"?

mfojtik: @bparees @gabemontero i don't think users will like to write complex regexp here... maybe an explicit option would be  better... like `--exclude-git`, etc.

bparees: they don't have to write complex regex.  --exclude="" will be the primary use case for users who want to include git (at least that better work :)  if it matches everything that's a problem)

mfojtik: @bparees how about just `--include-git` :-) do we have other use-case where we exclude files?

bparees: no, but the idea here was to kill two birds w/ one stone:
1) give users a way to include .git
2) give users a way to exclude other arbitrary files

given that this seems to be a relatively rare use case, i'm ok with the solution to (1) being slightly more complicated, in exchange for solving both problems, rather than just solving the single problem.  But i do understand your concern and share it...  maybe we need a tie breaker.  @gabemontero ?

gabemontero: hey @bparees @mfojtik - was reading the exchange at the doctor's this AM, but decided not to chime in until I got in front of the computer (that also allowed me to think a bit about it).
1) with the --*-git version, I can't help but circle back to the "what happens when we support other SCMs", and yes there is the "how likely is that" rebuttal, but my gut tells me it is better to be generic here ... though it adds more to-do's (see 2) and 3) )
2) that said, I think the "" use case isn't at the way we want it (since not specifying a string is "" today with golang) ... I believe I need to rework the impl so the actual default regexp is the default value for --exclude, and I update the help to a) avoid the tar stream ref as @bparees noted, but b) add a bit to explain the default excludes .git, but still note the golang ref
3) will add test case to confirm the "" behavior ... if it matches everything, will need to special case it

And yes, at least for now, even with 2) and 3), I'm still of the mind to keep it generic, but understand my footing there may get too problematic.  I'll update in a bit with findings.

gabemontero:
OK - I've pushed the updates per my last comments regarding handling of "".  As an FYI, the golang regexp package will match anything when "" is supplied (hence we would exclude everything).  The special case logic to invert that was trivial though.  The updated help doc to clarify this.

pkg/build/strategies/layered/layered.go
```diff
@@ -35,11 +36,15 @@ func New(config *api.Config, scripts build.ScriptsHandler, overrides build.Overr
 	if err != nil {
 		return nil, err
 	}
+	tarHandler := tar.New()
+	if config.ExcludeRegExp != tar.DefaultExclusionPattern.String() {
```
bparees: this check seems unnecessary, might as well just always call the Set?

gabemontero: I was thinking the check was cheaper than the construction of the regexp, but sure, both are cheap most likely... old habits :-) ... will remove / simplify code

gabemontero:
pushed update that just calls set of the exclusion pattern in both places

gabemontero:
pushed update that just calls set of the exclusion pattern in both places

bparees:
lgtm, will let @mfojtik put in a final protest about --include-git.

mfojtik:
LGTM (let my protest be recorded :-)

[merge]

gabemontero:
Absolutely:-).  Thx

Issue #416, opened 2016-02-25
Provide ability to mount volumes during builds


elyscape:
There are scenarios where it would be useful to be able to mount a host directory as a volume in the builder that aren't served by [file/directory injection](https://github.com/openshift/source-to-image/pull/380). For example: npm [supports using Git URLs for dependencies](https://docs.npmjs.com/files/package.json#git-urls-as-dependencies); if a private repository is specified, some sort of authentication needs to be provided. The easiest way to do do that without hardcoding credentials in the package.json file (which is, for obvious reasons, a bad idea) is to invoke SSH agent and [mount the socket inside the volume](https://gist.github.com/d11wtq/8699521). That is not currently possible with s2i.

PR #461, opened 2016-04-07
Add support for bind mounts

mfojtik:
Fixes: https://github.com/openshift/source-to-image/issues/416

This PR add ability to define [bind mounts](https://docs.docker.com/engine/userguide/containers/dockervolumes/#adding-a-data-volume) for container where the assemble runs.

Note that this is different from "inject", because specifying a bind mount will replace the content of the directory inside the container, where "inject" will only add/overwrite the files.

One cool side-effect here is that you are able to "build" sources using S2I, but you don't have to care about the output image (allows to build jars, binary artifacts, etc...)

Example:

```console
$ s2i build --loglevel=5 -v $(pwd):/opt/app-root/src . centos/ruby-22-centos7 hello-world-app
```

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -88,7 +88,8 @@ that image and add them to the tar streamed to the container into `/artifacts`.
 | `-s (--scripts-url)`       | URL of S2I scripts (see [S2I Scripts](https://github.com/openshift/source-to-image/blob/master/docs/builder_image.md#s2i-scripts)) |
 | `--recursive`              | Perform recursive git clone when getting sources using git|
 | `-q (--quiet)`             | Operate quietly, suppressing all non-error output |
-| `--inject`                 | Inject the content of the specified directory into the path in the container that runs the assemble script |
+| `-i (--inject)`            | Inject the content of the specified directory into the path in the container that runs the assemble script |
+| `-v (--volume)`            | Bind mounts a local directory into the container that runs the assemble script|
 
 #### Context directory
 
--- a/pkg/api/types.go
+++ b/pkg/api/types.go
@@ -177,7 +177,7 @@ type Config struct {
 	// Injections specifies a list source/destination folders that are injected to
 	// the container that runs assemble.
 	// All files we inject will be truncated after the assemble script finishes.
-	Injections InjectionList
+	Injections VolumeList
 
 	// CGroupLimits describes the cgroups limits that will be applied to any containers
 	// run by s2i.
@@ -202,6 +202,10 @@ type Config struct {
 
 	// HasOnBuild will be set to true if the builder image contains ONBUILD instructions
 	HasOnBuild bool
+
+	// BuildVolumes specifies a list of volumes to mount to container running the
+	// build.
+	BuildVolumes VolumeList
 }
 
 // EnvironmentSpec specifies a single environment variable.
@@ -226,14 +230,14 @@ type CGroupLimits struct {
 	MemorySwap       int64
 }
 
-// InjectPath contains definition of source directory and the injection path.
-type InjectPath struct {
-	SourcePath     string
-	DestinationDir string
+// Volume represents a single volume mount point
+type VolumeSpec struct {
+	Source      string
+	Destination string
 }
 
-// InjectionList contains list of InjectPath.
-type InjectionList []InjectPath
+// VolumeList contains list of VolumeSpec.
+type VolumeList []VolumeSpec
 
 // DockerConfig contains the configuration for a Docker connection.
 type DockerConfig struct {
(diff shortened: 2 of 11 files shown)


mfojtik:
@bparees @rhcarvalho PTAL

mfojtik:
@elyscape FYI

rhcarvalho:
@mfojtik this is just for the `s2i` tool, and won't be used by OpenShift, right?

mfojtik:
@rhcarvalho correct.

also missing tests (working on them)

rhcarvalho:
so far lgtm :) nice, code looks simpler than what we need to support the `--run` flag :)

mfojtik:
@rhcarvalho it is really just passthrough for Binds option to Docker ;-) And I can re-use the injection parsing as they use the same format.

mfojtik:
[test]

mfojtik:
@csrwng I was thinking if we can make this work for Origin and one idea I've got is that we can use `VolumeFrom` and use the builder container as a source for volumes. That means, you can specify volumes in the BuildConfig, which gets mounted into the builder pod and then S2I will mount the volumes from it (iow. container<->container volumes).

csrwng:
@mfojtik yup, that would be doable. I like it :-) I guess we'd need to understand if that gives us any additional security exposure.

docs/cli.md
```diff
@@ -88,7 +88,8 @@ that image and add them to the tar streamed to the container into `/artifacts`.
 | `-s (--scripts-url)`       | URL of S2I scripts (see [S2I Scripts](https://github.com/openshift/source-to-image/blob/master/docs/builder_image.md#s2i-scripts)) |
 | `--recursive`              | Perform recursive git clone when getting sources using git|
 | `-q (--quiet)`             | Operate quietly, suppressing all non-error output |
-| `--inject`                 | Inject the content of the specified directory into the path in the container that runs the assemble script |
+| `-i (--inject)`            | Inject the content of the specified directory into the path in the container that runs the assemble script |
+| `-v (--volume)`            | Bind mounts a local directory into a container that run the assemble script|
```
bparees: into the container that runs the assemble script

pkg/util/injection.go
```diff
@@ -39,24 +39,24 @@ func FixInjectionsWithRelativePath(workdir string, injections *api.InjectionList
...
 	result := []string{}
 	for _, s := range injections {
-		info, err := os.Stat(s.SourcePath)
+		info, err := os.Stat(s.Source)
 		if err != nil {
 			return nil, err
 		}
 		if !info.IsDir() {
-			return nil, fmt.Errorf("the %q must be a valid directory", s.SourcePath)
+			return nil, fmt.Errorf("the %q must be a valid directory", s.Source)
```
bparees: I thought the destination had to be a dir, why does the source have to be a dir?

mfojtik: good catch

mfojtik:
@bparees fixed.

pkg/util/injection.go
```diff
@@ -39,24 +39,24 @@ func FixInjectionsWithRelativePath(workdir string, injections *api.InjectionList
...
 	result := []string{}
 	for _, s := range injections {
-		info, err := os.Stat(s.SourcePath)
+		info, err := os.Stat(s.Source)
 		if err != nil {
 			return nil, err
 		}
 		if !info.IsDir() {
-			return nil, fmt.Errorf("the %q must be a valid directory", s.SourcePath)
+			return nil, fmt.Errorf("the %q must be a valid file or directory", s.Source)
```
bparees: you're still enforcing that the file itself be a dir, which seems wrong.  (at least before the error message matched the check)

mfojtik: @bparees I think that is right, we never supported injecting a single file, it always must be a directory... i will revisit the approach here, I think there was a reason for that, but I can't remember now :-)

elyscape: It looks like either the message should be changed back or this `if` block should be removed entirely.

mfojtik: I believe the check can be removed entirely, will run some tests to confirm it.

mfojtik:
re [test] (ec2 is not having its great day today)

mfojtik:
@bparees merge?

pkg/api/types.go
```diff
@@ -413,25 +417,25 @@ func (il *InjectionList) Set(value string) error {
 	default:
 		return fmt.Errorf("invalid source:path definition")
 	}
-	s := InjectPath{SourcePath: filepath.Clean(mount[0]), DestinationDir: filepath.Clean(mount[1])}
-	if IsInvalidFilename(s.SourcePath) || IsInvalidFilename(s.DestinationDir) {
-		return fmt.Errorf("invalid characters in filename: %q", value)
+	s := VolumeSpec{Source: filepath.Clean(mount[0]), Destination: filepath.Clean(mount[1])}
+	if IsInvalidFilename(s.Source) || IsInvalidFilename(s.Destination) {
+		return fmt.Errorf("invalid characters in flename: %q", value)
```
bparees: filename.

bparees:
one nit and then lgtm.

mfojtik:
@bparees fixed, waiting for tests to pass and will merge.

mfojtik:
i think @bparees is breaking me on purpose :-))

mfojtik:
[merge]

PR #462, opened 2016-04-07
Fix govet

mfojtik:
See: https://github.com/openshift/origin/pull/8398

### DIFF
--- a/hack/install-tools.sh
+++ b/hack/install-tools.sh
@@ -8,4 +8,4 @@ source "${STI_ROOT}/hack/common.sh"
 GO_VERSION=($(go version))
 echo "Detected go version: $(go version)"
 
-go get golang.org/x/tools/cmd/cover github.com/golang/lint/golint golang.org/x/tools/cmd/vet
+go get golang.org/x/tools/cmd/cover github.com/tools/godep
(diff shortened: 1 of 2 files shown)


soltysh:
LGTM

mfojtik:
[merge]

hack/verify-govet.sh
```diff
@@ -21,8 +21,7 @@ FAILURE=false
 test_dirs=$(find_files | cut -d '/' -f 1-2 | sort -u)
 for test_dir in $test_dirs
 do
-  go tool vet -shadow=false \
-              $test_dir
+  go vet -shadow=false $test_dir
```
kargakis: It doesn't seem that `go vet` can use its flags - you still need to use `go tool vet`. Also -shadow is by default false.

kargakis: Meaning you can drop the flag and still use `go vet`?

mfojtik: sure :-)

hack/install-tools.sh
```diff
@@ -8,4 +8,4 @@ source "${STI_ROOT}/hack/common.sh"
 GO_VERSION=($(go version))
 echo "Detected go version: $(go version)"
 
-go get golang.org/x/tools/cmd/cover github.com/golang/lint/golint golang.org/x/tools/cmd/vet
+go get golang.org/x/tools/cmd/cover github.com/tools/godep
```
rhcarvalho: Why was golint removed?

Issue #417, opened 2016-02-29
The --env option should be allowed to be used more than once.


GrahamDumpleton:
At present the ``--env`` option has that if wanting to set more than one variable, they need to be squashed into the one string and ``--env`` supplied only once. If you supply ``--env`` more than once it results in anything attempted to be set by prior ``--env`` being dropped and ignored

With ``.s2i/bin/assemble`` containing:

```
#!/bin/bash

env
```

and grabbing a handy S2I builder and running:

```
$ s2i build . grahamdumpleton/warp0-debian8-python27 xxx --env VARIABLE1=XXXXXXXX --env VARIABLE2=YYYYYYYYY | grep VARIABLE
I0229 20:40:25.623001 04019 install.go:236] Using "assemble" installed from "<source-dir>/.s2i/bin/assemble"
I0229 20:40:25.623144 04019 install.go:236] Using "run" installed from "image:///usr/local/s2i/bin/run"
I0229 20:40:25.623177 04019 install.go:236] Using "save-artifacts" installed from "image:///usr/local/s2i/bin/save-artifacts"
VARIABLE2=YYYYYYYYY
```

you can see that only the variable set by the last ``--env`` is passed through.

A related issue is that the fact that multiple values have to be passed in with one ``--env`` means you can't have values for variables which include a comma.

In fact attempting to do so triggers a bug in s2i, with it exiting straight away before doing anything and not giving any sort of message as to what was wrong.

```
$ s2i build . grahamdumpleton/warp0-debian8-python27 xxx --env VARIABLE1=XXXXXXXX --env VARIABLE2="YYYYYYYYY,ZZZZZ"
```

It just exits and dies and gives back prompt straight away.

Another parsing stuff is where you have '=' in the value.

```
$ s2i build . grahamdumpleton/warp0-debian8-python27 xxx --env VARIABLE1=XXXXXXXX --env VARIABLE2="=YYYYYYYYY"
```

Again also just exits straight away without a message.

It is going to be hard to fix this broken behaviour with embedded '=' and ',' know because of reliance on squashing more than one variable together, rather than using multiple ``--env`` options.

This issue likely relates to https://github.com/openshift/origin/issues/7326. So probably whole host of characters in values that are not allowed including quotes as well like in other issue.

mfojtik:
agree, this is a bug. The env should be list. @GrahamDumpleton Have you tried:

`s2i build . foo xxx --env FOO=bar,BLA=foo` ?

[...]

PR #463, opened 2016-04-07
Use cobra for parsing multiple env variables

mfojtik:
Fixes: https://github.com/openshift/source-to-image/issues/417

@bparees this one is tricky :-) Basically, you can specify a multiple environment variables, by providing multiple `-e FOO=BAR` flags. The syntax we supported before was that you can use something like: `-e FOO=BAR,TEST=FOO`. 

So this is introducing a backward incompatible change. I wonder if we want to make it backward compatible (basically do additional string.Split() in the `Set()` function) or leave it as this (which is consistent with other flags).

### DIFF
--- a/pkg/api/types.go
+++ b/pkg/api/types.go
@@ -7,6 +7,7 @@ import (
 	"strings"
 
 	docker "github.com/fsouza/go-dockerclient"
+	"github.com/golang/glog"
 
 	"github.com/openshift/source-to-image/pkg/util/user"
 )
@@ -117,7 +118,7 @@ type Config struct {
 	RemovePreviousImage bool
 
 	// Environment is a map of environment variables to be passed to the image.
-	Environment map[string]string
+	Environment EnvironmentList
 
 	// EnvironmentFile provides the path to a file with list of environment
 	// variables.
@@ -198,6 +199,15 @@ type Config struct {
 	// the container for images that don't have a tar binary available.
 	DisableImplicitBuild bool
 }
+
+// EnvironmentSpec specifies a single environment variable.
+type EnvironmentSpec struct {
+	Name  string
+	Value string
+}
+
+// EnvironmentList contains list of environment variables.
+type EnvironmentList []EnvironmentSpec
 
 type ProxyConfig struct {
 	HTTPProxy  *url.URL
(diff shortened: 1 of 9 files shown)


mfojtik:
@GrahamDumpleton FYI

mfojtik:
[test]

bparees:
bleah.  I don't like our options here :)

but yes, maybe continuing to support comma as well is the best thing to do.  though it means you can't have a comma in your env variable value still, right?

mfojtik:
@bparees yes that was always the case... you can workaround that by using `-e FOO=bar, -e BOO=far`

mfojtik:
@bparees updated :-)

pkg/api/types.go
```diff
@@ -415,3 +425,33 @@ func (il *InjectionList) String() string {
 func (il *InjectionList) Type() string {
 	return "string"
 }
+
+// Set implements the Set() function of pflags.Value interface.
+func (e *EnvironmentList) Set(value string) error {
+	parts := strings.SplitN(value, "=", 2)
+	if len(parts) != 2 || len(parts[0]) == 0 {
+		return fmt.Errorf("invalid environment format %q, must be NAME=VALUE", value)
+	}
```
bparees: doesn't this prevent me from having an env variable who's value includes an = sign?

MYVAR=FOO=2

where i want "FOO=2" to be the value?

bparees: oh nm, SplitN.

pkg/build/strategies/sti/sti_test.go
```diff
@@ -735,10 +735,10 @@ func equalArrayContents(a []string, b []string) bool {
...
 func TestGenerateConfigEnv(t *testing.T) {
 	rh := newFakeSTI(&FakeSTI{})
-	testEnv := map[string]string{
-		"Key1": "Value1",
-		"Key2": "Value2",
-		"Key3": "Value3",
+	testEnv := api.EnvironmentList{
+		{Name: "Key1", Value: "Value1"},
+		{Name: "Key2", Value: "Value2"},
+		{Name: "Key3", Value: "Value3"},
```
bparees: add a value that contains a comma, and a value that contains an =

mfojtik: added, thanks

mfojtik:
@bparees bump

bparees:
[merge]

PR #464, opened 2016-04-11
Improve signal handling util package

rhcarvalho:
- Move code to a separate package, mirroring upstream, and separating
  code out of the generic util package.
- Default to os.Exit(2) instead of os.Exit(0), matching the default
  behavior of the Go runtime.
- Unify list of termination signals across supported platforms.


rhcarvalho:
@bparees PTAL

bparees:
waiting on resolution of:
https://github.com/kubernetes/kubernetes/pull/23865
https://github.com/kubernetes/kubernetes/pull/23752
https://github.com/kubernetes/kubernetes/pull/23643

bparees:
related PRs have merged. [merge]

Issue #467, opened 2016-04-12
index.html no such file error in instructions on "Creating a basic S2I builder image"


alberttwong:
Hmm... another error I ran into.

```
[root@localhost wip]# s2i build test/ nginx-centos7 nginx-app
I0411 21:27:10.379144 04815 install.go:251] Using "assemble" installed from "image:///usr/libexec/s2i/assemble"
I0411 21:27:10.379206 04815 install.go:251] Using "run" installed from "image:///usr/libexec/s2i/run"
I0411 21:27:10.379219 04815 install.go:251] Using "save-artifacts" installed from "image:///usr/libexec/s2i/save-artifacts"
---> Building and installing application from source...
E0411 21:27:10.651792 04815 util.go:91] cp: cannot stat '/tmp/src/test/index.html': No such file or directory
E0411 21:27:12.349268 04815 main.go:357] An error occurred: non-zero (13) exit code from nginx-centos7
```

bparees:
i believe that is output from your assemble script.  running with --loglevel=5 might reveal more.

alberttwong:
```
[root@localhost wip]# s2i build test/ nginx-centos7 nginx-app
I0411 21:27:10.379144 04815 install.go:251] Using "assemble" installed from "image:///usr/libexec/s2i/assemble"
I0411 21:27:10.379206 04815 install.go:251] Using "run" installed from "image:///usr/libexec/s2i/run"
I0411 21:27:10.379219 04815 install.go:251] Using "save-artifacts" installed from "image:///usr/libexec/s2i/save-artifacts"
---> Building and installing application from source...
E0411 21:27:10.651792 04815 util.go:91] cp: cannot stat '/tmp/src/test/index.html': No such file or directory
E0411 21:27:12.349268 04815 main.go:357] An error occurred: non-zero (13) exit code from nginx-centos7
[root@localhost wip]# s2i build test/ nginx-centos7 nginx-app --loglevel=5
I0411 21:30:43.884068 04882 main.go:62] Running S2I version "v1.0.6"
I0411 21:30:43.887026 04882 docker.go:356] Using locally available image "nginx-centos7:latest"

Builder Name:			Nginx 1.6.3
Builder Image:			nginx-centos7
Source:				test/
Output Image Tag:		nginx-app
Incremental Build:		disabled
Remove Old Build:		disabled
Builder Pull Policy:		if-not-present
Previous Image Pull Policy:	if-not-present
Quiet:				disabled
[...]

PR #468, opened 2016-04-12
Update assemble

alberttwong:
fixed where the copy from location of source code `cp -Rf /tmp/src/index.html /usr/share/nginx/html/index.html`

Fix for https://github.com/openshift/source-to-image/issues/467


bparees:
lgtm
[merge]

PR #471, opened 2016-04-21
Handle user:group specification in Dockerfile

csrwng:
When checking whether a user is in the allowed range, a USER directive with uid:group is now allowed.

### DIFF
--- a/pkg/docker/util.go
+++ b/pkg/docker/util.go
@@ -1,13 +1,13 @@
 package docker
 
 import (
+	"bufio"
 	"fmt"
 	"io"
 	"os"
 	"path/filepath"
+	"regexp"
 	"strings"
-
-	"bufio"
 
 	client "github.com/fsouza/go-dockerclient"
 	"github.com/golang/glog"
@@ -202,10 +202,11 @@ func CheckAllowedUser(d Docker, imageName string, uids user.RangeList, isOnbuild
 	if uids == nil || uids.Empty() {
 		return nil
 	}
-	imageUser, err := d.GetImageUser(imageName)
+	imageUserSpec, err := d.GetImageUser(imageName)
 	if err != nil {
 		return err
 	}
+	imageUser := extractUser(imageUserSpec)
 	if !user.IsUserAllowed(imageUser, &uids) {
 		return errors.NewBuilderUserNotAllowedError(imageName, false)
 	}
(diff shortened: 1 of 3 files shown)


csrwng:
@bparees ptal

pkg/docker/util.go
```diff
@@ -214,13 +215,40 @@ func CheckAllowedUser(d Docker, imageName string, uids user.RangeList, isOnbuild
...
+// IsOnbuildAllowed checks a list of Docker ONBUILD instructions for
+// user directives. It ensures that any users specified by the directives
+// falls within the specified range list of users.
+func isOnbuildAllowed(directives []string, allowed *user.RangeList) bool {
+	for _, line := range directives {
+		parts := dockerLineDelim.Split(line, 2)
+		if strings.ToLower(parts[0]) != "user" {
+			continue
+		}
+		uname := extractUser(parts[1])
```
mfojtik: is this also trimming the uname?

csrwng: it's not... it should. Will fix

csrwng: fixed

bparees:
lgtm.

pkg/docker/util_test.go
```diff
@@ -78,6 +78,32 @@ func TestCheckAllowedUser(t *testing.T) {
 			onbuild:     []string{"USER 501", "VOLUME /data"},
 			expectErr:   false,
 		},
+		{
+			name:        "AllowedUIDs is set, numeric user with group",
+			allowedUIDs: rangeList("1-"),
+			user:        "5:5000",
+			expectErr:   false,
```
GrahamDumpleton: Should add a test for:

```
5:users
```

That is, named group if want to ensure coverage of all cases.

mfojtik:
LGTM, pending fixing @GrahamDumpleton check

GrahamDumpleton:
Hold off merging this. As per comments on original issue, this may be pointless as deployer may not use the group anyway. If that is the case, need to work if deployer should honour the group in Docker image data.

csrwng:
@GrahamDumpleton I believe we still need the fix for s2i not to reject images unnecessarily. We're simply not interpreting the user spec in a Dockerfile correctly right now.

pkg/docker/util_test.go
csrwng: added, thx

bparees:
@csrwng +1, i think this fix is as good as it can get from an s2i perspective.  lgtm.

GrahamDumpleton:
I have created https://github.com/openshift/origin/issues/8618 for wider issue of changing Kubernetes/OpenShift to honour the group. This change in this patch is still useful in context of standalone S2I usage, but images produced still aren't going to necessarily run properly on OpenShift.

bparees:
@csrwng @GrahamDumpleton are we in agreement that this PR is as complete as it's going to get?

csrwng:
yes

bparees:
[merge]

Issue #475, opened 2016-04-27
ENTRYPOINT in S2I builder image can interfere with assemble run.


GrahamDumpleton:
If one is making an S2I builder by extending an existing Docker image, if the base image has an existing `ENTRYPOINT` specified (and which has to be retained to allow direct use of the image still), this may interfere with the `assemble` phase and specifically the ability to send files into the container for `assemble` using `tar`.

An example of how it can be broken is if the `ENTRYPOINT` command took the `CMD` value and used it to the configure some sort of mini supervisor and then ran the supervisor, which in turn ran the command. This could die because of the expectation that what is being run is a long running demonised process, or simply because the `CMD` would effectively be run in background, it would have its standard input closed, which for the case of `tar` means it could not read the files being sent in.

To be safe, `s2i` when running the `assemble` phase should override any existing `ENTRYPOINT`.

When running `docker run`, if it has an existing `ENTRYPOINT` you can't from memory just supply `--entrypoint ''` as it will actually try and run the supplied argument (empty string) still.

It may therefore be necessary to use `--entrypoint env` as a way of chaining the execution of the command you actually want run.

bparees:
@mfojtik can you remember any reason we aren't doing this today when we run the assemble container? i'm a bit surprise we aren't.

mfojtik:
@bparees i thought we already override entrypoint to /bin/sh, if we don't then it is a bug. I will have a look.

GrahamDumpleton:
The only way I could solve my problem was to comment out ``ENTRYPOINT`` in the ``Dockerfile``, so doesn't appear to be overridden.

mfojtik:
@GrahamDumpleton thanks! I will send a PR to fix this today.

bparees:
thanks @mfojtik!

bparees:
i had to revert this here https://github.com/openshift/source-to-image/pull/482

because we were committing images with "env" as the entrypoint which resulted in our post-commit tests failing because they couldn't find bundle because the scl enablement stuff didn't happen with that entrypoint.

we can revisit this after the merge queue is fixed, but basically we need to not commit the changed entrypoint.

GrahamDumpleton:
Using ``/bin/env`` on this was wrong choice. Should have been ``/usr/bin/env``. There is no ``/bin/env`` on Debian systems. Generally ``/usr/bin/env`` exists everywhere though. See #559

bparees:
[...]

PR #476, opened 2016-04-27
Set entrypoint to /bin/env

mfojtik:
Fixes: https://github.com/openshift/source-to-image/issues/475

### DIFF
--- a/pkg/build/strategies/sti/sti.go
+++ b/pkg/build/strategies/sti/sti.go
@@ -417,6 +417,7 @@ func (b *STI) Save(config *api.Config) (err error) {
 	opts := dockerpkg.RunContainerOptions{
 		Image:           image,
 		User:            user,
+		Entrypoint:      []string{"/bin/env"},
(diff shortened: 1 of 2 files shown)


mfojtik:
[test]

@GrahamDumpleton @bparees PTAL

bparees:
lgtm [merge]

(we'll want to do a bump to origin of course)

mfojtik:
@bparees will wait for https://github.com/openshift/source-to-image/pull/471 and then do a bump :-) we can do some minor release as well.

soltysh:
@mfojtik out of curiosity can't we set entrypoint to be run, but only when one is set? I dislike the `/bin/env`, somehow...

mfojtik:
@soltysh we can't as we chain commands there (we pipe tar, doing magic with injections, etc..) I tried /bin/sh and other things but /bin/env works the best...

soltysh:
@mfojtik thx :smile:

PR #480, opened 2016-04-29
Resave godeps

rhcarvalho:
I ran `godep save ./...` and found some leftovers in `Godeps/_workspace` that are either not dependencies anymore, or are indirect dependencies already vendored within go-dockerclient.

--

* code.google.com/p/go/src/pkg/archive/tar: no occurences of this import
  path in the source code, only references to archive/tar (stdlib)
* github.com/docker/docker/pkg/units: not imported anywhere

For all others, they're vendored with import path rewrite in
github.com/fsouza/go-dockerclient, namely:

* github.com/docker/docker/pkg/archive
* github.com/docker/docker/pkg/fileutils
* github.com/docker/docker/pkg/ioutils
* github.com/docker/docker/pkg/pools
* github.com/docker/docker/pkg/promise
* github.com/docker/docker/pkg/system

--

* github.com/Sirupsen/logrus: source-to-image does not depend directly on this package. The same repo is vendored with rewritten imports in github.com/fsouza/go-dockerclient/external/github.com/Sirupsen/logrus.

--

* github.com/fsouza/go-dockerclient/external/*: all packages here were missing from Godeps.json

PR #481, opened 2016-04-29
Bump github.com/docker/docker

### LINKED ISSUE #398: Bump vendored docker to match origin
If we bump it to https://github.com/docker/docker/commit/2b27fe17a1b3fb8472fde96d768fa70996adf201, we can get rid of [some code](https://github.com/openshift/origin/blob/90ef4d7c23a5b942e01d85411f3345e904bb5bc2/Godeps/_workspace/src/github.com/openshift/source-to-image/pkg/docker/util.go#L148-L163) that [is implemented upstream](https://github.com/docker/docker/blob/2b27fe17a1b3fb8472fde96d768fa70996adf201/pkg/parsers/parsers.go#L71-L89).

I would like to do that, given some guidelines on how we use Godeps and knowing whether we're carrying any patches for docker/docker in S2I.

rhcarvalho:
Closes #398.

Eventually, the code I pointed out in #398 that could be removed has a slightly different API upstream, not worth adding the extra dependency (whole package for a simple function that we already have).

This makes the version of `github.com/docker/docker/*` to match the version in origin.

The files in `github.com/docker/docker/builder/parser/testfiles*` are being removed because godep does not include files outside of Go packages. Those files are not need/used by us.

We might manually want to keep them if we want to be able to run parser tests using the vendored code. I can update the commit if that's the case.

---

The first 3 commits are part of #480 that I'm building on -- will rebase once those get merged.


rhcarvalho:
@bparees PTAL / opinion on `docker/parser` test files

bparees:
i'm good w/ it as is, thanks!
[merge]

PR #483, opened 2016-05-03
pkg/build/strategies: rename b to builder

php-coder:
Rename `b` to more meaningful name `builder`.

PTAL @mfojtik @bparees
CC @PI-Victor


bparees:
[merge]

PR #484, opened 2016-05-04
when onbuild is blocked, tolerate onbuild images but don't docker build them

bparees:


bparees:
[test]

bparees:
@mfojtik ptal

pkg/build/strategies/layered/layered.go
```diff
@@ -126,6 +127,10 @@ func (builder *Layered) SourceTar(config *api.Config) (io.ReadCloser, error) {
 
 //Build handles the `docker build` equivalent execution, returning the success/failure details
 func (builder *Layered) Build(config *api.Config) (*api.Result, error) {
+	if config.HasOnBuild && config.BlockOnBuild {
+		return nil, fmt.Errorf("builder image uses ONBUILD instructions but ONBUILD is blocked.")
```
mfojtik: "is blocked" sounds pretty hard to me :-) I would say "is not allowed" ?

pkg/api/types.go
```diff
@@ -199,6 +199,9 @@ type Config struct {
 	// the container for images that don't have a tar binary available, if the
 	// image contains ONBUILD commands that would be executed.
 	BlockOnBuild bool
+
+	// HasOnBuild will be set to true if the builder image contains ONBUILD instructions
+	HasOnBuild bool
```
mfojtik: I think in the future we should stop bloating the Config and refactor the code ;-) But I'm OK wit this for now.

mfojtik:
@bparees one rephrasing suggestion, otherwise LGTM

pkg/build/strategies/layered/layered.go
bparees: updated.

pkg/api/types.go
bparees: yeah, i didn't feel great about it, but it's sort of part of the config :)

bparees:
[merge]

PR #485, opened 2016-05-04
README.md: Add some extra backtick quotation

cgwalters:
I think it looks better to quote literals.


bparees:
lgtm, thanks @cgwalters 
[merge]

PR #486, opened 2016-05-04
report actual error on commit failure

bparees:


bparees:
@csrwng @mfojtik ptal, we were masking the actual commit error since we don't expose the Details up the stack anywhere, which makes this issue impossible to debug:
https://github.com/openshift/origin/issues/8737

csrwng:
LGTM

bparees:
[merge]

mfojtik:
lgtm, nice catch

PR #490, opened 2016-05-13
Add commit.author label only when author is non-empty.

php-coder:
When we're using `--copy` we lose meta information from git.
Prior this change we put label `"io.openshift.s2i.build.commit.author": " <>"` to resulted image.

PTAL @bparees


bparees:
lgtm
[merge]

PR #491, opened 2016-05-15
Fix globbing in hack/rebase-to-origin.sh

rhcarvalho:
Previous version never expanded the * and consequently never deleted
files from $origin_s2i_godep_dir, and we got no warning because of '-f'.


rhcarvalho:
@bparees PTAL

I had to make this change yesterday to make an s2i bump in origin.
The files I expected to be deleted were still there, fortunately I noticed :)

bparees:
good catch, thanks
[merge]

PR #492, opened 2016-05-15
Apply several small fixes from go vet and others

rhcarvalho:
Includes:
- Changes to commentaries to follow godoc conventions;
- Fix shadowing of identifiers;
- Missing error checks;
- Fix incorrect usages of fmt.Print* functions;
- Initialize struct literals from other packages using explicit field
  names;
- Rename variables (not used in origin) to follow Go conventions.

### DIFF
--- a/pkg/api/types.go
+++ b/pkg/api/types.go
@@ -217,11 +217,13 @@ type EnvironmentSpec struct {
 // EnvironmentList contains list of environment variables.
 type EnvironmentList []EnvironmentSpec
 
+// ProxyConfig holds proxy configuration.
 type ProxyConfig struct {
 	HTTPProxy  *url.URL
 	HTTPSProxy *url.URL
 }
 
+// CGroupLimits holds limits used to constrain container resources.
 type CGroupLimits struct {
 	MemoryLimitBytes int64
 	CPUShares        int64
@@ -230,7 +232,7 @@ type CGroupLimits struct {
 	MemorySwap       int64
 }
 
-// Volume represents a single volume mount point
+// VolumeSpec represents a single volume mount point.
 type VolumeSpec struct {
 	Source      string
 	Destination string
--- a/pkg/build/strategies/layered/layered.go
+++ b/pkg/build/strategies/layered/layered.go
@@ -23,6 +23,11 @@ import (
 
 const defaultDestination = "/tmp"
 
+// A Layered builder builds images by first performing a docker build to inject
+// (layer) the source code and s2i scripts into the builder image, prior to
+// running the new image with the assemble script. This is necessary when the
+// builder image does not include "sh" and "tar" as those tools are needed
+// during the normal source injection process.
 type Layered struct {
 	config     *api.Config
 	docker     docker.Docker
@@ -32,6 +37,7 @@ type Layered struct {
 	hasOnBuild bool
 }
 
+// New creates a Layered builder.
 func New(config *api.Config, scripts build.ScriptsHandler, overrides build.Overrides) (*Layered, error) {
 	d, err := docker.New(config.DockerConfig, config.PullAuthentication)
 	if err != nil {
@@ -48,7 +54,7 @@ func New(config *api.Config, scripts build.ScriptsHandler, overrides build.Overr
 	}, nil
 }
 
-//getDestination returns the destination directory from the config
+// getDestination returns the destination directory from the config.
 func getDestination(config *api.Config) string {
 	destination := config.Destination
 	if len(destination) == 0 {
@@ -57,7 +63,8 @@ func getDestination(config *api.Config) string {
 	return destination
 }
 
-// checkValidDirWithContents will return true if the parameter provided is a valid, accessible directory that has contents (i.e. is not empty)
+// checkValidDirWithContents returns true if the parameter provided is a valid,
+// accessible and non-empty directory.
 func checkValidDirWithContents(name string) bool {
 	items, err := ioutil.ReadDir(name)
 	if os.IsNotExist(err) {
@@ -66,8 +73,8 @@ func checkValidDirWithContents(name string) bool {
 	return !(err != nil || len(items) == 0)
 }
 
-//CreateDockerfile takes the various inputs and creates the Dockerfile used by the docker cmd
-// to create the image produces by s2i
+// CreateDockerfile takes the various inputs and creates the Dockerfile used by
+// the docker cmd to create the image produced by s2i.
 func (builder *Layered) CreateDockerfile(config *api.Config) error {
 	buffer := bytes.Buffer{}
 
@@ -114,8 +121,8 @@ func (builder *Layered) CreateDockerfile(config *api.Config) error {
 	return nil
 }
 
+// SourceTar returns a stream to the source tar file.
 // TODO: this should stop generating a file, and instead stream the tar.
-//SourceTar returns a stream to the source tar file
 func (builder *Layered) SourceTar(config *api.Config) (io.ReadCloser, error) {
 	uploadDir := filepath.Join(config.WorkingDir, "upload")
 	tarFileName, err := builder.tar.CreateTarFile(builder.config.WorkingDir, uploadDir)
@@ -125,10 +132,11 @@ func (builder *Layered) SourceTar(config *api.Config) (io.ReadCloser, error) {
 	return builder.fs.Open(tarFileName)
 }
 
-//Build handles the `docker build` equivalent execution, returning the success/failure details
+// Build handles the `docker build` equivalent execution, returning the
+// success/failure details.
 func (builder *Layered) Build(config *api.Config) (*api.Result, error) {
 	if config.HasOnBuild && config.BlockOnBuild {
-		return nil, fmt.Errorf("builder image uses ONBUILD instructions but ONBUILD is not allowed.")
+		return nil, fmt.Errorf("builder image uses ONBUILD instructions but ONBUILD is not allowed")
 	}
 
 	if err := builder.CreateDockerfile(config); err != nil {
@@ -142,7 +150,7 @@ func (builder *Layered) Build(config *api.Config) (*api.Result, error) {
 	}
 	defer tarStream.Close()
 
-	dockerImageReference, err := docker.ParseDockerImageReference(builder.config.BuilderImage)
+	dockerImageReference, err := docker.ParseImageReference(builder.config.BuilderImage)
 	if err != nil {
 		return nil, err
 	}
--- a/pkg/build/strategies/layered/layered_test.go
+++ b/pkg/build/strategies/layered/layered_test.go
@@ -152,7 +152,7 @@ func TestBuildErrorCreateTarFile(t *testing.T) {
 	l.tar.(*test.FakeTar).CreateTarError = errors.New("CreateTarError")
 	_, err := l.Build(l.config)
 	if err == nil || err.Error() != "CreateTarError" {
-		t.Error("An error was expected for CreateTar, but got different: %v", err)
+		t.Errorf("An error was expected for CreateTar, but got different: %v", err)
 	}
 }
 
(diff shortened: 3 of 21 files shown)


pkg/build/strategies/layered/layered.go
```diff
@@ -23,6 +23,8 @@ import (
 
 const defaultDestination = "/tmp"
 
+// A Layered builder ...
+// REVIEW: improve commentary.
```
rhcarvalho: @bparees could you help me with a proper definition of a "layered build" here?

rhcarvalho:
[test]

pkg/docker/util.go
```diff
@@ -16,8 +16,8 @@ import (
 	"github.com/openshift/source-to-image/pkg/util/user"
 )
 
-// DockerImageReference points to a Docker image.
-type DockerImageReference struct {
+// ImageReference points to a Docker image.
+type ImageReference struct {
```
bparees: Most of this PR is nice cleanup, but changes like this one strike me as arbitrary and unnecessary code churn.  what's the motivation to rename this?

pkg/scripts/download.go
```diff
@@ -44,14 +44,14 @@ func NewDownloader(proxyConfig *api.ProxyConfig) Downloader {
 // Returns information a boolean flag informing whether any download/copy operation
 // happened and an error if there was a problem during that operation
 func (d *downloader) Download(url *url.URL, targetFile string) (*api.SourceInfo, error) {
-	schemeReader := d.schemeReaders[url.Scheme]
+	r := d.schemeReaders[url.Scheme]
```
bparees: rename makes this code less readable, what's the motivation?

pkg/scripts/download_test.go
```diff
@@ -30,8 +30,8 @@ func (f *FakeHTTPGet) get(url string) (*http.Response, error) {
 	}, f.err
 }
 
-func getHTTPReader() (*HttpURLReader, *FakeHTTPGet) {
-	sr := &HttpURLReader{}
+func getHTTPReader() (*HTTPURLReader, *FakeHTTPGet) {
```
bparees: since you renamed it to NewHTTPURLReader, shouldn't this be getHTTPURLReader?

pkg/build/strategies/layered/layered.go
bparees: A layered build is a build which first performs a docker build in order to inject(layer) the source/etc into the builder image, prior to running the new image with assemble.   This is necessary when the builder image does not include "sh" and "tar" as those tools are needed during normal source injection process.

pkg/docker/util.go
rhcarvalho: Stuttering:

```console
$ cat stutter.go 
package stutter

type StutterMasterExample struct {}
$ golint stutter.go 
stutter.go:3:6: exported type StutterMasterExample should have comment or be unexported
stutter.go:3:6: type name will be used as stutter.StutterMasterExample by other packages, and that stutters; consider calling this MasterExample
```

I'm also reluctant to renames, but since this is really just internal functionality I judge it was a fine cleanup (but still a long way to go to get rid of all reasonable/actionable lint suggestions).

What we could and I don't see us using is `internal/` directories to hold internal-only packages, would maybe make our life more in peace here when refactoring internal stuff that needs to be exported for use in other packages, while not meant to be part of a public API.

pkg/build/strategies/layered/layered.go
rhcarvalho: Thanks!

pkg/build/strategies/layered/layered.go
```diff
@@ -23,6 +23,11 @@ import (
 
 const defaultDestination = "/tmp"
 
+// A Layered builder builds images by first performing a docker build to inject
+// (layer) the source code and s2i scripts into the builder image, prior to
+// running the new image with the assemble script. This is necessary when the
+// builder image does not include "sh" and "tar" as those tools are needed
+// during the normal source injection process.
```
rhcarvalho: @bparees sounds accurate?

pkg/scripts/download.go
rhcarvalho: Shadowing. `schemeReader` is a top-level declared type.

Within this short context, calling a thing of type `schemeReader`, coming out of `d.schemeReaders[url.Scheme]`, doesn't seen to me "unreadable" when looking at it with a "Go-philosophy hat".

https://github.com/golang/go/wiki/CodeReviewComments#variable-names

pkg/scripts/download_test.go
rhcarvalho: I renamed `NewHTTPURLReader` because that has a special semantic value: `godoc` understands that that is a "constructor" for the type and places the function under the type.

See for instance https://golang.org/pkg/bytes/#Reader. It starts with the type definition, then the `NewReader` function, then all methods of the `Reader` type.

This test function doesn't carry the same value, I didn't intend to go rewrite huge portions of the code base in one shot. Perhaps this fits in the "less churn" you noted?

rhcarvalho: If you think this is important I can rename it, no problem, of course :)

pkg/api/types.go
```diff
@@ -378,7 +380,7 @@ func (p *PullPolicy) String() string {
 
 // Type implements the Type() function of pflags.Value interface
 func (p *PullPolicy) Type() string {
-	return "string"
+	return "PullPolicy"
```
php-coder: Are you sure that such changes won't break something?

pkg/build/strategies/layered/layered.go
```diff
@@ -48,7 +54,7 @@ func New(config *api.Config, scripts build.ScriptsHandler, overrides build.Overr
 	}, nil
 }
 
-//getDestination returns the destination directory from the config
+// getDestination returns the destination directory from the config.
 func getDestination(config *api.Config) string {
```
php-coder: I'm missing equivalent of `COALESCE` (from SQL) here :-( Or something like `strings.FirstNonEmpty()` method because in many places we're using code like this:
```go
result := cfg.Property
if len(result) > 0 {
    result = "default value"
}
```
I would like to write it like this:
```
result := strings.firstNonEmpty(cfg.Property, "default value")
```

(It's just an idea.)

pkg/build/strategies/layered/layered_test.go
```diff
@@ -152,7 +152,7 @@ func TestBuildErrorCreateTarFile(t *testing.T) {
 	l.tar.(*test.FakeTar).CreateTarError = errors.New("CreateTarError")
 	_, err := l.Build(l.config)
 	if err == nil || err.Error() != "CreateTarError" {
-		t.Error("An error was expected for CreateTar, but got different: %v", err)
+		t.Errorf("An error was expected for CreateTar, but got different: %v", err)
```
php-coder: Wow

pkg/scripts/download.go
php-coder: Why not `reader` BTW?

pkg/api/types.go
rhcarvalho: This is used to tell [pflag](https://godoc.org/github.com/spf13/pflag) the name of the type of a custom flag.
See the example in https://godoc.org/github.com/spf13/pflag#example-package.

Calling it "string" in all three cases seems a bit misleading, but it does change how the default values are printed in the command-line help. Perhaps it was intentional to leverage the treatment pflag gives to strings...
[spf13/pflag/blob/master/flag.go#L502-L511](https://github.com/spf13/pflag/blob/cb88ea77998c3f024757528e3305022ab50b43be/flag.go#L502-L511)

I got here originally via `goconst` suggesting that `"string"` should be a constant because it appears 3 times...

I will revert the changes to the output of the `Type` methods, no changes to CLI help messages.
@php-coder thanks for asking!

pkg/scripts/download.go
rhcarvalho: > Why not reader BTW?

Copying the contents of the link I posted above:

> The basic rule: the further from its declaration that a name is used, the more descriptive the name must be. For a method receiver, one or two letters is sufficient. **Common variables such as** loop indices and **readers** can be a single letter (i, **r**). More unusual things and global variables need more descriptive names.

I don't think `reader.Read(url)` in this context is any better than `r.Read(url)`.

pkg/docker/util.go
bparees: that feels like an overly obsessive lint suggestion, i'd rather not see us jump through hoops for lint over things like that.  among other things, there's no guarantee the package even gets imported as "docker", (especially with a package name like docker that's likely to be reused by other packages)

pkg/scripts/download.go
bparees: it's not really shadowing anything since they are two different syntactical elements.  and obviously that rule is pretty subjective. (how far away qualifies as further away, and is this really a generic "reader" given that it can hold different types of readers depending on the scheme it's being used for?  all these things make it a little more special imho and deserving of a unique name.

pkg/scripts/download_test.go
bparees: it's not that important (so i'm not asking you do it, it's fine as is) but it does bother me a little that they are out of sync now.  it may be more critical to have renamed the constructor, but that doesn't make it unimportant to align other stuff.

pkg/build/strategies/layered/layered.go
bparees: :+1:

pkg/scripts/download_test.go
rhcarvalho: They were already "out of sync", see the red/old part of the diff:

```diff
-func getHTTPReader() (*HttpURLReader, *FakeHTTPGet) {
-	sr := &HttpURLReader{}
```

`getHTTPReader` returns `*HttpURLReader` (and also a `*FakeHTTPGet`?!)

pkg/build/strategies/layered/layered_test.go
mfojtik: you would be surprised how much of there are for example in kube :-D

pkg/docker/util.go
rhcarvalho: This gets used [only once](https://github.com/openshift/source-to-image/pull/492/files#diff-7ae9cffe2e2079019c59cc3416502721L145) as:

```diff
- dockerImageReference, err := docker.ParseDockerImageReference(builder.config.BuilderImage)
+ dockerImageReference, err := docker.ParseImageReference(builder.config.BuilderImage)
```

Sorry not clear if you want me to undo the renaming or was just a rant?

bparees: just a rant, you can leave it.

pkg/scripts/download.go
rhcarvalho: The tools that shout the shadowing are not buggy in this case, it's just that it is a very subtle thing. Did you know about Method Values? We talked about this in one of our coding dojos some time ago, @php-coder must remember.

https://play.golang.org/p/C3DZmpy77S


Not saying we should blindly follow what linters suggest. In some cases shadowing is indeed the intended behavior. For example, when you have a variable in a closure with the same name as in the outer function, and they should be distinct variables -- I've seen it happen with the ubiquitous `err` variable name.

@bparees if you want this local variable stay as `schemeReader` so be it.

rhcarvalho:
@bparees could you please sum up what changes you'd like me to do here?

bparees:
@rhcarvalho i think it's all fine, none of my concerns w/ the changes warrant reverting it, just things that might be worth keeping in mind during future refactors.  I think i just lean harder towards "if it ain't broke don't fix it" over obsessive de-linting, but certainly if we can get to a clean lint state, that's got value, so I appreciate you keeping us honest. :+1:

rhcarvalho:
These changes were really to validate #495 when I realized that even though we had "verify-govet.sh" it was not playing any purpose in our CI.
Just a Sunday afternoon thing anyway....

bparees:
[merge]

pkg/scripts/download.go
rhcarvalho: > it's not really shadowing anything since they are two different syntactical elements.

@bparees @php-coder look what I came across today from golang-nuts:
https://groups.google.com/d/topic/golang-nuts/QOH-9wywB38/discussion

- Variables and types live in the same namespace, so one can shadow the
other.
- Types can also shadow variables.
- And variables can shadow packages.

bparees: interesting, thanks.

PR #493, opened 2016-05-15
Fix ldflag logic

rhcarvalho:
The condition should test for `go1.4` and use `${key} ${val}`, while `go1.5` and `go1.6` should use `${key}=${val}`.

The actual test could consider versions prior to go1.4, but I think we don't care about them, i.e., we require go1.4+?


rhcarvalho:
[test]

@bparees PTAL

bparees:
[merge]

PR #495, opened 2016-05-15
Fix hack/verify-*.sh and improve other scripts

rhcarvalho:
Summary:
- Add 'make verify' to Travis.
- Fix verify-golint.sh: old invocation would not work unless GOPATH is a
  single directory.
- Fix verify-govet.go: 'go vet' does not take flags, should use 'go tool
  vet'.
- Fix early termination of hack/verify-govet.sh.
- Enable 'go vet' in 'make verify'; will prevent introducing problems
  detected by the vet tool in future PRs.
- Make sure golint is installed in Travis.
- Improve install-tools.sh, parity with the one in openshift/origin.
- s/os::util::sed/s2i::util::sed for consistency.
- Reimplement/simplify find_files, now called s2i::util::find_files, and
  use it in all scripts consistently.
- Suggest autocorrection for gofmt, as seen in openshift/origin.
- Do not check Go version (verify-*.sh should be run no matter the Go
  version?!).
- Do not create directory _output/govet (not used?!).

Also:
- Bump minor versions of Go 1.5 and 1.6.

### DIFF
--- a/hack/install-tools.sh
+++ b/hack/install-tools.sh
@@ -1,11 +1,13 @@
 #!/bin/bash
 
-set -e
+set -o errexit
+set -o nounset
+set -o pipefail
 
-S2I_ROOT=$(dirname "${BASH_SOURCE}")/..
-source "${S2I_ROOT}/hack/common.sh"
+STARTTIME=$(date +%s)
 
-GO_VERSION=($(go version))
-echo "Detected go version: $(go version)"
+echo $(go version)
 
-go get golang.org/x/tools/cmd/cover github.com/tools/godep
+go get github.com/tools/godep github.com/golang/lint/golint
+
+ret=$?; ENDTIME=$(date +%s); echo "$0 took $(($ENDTIME - $STARTTIME)) seconds"; exit "$ret"
--- a/hack/util.sh
+++ b/hack/util.sh
@@ -71,7 +71,7 @@ s2i::log::error_exit() {
   exit "${code}"
 }
 
-os::util::sed() {
+s2i::util::sed() {
   if [[ "$(go env GOHOSTOS)" == "darwin" ]]; then
     sed -i '' $@
   else
@@ -79,13 +79,6 @@ os::util::sed() {
   fi
 }
 
-find_files() {
-  find . -not \( \
-      \( \
-        -wholename './_output' \
-        -o -wholename './_tools' \
-        -o -wholename './.*' \
-        -o -wholename '*/Godeps/*' \
-      \) -prune \
-    \) -name '*.go' | sort -u
+s2i::util::find_files() {
+  find . -type d -name Godeps -prune -o -name '*.go' -print
 }
--- a/hack/verify-gofmt.sh
+++ b/hack/verify-gofmt.sh
@@ -4,34 +4,18 @@ set -o errexit
 set -o nounset
 set -o pipefail
 
-GO_VERSION=($(go version))
-
-if [[ -z $(echo "${GO_VERSION[2]}" | grep -E 'go1.4|go1.5') ]]; then
-  echo "Unknown go version '${GO_VERSION[2]}', skipping gofmt."
-  exit 0
-fi
+echo $(go version)
 
 S2I_ROOT=$(dirname "${BASH_SOURCE}")/..
-source "${S2I_ROOT}/hack/common.sh"
+source "${S2I_ROOT}/hack/util.sh"
 
 cd "${S2I_ROOT}"
 
-find_files() {
-  find . -not \( \
-      \( \
-        -wholename './output' \
-        -o -wholename './_output' \
-        -o -wholename './release' \
-        -o -wholename './target' \
-        -o -wholename '*/Godeps/*' \
-      \) -prune \
-    \) -name '*.go'
-}
-
-bad_files=$(find_files | xargs gofmt -s -l)
+bad_files=$(s2i::util::find_files | xargs gofmt -s -l)
 if [[ -n "${bad_files}" ]]; then
-  echo "!!! gofmt needs to be run on the following files: "
+  echo "!!! gofmt needs to be run on the following files: " >&2
   echo "${bad_files}"
-  echo "Try running 'gofmt -s -d [path]'"
+  echo "Try running 'gofmt -s -d [path]'" >&2
+  echo "Or autocorrect with 'hack/verify-gofmt.sh | xargs -n 1 gofmt -s -w'" >&2
   exit 1
 fi
--- a/hack/verify-golint.sh
+++ b/hack/verify-golint.sh
@@ -1,7 +1,10 @@
 #!/bin/bash
 
 set -o errexit
+set -o nounset
 set -o pipefail
+
+echo $(go version)
 
 if ! which golint &>/dev/null; then
   echo "Unable to detect 'golint' package"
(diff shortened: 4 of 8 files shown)


rhcarvalho:
[test]

@bparees PTAL

Note: Travis is failing on purpose -- I've enabled running `make verify` and made that require no errors from `go vet`. All current vet errors are fixed in #492, so once it merges Travis should go green after a re-run.

bparees:
lgtm but will wait for #492 to merge so we can confirm this goes green before merging it.

hack/install-tools.sh
```diff
@@ -1,11 +1,13 @@
...
+STARTTIME=$(date +%s)
 
-GO_VERSION=($(go version))
-echo "Detected go version: $(go version)"
+echo $(go version)
 
-go get golang.org/x/tools/cmd/cover github.com/tools/godep
+go get github.com/tools/godep github.com/golang/lint/golint
+
+ret=$?; ENDTIME=$(date +%s); echo "$0 took $(($ENDTIME - $STARTTIME)) seconds"; exit "$ret"
```
php-coder: I would rather write `exit $ret`.

hack/verify-gofmt.sh
```diff
@@ -6,32 +6,21 @@ set -o pipefail
...
-      \) -prune \
-    \) -name '*.go'
-}
-
-bad_files=$(find_files | xargs gofmt -s -l)
+bad_files=$(s2i::util::find_files | xargs gofmt -s -l)
 if [[ -n "${bad_files}" ]]; then
-  echo "!!! gofmt needs to be run on the following files: "
+  echo "!!! gofmt needs to be run on the following files: " >&2
   echo "${bad_files}"
```
php-coder: Why the list of files isn't printed to `stderr` too?

hack/util.sh
```diff
@@ -71,21 +71,14 @@ sti::log::error_exit() {
   exit "${code}"
 }
 
-os::util::sed() {
+s2i::util::sed() {
```
php-coder: This makes a sync between openshift and s2i more difficult, right? Do we really need this rename?

hack/verify-golint.sh
```diff
@@ -10,45 +11,21 @@ if ! which golint &>/dev/null; then
...
-echo "Detected go version: $(go version)"
 
-if [[ -z $(echo "${GO_VERSION[2]}" | grep -E 'go1.4|go1.5') ]]; then
-  echo "Unknown go version '${GO_VERSION}', skipping golint."
+if [[ -z $(echo "${GO_VERSION[2]}" | grep -E 'go1.[4-6]') ]]; then
+  echo "Unknown go version '${GO_VERSION[@]:2}', skipping golint."
   exit 0
 fi
 
 STI_ROOT=$(dirname "${BASH_SOURCE}")/..
```
php-coder: Maybe rename it to `S2I_ROOT` if you're improving this script? WDYT?

hack/verify-golint.sh
```diff
@@ -10,45 +11,21 @@ if ! which golint &>/dev/null; then
...
-        -o -wholename './test' \
-        -o -wholename '*/Godeps/*' \
-        -o -wholename '*/third_party/*' \
-        -o -wholename '*/_output/*' \
-      \) -prune \
-    \) -name '*.go' | sort -u | sed 's/^.{2}//' | xargs -n1 printf "${GOPATH}/src/${STI_GO_PACKAGE}/%s\n"
-  }
-  bad_files=$(find_files | xargs -n1 golint)
-fi
+bad_files=$(s2i::util::find_files | xargs -n1 golint)
```
php-coder: `s2i::util::find_files` and `find` with 7 different conditions look different. Are you sure that this won't add regression?

hack/verify-govet.sh
```diff
@@ -1,27 +1,26 @@
 #!/bin/bash
 
+set -o errexit
```
rhcarvalho: Before we merge I want to fix the behavior on errors.

Because we `source "${STI_ROOT}/hack/util.sh"`, the script becomes effectively exit on error, and will exit on the first dir that fails `go vet`.

hack/verify-golint.sh
rhcarvalho: Yes, before getting to a single implementation I had it changed in each file to something that makes sense.
`-wholename` is documented as less portable than `-path`, and we don't need all those rules (probably there just because it was copied from origin at some point).

hack/verify-gofmt.sh
rhcarvalho: See https://github.com/openshift/origin/pull/6820

If it was sent to `stderr` we wouldn't be able to use the `xargs` approach to autocorrect.

hack/util.sh
rhcarvalho: I'd opt for consistency with the rest?
We don't do a great job at keeping this stuff in sync (humans...), and in most cases it is not a problem, the projects can diverge.

hack/install-tools.sh
rhcarvalho: Just copying what is in origin :yum: 

If we will talk about it, I'd point out another problem is that when ret != 0 we never see that line printed because of `set -e`....

I didn't intend this PR to go fix that (this I would like to apply to all scripts, here and in origin). Feel free to fix it with more bash-foo.

hack/verify-golint.sh
rhcarvalho: That's a good improvement, but touches 49 lines in 14 files... I fear it will just make this PR fuzzier to understand.
I will not do it here because there's a risk of things outside of this repo dependent on the name.
And if we change STI_ROOT, then we can go ahead and change all STI_* variables... but that's really a riskier move.

hack/verify-govet.sh
rhcarvalho: Fixed, now the Travis output should include the message "FAILURE: go vet failed!", instead of "make: *** [verify] Error 1" ([old job output](https://travis-ci.org/openshift/source-to-image/jobs/130451051#L274)).

rhcarvalho:
And the new job run shows the message we expected to see :)

https://travis-ci.org/openshift/source-to-image/jobs/130560314#L272

bparees:
@rhcarvalho rebase and let's see if we can get travis green

bparees:
@rhcarvalho nm, forgot travis is going to test the merged pr.  i kicked it off, we'll see if it goes green

rhcarvalho:
@bparees Travis got yet another vet problem:

```
pkg/build/strategies/onbuild/onbuild_test.go:66: api.EnvironmentList composite literal uses unkeyed fields
```

Gonna fix it and push.

rhcarvalho:
Turns out `api.EnvironmentList` is a `[]EnvironmentSpec`.... so just like we disable checks for shadow, we will have to disable checks for unkeyed fields in composite literal (there's an upstream bug in Go vet -- https://github.com/golang/go/issues/11394)

rhcarvalho:
Cannot reproduce the vet error locally. Rebased and pushed, let's see what Travis will say.

rhcarvalho:
Omg.. Travis still failing... in my dev box I get:

```
$ ./hack/verify-govet.sh 
go version go1.6.2 linux/amd64
SUCCESS: go vet succeded!
```

rhcarvalho:
Conclusion: disabling the vet check for composite literals makes Travis green:

image

What I'm not happy about is that I could not reproduce the failure locally -- same Go version.

rhcarvalho:
I was able to reproduce it locally, the problem was stale packages in GOPATH/pkg/... :(

For now disabling the check of composite literals because of false positives.

bparees:
@rhcarvalho travis looks green, is this good to go?

rhcarvalho:
@bparees yes good to go. I was waiting for jenkins SUCCESS before pinging you.

bparees:
[merge]

PR #496, opened 2016-05-15
Name containers created by s2i explicitly

### LINKED ISSUE #395: Build containers are named randomly by the Docker daemon
We should be setting names for the containers we create, instead of letting the Docker daemon name it "lonely_thompson" and alike.

Faulty code: https://github.com/openshift/source-to-image/blob/9dab4e2fa154610243ace245b128892831bb0af2/pkg/docker/docker.go#L596

Discussion: https://github.com/openshift/origin/pull/6715/files#r50524359

rhcarvalho:
Instead of leaving the name blank and letting the Docker daemon name
containers with random strings like 'lonely_thompson', we give proper
names so that we can identify and filter containers created by S2I
later.

Names take a fixed prefix, 's2i_', and a random suffix. The builder
image is also used to compose the container name, replacing with
underscores characters that would be invalid for a container name.

Example:

```console
$ s2i build \
    git://github.com/openshift/ruby-hello-world \
    centos/ruby-22-centos7 \
    hello-world-app
```

The above creates a temporary container named
's2i_centos_ruby_22_centos7_397cfecd'.

-----------------------------------------

Fixes #395.

### DIFF
--- a/pkg/docker/docker.go
+++ b/pkg/docker/docker.go
@@ -3,6 +3,7 @@ package docker
 import (
 	"fmt"
 	"io"
+	"math/rand"
 	"os"
 	"path"
 	"path/filepath"
@@ -51,6 +52,25 @@ const (
 	// timeout is reached, certain Docker API calls might error out.
 	DefaultDockerTimeout = 20 * time.Second
 )
+
+// containerNamePrefix prefixes the name of containers launched by S2I. We
+// cannot reuse the prefix "k8s" because we don't want the containers to be
+// managed by a kubelet.
+const containerNamePrefix = "s2i"
+
+// containerName creates names for Docker containers launched by S2I. It is
+// meant to resemble Kubernetes' pkg/kubelet/dockertools.BuildDockerName.
+func containerName(image string) string {
+	uid := fmt.Sprintf("%08x", rand.Uint32())
+	// Replace invalid characters for container name with underscores.
+	image = strings.Map(func(r rune) rune {
+		if ('0' <= r && r <= '9') || ('A' <= r && r <= 'Z') || ('a' <= r && r <= 'z') {
+			return r
+		}
+		return '_'
+	}, image)
+	return fmt.Sprintf("%s_%s_%s", containerNamePrefix, image, uid)
+}
 
 // Docker is the interface between STI and the Docker client
 // It contains higher level operations called from the STI
@@ -171,7 +191,7 @@ func (rco RunContainerOptions) asDockerCreateContainerOptions() docker.CreateCon
 	config := rco.asDockerConfig()
 	hostConfig := rco.asDockerHostConfig()
 	return docker.CreateContainerOptions{
-		Name:       "",
+		Name:       containerName(rco.Image),
 		Config:     &config,
 		HostConfig: &hostConfig,
 	}
--- a/cmd/s2i/main.go
+++ b/cmd/s2i/main.go
@@ -4,10 +4,12 @@ import (
 	"bytes"
 	"flag"
 	"fmt"
+	"math/rand"
 	"os"
 	"path/filepath"
 	"runtime"
 	"strings"
+	"time"
 
 	"github.com/golang/glog"
 	"github.com/spf13/cobra"
@@ -355,6 +357,7 @@ func checkErr(err error) {
 }
 
 func main() {
+	rand.Seed(time.Now().UnixNano())
 	cfg := &api.Config{}
 	stiCmd := &cobra.Command{
 		Use: "s2i",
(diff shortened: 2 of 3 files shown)


bparees:
s2i prefix seems reasonable to me.. maybe s2i_\<builderimagename\>_randomuid ?

rhcarvalho:
> s2i prefix seems reasonable to me.. maybe s2i_<builderimagename>_randomuid ?

I thought maybe we could include the same info that openshift uses (build name, namespace, ...), this would allow for better debugging and perhaps more directed cleanup.

But that would require openshift to pass in that information to s2i.

bparees:
when running from openshift i agree it would be good to include that info, i thought we were just discussing the non-openshift scenario here.

bparees:
@rhcarvalho any chance of moving forward w/ this?  (trying to clear tech debt)

pkg/docker/docker.go
```diff
@@ -171,7 +188,7 @@ func (rco RunContainerOptions) asDockerCreateContainerOptions() docker.CreateCon
 	config := rco.asDockerConfig()
 	hostConfig := rco.asDockerHostConfig()
 	return docker.CreateContainerOptions{
-		Name:       "",
+		Name:       containerName("", ""),
```
php-coder: BTW, why we're generating names? Could we specify names explicitly? For example, "Builder container", "Runtime image container" etc.

rhcarvalho: We cannot have name collisions. Keep in mind there can be several concurrent builds.

rhcarvalho:
> @rhcarvalho any chance of moving forward w/ this? (trying to clear tech debt)

I scheduled some time towards the end half of this sprint to catch up on old PRs.

bparees:
thanks @rhcarvalho!

On Wed, Jun 15, 2016 at 8:53 AM, Rodolfo Carvalho <notifications@github.com>
wrote:

> @rhcarvalho <https://github.com/rhcarvalho> any chance of moving forward
> w/ this? (trying to clear tech debt)
>
> I scheduled some time towards the end half of this sprint to catch up on
> old PRs.
>
> —
> You are receiving this because you commented.
> Reply to this email directly, view it on GitHub
> <https://github.com/openshift/source-to-image/pull/496#issuecomment-226232206>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/AEvl3pxTFqrO7fQL2wlmWi_jk-NrzaSxks5qMCAXgaJpZM4Ie8N3>
> .
>



-- 
Ben Parees | OpenShift

rhcarvalho:
> s2i prefix seems reasonable to me.. maybe s2i_\<builderimagename\>_randomuid ?

@bparees we cannot use the image spec as is, because it may not always lead to a valid container name. E.g.:

```console
$ docker run --name 's2i_centos:latest_f4f4f4' centos:latest
Error response from daemon: Invalid container name (s2i_centos:latest_f4f4f4), only [a-zA-Z0-9][a-zA-Z0-9_.-] are allowed
```

Neither `/` nor `:` are allowed, which would naturally appear in the image spec.

Since I like the idea of using the image name, we could replace slashes and colons with underscores, wdyt?

rhcarvalho:
Updated description to reflect the current state.
Ready for review / merge.

[test]

pkg/docker/docker.go
```diff
@@ -111,6 +131,7 @@ type PullResult struct {
 
 // RunContainerOptions are options passed in to the RunContainer method
 type RunContainerOptions struct {
+	Name             string
```
rhcarvalho: Looked at how we'd leverage this from Origin, doesn't seem useful.

We never set `RunContainerOptions` directly, instead we call the `Save` and `Execute` methods defined in `pkg/build/strategies/sti/sti.go`, and they call `RunContainer` with options based on yet another `Config` struct.

If we don't care to correlate BuildConfigs and S2I containers, then the name generated by `containerName` should be enough. Otherwise we'd need to make changes to the `Config` struct. In that case, probably instead of providing a container name we should provide just metadata because S2I might create multiple containers and I guess we don't want to use a single name (for a given build).

bparees: i'm fine with not introducing it at this point.

bparees:
substituting w/ underscores works for me.

pkg/docker/docker.go
rhcarvalho: Will drop the second commit and we're ready to go then.

rhcarvalho:
@bparees ready to go.

bparees:
lgtm [merge]
thanks @rhcarvalho, great to have this one done :)

rhcarvalho:
> great to have this one done :)

I :heart: it too, let's kill some more bugs this week.

PR #497, opened 2016-05-16
Replace STI_* env vars with S2I_*

rhcarvalho:
This is a careful replacement of `STI_([A-Z_]+)` with `S2I_$1` in all
*.sh files, plus references to those env vars in other files.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -65,11 +65,11 @@ To run all tests with verbose output:
 
 To turn off or change the coverage mode, which is `-cover -covermode=atomic` by default, use:
 
-    $ STI_COVER="" hack/test-go.sh
+    $ S2I_COVER="" hack/test-go.sh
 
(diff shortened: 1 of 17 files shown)


rhcarvalho:
@php-coder this is what you were asking for in https://github.com/openshift/source-to-image/pull/495/files#r63330648

This change triggers my 'risk radar' as we might have things that we're unaware of that depend on the old names.
Somehow it would be best to handle both variants STI_* and S2I_* for "public facing" variables, but that's hard to determine, and would add some complications to our Bash scripts, unless there's an obvious simple way to do it? I don't know.

@bparees FYI

[test]

bparees:
@rhcarvalho agree it feels a little risky, but it seems like your change is thorough, we'll give it a shot and keep an eye out for weirdness.
[merge]

rhcarvalho:
@bparees let's pray together :pray:

PR #500, opened 2016-05-18
S2I needs to selectively log to stderr instead of glog

smarterclayton:
Continuation of work done in Origin, but ensures that we aren't using
glog for our core messages for end users in builds.

@mfojtik @bparees fixes part of https://github.com/openshift/origin/pull/8924

### DIFF
--- a/pkg/docker/docker.go
+++ b/pkg/docker/docker.go
@@ -12,7 +12,6 @@ import (
 	"time"
 
 	docker "github.com/fsouza/go-dockerclient"
-	"github.com/golang/glog"
 
 	"github.com/openshift/source-to-image/pkg/api"
 	"github.com/openshift/source-to-image/pkg/errors"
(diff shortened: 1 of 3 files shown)


bparees:
lgtm [merge]

php-coder:
It means that from now we should always use `glog.Infof()` and add prefix to the error/warning messages, right?

php-coder:
Another question: should we also update places where we're passing `glog.Error`/`glog.Info`?

```console
$ git grep -n StreamContainerIO | grep glog
pkg/build/strategies/sti/sti.go:430:	go dockerpkg.StreamContainerIO(errReader, nil, glog.Error)
pkg/build/strategies/sti/sti.go:588:	go dockerpkg.StreamContainerIO(errReader, &errOutput, glog.Error)
pkg/run/run.go:57:	go docker.StreamContainerIO(errReader, nil, glog.Error)
pkg/run/run.go:58:	go docker.StreamContainerIO(outReader, nil, glog.Info)
```

smarterclayton:
To the first question, yes.  To the second question, maybe, if it's output
that users get by default.   If not, then it's less important (this is
mostly to make builds at log level 0, 1, or 2 to be human readable).

PR #501, opened 2016-05-18
Refactor PostExecute() method

php-coder:
Extracts 6 methods from `PostExecute()` method to make it shorter, readable and easy to modify.

PTAL @bparees

PR #503, opened 2016-05-20
Fix error message about retrieving previous image's metadata for non-incremental build

php-coder:
Also include a tag of previous image to the message.

This is a fix for the small regression that was added in #501: it started to show error message:
```
error: Error retrieving previous image's metadata: no such image
```
But it didn't do anything else, so nothing was broken.

PTAL @bparees

### DIFF
--- a/pkg/build/strategies/sti/sti.go
+++ b/pkg/build/strategies/sti/sti.go
@@ -273,7 +273,10 @@ func mergeLabels(newLabels, existingLabels map[string]string) map[string]string 
 // finishes.
 func (builder *STI) PostExecute(containerID, location string) error {
 
-	previousImageID := builder.getPreviousImage()
+	var previousImageID string
+	if builder.incremental && builder.config.RemovePreviousImage {
+		previousImageID = builder.getPreviousImage()
+	}
 
 	buildEnv := builder.createBuildEnvironment()
 	runCmd := builder.createCommandForResultingImage(location)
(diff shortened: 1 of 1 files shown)


pkg/build/strategies/sti/sti.go
```diff
@@ -306,9 +306,13 @@ func (builder *STI) PostExecute(containerID, location string) error {
 }
 
 func (builder *STI) getPreviousImage() string {
+	if !builder.incremental || !builder.config.RemovePreviousImage {
+		return ""
+	}
```
bparees: maybe we shouldn't even be calling this if we're not doing an incremental build?  it seems weird to return "".

php-coder: In case of the incremental build && when we need to remove previous image [we are saving the metadata at the beginning of the method and trying to remove image at the end of the method](https://github.com/openshift/source-to-image/blob/619e8798fc33a5f684201c716a12c41e69e74550/pkg/build/strategies/sti/sti.go#L276-L302). Also `previousImageId` can be empty if we couldn't fetch metadata for the image.

So, as far I understand, we can only move this `if` outside of the method. But I wouldn't like to do so, because I was trying to make `PostExecute()` more clean and contain only high-level methods with human-readable names.

bparees: seems like this if-logic does belong in postexecute, because what postexecute does is based on whether this is an incremental build or not.  hiding that logic in two different methods that have to stay in sync seems like a bad idea.  (getPreviousImage and removePreviousImage are now tightly coupled in a way that might not be obvious to someone editing one or the other of them).

php-coder: Yes, you're right. I've just pushed a fix.

bparees:
[test]

bparees:
git seems confused, maybe do a force push or something.  it doesn't appear to be testing the actual current commit.

php-coder:
@bparees It's ready to merge now.

bparees:
lgtm [merge]

php-coder:
I couldn't reproduce it locally :-(

bparees:
[test]

bparees:
[merge]

PR #504, opened 2016-05-20
strategies/sti/sti.go: add firstNonEmpty() method

php-coder:
Because [Golang doesn't provide method similar to `COALESCE()`](https://github.com/golang/go/issues/14423) I introduced our own.

I hope that no one will object about this.

PTAL @bparees @mfojtik @rhcarvalho

### DIFF
--- a/pkg/build/strategies/sti/sti.go
+++ b/pkg/build/strategies/sti/sti.go
@@ -145,10 +145,7 @@ func (builder *STI) Build(config *api.Config) (*api.Result, error) {
 	}
 
 	if builder.incremental = builder.artifacts.Exists(config); builder.incremental {
-		tag := config.IncrementalFromTag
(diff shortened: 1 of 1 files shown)


bparees:
nice.
[merge]

pkg/build/strategies/sti/sti.go
```diff
@@ -648,3 +636,12 @@ func includes(arr []string, str string) bool {
 	}
 	return false
 }
+
+func firstNonEmpty(args ...string) string {
+	for _, value := range args {
+		if len(value) > 0 {
```
rhcarvalho: I still don't get why our "standard" of not writing `value != ""`. I see @php-coder had to carefully change this from the example he pointed to in https://github.com/golang/go/issues/14423.

PR #505, opened 2016-05-20
Extended build

php-coder:
This is the implementation of the extended build.

When we're specifying the `--runtime-image` and `--runtime-artifact` options, s2i:
- runs builder container and does a build as usual
- downloads artifacts (that were specified by `--runtime-artifact`) from the builder container and places them to a directory
- runs another container with runtime image
- uploads directory with artifacts into `WORKDIR` of a runtime image
- runs `assemble-runtime` script
- commits container with runtime image

The build flow as a scheme:

using-runtime-image

Example:
```console
$ cat .s2i/bin/assemble 
#!/bin/bash
set -eu
echo "It works!" >/tmp/i-am-runtime-artifact1.txt
chmod 666 /tmp/i-am-runtime-artifact1.txt
echo "It works!" >/tmp/.i-am-runtime-artifact3.txt
chmod 600 /tmp/.i-am-runtime-artifact3.txt
cd /tmp
ln -s i-am-runtime-artifact1.txt i-am-runtime-artifact2
printf '#!/bin/bash\necho "I am $0!"\n' >/tmp/run-me
chmod 755 /tmp/run-me
mkdir -p /tmp/data-dir/2
touch /tmp/data-dir/1
touch /tmp/data-dir/2/3

$ cat .s2i/bin/assemble-runtime 
#!/bin/bash
set -eu
echo "I'm assemble-runtime"
touch /tmp/assemble-runtime-done

$ cat .s2i/bin/run 
#!/bin/bash
set -eu
/opt/jboss/run-me

$ s2i build . openshift/wildfly-100-centos7 my-app --runtime-image jboss/base-jdk:8 --copy --assemble-user=1000 -a /tmp/i-am-runtime-artifact1.txt:app -a /tmp/i-am-runtime-artifact2:app -a /tmp/.i-am-runtime-artifact3.txt:app -a /tmp/run-me -a /tmp/data-dir:.
warning: Image c638f9a406496a819dcd49ec496b388704826382e834c417886702a15b3e49ee does not contain a value for the io.openshift.s2i.scripts-url label


I'm assemble-runtime
$ docker images | head -2
REPOSITORY                                           TAG                 IMAGE ID            CREATED             VIRTUAL SIZE
my-app                                               latest              6ee5b5f789a7        10 seconds ago      464 MB

$ docker run my-app
I am /opt/jboss/run-me!

$ docker run my-app find -ls
16871329    0 drwxr-xr-x   2 root     root          105 Jul  1 15:50 ./app
16871330    4 -rw-r--r--   1 root     root           10 Jul  1 15:50 ./app/.i-am-runtime-artifact3.txt
16871331    4 -rw-r--r--   1 root     root           10 Jul  1 15:50 ./app/i-am-runtime-artifact1.txt
16871332    0 lrwxrwxrwx   1 root     root           26 Jul  1 15:50 ./app/i-am-runtime-artifact2 -> i-am-runtime-artifact1.txt
16871333    4 -rwxr-xr-x   1 root     root           81 Jul  1 15:50 ./assemble-runtime
33578700    0 drwxr-xr-x   3 root     root           24 Jul  1 15:50 ./data-dir
33578701    0 -rw-r--r--   1 root     root            0 Jul  1 15:50 ./data-dir/1
50342803    0 drwxr-xr-x   2 root     root           15 Jul  1 15:50 ./data-dir/2
50342804    0 -rw-r--r--   1 root     root            0 Jul  1 15:50 ./data-dir/2/3
16871335    4 -rwxr-xr-x   1 root     root           28 Jul  1 15:50 ./run-me
67117612    0 drwxr-xr-x   1 root     root           38 Jul  1 15:50 ./scripts
16871334    4 -rwxr-xr-x   1 root     root           38 Jul  1 15:50 ./scripts/run

$ docker run my-app ls -l /tmp/assemble-runtime-done
-rw-r--r--. 1 root root 0 Jun 15 10:00 /tmp/assemble-runtime-done

$ docker inspect my-app
...
    "Config": {
        "Hostname": "",
        "Domainname": "",
        "User": "jboss",
        "AttachStdin": false,
        "AttachStdout": false,
        "AttachStderr": false,
        "Tty": false,
        "OpenStdin": false,
        "StdinOnce": false,
        "Env": [
            "PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin",
            "JAVA_HOME=/usr/lib/jvm/java"
        ],
        "Cmd": [
            "/opt/jboss/run"
        ],
        "Image": "",
        "Volumes": null,
        "WorkingDir": "/opt/jboss",
        "Entrypoint": null,
        "OnBuild": null,
        "Labels": {
            "build-date": "2016-03-31",
            "io.k8s.display-name": "my-app",
            "io.openshift.s2i.build.image": "openshift/wildfly-100-centos7",
            "io.openshift.s2i.build.source-location": "/data/src/github.com/openshift/source-to-image/mystamps",
            "license": "GPLv2",
            "name": "CentOS Base Image",
            "vendor": "CentOS"
        }
    },
...
```

Trello card: https://trello.com/c/MWUAjl6k/237-13-extended-builds-separate-mvn-vs-jboss-evg

PTAL @bparees @mfojtik 
CC @PI-Victor @rhcarvalho 

Documentation will be in follow-up PRs.


rhcarvalho:
@php-coder when you say "`cp /artifacts/dir/* /dst`", do you have in mind copying hidden files as well?

rhcarvalho:
> It's not clear where to put files (there many TODOs for that, right now I'm using /tmp)

@php-coder could we use docker named volumes?

rhcarvalho:
@php-coder does the `assemble` script really need to know about `S2I_ARTIFACTS_DIR`? Does it mean we'd have to go change all of our images to support this?

rhcarvalho:
> Should we fail when --runtime-image and --incremental were specified in the same time?

@php-coder why can't we do an incremental build, followed by this new runtime build?

php-coder:
> does the assemble script really need to know about S2I_ARTIFACTS_DIR?

Yes, it should (at least in the current implementation). Without this we won't know what files we need to copy to runtime image.

Actually I can imagine other ways to achieve the same goal:
- user provides mapping for files
- we get the output of `docker diff`, find new files, filter out some files (based on user list?) and copy the resulted files (but as far I can understand, we don't have analogue of `docker diff` in the API. Correct me if I'm wrong).

> Does it mean we'd have to go change all of our images to support this?

Yes. I don't see another way.

The only good news here is that we can use the same `assemble` script for the regular build and the new one (`if [ -n "${S2I_ARTIFACTS_DIR:-}" ]; then ...; fi`).

php-coder:
2 @rhcarvalho:

>when you say "cp /artifacts/dir/* /dst", do you have in mind copying hidden files as well?

Ah, I completely forgot about them. Probably, they aren't get copied now. I added this to the todo-list. Thank you!

>could we use docker named volumes?

Can't answer right now, I need time to look at it. Added to todo. Thanks!

php-coder:
> why can't we do an incremental build, followed by this new runtime build?

Good question. AFAIR I've heard something about this on our planing. But right now I don't see why they couldn't work together.

@bparees Ben, do you remember something about this? Probably, I understood or remembered this incorrectly.

soltysh:
I'd like to be included in the review process for this feature :smile:, @php-coder ping me when you'll be closing this.

bparees:
> mount this directory as volume to the builder container

how will this work in openshift?  we aren't going to be able to mount volumes into containers, you'd have to create volumes on the host and we're not going to be allowing that, it's a security problem.

Per the original card, the pattern for extracting the assembled artifacts was going to follow the pattern used by save-artifacts, which streams the content out of the container, not writes it to a volume.  Or we should follow the pattern used by the "input from an image" feature:
1) run the assemble container, commit it
2) run a second build that extracts input files from image in (1) and injects them into the runtime image and commits it.

As for how incremental build fits into this...  that's an interesting question.  If we do the second approach, then it's a natural fit:
1) retrieve previous "builder" image (the image that contains the built artifacts and the compiler)
2) run save-artifacts in that image, which will dump the saved artifacts into the working dir, along with the new source
3) run the base builder image assemble script with the context from (2), committing the new "builder" image.
4) run the runtime image assemble process, injecting the content from the image from (3).


> It's not clear where to put files (there many TODOs for that, right now I'm using /tmp)

the files should go to the same place we put save-artifacts files.  the working directory where we cloned the source repository.

> Should we use environment variables from the builder for the runtime image?

for now yes, let's assume env variables defined on the buildconfig get provided to both the build and runtime image.

> It's not clear should we have all these labels on runtime image or not? I'm going to remove io.openshift.tags but I'm not sure about other

which labels specifically are you concerned with?  and where/when are you removing them?  There's going to be a base runtime image and it will have its own appropriate io.openshift.tags label, independent of the io.openshift.tags label on the builder image.

>  I'm going to add io.openshift.s2i.runtime-image label to resulted image. Any objections?

yes, there's an objection.  There will be a runtime image explicitly created (eg an EAP image that does not contain maven).  That image should already have the io.openshift.s2i.runtime-image label, if we want to identify "runtime" images explicitly.  It's not something you should be adding when you build the application image.

> S2I_ARTIFACTS_DIR -- maybe rename it? Or it's ok?

per the start of this comment, i think we need to talk through how you're extracting the built artifacts from the builder container and injecting them into the runtime container, i don't think we're on the same page right now.

> Should we fail when --runtime-image and --incremental were specified in the same time?

not necessarily.  --incremental means "reuse the artifacts from a previous iteration of the builder image" which is possible if we actually commit the builder container.  Again we need to discuss the flow.  If we lose the ability to do incremental builds when doing extended builds, we're going to make a lot of people very unhappy.

> Should we support --use-config also?

i don't care about --use-config, pretty sure it's basically unused.

>  Should we show runtime image version in describer?

yes

> Permissions: currently artifacts owned by USER from the image but run script owned by root. Should we change something here?

we need to be running fix-permissions on anything we put into the runtime image, just like we do during the assemble process today, to ensure that everything is owned by the "root" group and the "root" group has rwx permissions.  That will ensure the runtime container uid can access the files/dirs.

pkg/build/strategies/sti/sti.go
```diff
@@ -196,6 +205,15 @@ func (builder *STI) Prepare(config *api.Config) error {
 		WorkingDir: config.WorkingDir,
 	}
 
+	if len(config.RuntimeImage) > 0 {
+		// For pulling RuntimeImage currently we're using the same PullPolicy as for BuilderImage
+		_, err := dockerpkg.PullImage(config.RuntimeImage, builder.docker, config.BuilderPullPolicy, false)
```
wanghaoran1988: 1. why we need pull this image at prepare function , is it possible pull this image after the first build complete ? If the first build failed, this pull is really no need.
2. Or if you want to pull this image at the prepare phase , is it possible for you to make this pull in a new gorouting in a asynchronous way ?

jorgemoralespou:
@bparees Looking at the PR I have some comments (not code related, of course):
- Can we make copy ENV from builder into runtime optional? I recall having issues with using PROXY ENV in build time that got injected into the runtime image and then image fail to be transferred to environments with different proxy setup.
- Commiting the first container should also be optional if incremental is true, as otherwise there's not any point into commit. Does commit also mean push to registry? As I understand it, yes to support incremental, hence only if Incremental we should commit/push as build times will be much better.
- We need to have different PulPolicys, add both images to triggers for builds. If builder changes, rebuild. If runtime changes, rebuild.
- How are we going to do actual assembly of the build and configuration files into the runtime? Will the runtime still have some assembly scripts? Will configuration, modules, etc... in git source repository be also copied from builder into runtime so application image can be properly assembled?
- Is io.openshift.s2i.runtime-image label required for the runtime image? We will not support having as runtime any legacy (non-s2i image)? If the assembly/copy of artifacts from source to image is done like in imagesource (selectively) we could use any non-s2i-runtime image. The user would require to add configuration however they wanted in this case.

bparees:
> Can we make copy ENV from builder into runtime optional? I recall having issues with using PROXY ENV in build time that got injected into the runtime image and then image fail to be transferred to environments with different proxy setup.

How would you feel about the env variables only applying to the builder image and not to the runtime image at all?  Today the fact that the env variables end up in the final application image is an unavoidable side-effect of us making them available during the build.  For extended builds we won't have that limitation, so we could declare the behavior of buildconfig env variables to be:

1) for non-extended builds, they'll do what they do today (end up in app image)
2) for extended builds, they will not end up in the runtime(app) image

or we can do what you're asking and make it optional to propagate them to the app image, but i kinda hate to add more options.  The expectation would be if you have env variables associated with an image, you configure them in the deploymentconfig rather than baking them into the image.

> Commiting the first container should also be optional if incremental is true, as otherwise there's not any point into commit. Does commit also mean push to registry? As I understand it, yes to support incremental, hence only if Incremental we should commit/push as build times will be much better.

both the commit and the push of the builder image should be skipped if the build is not specified as incremental.  if it is incremental, both would be done.

> We need to have different PulPolicys, add both images to triggers for builds. If builder changes, rebuild. If runtime changes, rebuild.

yes we'll probably need an additional pull policy field so we can have distinct policies for the build+runtime image.  As for triggers, today you can setup your own ICT for any image you want, but yes it would be good if the default ICT (the one where you don't specify the image and it fires based on the From defined in the buildconfig) would now fire when either the build or app image changed.  That should be easy enough.

> How are we going to do actual assembly of the build and configuration files into the runtime? Will the runtime still have some assembly scripts? Will configuration, modules, etc... in git source repository be also copied from builder into runtime so application image can be properly assembled?

yes the runtime image still has to have an assemble script.  As for what is copied, it depends what the builder image assemble process "exports".  It's much like save-artifacts works today.

> Is io.openshift.s2i.runtime-image label required for the runtime image?

no.

jorgemoralespou:
@bparees
I'm happy with all your answers, I think. 
Optional was just in case you wanted to propagate the ENVs. For me the behavior you describe is perfect. 
Default ICT is fine as long as if you define ICT on builder only or runtime only works. 
I think we could think in examples where assembly should not be required, as the target location could be specified in the BC configuration, much like image source does. If there's assembly we can run it as well after copying. And by default it could chippy to where assembly will look for by default. 
I think this could cover most cases.

php-coder:
@bparees Ready for another round of api and code review. PTAL.

@PI-Victor I've added `RuntimeArtifacts` member to the `Config` for specifying list of artifacts and their locations. JFYI.

pkg/build/strategies/sti/sti.go
```diff
@@ -128,7 +137,33 @@ func New(config *api.Config, overrides build.Overrides) (*STI, error) {
 	builder.ignorer = &ignore.DockerIgnorer{}
 	builder.artifacts = builder
 	builder.scripts = builder
+
 	builder.postExecutor = builder
+	builder.postExecutorStage = 0
```
rhcarvalho: 0 is the fields' zero value (type int), why do you need this line?

cmd/s2i/main.go
```diff
@@ -181,6 +181,8 @@ $ s2i build . centos/ruby-22-centos7 hello-world-app
 	buildCmd.Flags().StringVarP(&(oldDestination), "location", "l", "",
 		"DEPRECATED: Specify a destination location for untar operation")
 	buildCmd.Flags().BoolVarP(&(cfg.ForceCopy), "copy", "c", false, "Use local file system copy instead of git cloning the source url")
+	buildCmd.Flags().StringVar(&(cfg.RuntimeImage), "runtime-image", "", "Base image that will be used in the run time")
```
bparees: Image that will be used as the base for the runtime image

cmd/s2i/main.go
```diff
@@ -181,6 +181,8 @@ $ s2i build . centos/ruby-22-centos7 hello-world-app
...
+	buildCmd.Flags().VarP(&(cfg.RuntimeArtifacts), "runtime-artifact", "a", "Specify runtime artifact (file or directory) that will be copied from builder to a runtime image")
```
bparees: Specify a file or directory to be copied from the builder to the runtime image

pkg/api/types.go
```diff
@@ -59,6 +59,15 @@ type Config struct {
 	// BuilderBaseImageVersion provides optional version information about the builder base image.
 	BuilderBaseImageVersion string
 
+	// RuntimeImage specifies the image that will be a base for resulting image
+	// and will be used for running an application. By default, BuilderImage is
+	// used for building and running, but the latter may be overridden.
+	RuntimeImage string
+
+	// RuntimeArtifacts specifies a list source/destination folders that will
```
bparees: just folders or files and folders?  can both the source and destination be files?

pkg/build/strategies/sti/sti.go
bparees: the counter argument is that you don't want to do the build if the runtime image isn't available, so you should pull the runtime image first to make sure all the pre-reqs are available before you start building anything.  I think there's a decent argument for either ordering.

Starting the pull asynchronously and then waiting for it to complete (or checking it if it completed, and reporting any error) when we reach the point of needing it is a good idea, but i'll leave it up to @php-coder whether he thinks it would be too complicated to add that flow right now, it's definitely not a requirement.

pkg/build/strategies/sti/sti.go
```diff
@@ -338,12 +365,18 @@ func (builder *STI) createCommandForResultingImage(location string) string {
 }
 
 func (builder *STI) createLabelsForResultingImage() map[string]string {
+	generatedLabels := util.GenerateOutputImageLabels(builder.sourceInfo, builder.config)
+	if len(builder.config.RuntimeImage) > 0 {
+		// don't add labels from the builder image if resulting image is based on another image
+		return generatedLabels
+	}
```
bparees: are we still adding labels from the base runtime image to the resulting runtime image?

bparees: looks like no.  I think what should happen here is createLabelsForResultingImage should take an image argument and then the commitBuilder step would pass the builder image arg and the commitRuntime step would pass the runtime image.

pkg/build/strategies/sti/sti.go
```diff
@@ -620,11 +646,295 @@ func (builder *STI) Execute(command string, user string, config *api.Config) err
...
+	if err := builder.fs.Mkdir(artifactsDir); err != nil {
+		return fmt.Errorf("Couldn't create directory %q: %v", artifactsDir, err)
+	}
+
+	for _, artifact := range builder.config.RuntimeArtifacts {
+		if err := builder.downloadAndExtractFile(artifact.Source, artifactsDir, ctx.containerID); err != nil {
+			return err
+		}
+
+		// for mapping like "/tmp/foo.txt -> app" we should create "app" and move "foo.txt" to that directory
```
bparees: so the source can be a directory or a file, but the destination is always a directory?  (and the destination is optional if they want it to reside directly in the working dir?).  I think that's fine if so, but let's make sure it's clearly doc'd.  I think it matches what we did when we introduced the original "input content from an image" feature.

pkg/build/strategies/sti/sti.go
```diff
@@ -620,11 +646,295 @@ func (builder *STI) Execute(command string, user string, config *api.Config) err
...
+		dstSubDir := path.Clean(artifact.Destination)
+		if dstSubDir != "." && dstSubDir != "/" {
```
bparees: dstSubDir of "/" should be an error/warning.  no absolute paths allowed since we aren't going to put the files into an absolute path.

pkg/build/strategies/sti/sti.go
```diff
@@ -620,11 +646,295 @@ func (builder *STI) Execute(command string, user string, config *api.Config) err
...
+			new := filepath.Join(artifactsDir, dstSubDir, file)
+			glog.V(2).Infof("Renaming %q to %q", old, new)
+			if err := os.Rename(old, new); err != nil {
+				return fmt.Errorf("Couldn't rename %q -> %q: %v", old, new, err)
+			}
+		}
+
+	}
+
+	// We copy run script to the directory with artifacts to upload files in one shot
```
bparees: this doesn't seem like it belongs here.  what does it have to do with downloading files from the builder image?

pkg/build/strategies/sti/sti.go
```diff
@@ -620,11 +646,295 @@ func (builder *STI) Execute(command string, user string, config *api.Config) err
...
+
+	outReader, outWriter := io.Pipe()
+	defer outReader.Close()
+	defer outWriter.Close()
+
+	errReader, errWriter := io.Pipe()
+	defer errReader.Close()
+	defer errWriter.Close()
+
+	lastFileDstPath := "/tmp/upload-done"
```
bparees: generate a unique filename with some random number in the filename, we don't want to risk this file existing in the user's image.

pkg/docker/docker.go
```diff
@@ -274,21 +304,27 @@ func (d *stiDocker) UploadToContainer(src, dest, name string) error {
...
-			if err := t.StreamFileAsTar(src, filepath.Base(dest), w); err != nil {
+			if err := t.StreamFileAsTarWithCallback(src, filepath.Base(dest), w, walkFn, modifyInplace); err != nil {
 				glog.V(0).Infof("error: Uploading files to container failed: %v", err)
 			}
 		}()
 	}
 	glog.V(3).Infof("Uploading %q to %q ...", src, path)
 	opts := docker.UploadToContainerOptions{Path: path, InputStream: r}
-	return d.client.UploadToContainer(name, opts)
+	return d.client.UploadToContainer(container, opts)
```
bparees: given that this is how we are uploading the content into the container, do we actually still require a "tar" binary in the image?  We used to require tar to decode the tar stream we sent into the container, but it seems like we're no longer doing it that way?

pkg/tar/tar.go
```diff
@@ -64,11 +64,21 @@ type Tar interface {
 	// The file permissions in tar archive will change to 0666.
 	StreamFileAsTar(string, string, io.Writer) error
 
+	// StreamFileAsTarWithCallback streams a single file as a TAR archive into specified
+	// writer. By specifying walkFn you can modify file's permissions in arbitrary way.
+	// If modifyInplace is set to true, file will be copied into a temporary directory before changing its permissions.
```
bparees: if modifyinplace is set to false you mean?

pkg/tar/tar.go
```diff
@@ -64,11 +64,21 @@ type Tar interface {
...
+
 	// StreamDirAsTar streams a directory as a TAR archive into specified writer.
 	// The second argument is the name of the folder in the archive.
 	// All files in the source folder will have permissions changed to 0666 in the
 	// tar archive.
 	StreamDirAsTar(string, string, io.Writer) error
+
+	// StreamDirAsTarWithCallback streams a directory as a TAR archive into specified writer.
+	// By specifying walkFn you can modify files' permissions in arbitrary way.
+	// If modifyInplace is set to true, all the files will be copied into a temporary directory before changing their permissions.
```
bparees: same here

pkg/build/strategies/sti/sti.go
bparees: nm, i guess the ioutil.TempFile below will append a unique suffix?

pkg/tar/tar.go
```diff
@@ -106,36 +134,43 @@ func (t *stiTar) StreamDirAsTar(source, dest string, writer io.Writer) error {
...
-			if err != nil {
-				return err
-			}
-			if info.IsDir() {
-				return os.Chmod(path, 0777)
-			}
-			return os.Chmod(path, 0666)
-		})
+	destDir := source
+	if !modifyInplace {
```
bparees: any particular reason this was needed?

pkg/build/strategies/sti/sti.go
```diff
@@ -269,38 +313,21 @@ func mergeLabels(newLabels, existingLabels map[string]string) map[string]string
...
-	builder.result.Success = true
-	builder.result.ImageID = imageID
-
-	glog.V(1).Infof("Successfully built %s", firstNonEmpty(builder.config.Tag, imageID))
-
-	if builder.incremental && builder.config.RemovePreviousImage {
-		builder.removePreviousImage(previousImageID)
+	for _, step := range stageSteps {
+		if err := step(builder.postExecutorStepsContext); err != nil {
+			glog.V(0).Info("Execution of post execute step failed")
```
bparees: "error: Execution of ......."

pkg/build/strategies/sti/sti.go
```diff
@@ -620,11 +646,295 @@ func (builder *STI) Execute(command string, user string, config *api.Config) err
...
+		glog.V(2).Infof("Uploading file %q -> %q", lastFilePath, lastFileDstPath)
+		if err := builder.docker.UploadToContainerWithCallback(lastFilePath, lastFileDstPath, containerID, setStandardPerms, true); err != nil {
+			return fmt.Errorf("Couldn't upload file (%q -> %q) into container %s: %v", lastFilePath, lastFileDstPath, containerID, err)
+		}
+
+		return err
+	}
+
+	go dockerpkg.StreamContainerIO(outReader, nil, func(message ...interface{}) {
+		glog.V(0).Infof("[runtime image stdout] %s", message...)
```
bparees: v(1) should be the default logging level for non-error/warnings.

bparees:
i need to give this another review but on my first pass i like the refactoring and state machine approach, and the general direction looks good for handling file copying.

pkg/api/types.go
php-coder: Source could be file or folder and destination is always a folder.

pkg/build/strategies/sti/sti.go
php-coder: Ok, I'll remove it.

pkg/build/strategies/sti/sti.go
php-coder: Good suggestion, I'll add this. Thank you!

pkg/build/strategies/sti/sti.go
php-coder: > nm, i guess the ioutil.TempFile below will append a unique suffix?

Yes, but this file is the source file that will be copied into the container to the file with "static" name:
```
I0609 14:57:44.790619 03525 sti.go:877] Uploading file "/tmp/s2i-upload-done559194078" -> "/tmp/upload-done"
```
I'll randomize the name.

pkg/tar/tar.go
php-coder: Ah, yes. (In the first version this parameter had name `copyToTempDir`.)

pkg/api/types.go
jorgemoralespou: What if you want on the copy operation to rename?
This can be quite handy to make the built application named ROOT.[war/ear] in for example wildfly-sti.

pkg/tar/tar.go
php-coder: To be honest, no. I just tried to avoid extra copy of the files (because in my case, files already inside of the temporary directory).

bparees: well avoiding the copy would be the default behavior i'd expect, so the question is under what conditions is it necessary to make the copy?

php-coder: As far I understand, this method was created for uploading build secrets where we can't change permissions of the original files (and hence need to copy them to a temporary directory). @mfojtik will correct me if I'm wrong.

pkg/docker/docker.go
php-coder: > given that this is how we are uploading the content into the container, do we actually still require a "tar" binary in the image?

Great question!

Yes, we do:
1) we need `tar` binary for `save-artifacts` script (to produce tar stream from container)
2) our code for uploading sources and `run`/`assemble`/`usage` still need this, as far I can see (https://github.com/openshift/source-to-image/blob/41e42f9b75b27f2437e2abe9400ce980f58a9743/pkg/docker/docker.go#L513-L516)

We're using `UploadToContainer` method only for uploading build secrets and artifacts into runtime image (it means that runtime image doesn't need to have `tar` but builder image still needs it).

> We used to require tar to decode the tar stream we sent into the container, but it seems like we're no longer doing it that way?

Build process didn't change (that's why we can use incremental build with extended build, for example) and that's why builder image still should have `tar` binary.

pkg/api/types.go
php-coder: > What if you want on the copy operation to rename?

It's not supported by the current implementation. User should rename a file in his `assemble` script.

Also, it's not clear how to handle destination in this case: how I can distinguish between a new file name and directory? For example, in case of "app-1.0-SNAPSHOT.jar" -> "app" how s2i can decide whether "app" is a directory for the artifact or it's new name?

pkg/build/strategies/sti/sti.go
php-coder: You're right. I moved to `startRuntimeImageAndUploadFilesStep()` method.

pkg/build/strategies/sti/sti.go
php-coder: > so the source can be a directory or a file, but the destination is always a directory? (and the destination is optional if they want it to reside directly in the working dir?)

Yes, it's exactly as you said. I also enhanced comment for `RuntimeArtifacts` field and will add it to the documentation.

pkg/build/strategies/sti/sti.go
php-coder: In the current implementation absolute path for destination is allowed but they're treated as the relative path. That happens because we always prepend the destination with a local path.

@bparees WDYT, should we produce warning in this case? Or error a build? Or just document it?

pkg/build/strategies/sti/sti.go
php-coder: @wanghaoran1988 @bparees It's interesting idea but
1) I like the fail-fast approach
2) it looks very logical (at lest for me) to pull the image in `Prepare()` method
3) anyway you still have to pull the image. I wouldn't add async pull because it makes code less obvious but it will cover only this rare case (IMHO)

> i'll leave it up to @php-coder whether he thinks it would be too complicated to add that flow right now, it's definitely not a requirement

We can postpone it and implement later on demand.

pkg/build/strategies/sti/sti.go
bparees: what behavior do we do for the input from image flow?  we should align with that.

php-coder:
>> Can we make copy ENV from builder into runtime optional?
>
> How would you feel about the env variables only applying to the builder image and not to the runtime image at all?

@bparees In the current implementation runtime image has the same environment as builder image. Should I change it and don't copy build environment?

>> We need to have different PulPolicys, add both images to triggers for builds. If builder changes, rebuild. If runtime changes, rebuild.
>
> yes we'll probably need an additional pull policy field so we can have distinct policies for the build+runtime image.

@bparees What did we decide on that? Should I add this field or leave as-is?

>> How are we going to do actual assembly of the build and configuration files into the runtime? Will the runtime still have some assembly scripts? Will configuration, modules, etc... in git source repository be also copied from builder into runtime so application image can be properly assembled?
>
> yes the runtime image still has to have an assemble script. As for what is copied, it depends what the builder image assemble process "exports". It's much like save-artifacts works today.

In the current implementation, runtime image doesn't need to have an assemble script. If application requires additional configuration files and modules they should be specified as `--runtime-artifact` options.

CC @jorgemoralespou

bparees:
> @bparees In the current implementation runtime image has the same environment as builder image. Should I change it and don't copy build environment?

we still need a way to provide environment for the runtime steps.  i think for now at least, any environment variables provided by the user need to be made available to both the builder container and the runtime container.  Which means ultimately they end up in the committed image, for better or worse.

> @bparees What did we decide on that? Should I add this field or leave as-is?

let's add the field.

> In the current implementation, runtime image doesn't need to have an assemble script. 

hmm.  this is something i clearly missed in reviewing your PR.  If the runtime image doesn't have an assemble script, how do the artifacts get moved to the right location within the runtime image?

are you saying the assemble script is optional, or there is never one?

pkg/build/strategies/sti/sti.go
bparees: so it turns out we are now defaulting the loglevel to 0, not 2, so anything we want printed all the time does need to be v(0).  sorry.

(whether this is actually something we want to print all the time, though, is another question...)

php-coder:
>we still need a way to provide environment for the runtime steps.

User can set any environment variables in deployment config or in in the image itself. Is it not enough?

>>@bparees What did we decide on that? Should I add this field or leave as-is?
>
> let's add the field.

Ok, I will add it.

>> In the current implementation, runtime image doesn't need to have an assemble script.
>
> hmm. this is something i clearly missed in reviewing your PR. If the runtime image doesn't have an assemble script, how do the artifacts get moved to the right location within the runtime image?

s2i copies these artifacts. We have a mapping of the source file/directory (in the builder image) and destination directory (in the runtime image).

> are you saying the assemble script is optional, or there is never one?

assemble script required only for a building inside builder image (like it was before).

bparees:
> User can set any environment variables in deployment config or in in the image itself. Is it not enough?

no, i'm talking about needing to set an env variable to control the behavior of the assemble logic that's executing in the runtime container during the build process.  (the one that we apparently don't have in this implementation :) )

> s2i copies these artifacts. We have a mapping of the source file/directory (in the builder image) and destination directory (in the runtime image).


hm.  that means the runtime image cannot do anything to prepare/configure itself.  i think that's going to be an unacceptable limitation.  we need to:
1) start the runtime container
2) copy the artifacts in from the build image
3) invoke the assemble script
4) commit the runtime container

similar to what we do today for a normal s2i flow.

My assumption was you were going to dump the artifacts into a working directory within the runtime container, just like we do with the source code today for a normal s2i flow, and then let the assemble script move it to the right location and do other things.

php-coder:
> hm. that means the runtime image cannot do anything to prepare/configure itself.

Do you have cases in mind that demonstrate such behavior? Because from my point of view, most of the cases don't need it and just copying the artifacts should be enough (WAR file for JEE app or JAR file in case of using  Spring Boot app).

>we need to:
>1) start the runtime container
>2) copy the artifacts in from the build image
>3) invoke the assemble script
>4) commit the runtime container

We only miss the step 3 here.

A couple of question then:
- how the runtime assemble script will be named?
- will it be the same as assemble script? (and it could be obtained from the same locations?)
- is this script optional or mandatory?

bparees:
> Do you have cases in mind that demonstrate such behavior? Because from my point of view, most of the cases don't need it and just copying the artifacts should be enough (WAR file for JEE app or JAR file in case of using Spring Boot app).

if i'm reading the code correctly, you're always uploading to the image "workdir" so the artifacts would still need to be moved to the proper deployment directory.

but even if the user could control the absolute destination path, files may need to be renamed, config files may need to be modified from what is provided in the runtime image, secondary preparation steps may need to be performed on the build artifacts such as extraction, obsfucation, etc.

> how the runtime assemble script will be named?
will it be the same as assemble script? (and it could be obtained from the same locations?)
is this script optional or mandatory?

I think we keep the same assumptions we make for a builder image.  I should literally be able to use the runtime image as an s2i builder image.  (which is a likely use case for something like a wildfly runtime-only image...  someone shows up having built their .war file elsewhere, and they just want to build a new app image with that.  That's something they can do today w/ the existing wildfly s2i image, but they end up with an image that contains maven.)

The only exception is that the script could be optional, but for now i think we can keep things simple and leave it mandatory.

but it will not be the same content as the assemble script for the "builder" image, it's a separate assemble script, specifically written for the runtime image.

PI-Victor:
i hate to be that guy, but what about plural ```--runtime-artifacts``` instead?

php-coder:
>but what about plural --runtime-artifacts instead?

When you're using `--runtime-artifact` you're specifying mapping for just one file and that's why it named in a singular form. Also you can specify `--runtime-artifact` more than once: `-a /tmp/.i-am-runtime-artifact3.txt:app -a /tmp/run-me -a /tmp/data-dir:.`

php-coder:
About separate pull policy for runtime image: I added a field for that. I didn't add an option to s2i, because I'm not sure whether we need it and also because I don't know how to name this option (because `--pull-policy` already used for a builder image pull policy).

bparees:
> About separate pull policy for runtime image: I added a field for that. I didn't add an option to s2i, because I'm not sure whether we need it and also because I don't know how to name this option (because --pull-policy already used for a builder image pull policy).

sounds good.

jorgemoralespou:
@php-coder I've tried with given instructions on top, but fails:

````
$ s2i build . openshift/wildfly-100-centos7 my-app --runtime-image jboss/base-jdk:8 --copy --assemble-user=1000 -a /tmp/i-am-runtime-artifact1.txt:app -a /tmp/i-am-runtime-artifact2:app -a /tmp/.i-am-runtime-artifact3.txt:app -a /tmp/run-me -a /tmp/data-dir

warning: Image c638f9a406496a819dcd49ec496b388704826382e834c417886702a15b3e49ee does not contain a value for the io.openshift.s2i.scripts-url label
I0614 00:38:39.420572 28800 install.go:251] Using "assemble" installed from "<source-dir>/.s2i/bin/assemble"
I0614 00:38:39.420647 28800 install.go:251] Using "run" installed from "<source-dir>/.s2i/bin/run"
I0614 00:38:39.420671 28800 install.go:251] Using "save-artifacts" installed from "image:///usr/libexec/s2i/save-artifacts"
E0614 00:38:39.420926 28800 main.go:344] An error occurred: failed to install [assemble-runtime]
E0614 00:38:39.420935 28800 main.go:345] Suggested solution: provide URL with Source-To-Image scripts with -s flag or check the image if it contains io.openshift.s2i.scripts-url label set
E0614 00:38:39.420939 28800 main.go:351] If the problem persists consult the docs at https://github.com/openshift/source-to-image/tree/master/docs. Eventually reach us on freenode #openshift or file an issue at https://github.com/openshift/source-to-image/issues providing us with a log from your build using --loglevel=3
`````

I guess it's missing `assemble-runtime`

**Comment by php-coder** [comment]
@jorgemoralespou Thank you for your early testing!

Yes, after talking with @bparees I've added (just day ago) support for `assemble-runtime` script. Unfortunately, I didn't update description yet, because I'm still working on this.

How it will work? In the same was as we are using `assemble` for builder image, we will use `assemble-runtime` script for a preparation of the runtime image. After copy artifacts from a builder image to a runtime image we will invoke `assemble-runtime` and only after that we will commit a runtime image.

`assemble-runtime` script can be:
1) supplied with runtime image (like `assemble` script supplied with a builder image)
2) downloaded from URL
3) provided by user (`.s2i/bin/assemble-runtime`)

In time of writing only 3 is supported (other cases will be implemented shortly).

> I guess it's missing assemble-runtime

Right now I suggest to use a workaround by providing no-op `assemble-runtime`:
```sh
$ cd /path/to/repo.git
$ mkdir -p .s2i/bin
$ echo '#!/bin/sh' >.s2i/bin/assemble-runtime
$ chmod +x .s2i/bin/assemble-runtime
```

php-coder:
@jorgemoralespou PR's description updated.

php-coder:
I added support for getting files mapping from the `io.openshift.s2i.assemble-required-files` label on the runtime image.

@PI-Victor From now `config.RuntimeArtifacts` becomes optional value.

@bparees Are you agree on name of the label `io.openshift.s2i.assemble-required-files`? It should be on a runtime image because these files required by the `assemble-runtime` script that is also can be embedded into a runtime image. Also we can specify multiple files by using `;` as a separator: `/tmp/i-am-runtime-artifact1.txt:app;/tmp/run-me;/tmp/data-dir:.`

pkg/build/strategies/sti/postexecutorstep.go
```diff
@@ -0,0 +1,467 @@
...
+	workDir, err := step.docker.GetImageWorkdir(image)
+	if err != nil {
+		return fmt.Errorf("Couldn't get working dir of %q image: %v", image, err)
+	}
+
+	cmd := filepath.Join(workDir, api.Run)
+
+	ctx.labels = createLabelsForResultingImage(step.builder, step.docker, image)
+
+	ctx.imageID, err = commitContainer(step.builder, step.docker, ctx.containerID, cmd, user, step.builder.env, ctx.labels)
```
php-coder: Here we're committing runtime image with build environment.

@bparees As far I remember you talked that we shouldn't do that, right? Build environment should only be available during `assemble-runtime` script execution and not during the `run` script execution. Build environment should only be available during `assemble-runtime` script execution and not during the `run` script execution.

pkg/build/strategies/sti/postexecutorstep.go
```diff
@@ -0,0 +1,467 @@
...
+		}
+		commandBaseDir = strings.TrimPrefix(scriptsURL, "image://")
+	}
+
+	waitForFile := fmt.Sprintf("while [ ! -f %q ]; do sleep 0.5; done", lastFileDstPath)
+	runScript := fmt.Sprintf("%s/%s", commandBaseDir, api.AssembleRuntime)
+	rememberExitCode := "result=$?"
+	cleanup := fmt.Sprintf("rm -f -- %q", lastFileDstPath)
+	if useExternalAssembleScript {
+		cleanup += fmt.Sprintf(" %q", runScript)
```
php-coder: I decided to remove `assemble-runtime` (when it was external) script after its execution.

@bparees Are you agree with that?

pkg/build/strategies/sti/postexecutorstep.go
```diff
@@ -0,0 +1,467 @@
...
+		PullImage:       false, // The PullImage is false because we've already pulled the image
+		CommandExplicit: []string{"/bin/sh", "-c", cmd},
+		Stdout:          outWriter,
+		Stderr:          errWriter,
+		NetworkMode:     string(step.builder.config.DockerNetworkMode),
+		CGroupLimits:    step.builder.config.CGroupLimits,
+		CapDrop:         step.builder.config.DropCapabilities,
+		PostExec:        step.builder.postExecutor,
+		Env:             step.builder.env,
+		User:            "root", // without this we won't be able to remove last file
```
php-coder: @bparees under what user we should run runtime image with `assemble-runtime` script? Currently it runs under `root` which also allows us to remove temporary file (that is used as signal that uploading done).

pkg/build/strategies/sti/postexecutorstep.go
```diff
@@ -0,0 +1,467 @@
...
+	cleanup := fmt.Sprintf("rm -f -- %q", lastFileDstPath)
+	if useExternalAssembleScript {
+		cleanup += fmt.Sprintf(" %q", runScript)
+	}
+	exit := "exit $result"
+	cmd := strings.Join([]string{waitForFile, runScript, rememberExitCode, cleanup, exit}, "; ")
+
+	opts := dockerpkg.RunContainerOptions{
+		Image:           image,
+		PullImage:       false, // The PullImage is false because we've already pulled the image
```
wanghaoran1988: I think even if the images is pulled before , we should set true here in case of the image gc comes in the gaps before we use this image, WDYT ?

wanghaoran1988:
Seems incremental build have some problems:
[haoran@cheetah tmp.2uRZc1BpjA]$ ls -rlt .s2i/bin/
total 16
-rw-rw-r--. 1 haoran haoran 416 Jun 17 13:47 assemble
-rw-rw-r--. 1 haoran haoran  81 Jun 17 13:47 assemble-runtime
-rw-rw-r--. 1 haoran haoran  38 Jun 17 13:48 run
-rw-rw-r--. 1 haoran haoran  47 Jun 17 14:37 save-artifacts
[haoran@cheetah tmp.2uRZc1BpjA]$ cat .s2i/bin/save-artifacts 
#!/bin/bash
set -eu
echo "I am save artifact"

logs:https://paste.fedoraproject.org/380392/14661474/

When we do a incremental build, save-artifact script is not correctly located ,so the container for copy artifacts failed to start

I have a questions for the artifacts , will we copy the artifacts like jar package download from maven repo to the final app image ?

php-coder:
@wanghaoran1988 Thank you for testing, I'll take a look.

pkg/api/types.go
```diff
@@ -59,6 +62,24 @@ type Config struct {
...
+	// used for building and running, but the latter may be overridden.
+	RuntimeImage string
+
+	// RuntimeImagePullPolicy specifies when to pull a runtime image.
+	RuntimeImagePullPolicy PullPolicy
+
+	// RuntimeArtifacts specifies a list of source/destination pairs that will
+	// be copied from builder to a runtime image. Source can be a file or
+	// directory. Destination must be a directory. Regardless whether it
+	// absolute or relative path, it will be placed into image's WORKDIR.
```
PI-Victor: ```*it is an ```
i'm a grammar nazi, i know.

php-coder:
@PI-Victor 

>*it is an

Fixed, thank you!

>i'm a grammar nazi, i know.

Actually I was waiting for someone who will correct my English.

php-coder:
@wanghaoran1988 

>When we do a incremental build, save-artifact script is not correctly located ,so the container for copy artifacts failed to start

I've found a couple of issues here:

1) we're trying to run `save-artifacts` script on the wrong image:

> I0617 14:48:00.082180 02048 sti.go:379] Saving build artifacts from image my-app to path /tmp/s2i190813125/upload/artifacts

Here we're runing `my-app` image but it's based on runtime image and doesn't contain `save-artifacts` script. I've fixed this issue to use builder image for such cases.

2) But it continues to fail with the same error on a builder image. I suspect that the current version of S2I is broken for the case of using incremental build with `save-artifacts` from user's repo. Could you test this assumption on a version from master branch?

3) Also I see some design misses here.

@bparees Looks like, when we're using incremental build and we're specifying runtime image, we need to provide 2 tags: one for builder image and another one for a resulted image. Earlier we could use just one tag, because resulted image was based on a builder image, but from now we can't do it because it will break next incremental build then. Do you agree? Or I'm missing something?

> I have a questions for the artifacts , will we copy the artifacts like jar package download from maven repo to the final app image ? 

Probably no. We'll copy only what user told us to copy (`--runtime-artifact`) or what the runtime image expects to be present (`io.openshift.s2i.assemble-required-files` label).

wanghaoran1988:
@php-coder the latest s2i works well with incremental build, see the log:
https://paste.fedoraproject.org/381973/

php-coder:
@wanghaoran1988 Are you using the same `save-artifacts` script as before? Because it should print message "I am save artifact" and there is no such message in the build log.

wanghaoran1988:
@php-coder I am using https://github.com/wanghaoran1988/nodejs-ex/blob/master/.sti/bin/save-artifacts

pkg/build/strategies/sti/postexecutorstep.go
```diff
@@ -0,0 +1,467 @@
...
+}
+
+type removePreviousImageStep struct {
+	builder *STI
+	docker  dockerpkg.Docker
+}
+
+func (step *removePreviousImageStep) execute(ctx *postExecutorStepContext) error {
+	if step.builder.incremental && step.builder.config.RemovePreviousImage {
+		glog.V(1).Info("Executing step: remove previous image")
```
php-coder: @bparees After reading @smarterclayton logging rules, I'm starting to doubt whether we should show name of executed post step on level 1.

pkg/build/strategies/sti/sti.go
php-coder: @bparees:

>dstSubDir of "/" should be an error/warning. no absolute paths allowed since we aren't going to put the files into an absolute path

Now it's an error.

>> In the current implementation absolute path for destination is allowed but they're treated as the relative path. That happens because we always prepend the destination with a local path.
>
>what behavior do we do for the input from image flow? we should align with that.

I've copied behavior of input from image validation: https://github.com/openshift/origin/blob/d9dfa1005fe2684f722072bdd7e46a299d3c9a2d/pkg/build/api/validation/validation.go#L317-L325

bparees:
> @bparees Are you agree on name of the label io.openshift.s2i.assemble-required-files? It should be on a runtime image because these files required by the assemble-runtime script that is also can be embedded into a runtime image. Also we can specify multiple files by using ; as a separator: /tmp/i-am-runtime-artifact1.txt:app;/tmp/run-me;/tmp/data-dir:.

@php-coder so that label would take the place of needing to define the from/to files in the buildconfig?  but if you define them in the buildconfig (or in an s2i argument) it would override the value in this label?  I think i'd prefer "assemble-input-files".  Required seems like too strong a statement.

> @bparees Looks like, when we're using incremental build and we're specifying runtime image, we need to provide 2 tags: one for builder image and another one for a resulted image. Earlier we could use just one tag, because resulted image was based on a builder image, but from now we can't do it because it will break next incremental build then. Do you agree? Or I'm missing something?

I think you're correct.  If you want to be able to do an incremental flow, you need to tag the builder image output separately.  So we'd want an optional s2i flag which would be something like --builder-image-tag and we can tag the builder with that when building, and use that tag as the previous image when doing an incremental build.

I'm not overly concerned with supporting that in openshift, at least as a first pass, though.  but eventually we'll probably need the ability to let openshift buildconfigs also specify the "builder image tag" to push the builder image to.

pkg/build/strategies/sti/postexecutorstep.go
bparees: agree, probably level 3.

pkg/build/strategies/sti/postexecutorstep.go
bparees: i don't think you have any control over it, do you?  I thought whatever env was passed to the container when it was launched, is going to be in any image you commit from that container.

that's why our docs currently say:

> You can add environment variables to the sourceStrategy definition of the BuildConfig. The environment variables defined there are visible during the assemble script execution and will be defined in the output image, making them also available to the run script and application code.

but yes, if you can avoid having the assemble-time env variables end up in the final image, that would be preferred.

pkg/build/strategies/sti/postexecutorstep.go
bparees: we're not doing that for any other scripts, so i don't think we should start introducing unique behavior here.

pkg/build/strategies/sti/postexecutorstep.go
bparees: it needs to run as the default image user (and in openshift, just like we do for the builder image, we are going to need to confirm that the default image user is numeric and not zero).  We can't let users provide arbitrary scripts that get run as root.

if that means we can't remove the marker file, we can live with that, but how was this solved when @mfojtik used similar logic to inject and delete secrets?

bparees:
@php-coder i'd like to start seeing some examples of how a user uses this..  sample command lines and described behavior for things like:  incremental build.  overriding the assemble-runtime script. etc.

pkg/build/strategies/sti/postexecutorstep.go
```diff
@@ -0,0 +1,464 @@
...
+
+	image := step.builder.config.RuntimeImage
+	workDir, err := step.docker.GetImageWorkdir(image)
+	if err != nil {
+		return fmt.Errorf("Couldn't get working dir of %q image: %v", image, err)
+	}
+
+	commandBaseDir := workDir
+	if !useExternalAssembleScript {
+		// script already inside of the image
```
bparees: can i provide assemble-runtime in my git repo in the .s2i/bin directory, like i can with assemble and run?

php-coder: > can i provide assemble-runtime in my git repo in the .s2i/bin directory, like i can with assemble and run?

@bparees Yes, of course.

php-coder:
>> @bparees Are you agree on name of the label io.openshift.s2i.assemble-required-files? It should be on a runtime image because these files required by the assemble-runtime script that is also can be embedded into a runtime image.
>
>@php-coder so that label would take the place of needing to define the from/to files in the buildconfig? but if you define them in the buildconfig (or in an s2i argument) it would override the value in this label?

Yes, we're using labels from an image only when user didn't define them:
```go
// user didn't specify mapping, let's take it from the runtime image then
if len(builder.config.RuntimeArtifacts) == 0 {
	mapping, err := builder.docker.GetAssembleRequiredFiles(config.RuntimeImage)
	if err != nil {
		return err
	}
	if len(mapping) == 0 {
		return fmt.Errorf("No runtime artifacts to copy were specified (neither in the command line nor as image's label)")
	}
...
}
```

> I think i'd prefer "assemble-input-files". Required seems like too strong a statement.

Ok, I've updated the code and rename it.

>> @bparees Looks like, when we're using incremental build and we're specifying runtime image, we need to provide 2 tags: one for builder image and another one for a resulted image. [...] Do you agree?
>
>I think you're correct. If you want to be able to do an incremental flow, you need to tag the builder image output separately. So we'd want an optional s2i flag which would be something like --builder-image-tag and we can tag the builder with that when building, and use that tag as the previous image when doing an incremental build.

Maybe it make sense to introduce `--runtime-image-tag` option instead? It will be mandatory when `--runtime-image` is used. This allows us to minimize the changes in an existing workflow/code.

>@php-coder i'd like to start seeing some examples of how a user uses this.. sample command lines and described behavior for things like: incremental build. overriding the assemble-runtime script. etc.

They will be added as part of documentation. Or you would like to see something else?

pkg/build/strategies/sti/postexecutorstep.go
php-coder: >i don't think you have any control over it, do you?

I tested it and, indeed, I can commit with empty environment but the variables still available when we're running image.

pkg/build/strategies/sti/postexecutorstep.go
php-coder: >if that means we can't remove the marker file, we can live with that, but how was this solved when @mfojtik used similar logic to inject and delete secrets?

The secrets were created with world-writable permissions and once we're done, we're truncated these secrets. So, they're included in the resulting image but they're empty.

Ok, I will modify the logic as you said.

bparees:
> Maybe it make sense to introduce --runtime-image-tag option instead? It will be mandatory when --runtime-image is used. This allows us to minimize the changes in an existing workflow/code.

i don't think so, because as a user i would expect similar to the existing behavior, the the arg i specify as the output image name, is my final image tag.

today:
s2i repo builderimage outputimage

w/ extended:
s2i repo builderimage outputimage --runtime-image=runtimeimage
w/ extended and tag for incremental:
s2i repo builderimage outputimage --runtime-image=runtimeimage --builder-image-tag=sometag

(those are not exact commands of course, but hopefully you get the idea...  we want outputimage to retain the same meaning it had previously).

> They will be added as part of documentation. Or you would like to see something else?

i was just looking for a rough summary from you to help me understand what's implemented here because trying to visualize it by reviewing the code is proving difficult.

php-coder:
>>Maybe it make sense to introduce --runtime-image-tag option instead? It will be mandatory when --runtime-image is used. This allows us to minimize the changes in an existing workflow/code.
>
>i don't think so, because as a user i would expect similar to the existing behavior

@bparees Ok, I'll add `--builder-image-tag` option.

bparees:
fyi we discussed incremental extended builds on the arch call this morning and clayton was of the opinion, at least for openshift builds, that if you want incremental behavior, then you setup 2 builds and chain them, manually.  We won't make all that config part of a new api.

The same argument could be made for s2i.

jorgemoralespou:
@bparees but this is because Clayton says so or for a good reason?
I don't think that incremental by itself should stop being a first clad citizen of this new extended build neither in s2i or openshift as I see it key for local developer experience in CDK/ADB for speeding up respeated builds, which is something that could/should be enabled by BuildOverrides. It will not be used in cloud environments, where you want reproducibility for your builds and build time is not so important (with the exception of online). If we just make for the use case of incremental the need to define two chained builds we should then just kill incremental at all.

bparees:
@jorgemoralespou it can't be enabled via buildoverrides, because the user still would have to specify an explicit builder image tag to use for each buildconfig.  

and for local developer experience, I wouldn't even expect you to use an extended build...if i'm doing local dev, why do i even care if maven is in my final application image?  extended builds are a consideration for production builds.

php-coder:
>if you want incremental behavior, then you setup 2 builds and chain them, manually. We won't make all that config part of a new api.

@bparees Does it mean that I should add a code to detect the case of using `--incremental` and `--runtime-image` together and fail with a message?

bparees:
@php-coder no.  we should still look for a previous builder image if the user specifies incremental, we just don't necessarily need to explicitly offer the ability to tag the new builder image, for now.  

Since you already have implemented it, i'm not suggesting you remove it, but for openshift we won't expose it.

php-coder:
>we just don't necessarily need to explicitly offer the ability to tag the new builder image, for now. Since you already have implemented it, i'm not suggesting you remove it, but for openshift we won't expose it.

I implemented the logic but it works wrong because I didn't add `--builder-image-tag` so now I'm trying to understand a new way -- should I implement this option or I should remove the part of tagging builder image?

bparees:
@php-coder let's discuss on scrum.

jorgemoralespou:
@bparees I totally disagree with you. (If you allow me to).
I'm just saying that what should be enabled by build overrides is the ability to set the incremental flag on. Whether it's an extended build our not shouldn't matter. 
And regarding the "extended builds are for piod". I think that builds in Dev should be as close as period as possible and now extended make it more possible as the builder image and hence all the cached artifacts will never promote. This is why is not only a period but a developers thing, and very critical IMHO.
Inmutability can come even closer to the dev.

bparees:
> @bparees I totally disagree with you. (If you allow me to).

@jorgemoralespou i can't stop you from disagreeing with me :)

> I'm just saying that what should be enabled by build overrides is the ability to set the incremental flag on. 

i have no issue w/ a feature request to make the incremental flag something that can be either defaulted or overridden, in the same way that forcePull can be overridden today and proxies can be defaulted.

> And regarding the "extended builds are for piod". I think that builds in Dev should be as close as period as possible and now extended make it more possible as the builder image and hence all the cached artifacts will never promote. 

yes and no.  plenty of people are going to be doing dev builds w/ debugging/test libraries enabled/added, etc.  those are things they are going to turn off before they do the production build and those are not artifacts you will want to promote into production.

someone who really cares can also create a builder image that includes their dependencies, circumventing the need for incremental builds.

Finally, if you choose an incremental extended build, then every build must:
1) pull the previous builder image that was produced (because it may not exist on this node)
2) push the new builder image that this build has produced (to ensure the next build has somewhere to pull it from)

neither of those steps are needed if you don't choose incremental.

So whether that is cheaper than re-pulling your dependencies from maven/etc on each build is debatable anyway, especially if you have a local maven mirror.

The point here is this is not requisite function to deliver a useful extended builds feature, and given time/resource constraints, it doesn't make sense to try to pursue it right now.  We may revisit it in the future.

php-coder:
For the record: we decided to fail with error when both `--runtime-image` and `--incremental` options were specified.

jorgemoralespou:
@bparees I think that when you talk about developers your talking about development in openshift cloud while I'm talking about development on local laptop (cdk/adb) where until we don't have mechanisms to cache dependencies, incremental is the best way to provide an acceptable developer experience.
We need to keep focus on local development experience. 
That's why there's a trello card (request) to allow incremental as a build override, although not groomed yet.

bparees:
as i said, i have no issue w/ adding incremental build as something that can be configured as a buildefault or override.  but
1) that is a separate discussion from whether we support it for extended builds or not (and we are not going to resolve that question in this PR, we are quickly running out of time for 3.3 to get this feature landed, any optional additional functionality is being punted for now)
2) i don't think anything i described above is specific to the online experience, it is all just as applicable to someone using the CDK/ADB.  And I stand by the statement that if you are doing CDK/ADB development, you probably don't care about having maven in your final image.  You are not promoting an image out of the CDK and into production.

php-coder:
@bparees Prepared for merge. PTAL.

pkg/build/strategies/sti/postexecutorstep.go
```diff
@@ -0,0 +1,459 @@
...
+	// id of the previous image that we're holding because after committing the image, we'll lose it.
+	// Used only when build is incremental and RemovePreviousImage setting is enabled.
+	// See also: rememberPreviousImageStep and removePreviousImageStep
+	previousImageID string
+
+	// Container id that will be committed.
+	// See also: commitBuilderImageStep/commitRuntimeImageStep
+	containerID string
+
+	// See also: commitBuilderImageStep/commitRuntimeImageStep
```
bparees: location of what?

pkg/build/strategies/sti/postexecutorstep.go
```diff
@@ -0,0 +1,459 @@
...
+	location string
+
+	// Container id after committing.
```
bparees: image id created by committing the container.

pkg/build/strategies/sti/postexecutorstep.go
```diff
@@ -0,0 +1,459 @@
...
+	// See also: commitBuilderImageStep/commitRuntimeImageStep and reportAboutSuccessStep
+	imageID string
+
+	// Labels that will be passed to a callback.
+	// It's labels that is used during commit of builder image (or runtime image if it's used).
```
bparees: labels that are added to the builder or runtime image during commit?

pkg/build/strategies/sti/postexecutorstep.go
```diff
@@ -0,0 +1,459 @@
...
+	// See also: commitBuilderImageStep/commitRuntimeImageStep and invokeCallbackStep
+	labels map[string]string
+}
+
+type postExecutorStep interface {
+	execute(*postExecutorStepContext) error
+}
+
+type rememberPreviousImageStep struct {
```
bparees: storePreviousImageIDStep ?

bparees: (remember can refer to either storing, or recalling, a value)

pkg/build/strategies/sti/postexecutorstep.go
php-coder: Two steps (`commitBuilderImageStep` and `commitRuntimeImageStep`) use this field. So, which exactly labels will be hold here depends on the how workflow is configured.

How I can improve this comment then?...

bparees: sorry my question was intended as a suggestion.
"// these labels are added to the builder or runtime image during commit"

assuming that is accurate.

pkg/build/strategies/sti/sti.go
```diff
@@ -198,6 +221,47 @@ func (builder *STI) Prepare(config *api.Config) error {
 		WorkingDir: config.WorkingDir,
 	}
 
+	if len(config.RuntimeImage) > 0 {
+		if config.Incremental {
+			return fmt.Errorf("Incremental build with runtime image isn't supported")
```
bparees: this should have been sorted out earlier when validating the CLI options in main.go.

pkg/build/strategies/sti/postexecutorstep.go
php-coder: I'll add the comment and rename the field like this:
```go
	// Path to a directory in the image where scripts (for example, "run") will be placed.
	// This location will be used for generation of the CMD directive.
	// See also: commitBuilderImageStep/commitRuntimeImageStep
	destination string
```

pkg/build/strategies/sti/sti.go
php-coder: But in this case it wouldn't fail for invocation inside of OpenShift, right? I tried to catch a both cases and I didn't find a better place for that.

Are you still thinking that this check should be moved to `main.go`?

pkg/build/strategies/sti/postexecutorstep.go
bparees: it's also the directory where the source code will be placed, right? isn't it basically the working directory for assembly?

pkg/build/strategies/sti/sti.go
bparees: in openshift we need to be catching that in api validation.  we don't want to catch an error like this at runtime.

pkg/build/strategies/sti/postexecutorstep.go
php-coder: Yes, it is. It also the value of `io.openshift.s2i.destination` label.

>isn't it basically the working directory for assembly?

I'm not sure, maybe it is.

pkg/build/strategies/sti/sti.go
php-coder: >in openshift we need to be catching that in api validation. we don't want to catch an error like this at runtime.

Ok, done.

bparees:
> commits builder container (only when it's also increment build)

now that we're disallowing incremental, do we ever commit the builder container?

bparees:
i don't think i have any other comments, i'm going to fetch this branch and play with the cli this afternoon, but hopefully i'll apply the merge tag tonight.  the summary/sample output was super helpful, thank you.

php-coder:
To @bparees:

>> commits builder container (only when it's also increment build)

BTW, I removed this item from the description.

>now that we're disallowing incremental, do we ever commit the builder container?

We never committing a builder container in case of extended build.

The important part of the configuration code is the `STI.initPostExecutorSteps()` method, here we're set up our steps depending on whether this an extended build or not. For extended build we don't add `storePreviousImageStep`/`commitBuilderImageStep` and `removePreviousImageStep` method to the "chain" and hence they won't be executed.

php-coder:
@bparees I've added a picture with build flow. I'm going to add it to documentation. Meanwhile you can review it also :)

@soltysh I remember that you wanted to do a last review and now is a good time because this PR is on the final stage.

pkg/build/strategies/sti/postexecutorstep_test.go
```diff
@@ -0,0 +1,346 @@
...
+		t.Errorf("should commit container with User: %q, but commited with %q", expectedImageUser, commitOpts.User)
+	}
+
+	if !reflect.DeepEqual(commitOpts.Labels, expectedLabels) {
+		t.Errorf("should commit container with Labels: %v, but commited with %v", expectedLabels, commitOpts.Labels)
+	}
+}
+
+func TestDownloadFilesFromBuilderImageStep(t *testing.T) {
+	// FIXME
```
rhcarvalho: Will this go like this?

php-coder: I hope to that :) I can remove these templates but I believe that it's still good to have tests for these steps. In other hand, writing such tests is hard and I don't have time for them now. I decided to leave here these "reminders".

Do you think that it's really bad?

rhcarvalho: No opinion, just did a quick look and thought maybe this was forgotten here.

jorgemoralespou:
@bparees @php-coder I personally don't like the assemble-runtime name for the assemble script in the runtime image. Not sure if this is possible or there is some backwards compatibility that I'm missing but I would rather have a "build" or "make" script in the builder and assemble in the runtime, which sounds more intuitive for the developer.

bparees:
@jorgemoralespou i don't love it either, in fact my preference would be to call it assemble just like in the builder image, but that runs into problems when a user wants to provide both a builder assemble override script and a runtime assemble override script in the same repo (.s2i/bin directory).

But i'd prefer to keep the name associated with assemble because an assemble-runtime script has all the same responsibilities as the existing assemble scripts, so it is more logical imho to give it a similar name.

Trying to refactor the names at this point, especially if we're going to reuse "assemble" but for a different purpose would be extremely confusing to users.  Remember, all these scripts will continue to live in a user's .s2i/bin directory if they are overriding things, so now you'd have a the user's .s2i/bin/assemble script suddenly.... not used?  used for extended builds?  used as the builder script for normal builds and the runtime assemble script for runtime builds?

I think we can manage to educate users on the naming convention.

pkg/build/strategies/sti/postexecutorstep.go
```diff
@@ -0,0 +1,461 @@
...
+		glog.V(5).Infof("Uploading file %q -> %q", lastFilePath, lastFileDstPath)
+		if err := step.docker.UploadToContainerWithCallback(lastFilePath, lastFileDstPath, containerID, setStandardPerms, true); err != nil {
+			return fmt.Errorf("Couldn't upload file (%q -> %q) into container %s: %v", lastFilePath, lastFileDstPath, containerID, err)
+		}
+
+		return err
+	}
+
+	go dockerpkg.StreamContainerIO(outReader, nil, func(message ...interface{}) {
+		glog.V(0).Infof("[runtime image stdout] %s", message...)
```
bparees: if we're going to do this, we should do it on the builder image also for consistency.

pkg/build/strategies/sti/postexecutorstep.go
```diff
@@ -0,0 +1,461 @@
...
+	})
+
+	errOutput := ""
+	go dockerpkg.StreamContainerIO(errReader, &errOutput, func(message ...interface{}) {
+		glog.Errorf("[runtime image stderr] %s", message...)
```
bparees: same here

pkg/build/strategies/sti/postexecutorstep.go
bparees: (and i suspect @smarterclayton is going to dislike that prefix, it's going to make the logging kind of ugly.  let's remove it and just make sure we print a header that indicates the following output is from assemble-runtime (and same for assemble).

i think the fact that error output is prefixed with "ERROR" is probably sufficient to distinguish it (as opposed to saying "runtime image stderr")

bparees:
couple final nits and then lgtm.

jorgemoralespou:
I may have not expressed correctly (as usual). I would keep assemble for
the runtime which is the one really assembling the configuration and
artifact into the runtime  app (server or whatever) and I would add an
additional build that was the script executed in the first image (the
builder).
This allows current s2i images to do all in one step for backwards
compatibility and if spliting, then first image would execute just the
build.
I think this language is more natural. But maybe only to me.

bparees:
@jorgemoralespou what you're describing would actually prevent using the existing eap/wildfly/etc images from being used as "builder" images in an extended build scenario since they do not contain the "build" script, unless we introduced fallback logic to first check for a build script and then check for an assemble script, at which point the overriding logic becomes very confusing (what do you do if the image contains a build script and the user provided an assemble script?  which is more important?)

I suspect people will want to use those images in that way, at least until we have a more mature set of 'builder only" images.

I think that is the exact scenario you're describing, so i may be misunderstanding your point.

pkg/build/strategies/sti/postexecutorstep.go
php-coder: >if we're going to do this, we should do it on the builder image also for consistency.

Good idea. Do you want to have it as part of this PR?

>and i suspect @smarterclayton is going to dislike that prefix, it's going to make the logging kind of ugly.

I don't mind to change the prefix.

>let's remove it and just make sure we print a header that indicates the following output is from assemble-runtime

What if script won't produce any output? In this case we'll print useless header and other log message could be treated as assemble-runtime output.

Also this messages are printed asynchronously and one-by-one. It's possible that there will be one steder line, after that stdout line and then from stderr again. In this case I don't see a better way except adding a prefix.

> i think the fact that error output is prefixed with "ERROR" is probably sufficient to distinguish it

May be. But how user distinguish between in what container this error happened? And was it in the container or it's from S2I itself?

bparees: > What if a script won't produce any output? In this case we'll print useless header and other log messages could be treated as assemble-runtime output.

print a footer too.

> Also this messages are printed asynchronously and one-by-one. It's possible that there will be one stderr line, after that stdout line and then from stderr again. In this case I don't see a better way except adding a prefix.

the stderr output is still going to have the "ERROR:" prefix introduced by glog.

> May be. But how user distinguish between in what container this error happened? And was it in the container or it's from S2I itself?

because the message will appear in between a header/footer that say:
```
=== Begin Builder assemble output ====
i'm a stdout message from assemble
ERROR: i'm a stderr message from assemble
=== End Builder assemble output ====
```
or
```
=== Begin Runtime assemble output ====
i'm a stdout message from assemble-runtime
ERROR: i'm a stderr message from assemble-runtime
=== End Runtime assemble output ====
```

php-coder: >because the message will appear in between a header/footer that say

I'm afraid it could lead to a mess in case of, for example, running `mvn compile`.

(I'll test this case on Thursday.)

bparees: how so?

jorgemoralespou:
@bparees AFAIK current eap/wildfly images should work. If in single
builder+runtime image s2i would call the build script and as there's
nothing would not do anything, then the assemble would build + assemble as
now. If in extended build, builder would try to build, such if there's a
build script would then build otherwise use the default for the builder. In
case of maven,build the war file. This file would be streamed into the
runtime where assemble would run, as there's a binary would pick it and not
build again with current assemble across in wildfly/eap. Don't know other
languages so this might not work.

bparees:
I think i stand by my original arguments: assemble and assemble-runtime are similar in intent/mission, so giving them a shared name makes sense.

introducing a hierarchy of scripts that may or may not be called depending which mode you're in strikes me as harder for users to understand:  builder image w/ assemble script invoked using extended build:  run the assemble script?  run the user provided build script if any?  run both?

>  such if there's a build script would then build otherwise use the default for the builder. In case of maven,build the war file.

that is no "default for the builder" there must be a script to be executed.  so either there has to be an "assemble" script or there has to be a "build" script or the user has to be providing one.

pkg/build/strategies/sti/postexecutorstep.go
php-coder: I like idea of using prefix. Just need to change it. Maybe use script name for that? And each line will be like this:
```
ERROR: assemble: permission denied
```
or
```
assemble-runtime: compiling binaries
```
WDYT?

bparees: i still think the prefix will be ugly when the only output is the script output (which is basically all you get if you run s2i with default logging levels)

php-coder: Maybe we can omit the prefix when default logging level is used and add it for verbose levels?

php-coder: BTW, with default logging level there is 2 empty lines that wasn't here before. I suspect that they were added by last logging changes..

bparees: yes it was, to separate the script output from the s2i engine output.  a header/footer would do the job nicely too :)

i'm not that worried about cases where the script has no output.  does not seem likely to be a real world case.

as for omiting the prefix if the loglevel is < 1 , i could be convinced to go that route.

bparees:
@php-coder hitting an issue trying to use this:
```
E0704 23:08:21.508564   23398 main.go:365] An error occurred: Couldn't copy file ("/tmp/s2i380176718/upload/scripts/run" -> "/tmp/s2i380176718/upload/runtimeArtifacts/run"): open /tmp/s2i380176718/upload/scripts/run: no such file or directory
```

my source repo does not provide any script files.  both my builder image and my runtime image do contain a "run" script though of course my builder image should not need one.

it seems like maybe something in the code is now expecting that the repo itself includes a run script always.

bparees:
looks like there is an assumption in the runtime image construction flow that presumes the "run" script is coming from an external source and is not part of the runtime image itself.

php-coder:
>looks like there is an (invalid) assumption in the runtime image construction flow that presumes the "run" script is coming from an external source and is not part of the runtime image itself.

Thanks for testing and discovering this issue! It's fixed now.

pkg/build/strategies/sti/postexecutorstep.go
php-coder: I've pushed the changes where I removed prefixes from the messages to output them as-is. I think that we can improve logging later (maybe after some feedback). WDYT?

bparees:
[merge]

bparees:
[merge]

php-coder:
>Documentation will be in follow-up PRs.

https://github.com/openshift/source-to-image/pull/536

PR #509, opened 2016-05-27
Pass --env variables to the build container

mnagy:
This fixes a bug introduced in https://github.com/openshift/source-to-image/pull/501

@bparees @php-coder

### DIFF
--- a/pkg/build/strategies/sti/sti.go
+++ b/pkg/build/strategies/sti/sti.go
@@ -317,7 +317,6 @@ func (builder *STI) createBuildEnvironment() []string {
 	env, err := scripts.GetEnvironment(builder.config)
 	if err != nil {
 		glog.V(1).Infof("No user environment provided (%v)", err)
-		return nil
 	}
 
 	return append(scripts.ConvertEnvironment(env), builder.generateConfigEnv()...)
--- a/pkg/build/strategies/sti/sti_test.go
+++ b/pkg/build/strategies/sti/sti_test.go
@@ -758,6 +758,11 @@ func TestExecuteOK(t *testing.T) {
 	rh.config.WorkingDir = "/working-dir"
 	rh.config.BuilderImage = "test/image"
 	rh.config.BuilderPullPolicy = api.PullAlways
+	rh.config.Environment = api.EnvironmentList{
+		api.EnvironmentSpec{"Key1", "Value1"},
+		api.EnvironmentSpec{"Key2", "Value2"},
+	}
+	expectedEnv := []string{"Key1=Value1", "Key2=Value2"}
 	th := rh.tar.(*test.FakeTar)
 	th.CreateTarResult = "/working-dir/test.tar"
 	fd := rh.docker.(*docker.FakeDocker)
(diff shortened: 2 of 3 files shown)


pkg/build/strategies/sti/sti_test.go
```diff
@@ -808,6 +813,9 @@ func TestExecuteOK(t *testing.T) {
 		t.Errorf("PostExecutor not called with expected ID: %s",
 			pe.PostExecuteContainerID)
 	}
+	if !reflect.DeepEqual(ro.Env, expectedEnv) {
+		t.Errorf("Unexpected container environment passed to RunContainer: %v, should be %v", ro.Env, rh.config.Environment)
```
php-coder: Why we're comparing with `expectedEnv` but in the message showing `rh.config.Environment`? It's for readability or by mistake?

php-coder:
Thank you for finding and fixing it! LGTM.

pkg/build/strategies/sti/sti_test.go
mnagy: Mistake

mnagy:
Fixed the error message, good catch :)

pkg/build/strategies/sti/sti.go
```diff
@@ -317,7 +317,6 @@ func (builder *STI) createBuildEnvironment() []string {
 	env, err := scripts.GetEnvironment(builder.config)
 	if err != nil {
 		glog.V(1).Infof("No user environment provided (%v)", err)
-		return nil
```
bparees: What happens when a legitimate error occurs in GetEnvironment and "env" is returned as nil?  the append is going to fail, right?

you need to at least initialize env if it came back nil.

bparees: and print the error as a warning, if not an outright failure.

bparees: nm, i guess the error will be printed as info.  still need to init env though.

php-coder: AFAIR `scripts.ConvertEnvironment` is able to handle `nil` as an argument.

bparees: i guess maybe since ConvertEnvironment has a named return value, "result" will be initialized and returned and the range loop will no-op since env is nil.

feels side-effecty to me, but i guess you are correct that it'll work.

php-coder: I've just tested it and it works (see https://play.golang.org/p/4RiId5_kaV).

bparees:
lgtm [merge]

pkg/build/strategies/sti/sti.go
mnagy: The function createBuildEnvironment() was created by taking code that was duplicated, with the return nil added, I'm assuming by mistake. So my change just returns it into the previous state.

php-coder: Yes, it was by mistake and didn't discover it because on my test project I have `.s2i/environment` file.

PR #511, opened 2016-06-02
tar: fix panic when dir is symlink

csrwng:
Avoids writing header when directory being tarred itself is a symlink


csrwng:
@gabemontero ptal

pkg/tar/tar.go
```diff
@@ -198,6 +198,9 @@ func (t *stiTar) CreateTarStreamWithLogging(dir string, includeDirInPath bool, w
 		if !info.IsDir() && !t.shouldExclude(path) {
 			// if file is a link just writing header info is enough
 			if info.Mode()&os.ModeSymlink != 0 {
+				if dir == path {
```
bparees: is this only a problem when the root dir is a symlink?  or do we need to also address symlinks within the root dir that are pointing to dirs?

csrwng: those cause no issues - they get extracted on the other end as links.

bparees:
lgtm [merge]

PR #513, opened 2016-06-07
tar(StreamDirAsTar,StreamFileAsTar): fix possible fd leak and handle error from Stat()

php-coder:
PTAL @bparees

### DIFF
--- a/pkg/tar/tar.go
+++ b/pkg/tar/tar.go
@@ -91,17 +91,21 @@ func (t *stiTar) SetExclusionPattern(p *regexp.Regexp) {
 	t.exclude = p
 }
 
-// StreamFileAsTar streams the source file as a tar archive.
+// StreamFileAsTar streams the source directory as a tar archive.
 // The permissions of the file is changed to 0666.
 func (t *stiTar) StreamDirAsTar(source, dest string, writer io.Writer) error {
 	f, err := os.Open(source)
 	if err != nil {
 		return err
 	}
-	if info, _ := f.Stat(); !info.IsDir() {
+	defer f.Close()
+	info, err := f.Stat()
+	if err != nil {
+		return err
+	}
+	if !info.IsDir() {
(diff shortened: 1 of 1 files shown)


pkg/tar/tar.go
```diff
@@ -91,17 +91,21 @@ func (t *stiTar) SetExclusionPattern(p *regexp.Regexp) {
...
 func (t *stiTar) StreamDirAsTar(source, dest string, writer io.Writer) error {
 	f, err := os.Open(source)
 	if err != nil {
 		return err
 	}
-	if info, _ := f.Stat(); !info.IsDir() {
+	defer f.Close()
+	info, err := f.Stat()
+	if err != nil {
+		return err
```
rhcarvalho: I don't know how this error gets exposed to the user, maybe it's possible to wrap it like below to include some more context.

php-coder: I handled it in the same way as we did for `os.Open()`. I believe that it's enough.

rhcarvalho: I think a useful context would be something like:

```go
return fmt.Errorf("streaming directory as tar: %v", err)
```

But depends on how this pops up to the end user, and if we want to be able to handle it from callers of `StreamDirAsTar`.

bparees: i think it's reasonable to make callers of StreamFileAsTar do the logging of the error if they so choose, which is the current behavior for errors on open as @php-coder noted anyway.

bparees:
lgtm [merge]

mfojtik:
I would prefer to holding off merging this if there is not significant bug associated with this till the rebase lands.

bparees:
@mfojtik how does this affect the rebase?  this is in the s2i repo.

mfojtik:
@bparees omg, ignore me... I thought this is a bump PR :-( oh my, I should get on PTO.

php-coder:
[test] please

PR #518, opened 2016-06-16
Minor improvements

php-coder:
A set of minor improvements:
- `VolumeList.Set()`: raise error for an empty string
- `fs.Copy()`: fix possible fd leak and explicitly ignore err
- `docker.GetScriptsURL()`: unify error messages and show image id
- improves error messages in some tests
- replaces sti by s2i in some places
- fixes a couple of typos
- improves godoc

PTAL @bparees

### DIFF
--- a/pkg/api/types.go
+++ b/pkg/api/types.go
@@ -130,7 +130,7 @@ type Config struct {
 	// CallbackURL is a URL which is called upon successful build to inform about that fact.
 	CallbackURL string
 
-	// ScriptsURL is a URL describing the localization of STI scripts used during build process.
+	// ScriptsURL is a URL describing the localization of S2I scripts used during build process.
 	ScriptsURL string
 
 	// Destination specifies a location where the untar operation will place its artifacts.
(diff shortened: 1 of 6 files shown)


pkg/api/types.go
```diff
@@ -410,10 +410,11 @@ func IsInvalidFilename(name string) bool {
 // When the destination is not specified, the source get copied into current
 // working directory in container.
 func (l *VolumeList) Set(value string) error {
+	if len(value) == 0 {
+		return fmt.Errorf("invalid format, must be source:destination")
+	}
 	mount := strings.Split(value, ":")
 	switch len(mount) {
-	case 0:
-		return fmt.Errorf("invalid format, must be source:destination")
```
php-coder: It turns out that `strings.Split("", ":")` returns slice with single element (instead of empty slice).

pkg/docker/docker.go
```diff
@@ -463,20 +463,21 @@ func getScriptsURL(image *docker.Image) string {
 	if len(scriptsURL) == 0 {
 		scriptsURL = getLabel(image, "io.s2i.scripts-url")
 		if len(scriptsURL) > 0 {
-			glog.V(0).Infof("warning: The 'io.s2i.scripts-url' label is deprecated. Use %q instead.", ScriptsURLLabel)
+			glog.V(0).Infof("warning: Image %s uses deprecated label 'io.s2i.scripts-url', please migrate it to %s instead!",
+				image.ID, ScriptsURLLabel)
```
php-coder: Here we also include `image.ID` into the message. Unfortunately it will be SHA1 sum of the image, because on that level we don't have a human readable name, but only object with image data.

BTW it will be especially useful when we'll have not only builder image but also a runtime image.

bparees:
squash and lgtm
[test]

php-coder:
Squashed.

bparees:
[merge]

bparees:
@php-coder please look into this flake when you have a chance:
--- FAIL: TestLayeredBuildNoTar (0.86s)
	integration_test.go:316: An error occurred during the build: non-zero (13) exit code from 

[merge]

PR #519, opened 2016-06-19
Changing messaging to Clayton's spec

jupierce:
Attempting to address concerns from openshift/origin#9247

### DIFF
--- a/pkg/docker/docker.go
+++ b/pkg/docker/docker.go
@@ -347,16 +347,30 @@ func (d *stiDocker) GetOnBuild(name string) ([]string, error) {
 // and returns the image metadata
 func (d *stiDocker) CheckAndPullImage(name string) (*docker.Image, error) {
 	name = getImageName(name)
+	displayName := name
+
+	if !glog.Is(3) {
+		// For less verbose log levels (less than 3), shorten long iamge names like:
+		//     "centos/php-56-centos7@sha256:51c3e2b08bd9fadefccd6ec42288680d6d7f861bdbfbd2d8d24960621e4e27f5"
+		// to include just enough characters to differentiate the build from others in the docker repository:
+		//     "centos/php-56-centos7@sha256:51c3e2b08bd..."
+		// 18 characters is somewhat arbitrary, but should be enough to avoid a name collision.
+		split := strings.Split(name, "@")
+		if len(split) > 1 && len(split[1]) > 18 {
+			displayName = split[0] + "@" + split[1][:18] + "..."
+		}
+	}
+
 	image, err := d.CheckImage(name)
 	if err != nil && err.(errors.Error).Details != docker.ErrNoSuchImage {
 		return nil, err
 	}
 	if image == nil {
-		glog.V(0).Infof("Image %q not available locally, pulling ...", name)
+		glog.V(1).Infof("Image %q not available locally, pulling ...", displayName)
 		return d.PullImage(name)
 	}
 
-	glog.V(1).Infof("Using locally available image %q", name)
+	glog.V(1).Infof("Using locally available image %q", displayName)
 	return image, nil
 }
 
--- a/pkg/scripts/install.go
+++ b/pkg/scripts/install.go
@@ -248,7 +248,7 @@ func (i *DefaultScriptSourceManager) InstallOptional(scripts []string, dstDir st
 					result = append(result, *r)
 					installed = true
 					detected = true
-					glog.Infof("Using %q installed from %q", script, r.URL)
+					glog.V(4).Infof("Using %q installed from %q", script, r.URL)
 				}
 			}
 			if detected {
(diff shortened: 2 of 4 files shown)


pkg/docker/docker.go
```diff
@@ -347,16 +347,26 @@ func (d *stiDocker) GetOnBuild(name string) ([]string, error) {
 // and returns the image metadata
 func (d *stiDocker) CheckAndPullImage(name string) (*docker.Image, error) {
 	name = getImageName(name)
+	displayName := name
+
+	if glog.Is(3) == false {
```
php-coder: Why we don't need this on level 3? Looks like you've tried to write condition like `if logLevel < 3`. If it's true I see that you missed levels 4 and 5.

php-coder: BTW, what is "Clayton's spec" and where I can find it?

csrwng: @php-coder It's comments on this issue: https://github.com/openshift/origin/issues/9247

glog.Is(x) really means loglevel >= x ... not very well named, but it's used correctly above.

@jupierce why not  if !glog.Is(3) ?

csrwng:
We really need to use the imageprogress pull writer as used here:
https://github.com/openshift/origin/blob/master/pkg/util/docker/dockerfile/builder/client.go#L326-L327

We should add an option to the s2i builder config to specify the output writer and raw flag for image pulls and make it available here.

pkg/docker/docker.go
php-coder: @csrwng Thank you for explanation and a link. I was out of context but now it's clear.

pkg/docker/docker.go
```diff
@@ -347,16 +347,26 @@ func (d *stiDocker) GetOnBuild(name string) ([]string, error) {
...
+		// Summarize name for lower level log output
+		split := strings.Split(name, "@")
+		if len(split) > 1 && len(split[1]) > 18 {
```
csrwng: What's the purpose for checking len > 18 ?

csrwng: btw, there's a utility function here that parses the ref here:
https://github.com/openshift/origin/blob/master/pkg/image/api/helper.go#L133

We can't use it from here but maybe we can take the useful bits.

pkg/docker/docker.go
```diff
@@ -347,16 +347,26 @@ func (d *stiDocker) GetOnBuild(name string) ([]string, error) {
...
+			displayName = split[0] + "@" + split[1][:18] + "..."
```
bparees: i assume 18 refers to the id length, but a comment would be good.

bparees:
@jupierce can we get a sample of the default output?

pkg/docker/docker.go
jupierce: @csrwng  Will change to !glog in next commit

pkg/docker/docker.go
jupierce: @csrwng  @bparees  Clayton suggested a "summary" at this logging level. The choice of 18 was somewhat arbitrary: I wanted the user to see enough of the sha for it to be unlikely to correspond to existing images on the system, but shorter than the 80 chars Clayton believed was problematic.
So it changes:
**Image "centos/php-56-centos7@sha256:51c3e2b08bd9fadefccd6ec42288680d6d7f861bdbfbd2d8d24960621e4e27f5" not available locally, pulling ...**
to 
**Image "centos/php-56-centos7@sha256:51c3e2b08bd..." not available locally, pulling ...**

jupierce:
Working on resucitating my golang environment and I will provide an actual log.

pkg/docker/docker.go
bparees: i see.  even more reason for a comment in the code then.

smarterclayton: Note that docker also shortens these in some contexts

jupierce:
s2i runs at all log levels:
https://gist.github.com/jupierce/59413ff40e89127b5b1b9f16c9c62aa4

pkg/scripts/install.go
```diff
@@ -120,7 +120,7 @@ func (s *SourceScriptHandler) Get(script string) *api.InstallResult {
 	// and this should (and will) be removed soon.
 	location = strings.Replace(location, "s2i/bin", "sti/bin", 1)
 	if s.fs.Exists(location) {
-		glog.Info("DEPRECATED: Use .s2i/bin instead of .sti/bin")
+		glog.V(3).Infof("DEPRECATED: Use .s2i/bin instead of .sti/bin")
```
bparees: this one can stay, people should be told if they're using deprecated logic.

jupierce: Restored the deprecated output to use glog.Info.

jupierce:
Changed downloading and setting environment variables to V(1) per @bparees .

bparees:
lgtm [merge]

PR #523, opened 2016-06-23
Minor improvements

php-coder:
- fix typos
- sti -> s2i
- other minor code improvements

PTAL @bparees

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -24,7 +24,7 @@ at common flags that can be used with all of the subcommands.
 
 #### Log levels
 
-There are four log levels:
+There are six log levels:
 * Level `0` - produces output from containers running `assemble` script and all encountered errors
 * Level `1` - produces basic information about the executed process
 * Level `2` - produces very detailed information about the executed process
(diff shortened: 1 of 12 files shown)


pkg/build/strategies/onbuild/entrypoint.go
```diff
@@ -24,7 +24,7 @@ func GuessEntrypoint(fs util.FileSystem, sourceDir string) (string, error) {
 		return "", err
 	}
 	for _, f := range files {
-		if f.IsDir() || !f.Mode().IsRegular() {
+		if !f.Mode().IsRegular() {
```
bparees: why are you dropping this?  a dir isn't going to be a valid entrypoint.

php-coder: Because I decided that directory is not a regular file and single should cover a both cases. Do you agree?

bparees: ok

bparees:
squash and lgtm

php-coder:
Squashed.

bparees:
--- FAIL: TestGuessEntrypoint (0.00s)

	entrypoint_test.go:51: Expected error for ERR:run_is_dir, got run

php-coder:
It's great that we have tests!

bparees:
I don't think your change was necessarily wrong, I think you just needed to update the FakeFile implementation to also set the os.FileMode when IsDir is set to true.

php-coder:
Let's leave it as it was before.

bparees:
[merge]

PR #524, opened 2016-06-23
Migrating s2i to use glog mock library

jupierce:
@smarterclayton @bparees @csrwng 

- Added additional methods to glog mock util to limit impact when porting other libraries currently using standard glog
- Add flushing to file based loggers (doesn't really matter for stderr, but if you changed to use a real file, it could)

Question: I tied --loglevel to the level maintained by the toFile logger created by toFile. This obviates some of the logic of pkg/util/glog.go:V(int32) because file.level will always match vanilla golog's Level. That is, V(#) will never decide to send user output to standard glog instead of the file output.

The alternative, leaving toFile's file.level at a constant 2, didn't seem like a useful notion here *unless* you want golog's prefix to be included on Info and above. However, even if you did, the stack element would be useless, because it would always point to pkg/util/glog.go .

### DIFF
--- a/cmd/s2i/main.go
+++ b/cmd/s2i/main.go
@@ -11,7 +11,8 @@ import (
 	"strings"
 	"time"
 
-	"github.com/golang/glog"
+	log "github.com/golang/glog"
+	utilglog "github.com/openshift/source-to-image/pkg/util/glog"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
 
@@ -31,6 +32,10 @@ import (
 	"github.com/openshift/source-to-image/pkg/util"
 	"github.com/openshift/source-to-image/pkg/version"
 )
+
+// glog is a placeholder until the builders pass an output stream down
+// client facing libraries should not be using glog
+var glog = utilglog.StderrLog
 
 func newCmdVersion() *cobra.Command {
 	return &cobra.Command{
@@ -60,9 +65,7 @@ $ s2i build git://github.com/openshift/ruby-hello-world centos/ruby-22-centos7 h
 $ s2i build . centos/ruby-22-centos7 hello-world-app
 `,
 		Run: func(cmd *cobra.Command, args []string) {
-			if glog.V(1) {
-				glog.Infof("Running S2I version %q\n", version.Get())
-			}
+			glog.V(1).Infof("Running S2I version %q\n", version.Get())
 
 			// Attempt to restore the build command from the configuration file
 			if useConfig {
@@ -134,9 +137,7 @@ $ s2i build . centos/ruby-22-centos7 hello-world-app
 				cfg.Destination = oldDestination
 			}
 
-			if glog.V(2) {
-				fmt.Printf("\n%s\n", describe.DescribeConfig(cfg))
-			}
+			glog.V(2).Infof("\n%s\n", describe.DescribeConfig(cfg))
 
 			if !docker.IsReachable(cfg) {
 				glog.Fatalf("Unable to connect to Docker daemon. Please set the DOCKER_HOST or make sure the Docker socket %q exists", cfg.DockerConfig.Endpoint)
@@ -227,9 +228,7 @@ func newCmdRebuild(cfg *api.Config) *cobra.Command {
 				cfg.PreviousImagePullPolicy = api.DefaultPreviousImagePullPolicy
 			}
 
-			if glog.V(2) {
-				fmt.Printf("\n%s\n", describe.DescribeConfig(cfg))
-			}
+			glog.V(2).Infof("\n%s\n", describe.DescribeConfig(cfg))
 
 			builder, err := strategies.GetStrategy(cfg)
 			checkErr(err)
@@ -324,13 +323,16 @@ func newCmdUsage(cfg *api.Config) *cobra.Command {
 	return usageCmd
 }
 
+// setupGlog makes --loglevel reflect in glog's -v flag
 func setupGlog(flags *pflag.FlagSet) {
+
 	from := flag.CommandLine
 	if fflag := from.Lookup("v"); fflag != nil {
-		level := fflag.Value.(*glog.Level)
-		levelPtr := (*int32)(level)
-		flags.Int32Var(levelPtr, "loglevel", 0, "Set the level of log output (0-5)")
+		level := fflag.Value.(*log.Level)
+		loglevelPtr := (*int32)(level)
+		flags.Int32Var(loglevelPtr, "loglevel", 0, "Set the level of log output (0-5)")
 	}
+
 	// FIXME currently glog has only option to redirect output to stderr
 	// the preferred for S2I would be to redirect to stdout
 	flag.CommandLine.Set("logtostderr", "true")
@@ -358,8 +360,13 @@ func checkErr(err error) {
 
 func main() {
 	rand.Seed(time.Now().UnixNano())
+
+	// Applying partial glog flag initialization workaround from: https://github.com/kubernetes/kubernetes/issues/17162
+	// Without this fake command line parse, glog will compain its flags have not been interpreted
+	flag.CommandLine.Parse([]string{})
+
 	cfg := &api.Config{}
-	stiCmd := &cobra.Command{
+	s2iCmd := &cobra.Command{
 		Use: "s2i",
 		Long: "Source-to-image (S2I) is a tool for building repeatable docker images.\n\n" +
 			"A command line interface that injects and assembles source code into a docker image.\n" +
@@ -369,18 +376,16 @@ func main() {
 		},
 	}
 	cfg.DockerConfig = docker.GetDefaultDockerConfig()
-	stiCmd.PersistentFlags().StringVarP(&(cfg.DockerConfig.Endpoint), "url", "U", cfg.DockerConfig.Endpoint, "Set the url of the docker socket to use")
-	stiCmd.PersistentFlags().StringVar(&(cfg.DockerConfig.CertFile), "cert", cfg.DockerConfig.CertFile, "Set the path of the docker TLS certificate file")
-	stiCmd.PersistentFlags().StringVar(&(cfg.DockerConfig.KeyFile), "key", cfg.DockerConfig.KeyFile, "Set the path of the docker TLS key file")
-	stiCmd.PersistentFlags().StringVar(&(cfg.DockerConfig.CAFile), "ca", cfg.DockerConfig.CAFile, "Set the path of the docker TLS ca file")
-
-	stiCmd.AddCommand(newCmdVersion())
-	stiCmd.AddCommand(newCmdBuild(cfg))
-	stiCmd.AddCommand(newCmdRebuild(cfg))
-	stiCmd.AddCommand(newCmdUsage(cfg))
-	stiCmd.AddCommand(newCmdCreate())
-	setupGlog(stiCmd.PersistentFlags())
-
+	s2iCmd.PersistentFlags().StringVarP(&(cfg.DockerConfig.Endpoint), "url", "U", cfg.DockerConfig.Endpoint, "Set the url of the docker socket to use")
+	s2iCmd.PersistentFlags().StringVar(&(cfg.DockerConfig.CertFile), "cert", cfg.DockerConfig.CertFile, "Set the path of the docker TLS certificate file")
+	s2iCmd.PersistentFlags().StringVar(&(cfg.DockerConfig.KeyFile), "key", cfg.DockerConfig.KeyFile, "Set the path of the docker TLS key file")
+	s2iCmd.PersistentFlags().StringVar(&(cfg.DockerConfig.CAFile), "ca", cfg.DockerConfig.CAFile, "Set the path of the docker TLS ca file")
+	s2iCmd.AddCommand(newCmdVersion())
+	s2iCmd.AddCommand(newCmdBuild(cfg))
+	s2iCmd.AddCommand(newCmdRebuild(cfg))
+	s2iCmd.AddCommand(newCmdUsage(cfg))
+	s2iCmd.AddCommand(newCmdCreate())
+	setupGlog(s2iCmd.PersistentFlags())
 	basename := filepath.Base(os.Args[0])
 	// Make case-insensitive and strip executable suffix if present
 	if runtime.GOOS == "windows" {
@@ -391,8 +396,9 @@ func main() {
 		glog.Warning("sti binary is deprecated, use s2i instead")
 	}
 
-	stiCmd.AddCommand(newCmdGenBashCompletion(stiCmd))
-	err := stiCmd.Execute()
+	s2iCmd.AddCommand(newCmdGenBashCompletion(s2iCmd))
+
+	err := s2iCmd.Execute()
 	if err != nil {
 		os.Exit(1)
 	}
--- a/pkg/build/strategies/onbuild/entrypoint.go
+++ b/pkg/build/strategies/onbuild/entrypoint.go
@@ -5,9 +5,11 @@ import (
 	"path/filepath"
 	"regexp"
 
-	"github.com/golang/glog"
 	"github.com/openshift/source-to-image/pkg/util"
+	utilglog "github.com/openshift/source-to-image/pkg/util/glog"
 )
+
+var glog = utilglog.StderrLog
 
 var validEntrypoints = []*regexp.Regexp{
 	regexp.MustCompile(`^run(\.sh)?$`),
--- a/pkg/build/strategies/sti/sti.go
+++ b/pkg/build/strategies/sti/sti.go
@@ -11,7 +11,6 @@ import (
 	"strings"
 	"sync"
 
-	"github.com/golang/glog"
 	"github.com/openshift/source-to-image/pkg/api"
 	"github.com/openshift/source-to-image/pkg/build"
 	"github.com/openshift/source-to-image/pkg/build/strategies/layered"
@@ -23,7 +22,10 @@ import (
 	"github.com/openshift/source-to-image/pkg/scripts"
 	"github.com/openshift/source-to-image/pkg/tar"
 	"github.com/openshift/source-to-image/pkg/util"
+	utilglog "github.com/openshift/source-to-image/pkg/util/glog"
 )
+
+var glog = utilglog.StderrLog
 
 var (
 	// List of directories that needs to be present inside working dir
@@ -293,7 +295,7 @@ func (builder *STI) PostExecute(containerID, location string) error {
 	builder.result.Success = true
 	builder.result.ImageID = imageID
 
-	glog.V(1).Infof("Successfully built %s", firstNonEmpty(builder.config.Tag, imageID))
+	glog.V(3).Infof("Successfully built %s", firstNonEmpty(builder.config.Tag, imageID))
 
 	if builder.incremental && builder.config.RemovePreviousImage {
 		builder.removePreviousImage(previousImageID)
@@ -316,7 +318,7 @@ func (builder *STI) getPreviousImage() string {
 func (builder *STI) createBuildEnvironment() []string {
 	env, err := scripts.GetEnvironment(builder.config)
 	if err != nil {
-		glog.V(1).Infof("No user environment provided (%v)", err)
+		glog.V(3).Infof("No user environment provided (%v)", err)
 	}
 
 	return append(scripts.ConvertEnvironment(env), builder.generateConfigEnv()...)
@@ -584,12 +586,14 @@ func (builder *STI) Execute(command string, user string, config *api.Config) err
 
 	go func(reader io.Reader) {
 		scanner := bufio.NewReader(reader)
+		// Precede build output with newline
+		glog.Info()
 		for {
 			text, err := scanner.ReadString('\n')
 			if err != nil {
 				// we're ignoring ErrClosedPipe, as this is information
 				// the docker container ended streaming logs
-				if glog.V(2) && err != io.ErrClosedPipe && err != io.EOF {
+				if glog.Is(2) && err != io.ErrClosedPipe && err != io.EOF {
 					glog.Errorf("Error reading docker stdout, %v", err)
 				}
 				break
@@ -598,13 +602,11 @@ func (builder *STI) Execute(command string, user string, config *api.Config) err
 			if config.Quiet {
 				continue
 			}
-			// The log level > 3 forces to use glog instead of printing to stdout
-			if glog.V(3) {
-				glog.Info(text)
-				continue
-			}
-			fmt.Fprintf(os.Stdout, "%s\n", strings.TrimSpace(text))
+			glog.Info(strings.TrimSpace(text))
 		}
+		// Terminate build output with new line
+		glog.Info()
+
 	}(outReader)
 
 	go dockerpkg.StreamContainerIO(errReader, &errOutput, glog.Error)
--- a/pkg/run/run.go
+++ b/pkg/run/run.go
@@ -5,11 +5,13 @@ package run
 import (
 	"io"
 
-	"github.com/golang/glog"
 	"github.com/openshift/source-to-image/pkg/api"
 	"github.com/openshift/source-to-image/pkg/docker"
 	"github.com/openshift/source-to-image/pkg/errors"
+	utilglog "github.com/openshift/source-to-image/pkg/util/glog"
 )
+
+var glog = utilglog.StderrLog
 
 // A DockerRunner allows running a Docker image as a new container, streaming
 // stdout and stderr with glog.
--- a/pkg/scm/git/clone.go
+++ b/pkg/scm/git/clone.go
@@ -3,7 +3,6 @@ package git
 import (
 	"path/filepath"
 
-	"github.com/golang/glog"
 	"github.com/openshift/source-to-image/pkg/api"
 	"github.com/openshift/source-to-image/pkg/errors"
 	"github.com/openshift/source-to-image/pkg/util"
--- a/pkg/util/glog/glog.go
+++ b/pkg/util/glog/glog.go
@@ -1,101 +1,176 @@
 package glog
 
 import (
+	"bufio"
 	"fmt"
 	"io"
+	"os"
 	"strings"
 
-	"github.com/golang/glog"
+	log "github.com/golang/glog"
 )
 
 // Logger is a simple interface that is roughly equivalent to glog.
 type Logger interface {
-	Is(level int) bool
-	V(level int) Logger
+	Is(level int32) bool
+	V(level int32) VerboseLogger
 	Infof(format string, args ...interface{})
+	Info(args ...interface{})
+	Warningf(format string, args ...interface{})
+	Warning(args ...interface{})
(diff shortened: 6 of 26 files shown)


pkg/build/strategies/onbuild/entrypoint.go
```diff
@@ -2,13 +2,14 @@ package onbuild
 
 import (
 	"errors"
+	"github.com/openshift/source-to-image/pkg/util"
+	utilglog "github.com/openshift/source-to-image/pkg/util/glog"
```
bparees: keep these imports separated, our pattern is:

golang system imports

thirdparty imports

k8s imports (not applicable in s2i)

local project imports

eg:
```
"fmt"
	"net"
	"path/filepath"
	"strconv"
	"strings"

	"github.com/golang/glog"

	kapi "k8s.io/kubernetes/pkg/api"
	"k8s.io/kubernetes/pkg/api/resource"
	utilerrs "k8s.io/kubernetes/pkg/util/errors"
	"k8s.io/kubernetes/pkg/util/sets"

	deployapi "github.com/openshift/origin/pkg/deploy/api"
	"github.com/openshift/origin/pkg/generate/app"
	"github.com/openshift/origin/pkg/generate/git"
	templateapi "github.com/openshift/origin/pkg/template/api"
	dockerfileutil "github.com/openshift/origin/pkg/util/docker/dockerfile"
	"github.com/openshift/origin/third_party/github.com/docker/libcompose/project"
```

pkg/build/strategies/sti/sti.go
```diff
@@ -584,12 +586,18 @@ func (builder *STI) Execute(command string, user string, config *api.Config) err
 
 	go func(reader io.Reader) {
 		scanner := bufio.NewReader(reader)
+		// Preceed build output with newline
```
bparees: precede

pkg/build/strategies/sti/sti.go
```diff
@@ -599,12 +607,20 @@ func (builder *STI) Execute(command string, user string, config *api.Config) err
 				continue
 			}
 			// The log level > 3 forces to use glog instead of printing to stdout
-			if glog.V(3) {
+			if glog.Is(3) {
```
bparees: since "glog" is now the wrapper, i don't think this logic is needed anymore..you can just send everything to the glog wrapper and let it decide what to do with it.
(same for the other if checks in this file)

pkg/run/run.go
```diff
@@ -3,14 +3,15 @@
...
 
 import (
-	"io"
-
-	"github.com/golang/glog"
 	"github.com/openshift/source-to-image/pkg/api"
 	"github.com/openshift/source-to-image/pkg/docker"
 	"github.com/openshift/source-to-image/pkg/errors"
+	utilglog "github.com/openshift/source-to-image/pkg/util/glog"
+	"io"
```
bparees: reorder imports

pkg/scm/git/clone.go
```diff
@@ -1,12 +1,10 @@
 package git
 
 import (
-	"path/filepath"
-
-	"github.com/golang/glog"
 	"github.com/openshift/source-to-image/pkg/api"
 	"github.com/openshift/source-to-image/pkg/errors"
 	"github.com/openshift/source-to-image/pkg/util"
+	"path/filepath"
```
bparees: reorder imports

pkg/util/glog/glog.go
```diff
@@ -1,101 +1,224 @@
...
+	Is(level int32) bool
+	V(level int32) VerboseLogger
 	Infof(format string, args ...interface{})
+	Info(args ...interface{})
+	Warningf(format string, args ...interface{})
+	Warning(args ...interface{})
+	Errorf(format string, args ...interface{})
+	Error(args ...interface{})
+	Fatalf(format string, args ...interface{})
+	Fatal(args ...interface{})
```
bparees: i'm guessing there's a reason @smarterclayton didn't do this when he first implemented it, so i'd like him to weigh in before we add these utility methods.

pkg/util/glog/glog.go
```diff
@@ -1,101 +1,224 @@
...
+
+func (f *FileLogger) outputf(format string, args ...interface{}) {
+	f.writeln(fmt.Sprintf(format, args...))
+}
+
+func (f *FileLogger) output(args ...interface{}) {
+	f.writeln(fmt.Sprint(args...))
+}
+
+// Infof records an info log entry
```
php-coder: AFAIR godoc comments should end with dot.

bparees: while technically true per the go guidelines, it is one my least favorite items to nitpick people on.

pkg/util/glog/glog.go
```diff
@@ -1,101 +1,224 @@
...
+type FileLogger struct {
+	w     *bufio.Writer
+	level int32
 }
 
-func (f file) Is(level int) bool {
-	return level <= f.level
+// Is returns whether the current logging level is greater than or equal to the parameter
+func (f *FileLogger) Is(level int32) bool {
+	return level <= atomic.LoadInt32(&f.level)
```
bparees: are we expecting this to be changing? why is the atomic load needed?  (here and elsewhere)

pkg/util/glog/glog.go
```diff
@@ -1,101 +1,224 @@
...
+func (f *FileLogger) writeln(line string) {
+	defer f.w.Flush()
+	if atomic.LoadInt32(&f.level) >= 100 {
+		pc := make([]uintptr, 10)
+		runtime.Callers(4, pc)
+		function := runtime.FuncForPC(pc[0])
+		file, line := function.FileLine(pc[0])
+		s := strings.Split(file, "/") // remove absolute path from file
+		file = s[len(s)-1]
+		f.w.WriteString(fmt.Sprintf("[%s:%d] ", file, line-1))
```
bparees: where did this come from?

pkg/util/glog/glog.go
```diff
@@ -1,101 +1,224 @@
...
+	g.Verbose.Info(args...)
+}
+
+// FileLogger logs the provided messages at level or below to the writer, or delegates
 // to glog.
-type file struct {
-	w     io.Writer
-	level int
+type FileLogger struct {
+	w     *bufio.Writer
```
bparees: why did this need to change?

pkg/util/glog/glog.go
```diff
@@ -1,101 +1,224 @@
...
+	f.outputf(format, args...)
+}
+
+// Info records an info log entry
+func (f *FileLogger) Info(args ...interface{}) {
+	f.output(args...)
+}
+
+// Warningf records an warning log entry
+func (f *FileLogger) Warningf(format string, args ...interface{}) {
```
bparees: if you're going to add these helpers, they should prefex "warning" and "error" appropriately, so that lines like this:
`glog.V(0).Infof("warning: Image %s uses deprecated label 'io.s2i.scripts-url', please migrate it to %s instead!",a,b)`
can be turned into:
`glog.Warningf("Image %s uses deprecated label 'io.s2i.scripts-url', please migrate it to %s instead!",a,b)`

overall i'm inclined to just punt on these helpers and stick w/ the path @smarterclayton took in the original implementation.. if you want to log a warning or an error, log at level 0 and prefix your message with "error" or "warning"

bparees: (that ensures the message content is the same whether it is logged by the file logger, or delegated to glog, aside from glog's timestamp+filename prefixing)

pkg/util/glog/glog.go
php-coder: I'm sorry, I didn't want to nitpick of course. (I just remember this rule from one of my first code reviews.)

I even don't mind to leave it as is.

cmd/s2i/main.go
```diff
@@ -356,28 +358,31 @@ func checkErr(err error) {
...
+	s2iCmd := &cobra.Command{
 		Use: "s2i",
 		Long: "Source-to-image (S2I) is a tool for building repeatable docker images.\n\n" +
 			"A command line interface that injects and assembles source code into a docker image.\n" +
 			"Complete documentation is available at http://github.com/openshift/source-to-image",
+		PersistentPreRun: func(cmd *cobra.Command, args []string) {
+			if fileLogger, ok := glog.(*utilglog.FileLogger); ok {
+				fileLogger.SetLoggerLevel(*loglevelPtr)
+			}
+		},
```
bparees: if i understand this correctly, you're setting the level at which messages go to the file logger (instead of glog) to the loglevelPtr value (which is the --loglevel value).  I don't think that's correct.  The original implementation split the messages at level 2.  (level 2 and below went to the file logger, level 3 and above went to glog).  that should still be the behavior.

however @smarterclayton and I did discuss that rather than splitting where some messages get reported via glog and some get reported via filelogger, it would be better if the wrapper logic said "if the loglevel is higher than 2, send everything to glog.  if the loglevel is <=2, send everything to filelogger", that way you don't have some messages getting glogged and some getting filelogged, in a single execution.  So if you want to implement that behavior, that would be good.  If not, we can do that in a follow up to this PR.

pkg/util/glog/glog.go
bparees: well, you've proven you have good eyesight if nothing else :)

bparees:
sorry probably should have read your PR summary before reviewing, it answers some of the questions i had.

pkg/util/glog/glog.go
bparees: guessing you did this so you can invoke flush, but let me know if there is another reason.

bparees:
> Question: I tied --loglevel to the level maintained by the toFile logger created by toFile. This obviates some of the logic of pkg/util/glog.go:V(int32) because file.level will always match vanilla golog's Level. That is, V(#) will never decide to send user output to standard glog instead of the file output.

> The alternative, leaving toFile's file.level at a constant 2, didn't seem like a useful notion here unless you want golog's prefix to be included on Info and above. 

right idea but wrong choice I think.  what you've done is always send everything to the filelogger.  what we want to do (I think, based on @smarterclayton's original implementation) is say "if the user requested loglevel 2 or below, send all output to the filelogger, if they requested 3 or above, send all output to glog".  Which basically means "if i request level 3 or above, i want glog prefixing on everything"

As you note, the existing implementation doesn't quite do that today, but the right implementation should.

pkg/build/strategies/sti/sti.go
jupierce: Agree in practice, but for this particular case, the Is(3) also protects a continue statement which changes the behavior of the loop.

pkg/util/glog/glog.go
jupierce: Dotted! :-)

pkg/build/strategies/sti/sti.go
bparees: yeah but the only point of the continue statement is to avoid having an "else log to stdout".
right now the code is effectively:
if loglevel==3 {
  log to glog
} else {
  log to stdout
}

where it can just be "log to wrapper"

pkg/util/glog/glog.go
jupierce: Not expected, but not prohibited. If the library found wider/additional use, I wanted to make it threadsafe from the outset.

bparees: premature optimization something something.

seems like we're complicating the code unnecessarily (and possibly performance implications of the atomic operations on every logging call)

pkg/util/glog/glog.go
jupierce: I left it in pending the answer to the glog usage question. If we want *real* filenames and line numbers for verbose logging statements, we can't defer to golog since it will only report pkg/util/glog/glog.go as the source of logging statements.

pkg/util/glog/glog.go
jupierce: Yep, just for the flush.

pkg/util/glog/glog.go
bparees: yeah i kinda like it, i was just curious if it was borrowed from somewhere (ie that this was the accepted right way to get at the information).

pkg/util/glog/glog.go
jupierce: If this is going to be a (near) drop-in replacement for glog, it seems like a shame to lose semantic information like glog.Warning versus glog.Error. If we control the logging implementation, we can enforce consistency with or divergence from glog without any difficulty (we can change filtering, output routing, prefixes, add a timestamp/file/line info, etc, based on our internal needs). Delegation to glog seems unnecessary as it appears to actually interfere with the current desired output format.

To a lesser extent, I think developer convenience is something to consider. glog.V(0).Info("Warning..." ) takes more thought than glog.Warning, will lead to inconsistency ("WARNING:", "Warning:", "warning:", or forgetting it altogether) .. and again, valuable meaning on how to handle the message that we might otherwise leverage in the future will be lost. 

It is also worth considering, barring some complex handling of command line flags, glog is made mostly useless when cobra is used in the client program (-v, -vmodules, etc, don't work).

cmd/s2i/main.go
jupierce: That makes a bit more sense, but I remain concerned on three fronts:

1. There is an override in s2i which forces glog to output to stderr. The comments around it complain that it should be redirected to stdout. https://github.com/jupierce/source-to-image/blob/master/cmd/s2i/main.go#L332-L334 . This again suggest to me that we want control of our own destiny here instead of delegating to glog. 
2. In s2i, glog's capabilities almost completely suppressed because cobra's CLI library prevents important flags from reaching glog (vmodule can't be used, for example, because it is not registered with cobra and would not be passed to glog even if it were). This is why the original code bothered blowing --loglevel into -v: https://github.com/jupierce/source-to-image/blob/master/cmd/s2i/main.go#L327-L331 .
3. utilglog interferes with glog's ability to report the correct file/line number.

Combining the above means glog is mostly useless for delegation. It can only log to stderr (which filelogger is already doing), it can't be configured from the command line, and its prefix is incorrect if delegation is triggered.

jupierce:
I can definitely unwind the choice, but please do consider my longer explanation on why glog delegation seems broken in this model: https://github.com/openshift/source-to-image/pull/524#discussion_r68414014

pkg/build/strategies/sti/sti.go
jupierce: Gotcha. Thanks.

pkg/util/glog/glog.go
jupierce: There are already go routines using the logging library. Without having researched go's memory consistency contracts, I can't promise the outcome will be deterministic. That said, I will remove the guard.
If the desired outcome is the fixed > 2 implementation, this issue is moot since the value will no longer be dynamic.

jupierce:
New push should address:
- formatting imports
- spelling
- comments
- atomic I/O in utilglog

bparees:
delegation breaking the glog filename prefixing is a big problem.  but i'm not clear what you're suggesting to address it?

bparees:
as discussed on irc:
(01:50:00 PM) bparees: claytonc: jupierce: what do we want to do about jupierce's delegation observation?  (that all logging that gets delegated to glog, prints "glog.go:50" as the source line)
(01:50:39 PM) claytonc: bparees: there's a method on glog that should let us call the correct one
(01:50:41 PM) claytonc: depth or something
(01:50:50 PM) claytonc: we should def. fix
(01:52:43 PM) bparees: claytonc: jupierce: looks like https://godoc.org/github.com/golang/glog#InfoDepth

jupierce:
The suggestion is just to handle it ourselves in utilglog (e.g. level>100 snippet). If we can't control glog via command line parameters and don't like that it outputs to stderr, I don't currently understand the utility of delegating. But I'll change things back to the static 2 threshold for filelogger.

cmd/s2i/main.go
bparees: The original goal behind @smarterclayton doing the wrapper was to give us the ability to log without timestamps when s2i is being invoked from within openshift.  Ultimately we'll pick another logging framework that does what we want (ie allows us to better control the prefixing and output target), but i don't think we want to head down the path of implementing our own full logger to get there.

pkg/util/glog/glog.go
bparees: > If this is going to be a (near) drop-in replacement for glog

it's not intended to be that, it's just a hack to deal with getting prefix-less output for our logging statements.

>  it seems like a shame to lose semantic information like glog.Warning versus glog.Error

not sure where you feel we lose that.  per my suggestion, one way or another, "warning" and "error" will be part of the output.  (either because you're going to prefix it in these methods, or because we'll rely on the author writing it as part of their message).

> If we control the logging implementation, we can enforce consistency with or divergence from glog without any difficulty (we can change filtering, output routing, prefixes, add a timestamp/file/line info, etc, based on our internal needs).

we don't want to own the maintenance of a logging framework.

> Delegation to glog seems unnecessary as it appears to actually interfere with the current desired output format.

A good point but addressed via @smarterclayton's Depth suggestion

> To a lesser extent, I think developer convenience is something to consider. glog.V(0).Info("Warning..." ) takes more thought than glog.Warning, will lead to inconsistency ("WARNING:", "Warning:", "warning:", or forgetting it altogether) .. and again, valuable meaning on how to handle the message that we might otherwise leverage in the future will be lost. 

I agree, but my suggestion was that either you need to do the (consistent) prefixing in your wrapper methods, or we make developers do it.  right now neither is happening (Warningf, Errorf, and Logf are outputting the exact same thing)  or am I missing something about your implementation?

> It is also worth considering, barring some complex handling of command line flags, glog is made mostly useless when cobra is used in the client program (-v, -vmodules, etc, don't work). 

it's a good point but we do have ways to expose those flags if we decided it's needed.

jupierce:
- Replaced hardcoding for level 2 in filelogger. 
- Waiting for glog bump in source-to-image in order to leverage Depth golog calls.

jupierce:
Pulled new glog library. Standard usage of the library resulted in a log prefix like:
ERROR: logging before flag.Parse: I0627 10:20:06.100980    3539 .....

I found a hack here: https://github.com/kubernetes/kubernetes/issues/17162 to make glog happy. It results in glog's parameters being accessible on the command line which will likely result in some confusion between specifying -v=<volume> and --v=<glog-level>. 

Using this approach, --loglevel is just a synonym for --v . The command line help for build verb looks like:
https://gist.github.com/jupierce/94b18abeed3e7af641eb7b641a41a37a#file-gistfile1-txt-L38-L45

Notice the highlighted glog parameters which are added into the mix. 

The alternative is simply to trick glog into thinking its parameters have been parsed (and not expose them through to the command line user). 

In summary, I think there are three possible approaches, and only (1) really makes sense to me:
(1) Remove glog delegation and insert our own file/line prefixes as part of utilglog
(2) Add glog parameters to the s2i command line parameter set, confusing -v|--v|--loglevel   (this approach is reflected in the current PR state)
(3) Trick glog into thinking its parameters have been parsed and prevent the command line user from seeing/specifying --v,--vmodule,etc. This removes virtually all utility from glog since it cannot be configured.

bparees:
> Remove glog delegation and insert our own file/line prefixes as part of utilglog

i still really don't want to adopt our own custom logging framework, though i realize you've already implemented 99% of it.

>  Add glog parameters to the s2i command line parameter set, confusing -v|--v|--loglevel (this approach is reflected in the current PR state)

yeah i don't like that.  can't we at least hide the --v flag?  I think i could accept it if we did that, assuming there are no other overlapping names.  but i still don't like that we have a set of flags we have no control over.

> Trick glog into thinking its parameters have been parsed and prevent the command line user from seeing/specifying --v,--vmodule,etc. This removes virtually all utility from glog since it cannot be configured. 

which is the state we were already in, with just loglevel being exposed.  seems ok to me?  it's also what we do in origin (and what k8s is doing).

jupierce:
PR now reflects loglevel control only. Here is the output at each level currently:
[what.txt](https://github.com/openshift/source-to-image/files/335378/what.txt)

bparees:
can you update it to send all logs to glog if the requested loglevel is >2 ? (so we get consistent timestamping)

jupierce:
Absolutely. New output: https://gist.github.com/jupierce/73f2707c8b121204244943ae7da4bc37 . 
To be clear, the logic is effectively:
- Always log to stderr
- The loglevel parameter is tied directly to glog's -v 
- If loglevel > 2, delegate to glog in order to receive file/line/timestamp prefix

bparees:
output looks good, will review the code tonight(need to run an errand)

pkg/build/strategies/sti/sti.go
```diff
@@ -598,13 +602,11 @@ func (builder *STI) Execute(command string, user string, config *api.Config) err
 			if config.Quiet {
 				continue
 			}
-			// The log level > 3 forces to use glog instead of printing to stdout
-			if glog.V(3) {
-				glog.Info(text)
-				continue
-			}
-			fmt.Fprintf(os.Stdout, "%s\n", strings.TrimSpace(text))
+			glog.Info(strings.TrimSpace(text))
```
bparees: does this need a .V(0) so it gets delegated to glog at higher verbose levels?  (same for glog.Info() above and below)

pkg/util/glog/glog.go
```diff
@@ -1,101 +1,225 @@
...
+
+// Info records an info log entry.
+func (f *FileLogger) Info(args ...interface{}) {
+	f.output(args...)
+}
+
+// Warningf records an warning log entry.
+func (f *FileLogger) Warningf(format string, args ...interface{}) {
+	f.outputf(format, args...)
+}
```
bparees: i'm still struggling to understanding why you aren't prefixing "WARNING" here and ERROR in Errorf, etc in these implementations.

jupierce: With the delegation model we are zeroing in on, my thought was that it would lead to inconsistency. If a developer wanted to be sure a warning was clear even to a casual user, they would be tempted to do : glog.Warning( "WARNING: something went wrong" )
- When delegating to glog proper, this would result in the expected outcome: "*W* [normal glog prefix] WARNING: something went wrong"
- When the file logger was in operation, however, it would result in "WARNING: WARNING: something when wrong"

If you are liking the prefix idea, I think the options are:
(1) Inject the prefix only in the file logger path, encourage devs to never use their own prefix, and accept that the glog and file logger output will be a little different. I think this reduces the searchability of logs since a developer won't be sure if they are searching for "W " or "WARNING" unless they are aware of this subtlety. 
(2) Inject the prefix when using the file logger *and* prior to handing the message to glog, so log output will always feel consistent (only the presence or absence of a header would be obvious). I hesitate to do this without your input since it seemed to be edging us closer to our own logging implementation.

pkg/build/strategies/sti/sti.go
jupierce: Great catch! However, I would be more inclined to achieve this consistency by allowing Info & others to delegate to glog even if V() is not used. My hope being that we can have a near drop in replacement for glog and not require the extra .V() call on all logging statements.

bparees: well, either way you want to implement it i suppose.  but if Info, Warning, etc are going to do the delegation, then it really needs to be done in the writeln implementation, so that each of those methods don't have to repeat the delegation.

pkg/util/glog/glog.go
bparees: so everywhere that we do glog.Warningf today in origin code, we do not prefix the message ourselves, eg:

`glog.Warningf("x509: subject with cn=%s is not in the allowed list: %v", subject.CommonName, a.allowedCommonNames.List())`

so we are relying on the underlying logging framework to make it clear that a particular item is a warning, in whatever way that framework chooses.

So if you're gong to implement your own Warningf/Errorf methods, I think you need to also provide that behavior.  Which basically means path(1).

I'm less worried about a consistent prefix than I am about just having a prefix in both cases, so i don't think (2) is needed.

pkg/build/strategies/sti/sti.go
jupierce: Agreed. That refactor is underway.

pkg/util/glog/glog.go
jupierce: Sounds good. I had just seen a few places where prefixes were mixed in:

// HandleDockerError handles an an error from the docker daemon
func (c *NodeConfig) HandleDockerError(message string) {
	if !c.AllowDisabledDocker {
		glog.Fatalf("error: %s", message)
	}
	glog.Errorf("WARNING: %s", message)
	c.DockerClient = &dockertools.FakeDockerClient{VersionInfo: dockertypes.Version{APIVersion: "1.18"}}
}

bparees: well i certainly can't claim we've been consistent :)

jupierce:
New output: https://gist.github.com/jupierce/db4db1d892431e6c1a4cd8854ca88baf
- Adds prefixes for Warning/f, Error/f, and Fatal/f logs.
- Ensures all usage patterns can be delegated to glog.

pkg/util/glog/glog.go
```diff
@@ -1,101 +1,234 @@
...
+type gverbose struct {
+	log.Verbose
 }
 
 func (g gverbose) Infof(format string, args ...interface{}) {
 	g.Verbose.Infof(format, args...)
 }
 
-// file logs the provided messages at level or below to the writer, or delegates
+func (g gverbose) Info(args ...interface{}) {
```
bparees: is this and Infof needed, given that you're embedding log.Verbose already?

pkg/util/glog/glog.go
```diff
@@ -1,101 +1,234 @@
...
+	prefix     string
+	delegateFn elevated
+}
+
+var severities = []severityDetail{
+	infoLog:    {"", log.InfoDepth},
+	warningLog: {"WARNING: ", log.WarningDepth},
+	errorLog:   {"ERROR: ", log.ErrorDepth},
+	fatalLog:   {"FATAL: ", log.FatalDepth},
+}
```
bparees: reorder this block so the entire FileLogger struct + method implementations are grouped together.

pkg/util/glog/glog.go
```diff
@@ -1,101 +1,234 @@
...
+
+func (f *FileLogger) writeln(sev severity, line string) {
+	severity := severities[sev]
+
+	// If the loglevel has been elevated above this file logger's verbosity (generally set to 2)
+	// then delegate ALL messages to elevated logger in order to leverage its file/line/timestamp
+	// prefix information.
+	if log.V(log.Level(f.level + 1)) {
+		severity.delegateFn(3, line)
```
bparees: now that you've done this in this way, is the glogger struct even necessary anymore?

jupierce: Not in this context. I left it in because it was public via utilglog.Log . Since it was public, I assumed it might have some utility outside of s2i, but I am happy to remove it.

jupierce: glogger removed

bparees:
lgtm, ready for merge?

bparees:
[merge]

bparees:
fyi i don't get notified on emoji's either :)

bparees:
@jupierce after this merges, we'll need a bump commit to origin to pick up these changes there.  let me know if you want a hand w/ the bump process.

PR #528, opened 2016-06-30
Fixing race condition in s2i's new glog replacement

jupierce:
Addressing openshift/origin#9645


jupierce:
@bparees

bparees:
lgtm [merge]

bparees:
seems we now have flakes in our s2i merge queue too, that'll need to be investigated in the near future.
[merge]

Issue #530, opened 2016-07-05
The `--recursive` has the opposite action as expected


lbriais:
The default config of `s2i` to clone repositories recursively (`--recursive`) is supposed to be set to true.  
Actually it is not the case, even passing explicitly `--recursive=true`.

But if you pass  `--recursive=false`, then __it is actually recursive__ !

I believe this option should be named `--disable-submodules` with a default to false. It would be much more consistent with the behaviour and the code too (as in the code it looks like it is referred to as `DisableRecursive`)...

bparees:
i agree we should introduce a --disable-submodules flag w/ a default of false, when set to true we'll either use --recursive or submodule init logic depending whether --ref was also specified, and we'll remove the recursive flag.

PR #531, opened 2016-07-06
replace broken --recursive with a working --ignore-submodules option

bparees:
fixes https://github.com/openshift/source-to-image/issues/530


bparees:
@paralin @mfojtik @soltysh ptal.

bparees:
[test]

paralin:
lgtm

cmd/s2i/main.go
```diff
@@ -165,7 +165,7 @@ $ s2i build . centos/ruby-22-centos7 hello-world-app
 	cmdutil.AddCommonFlags(buildCmd, cfg)
 
 	buildCmd.Flags().BoolVar(&(cfg.RunImage), "run", false, "Run resulting image as part of invocation of this command")
-	buildCmd.Flags().BoolVar(&(cfg.DisableRecursive), "recursive", true, "Fetch all git submodules when cloning application repository")
+	buildCmd.Flags().BoolVar(&(cfg.IgnoreSubmodules), "ignore-submodules", false, "Ignore all git submodules when cloning application repository")
```
soltysh: You want to get away without keeping backwards compatibility entirely?

soltysh:
LGTM, just one doubt.

cmd/s2i/main.go
bparees: given that the flag was completely broken for anyone using it, yes :)

bparees: (I really don't want to retain deprecated backwards compatibility for something that has confusing behavior.  better to just get people to move to the right flag)

bparees:
[merge]

php-coder:
Looks like you forgot to update the docs: https://github.com/openshift/source-to-image/blob/master/docs/cli.md#build-flags

PR #536, opened 2016-07-12
Document extended build

### LINKED ISSUE #515: Example of multi-step s2i build is needed
From the README:

> For compiled languages like Java, C, or Golang, the dependencies necessary for compilation might dramatically outweigh the actual runtime artifacts. To keep runtime images slim, S2I enables multiple-step build processes, where a binary artifact such as an executable or Java WAR file is created in the first builder image, extracted, and injected into an second image that simply places the executable in the correct location for execution.

Please provide an example how to take advantage of this possibility (including the needed BuildConfig).

php-coder:
Documentation for extended build support in S2I (https://trello.com/c/MWUAjl6k/237-13-extended-builds-separate-mvn-vs-jboss-evg).

PTAL @bparees @PI-Victor I will highly appreciate any feedback from you!

This PR also fixes #515


README.md
```diff
@@ -29,7 +29,7 @@ For example, to create a reproducible build pipeline for Tomcat (the popular Jav
 
 By placing our build logic inside of images, and by combining the images into multiple steps, we can keep our runtime environment close to our build environment (same JDK, same Tomcat JARs) without requiring build tools to be deployed to production.
 
-Learn more about [building your own images](#getting-started).
+Learn more about [building your own images](#getting-started) and [how to use non-builder image for a resulting image](https://github.com/openshift/source-to-image/blob/master/docs/runtime_image.md).
```
bparees: "how to use a non-builder image for the final application image"

docs/cli.md
```diff
@@ -90,6 +90,8 @@ that image and add them to the tar streamed to the container into `/artifacts`.
 | `-q (--quiet)`             | Operate quietly, suppressing all non-error output |
 | `-i (--inject)`            | Inject the content of the specified directory into the path in the container that runs the assemble script |
 | `-v (--volume)`            | Bind mounts a local directory into the container that runs the assemble script|
+| `--runtime-image`          | Image that will be used as the base for the runtime image (see [How to use non-builder image for a resulting image](https://github.com/openshift/source-to-image/blob/master/docs/runtime_image.md)) |
```
bparees: same text change here

docs/cli.md
```diff
@@ -90,6 +90,8 @@ that image and add them to the tar streamed to the container into `/artifacts`.
...
+| `-a (--runtime-artifact)`  | Specify a file or directory to be copied from the builder to the runtime image  (see [How to use non-builder image for a resulting image](https://github.com/openshift/source-to-image/blob/master/docs/runtime_image.md)) |
```
bparees: same text change here

docs/runtime_image.md
```diff
@@ -0,0 +1,83 @@
+# Runtime image
+
+## Overview
+
+For a dynamic languages like PHP, Python or Ruby, the build-time and run-time environments are the same. In this case using builder as a base image for a resulting image is natural.
```
bparees: resulting application image

docs/runtime_image.md
```diff
@@ -0,0 +1,83 @@
...
+
+For compiled languages like C/C++, Java or Golang, the dependencies necessary for compilation might dramatically outweigh the actual runtime artifacts. To keep runtime images slim, S2I enables multiple-step build processes, where a binary artifact such as an executable or Java WAR file is created in the first builder image, extracted, and injected into an second image that simply places the executable in the correct location for execution. To give you even more abilities to customize a resulting image, S2I is also executing `assemble-runtime` script inside of this (runtime) image. In this way you may do a final adjustments by modifying files before an image will be committed.
```
bparees: "might dramatically outweigh the size of the actual runtime artifacts, or provide attack surface areas that are undesirable in an application image."

bparees: "you may do final adjustments by modifying....."

docs/runtime_image.md
```diff
@@ -0,0 +1,83 @@
...
+
+The following scheme illustrates the build workflow:
```
bparees: s/scheme/diagram/

docs/runtime_image.md
```diff
@@ -0,0 +1,83 @@
...
+
+![s2i workflow](./runtime-image-flow.png "s2i workflow")
+
+
+## How it works?
```
bparees: remove the ?

docs/runtime_image.md
```diff
@@ -0,0 +1,83 @@
...
+
+To make this work S2I needs to know the following:
+
+* builder image
+* mapping of the artifacts
+* runtime image
+
+And all these information you should specify during S2I invocation:
```
bparees: "This information is specified during S2I invocation:"

docs/runtime_image.md
```diff
@@ -0,0 +1,83 @@
...
+
+    s2i build <repo> <builder-image> <my-app> --runtime-image <runtime-image> --runtime-artifact </path/to/artifact>
+
+The only new here are `--runtime-image` and `--runtime-artifact` options. The latter option is used to specify a full path to file (or directory) inside of the builder image and a corresponding directory on the runtime image where S2I will copy these files. Note that you can specify this option multiple times.
```
bparees: "The first option specifies the image that will be used as the base image for the final application image.  The second option specifies a full path to a file (or directory) that will exist in the builder image after assembly, and a corresponding directory on the runtime image where S2I will copy these files, separated by a colon(?).  Note that you can specify this option multiple times."

Also you need to update the example invocation to specify the destination path.  Or otherwise do a better job of explaining that the destination path is optional, and how to specify it if desired.

docs/runtime_image.md
```diff
@@ -0,0 +1,83 @@
...
+
+For our example S2I will do the following steps:
+
+1. runs builder container and performs a build (as usual)
```
bparees: "run a builder container and invoke the assemble script (as usual)"

docs/runtime_image.md
```diff
@@ -0,0 +1,83 @@
...
+1. after build finishes but before stopping runtime container, it downloads artifacts from the builder and places them to a temporary directory on the host
```
bparees: "after the builder finishes but before stopping the builder container, download the requested artifacts from the builder and place them in a temporary directory on the host"

docs/runtime_image.md
```diff
@@ -0,0 +1,83 @@
...
+1. runs container with runtime image
```
bparees: "start a container using the specified runtime image"

docs/runtime_image.md
```diff
@@ -0,0 +1,83 @@
...
+1. uploads scripts and build artifacts from a temporary directory into `WORKDIR` of a runtime container
```
bparees: "upload the scripts and build artifacts from the temporary directory into `WORKDIR` of the runtime container"

docs/runtime_image.md
```diff
@@ -0,0 +1,83 @@
...
+1. runs `assemble-runtime` script in a runtime container
```
bparees: "run the `assemble=runtime` script in the runtime container"

docs/runtime_image.md
```diff
@@ -0,0 +1,83 @@
...
+1. commits container with runtime image
```
bparees: "commit the runtime container and tag it as the new application image"

docs/runtime_image.md
```diff
@@ -0,0 +1,83 @@
...
+1. stops build and runtime containers
```
bparees: you can leave this off, not that interesting to users.

docs/runtime_image.md
```diff
@@ -0,0 +1,83 @@
...
+
+## Details
+
+### Format and behavior of `--runtime-artifact` option
```
bparees: s/of/of the/

docs/runtime_image.md
```diff
@@ -0,0 +1,83 @@
...
+
+`--runtime-artifact` option have a format `<source>[:<destination>]`. Here is an example of sample values:
```
bparees: s/have a/has the/

bparees: "Here are some example values:"

docs/runtime_image.md
```diff
@@ -0,0 +1,83 @@
...
+
+| Value                   | Meaning |
+|-------------------------|---------|
+| `/tmp/app.war`          | `/tmp/app.war` file will be extracted from a builder image and uploaded into the `WORKDIR` of a runtime image |
```
bparees: s/of a/of the/

docs/runtime_image.md
```diff
@@ -0,0 +1,83 @@
...
+| `/tmp/app.war:.`        | the same as above |
+| `/tmp/app.war:app/dist` | `/tmp/app.war` file will be copied from a builder image into into the `app/dist` subdirectory of the `WORKDIR` of a runtime image |
```
bparees: s/of a/of the/

docs/runtime_image.md
```diff
@@ -0,0 +1,83 @@
...
+| `/opt/data`             | `/opt/data` directory will be copied from a builder image into the `WORKDIR` of a runtime image |
```
bparees: can I specify /opt/data/* ?

is this going to result in a directory named "data" in WORKDIR the runtime image, or is it going to result in the contents of the data directory being placed in WORKDIR?

docs/runtime_image.md
```diff
@@ -0,0 +1,83 @@
...
+
+Keep in mind that the `source` must be an absolute path. The `destination` must be a relative path to directory, it must exist and path must not start from `..`
```
bparees: s/Keep in mind that the/The/
/path to directory,/path,/
s/must not start from/must not start with/

docs/runtime_image.md
```diff
@@ -0,0 +1,83 @@
...
+| Value                   | Meaning |
+|-------------------------|---------|
+| `/tmp/app.war`          | `/tmp/app.war` file will be extracted from a builder image and uploaded into the `WORKDIR` of a runtime image |
+| `/tmp/app.war:.`        | the same as above |
+| `/tmp/app.war:app/dist` | `/tmp/app.war` file will be copied from a builder image into into the `app/dist` subdirectory of the `WORKDIR` of a runtime image |
+| `/opt/data`             | `/opt/data` directory will be copied from a builder image into the `WORKDIR` of a runtime image |
+
+Keep in mind that the `source` must be an absolute path. The `destination` must be a relative path to directory, it must exist and path must not start from `..` 
+
+During copying the artifacts, S2I will modify their permissions. All directories and files with executable bit will be uploaded with `0755` mode. Other files will have `0644` mode.
```
bparees: s/During/When/

docs/runtime_image.md
```diff
@@ -0,0 +1,83 @@
...
+
+### `assemble-runtime` script requirements
+
+`assemble-runtime` script can be written in any programming language. S2I searches the following locations for this script in the following order:
```
bparees: "The `assemble-runtime` script"

and it's not really a script if it can be written in any language, it just has to be executable, right?

docs/runtime_image.md
```diff
@@ -0,0 +1,83 @@
...
+
+1. A script found at the `--scripts-url` URL
+1. A script found in the application source `.s2i/bin` directory
+1. A script found at the default image URL (`io.openshift.s2i.scripts-url` label)
+
+`assemble-runtime` script is always gets executed under `USER` from a runtime image.
```
bparees: "The `assemble-runtime` script is always executed as the runtime image `USER`"

docs/runtime_image.md
```diff
@@ -0,0 +1,83 @@
...
+
+### Runtime image requirements
+
+In most of the cases you can use any images as a runtime image. Only if you are using `--allowed-uids` option then image should have numeric `USER` and this user should be within the range of allowed users.
```
bparees: "In most cases you can use any image as a runtime image.  However, if you are using the `--allowed-uids` option then the image must have a numeric `USER` specified and the value must be within the range of allowed user ids."

docs/runtime_image.md
```diff
@@ -0,0 +1,83 @@
...
+
+To simplify build workflow and to provide some reasonable defaults, author of the runtime image can use the following techniques:
```
bparees: "To simplify the build workflow and provide some reasonable defaults, the author of the runtime image can....."

docs/runtime_image.md
```diff
@@ -0,0 +1,83 @@
...
+
+* `run` and `assemble-runtime` scripts can be placed inside of the runtime image. Scripts from the image will be used as a fallback when user didn't have any in the `.s2i/bin` directory of his repository and when he didn't specify alternative location with `--scripts-url` option. The location of the scripts is defined by the value of the `io.openshift.s2i.scripts-url` label that should be presented on the image. For example, you can set it to `image:///usr/libexec/s2i`
```
bparees: `run` and `assemble-runtime` scripts can be placed inside of the runtime image. Scripts from the image will be used as a fallback when user does not provide them in the `.s2i/bin` directory of the source repository and an alternative location is not specified with the `--scripts-url` option.

docs/runtime_image.md
```diff
@@ -0,0 +1,83 @@
...
+* default mapping for the files can be specified by adding `io.openshift.s2i.assemble-input-files` label to a runtime image. This mapping will be used as a fallback when user didn't provide it explicitly with `--runtime-artifact` option.
```
bparees: "adding the `io.openshift.s2i.assemble-input-files` label to the runtime image"

"This mapping will be used as a fallback when the user does not specify the artifacts explicitly with the `--runtime-artifact` option."

docs/runtime_image.md
```diff
@@ -0,0 +1,83 @@
...
+
+To specify mapping for multiple files separate them by a semicolon. For example: `/tmp/app.war:app;/opt/data`
```
bparees: "To specify mappings for multiple files, separate them with a semicolon."

docs/runtime_image.md
```diff
@@ -0,0 +1,83 @@
...
+
+### Build and runtime environments
+
+Builder and runtime containers have the same environment. In other words `assemble` and `assemble-runtime` scripts are able to use environment variables defined with `--env` and `--environment-file` options along with the values from `.s2i/environment` file in the repo.
```
bparees: s/in the repo./in the source repository./

docs/runtime_image.md
```diff
@@ -0,0 +1,83 @@
...
+
+### Extended build and incremental build
+
+In the current implementation it's impossible to do an exteneded incremental build. This combination is invalid and build will fail.
```
bparees: s/it's impossible/it is not possible/
s/exteneded/extended/
s/and build will fail/and the build will fail/

bparees:
@php-coder nice thorough doc, just a bunch of grammatical cleanup.

docs/runtime_image.md
rhcarvalho: @bparees I have the impression we refer to all other executables as scripts (assemble script, run script), though technically it's as you said :-)
https://docs.openshift.com/enterprise/3.2/creating_images/s2i.html#s2i-scripts

bparees: yeah, this is certainly not the only place we do it..the reason it caught my eye is because of the "any programming language" comment which feels like a weird thing to say after "script".

maybe "assemble-runtime can be any executable script or binary" would be less confusing?

docs/runtime_image.md
php-coder: >The second option specifies a full path to a file (or directory) that will exist in the builder image after assembly, and a corresponding directory on the runtime image where S2I will copy these files, separated by a colon(?). Note that you can specify this option multiple times."

In the "How it works" chapter I tried to hide some details to show a full picture. And that's why I'm not talking about optional parameter of the `--runtime-artifact` option here. Also I moved hint about using option multiple times to section below and added an example:

- https://github.com/php-coder/source-to-image/blob/71dfcd496510cb4d8e8a3153614bbe46c1b2225a/docs/runtime_image.md#how-it-works
- https://github.com/php-coder/source-to-image/blob/71dfcd496510cb4d8e8a3153614bbe46c1b2225a/docs/runtime_image.md#format-and-behavior-of-the---runtime-artifact-option

>Also you need to update the example invocation to specify the destination path. Or otherwise do a better job of explaining that the destination path is optional, and how to specify it if desired.

Would you mind to leave this explanation in the `Format and behavior of the --runtime-artifact option` chapter? It's very rare case as I think and (probably) beginners won't use it at all.

(Moreover, I started to think that we can remove it. If user wants to move the files, he can do it in the `assemble-runtime` script. So, less code, less misunderstanding and optional parameters, less documentation but user still will be able to achieve his goal. WDYT?)

docs/runtime_image.md
php-coder: >maybe "assemble-runtime can be any executable script or binary" would be less confusing?

Updated to use this variant.

docs/runtime_image.md
php-coder: > can I specify /opt/data/* ?

Interesting question, let's see..

> is this going to result in a directory named "data" in WORKDIR the runtime image, or is it going to result in the contents of the data directory being placed in WORKDIR?

It's going to be an error:
```
Error: invalid argument "/tmp/data-dir/*" for a: invalid characters in filename: "/tmp/data-dir/*"
```

docs/runtime_image.md
bparees: > In the "How it works" chapter I tried to hide some details to show a full picture. And that's why I'm not talking about optional parameter of the --runtime-artifact option here.

that's fine but then you should fix the sentence that says "The latter option is used to specify a full path to file (or directory) inside of the builder image and a corresponding directory on the runtime image where S2I will copy these files." to just say "The latter option is used to specify a full path to file (or directory) inside of the builder image that will be copied into the `WORKDIR` of the runtime image"

because right now you're referring to a "corresponding directory" that you aren't actually specifying in that example, so it's confusing.

> Would you mind to leave this explanation in the Format and behavior of the --runtime-artifact option chapter? It's very rare case as I think and (probably) beginners won't use it at all.

yeah I hadn't gotten to your more detailed section when I made those comments...  the detailed section definitely helps provide the needed specifics.

> (Moreover, I started to think that we can remove it. If user wants to move the files, he can do it in the assemble-runtime script. So, less code, less misunderstanding and optional parameters, less documentation but user still will be able to achieve his goal. WDYT?)

I think we still need it.  I don't want to have to override the assemble-runtime script just because my maven build produces a file named "foo.war" and my assemble-runtime script only looks for "ROOT.war" or something.  (There are probably better examples).

bparees: also it makes that argument consistent with what we already do for injections.

docs/runtime_image.md
bparees: sorry it wasn't clear, my second questions were referring to your exampe, not the wild card invocation:  if i specify "/opt/data" as my source path, to i get a "data" directory under WORKDIR?  or do i just get the contents of data under WORKDIR?

as for the wildcarding, we should make a note that wildcards are not supported.  (and be prepared for users to complain that they can't specify a path like "/opt/data/*.txt", but i'm not too worried about that)

docs/runtime_image.md
rhcarvalho: s/For a dynamic languages/For dynamic languages/

docs/runtime_image.md
php-coder: >because right now you're referring to a "corresponding directory" that you aren't actually specifying in that example, so it's confusing.

Ah, yes, now I see what do you mean. Fixed: https://github.com/openshift/source-to-image/pull/536/commits/f38a5b1310885a96da172aa32c76170a18355204

>I don't want to have to override the assemble-runtime script just because my maven build produces a file named "foo.war" and my assemble-runtime script only looks for "ROOT.war" or something. (There are probably better examples).

Yes, but this option doesn't help in provided case because it only affects destination directory and not file name. In other words, you only can choose whether foo.war will be uploaded as `WORKDIR/foo.war` or `WORKDIR/you-app/foo.war`.

>also it makes that argument consistent with what we already do for injections.

Yes, now I agree. I also realized this point later.

docs/runtime_image.md
php-coder: >/path to directory,/path,/

Why do you want to omit the fact that it should be a directory, btw? I think that it's important and we even should write it bold :)

bparees: sorry you're correct, "The destination must be a relative path to a directory (not a file)".

I think i was thinking that "path to a directory" was redundant, but we hadn't explicitly stated earlier that the destination must be a directory.  Perhaps we should.

docs/runtime_image.md
php-coder: > if i specify "/opt/data" as my source path, do i get a "data" directory under WORKDIR? or do i just get the contents of data under WORKDIR?

Yes, you will "data" directory under WORKDIR.

> as for the wildcarding, we should make a note that wildcards are not supported

I tried to address all these questions in the following commits:
- https://github.com/openshift/source-to-image/pull/536/commits/04db791d42f67dabb0ca55694d7bccdc63f83c82
- https://github.com/openshift/source-to-image/pull/536/commits/e5c6c7b1774ac405ae7ec4ba8e07cd9d869c0b9e

bparees: what about /opt/data/ (trailing slash).  I assume then i don't get a "data" directory.

php-coder: > what about /opt/data/ (trailing slash). I assume then i don't get a "data" directory.

It will works exactly the same as without trailing slash because we're invoking `filepath.Clean()` on `source` and `destination`. I added this example to doc.

docs/runtime_image.md
```diff
@@ -0,0 +1,87 @@
+# How to use a non-builder image for the final application image
+
+## Overview
+
+For dynamic languages like PHP, Python or Ruby, the build-time and run-time environments are the same. In this case using builder as a base image for a resulting application image is natural.
```
bparees: "PHP, Python, or Ruby"  (oxford comma)
"using the builder"

docs/runtime_image.md
```diff
@@ -0,0 +1,87 @@
...
+
+For compiled languages like C/C++, Java or Golang, the dependencies necessary for compilation might dramatically outweigh the size of the actual runtime artifacts, or provide attack surface areas that are undesirable in an application image. To keep runtime images slim, S2I enables multiple-step build processes, where a binary artifact such as an executable or Java WAR file is created in the first builder image, extracted, and injected into an second image that simply places the executable in the correct location for execution. To give you even more abilities to customize a resulting image, S2I is also executing `assemble-runtime` script inside of this (runtime) image. In this way you may do final adjustments by modifying files before an image will be committed.
```
bparees: "C/C++, Java, or Golang"

bparees: "S2I enables a multiple-step build process"

bparees: "S2I is also executing an `assemble-runtime` script..."

bparees:
couple final nits, squash, and lgtm.

php-coder:
>couple final nits, squash, and lgtm.

@bparees Done. Ready to merge.

bparees:
[merge]

great work @php-coder

php-coder:
@bparees Thank you for your patience, Ben!

README.md
```diff
@@ -18,7 +18,7 @@ For a dynamic language like Ruby, the build-time and run-time environments are t
...
 1. Commit the new container and set the image entrypoint to be a script (provided by the builder image) that will start Apache to host the Ruby application.
 
-For compiled languages like Java, C, or Golang, the dependencies necessary for compilation might dramatically outweigh the actual runtime artifacts. To keep runtime images slim, S2I enables multiple-step build processes, where a binary artifact such as an executable or Java WAR file is created in the first builder image, extracted, and injected into an second image that simply places the executable in the correct location for execution.
+For compiled languages like Java, C, or Golang, the dependencies necessary for compilation might dramatically outweigh the actual runtime artifacts. To keep runtime images slim, S2I enables a multiple-step build processes, where a binary artifact such as an executable or Java WAR file is created in the first builder image, extracted, and injected into a second runtime image that simply places the executable in the correct location for execution.
```
rhcarvalho: Actually the name of the language is "Go". The website is golang.org.

https://groups.google.com/d/topic/golang-nuts/pR5cyDaZd1E/discussion

docs/runtime_image.md
```diff
@@ -0,0 +1,87 @@
+# How to use a non-builder image for the final application image
+
+## Overview
+
+For dynamic languages like PHP, Python, or Ruby, the build-time and run-time environments are the same. In this case using the builder as a base image for a resulting application image is natural.
+
+For compiled languages like C/C++, Java, or Golang, the dependencies necessary for compilation might dramatically outweigh the size of the actual runtime artifacts, or provide attack surface areas that are undesirable in an application image. To keep runtime images slim, S2I enables a multiple-step build processes, where a binary artifact such as an executable or Java WAR file is created in the first builder image, extracted, and injected into an second image that simply places the executable in the correct location for execution. To give you even more abilities to customize a resulting image, S2I is also executing an `assemble-runtime` script inside of this (runtime) image. In this way you may do final adjustments by modifying files before an image will be committed.
```
rhcarvalho: Same here, s/Golang/Go/.

README.md
rhcarvalho: Its funny that this paragraph seems to be copy-pasted into `docs/runtime_image.md`, but the contents are not exactly the same. In particular, there seems to be an arbitrary difference in the programming languages mentioned:

```
For compiled languages like {+C/C++,+} Java,[-C,-] or Golang, the dependencies
necessary for compilation might dramatically outweigh the {+size of the+} actual
runtime [-artifacts.-]{+artifacts, or provide attack surface areas that are
undesirable in an application image.+} To keep runtime images slim, S2I enables
a multiple-step build processes, where a binary artifact such as an executable
or Java WAR file is created in the first builder image, extracted, and injected
into [-a-]{+an+} second[-runtime-] image that simply places the executable in
the correct location for execution. {+To give you even more abilities to
customize a resulting image, S2I is also executing an `assemble-runtime` script
inside of this (runtime) image. In this way you may do final adjustments by
modifying files before an image will be committed.+}
```

(output from `git diff --word-diff`)

rhcarvalho:
@php-coder, sorry, I just noticed that I commented after merge...

rhcarvalho:
@php-coder nevermind, I've address my own comments in #538.

PR #537, opened 2016-07-13
Extended build minor improvements

php-coder:
Minor improvements:
- update command line options descriptions to mention `assemble-runtime` script
- simplify code a bit (before we were extracting `USER` from the runtime image and passing this user to the Docker during runtime container creation. It looks useless, because we can just omit user and Docker will use default `USER` for us)
- move `assemble-runtime` script to the `scripts` sub-directory in the resulting application image (`run` script is already here, so it looks strange that they're in a different places)

@bparees WDYT about these changes?


pkg/build/strategies/sti/postexecutorstep.go
```diff
@@ -276,7 +270,6 @@ func (step *startRuntimeImageAndUploadFilesStep) execute(ctx *postExecutorStepCo
 		CapDrop:         step.builder.config.DropCapabilities,
 		PostExec:        step.builder.postExecutor,
 		Env:             step.builder.env,
-		User:            user,
```
bparees: is this just not needed because the container will default to that user anyway?

php-coder: Yes, that was my point. (Probably, I wrote it vague in the description.)

bparees: sorry, i don't tend to read PR summaries.  I should, my bad.

bparees:
lgtm, squash

php-coder:
@bparees Squashed and prepared for merge.

bparees:
[merge]

PR #543, opened 2016-07-15
override entrypoint to env during assemble

bparees:
fixes https://github.com/openshift/source-to-image/issues/475

this basically reimplements https://github.com/openshift/source-to-image/pull/476 but ensures the entrypoint in the final committed image is not changed to "env".

### DIFF
--- a/pkg/build/strategies/sti/postexecutorstep.go
+++ b/pkg/build/strategies/sti/postexecutorstep.go
@@ -13,6 +13,7 @@ import (
 	"github.com/openshift/source-to-image/pkg/api"
 	dockerpkg "github.com/openshift/source-to-image/pkg/docker"
 	"github.com/openshift/source-to-image/pkg/errors"
+	"github.com/openshift/source-to-image/pkg/run"
 	"github.com/openshift/source-to-image/pkg/tar"
 	"github.com/openshift/source-to-image/pkg/util"
 )
@@ -53,10 +54,10 @@ type storePreviousImageStep struct {
 
 func (step *storePreviousImageStep) execute(ctx *postExecutorStepContext) error {
 	if step.builder.incremental && step.builder.config.RemovePreviousImage {
-		glog.V(3).Info("Executing step: remember previous image")
+		glog.V(3).Info("Executing step: store previous image")
 		ctx.previousImageID = step.getPreviousImage()
 	} else {
-		glog.V(3).Info("Skipping step: remember previous image")
+		glog.V(3).Info("Skipping step: store previous image")
 	}
 	return nil
 }
@@ -114,7 +115,19 @@ func (step *commitImageStep) execute(ctx *postExecutorStepContext) error {
 
 	ctx.labels = createLabelsForResultingImage(step.builder, step.docker, step.image)
 
-	ctx.imageID, err = commitContainer(step.docker, ctx.containerID, cmd, user, step.builder.config.Tag, step.builder.env, ctx.labels)
+	// Set the image entrypoint back to its original value on commit, the running
+	// container has "env" as its entrypoint and we don't want to commit that.
+	entrypoint, err := step.docker.GetImageEntrypoint(step.image)
+	if err != nil {
+		return fmt.Errorf("Couldn't get entrypoint of %q image: %v", step.image, err)
+	}
+	// If the image has no explicit entrypoint, set it to an empty array
+	// so we don't default to leaving the entrypoint as "env" upon commit.
+	if entrypoint == nil {
+		entrypoint = []string{}
+	}
+
+	ctx.imageID, err = commitContainer(step.docker, ctx.containerID, cmd, user, step.builder.config.Tag, step.builder.env, entrypoint, ctx.labels)
 	if err != nil {
 		return err
 	}
@@ -261,6 +274,7 @@ func (step *startRuntimeImageAndUploadFilesStep) execute(ctx *postExecutorStepCo
 
 	opts := dockerpkg.RunContainerOptions{
 		Image:           image,
+		Entrypoint:      run.DefaultEntrypoint,
 		PullImage:       false, // The PullImage is false because we've already pulled the image
 		CommandExplicit: []string{"/bin/sh", "-c", cmd},
 		Stdout:          outWriter,
@@ -369,10 +383,11 @@ func (step *invokeCallbackStep) execute(ctx *postExecutorStepContext) error {
 
 // shared methods
 
-func commitContainer(docker dockerpkg.Docker, containerID, cmd, user, tag string, env []string, labels map[string]string) (string, error) {
+func commitContainer(docker dockerpkg.Docker, containerID, cmd, user, tag string, env, entrypoint []string, labels map[string]string) (string, error) {
 	opts := dockerpkg.CommitContainerOptions{
 		Command:     []string{cmd},
 		Env:         env,
+		Entrypoint:  entrypoint,
 		ContainerID: containerID,
 		Repository:  tag,
 		User:        user,
--- a/pkg/docker/docker.go
+++ b/pkg/docker/docker.go
@@ -95,6 +95,7 @@ type Docker interface {
 	CheckAndPullImage(name string) (*docker.Image, error)
 	BuildImage(opts BuildImageOptions) error
 	GetImageUser(name string) (string, error)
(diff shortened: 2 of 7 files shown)


gabemontero:
ahh ... more merge pain for me :-)

pkg/run/run.go
```diff
@@ -44,14 +44,15 @@ func (b *DockerRunner) Run(config *api.Config) error {
 
 	opts := docker.RunContainerOptions{
 		Image:        config.Tag,
+		Entrypoint:   []string{"/bin/env"},
```
php-coder: If we always (?) use `/bin/env`, then maybe make it default? Or it's impossible/undesirable?

(BTW I've never heard about `/bin/env` before and in many examples  `/usr/bin/env` is used.)

pkg/docker/docker.go
```diff
@@ -287,6 +291,15 @@ func (d *stiDocker) GetImageWorkdir(name string) (string, error) {
 	return workdir, nil
 }
 
+// GetImageEntrypoint returns the ENTRYPOINT property for the given image name.
+func (d *stiDocker) GetImageEntrypoint(name string) ([]string, error) {
+	image, err := d.client.InspectImage(name)
+	if err != nil {
+		return nil, err
```
php-coder: In most of functions we're throwing `errors.NewInspectImageError()` but in some we don't do it. Do you know why? Do we have some rule about it that I don't know?

pkg/build/strategies/sti/postexecutorstep.go
```diff
@@ -114,7 +114,19 @@ func (step *commitImageStep) execute(ctx *postExecutorStepContext) error {
...
+	// If the image has no explicit entrypoint, set it to an empty array
+	// so we don't default to leaving the entrypoint as "env" upon commit.
+	if entrypoint == nil {
+		entrypoint = []string{}
+	}
+	if err != nil {
+		return fmt.Errorf("Couldn't get entrypoint of %q image: %v", step.image, err)
+	}
+
+	ctx.imageID, err = commitContainer(step.docker, ctx.containerID, cmd, user, step.builder.config.Tag, step.builder.env, entrypoint, ctx.labels)
```
php-coder: This method has unit test and would be great if its coverage won't go down :-)

pkg/docker/docker.go
bparees: I don't know...i'll have to look into how we're using that error to see if we should convert the rest of the cases to also throw that.

pkg/run/run.go
bparees: make it a default where?

but now i'm wondering if doing this is also going to break images that don't have /bin/env for one reason or another.  (this originally came from @mfojtik, i'm just re-implementing his change because the original one had to be reverted)

pkg/build/strategies/sti/postexecutorstep.go
bparees: is that a request to add tests? :)

bparees:
@gabemontero yeah, sorry :)  Consider it "motivation to finish the migration to docker-engine sooner"

gabemontero:
yep :-) .... i'm actually down to the unit tests and replacing the fake
client stuff ... shouldn't be too much longer

pkg/build/strategies/sti/postexecutorstep.go
php-coder: It was a wish to see an about 5 lines in the the existing unit test ;-)

pkg/run/run.go
php-coder: >make it a default where?

In the methods that are using this (docker.CommitContainer/docker.RunContainer).

pkg/build/strategies/sti/postexecutorstep.go
bparees: "see an about"?

php-coder: > "see an about"?

I mean that I'm expecting that there is no more than 5 lines of code to be added.

bparees:
@php-coder i've updated the unit test and added a constant for the entrypoint.

i'm still not sure what to do about the fact that this introduces a requirement that all images (including runtime images) have a "/bin/env" binary available. For the most part i'm not worried about it, but if someone wants to do an extended build with a minimal target image, they can't do it here.

maybe i'll update the code to only set(override) the entrypoint if your image already sets an entrypoint.

php-coder:
>if someone wants to do an extended build with a minimal target image, they can't do it here

Is it possible to handle such a case like we're handling case of missing sh/tar? As I understand because we're always run assemble script as `env assemble` it will fail with `env not found`, right? And we can catch it in the same way.

(I don't have an answer, just suggesting idea and I'm not sure that it's possible..)

php-coder:
@bparees

>i've updated the unit test and added a constant for the entrypoint.

... but looks like you forgot to push your changes, did you? :)

bparees:
actually I did push, but what i didn't do was commit them :)

but not worth you looking now, i'm going to update again to do what i said (not set the entrypoint unless the image explicitly declares one)

bparees:
@php-coder ok, ptal again.

pkg/docker/docker.go
```diff
@@ -138,13 +139,17 @@ type PullResult struct {
...
+	PullImage       bool
+	PullAuth        docker.AuthConfiguration
+	ExternalScripts bool
+	ScriptsURL      string
+	Destination     string
+	Env             []string
+	// Entrypoint will be used to override the default entrypoint
+	// for the image if it has one.  If the image has no entrypoint,
+	// this value is ignored.
+	Entrypoint       []string
```
php-coder: What if later we'll want to always set the `Entrypoint`? What we'll do then?

mfojtik:
@bparees are we setting the entrypoint back to the original value after the assemble finish to run?

pkg/docker/docker.go
bparees: we'll change the behavior and/or introduce another field w/ that behavior?

bparees:
> @bparees are we setting the entrypoint back to the original value after the assemble finish to run?

@mfojtik yes, that's the primary new functionality over your original PR for this issue.

mfojtik:
@bparees at that time I thought `/bin/env` must be enough for everybody :-)

pkg/build/strategies/sti/postexecutorstep.go
```diff
@@ -114,7 +115,19 @@ func (step *commitImageStep) execute(ctx *postExecutorStepContext) error {
...
-	ctx.imageID, err = commitContainer(step.docker, ctx.containerID, cmd, user, step.builder.config.Tag, step.builder.env, ctx.labels)
+	// Set the image entrypoint back to its original value on commit, the running
+	// container has "env" as its entrypoint and we don't want to commit that.
+	entrypoint, err := step.docker.GetImageEntrypoint(step.image)
+	// If the image has no explicit entrypoint, set it to an empty array
+	// so we don't default to leaving the entrypoint as "env" upon commit.
+	if entrypoint == nil {
+		entrypoint = []string{}
+	}
+	if err != nil {
```
mfojtik: move this up, no point of setting the entrypoint if you get the error.

pkg/docker/docker.go
mfojtik: +1 (i think the godoc is enough)

pkg/run/run.go
```diff
@@ -11,6 +11,9 @@ import (
 	utilglog "github.com/openshift/source-to-image/pkg/util/glog"
 )
 
+// DefaultEntrypoint is the default entry point used when starting containers
+const DefaultEntrypoint = "/bin/env"
```
mfojtik: `var DefaultEntrypoint = []string{"/bin/env"}` ?

(you don't have to create the arrays when calling this)

mfojtik:
minor suggestions, but LGTM!

pkg/build/strategies/sti/postexecutorstep.go
bparees: good catch

pkg/run/run.go
bparees: good idea.

bparees:
@mfojtik updated, will let @php-coder have a final word before merging.

php-coder:
LGTM too.

bparees:
[merge]

PR #547, opened 2016-07-19
Fix directory delete in Windows

csrwng:
When running on windows, calls out to system to delete a directory to avoid golang issue https://github.com/golang/go/issues/9606

When we switch to golang 1.7, we can remove the workaround.


csrwng:
@bparees ptal

bparees:
latm (looks acceptable to me, i can't claim it's good :) )
[merge]

PR #548, opened 2016-07-20
add k8s docker client and docker engine-api dependencies

gabemontero:
@bparees - so, at least via my interpretation of how the `git mergetool` and `git rebase` went, I was able to merge the Godeps changes and s2i changes that occurred since I started to migration off of fsouza to k8s/engine-api.

And I've been able to build s2i and run the test suites, with fsouza now totally removed from s2i (both in s2i code and Godeps).  I minimally remember seeing you updated fsouza recently, but don't recall if other Godeps were updated.

That said, given all the files related to the Godeps I had to pull from origin, the file list is as you can see huge, and git does not even list them all.  The list I've pulled so far seems to line up with what was added, but can't say for sure simply from the normal viewing of this pull.  But I'm trying to pull together a local html doc with the complete git diff for cross examination.

But bottom line:
1) is the PR reviewable given the number of files ... is there a way for you to say move past the Godeps and to the s2i changes?
2) pending my ability to fully verify that all the Godeps changes are from my additions of k8s/engine-api, some concern seem warranted.


soltysh:
@gabemontero at least split the changes into two commits, one updating the godeps and 2nd changing s2i sources. Unfortunately, there's no magic trick to do that and reviewer has to go through both Godeps changes (more or less :stuck_out_tongue_winking_eye:) and through s2i changes.

gabemontero:
pending any rework of the commits, the `git diff` did show (via a `grep "\-\-\-" on the saved output) that only new files (i.e. a `--- /dev/null`) and Godeps.json were part of the Godeps portion of the commit.

If worthwhile I can make the output from `git diff` available.

I'm on PTO until Monday but will be monitoring the discussion here periodically, whether comments in the `s2i` code I changed or further calls to change the commit structure.

bparees:
@gabemontero yeah we're going to want this as two commits, one for the godeps changes, and the other for the actual s2i changes.  do a "git reset HEAD~1" to unstage things, then (i'm guessing) you can do a git add with a Godeps path and commit to commit those changes, then a second git add to add everything else and commit that.

obviously back things up before doing this :)

soltysh:
> obviously back things up before doing this :)

Actually you don't have to, since you'll be modifying locally, the backup is already at github, as long as you don't push ;)

bparees:
> Actually you don't have to, since you'll be modifying locally, the backup is already at github, as long as you don't push ;) 

yeah i know, but belt+suspenders.......

gabemontero:
Roger that @bparees and @soltysh.  Already got a non-git backup :-) but
yeah I'll play around with git and further my git education a bit :-)

PR #552, opened 2016-07-22
Added checks for existing content is git repo

### LINKED ISSUE #298: Unhelpful error message if .git project repository is empty
e.g. I have a not-empty project repository with my Ruby app, but with an empty .git repository. Then the following error message is misleading since I as a user have nothing to do with `/tmp/src`.
```
s2i build file:///$PWD openshift/ruby-22-centos7 newapp
I1002 08:32:59.352893 20229 sti.go:406] ---> Installing application source ...
E1002 08:32:59.358128 20229 util.go:85] cp: cannot stat '/tmp/src/.': No such file or directory
``

### PR DESCRIPTION
**Pull request by oatmealraisin**
This fixes an error when the git repo is empty, e.g. the user only ran
`git init`.  If the repo is empty, we use system copy commands instead
of git commands, and warn the user.

Also added warnings for if there is a .git directory in the project, but
a Git binary is not installed.

Currently working on tests, just posting this for discussion
@bparees 

Old output, git installed:
```

---> Installing application source ...
ERROR: cp: cannot stat '/tmp/src/.': No such file or directory

ERROR: An error occurred: non-zero (13) exit code from openshift/ruby-22-centos7
```

Old output, git not installed:
```

---> Installing application source ...
---> Building your Ruby application from source ...
WARNING: Rubygem Rack is not installed in the present image.
Add rack to your Gemfile in order to start the web server.

```

New output, git installed:
```
Warning: file:///home/rymurphy/documents/projects/go/src/github.com/openshift/source-to-image/asdf is a Git repository, but is empty.

---> Installing application source ...
---> Building your Ruby application from source ...
WARNING: Rubygem Rack is not installed in the present image.
Add rack to your Gemfile in order to start the web server.

```

New output, git not installed
```
Warning: file:///home/rymurphy/documents/projects/go/src/github.com/openshift/source-to-image/asdf is a Git repository, but you do not have git installed.
Warning: file:///home/rymurphy/documents/projects/go/src/github.com/openshift/source-to-image/asdf is a Git repository, but is empty.

---> Installing application source ...
---> Building your Ruby application from source ...
WARNING: Rubygem Rack is not installed in the present image.
Add rack to your Gemfile in order to start the web server.

```
Fixes #298

### DIFF
--- a/pkg/errors/errors.go
+++ b/pkg/errors/errors.go
@@ -21,6 +21,8 @@ const (
 	STIContainerError
 	SourcePathError
 	UserNotAllowedError
+	MissingGitBinaryError
+	EmptyGitRepositoryError
 )
 
 // Error represents an error thrown during STI execution
@@ -230,5 +232,26 @@ func NewUserNotAllowedError(image string, onbuild bool) error {
 		Message:    msg,
 		ErrorCode:  UserNotAllowedError,
 		Suggestion: fmt.Sprintf("modify image %q to use a numeric user within the allowed range or build without the --allowed-uids flag", image),
+	}
+}
+
+// NewMissingGitBinaryError returns a new error that indicates that we couldn't
+// find a Git binary in the path.
+func NewMissingGitBinaryError() error {
+	return Error{
+		Message:    "Git was not found in the PATH",
+		ErrorCode:  MissingGitBinaryError,
+		Suggestion: "Install Git from your distribution's vendor, or use --copy to ignore the repository.",
+	}
+}
+
+// NewEmptyGitRepositoryError returns a new error which indicates that a found
+// .git directory has no tracking information, e.g. if the user simply used
+// `git init` and forgot about the repository
+func NewEmptyGitRepositoryError(source string) error {
+	return Error{
+		Message:    fmt.Sprintf("The git repository \"%s\" has no tracking information or commits", source),
+		ErrorCode:  EmptyGitRepositoryError,
+		Suggestion: "Either commit files to the Git repository, remove the .git directory from the project, or use --copy to ignore the repository.",
 	}
 }
--- a/pkg/scm/git/clone.go
+++ b/pkg/scm/git/clone.go
@@ -23,7 +23,12 @@ func (c *Clone) Download(config *api.Config) (*api.SourceInfo, error) {
 	hasRef := len(config.Ref) > 0
 	cloneConfig := api.CloneConfig{Quiet: true, Recursive: !config.IgnoreSubmodules && !hasRef}
 
-	if c.ValidCloneSpec(config.Source) {
+	ok, err := c.ValidCloneSpec(config.Source)
+	if err != nil {
+		return nil, err
+	}
+
+	if ok {
 		if len(config.ContextDir) > 0 {
 			targetSourceDir = filepath.Join(config.WorkingDir, api.ContextTmp)
 			glog.V(1).Infof("Downloading %q (%q) ...", config.Source, config.ContextDir)
--- a/pkg/scm/git/git.go
+++ b/pkg/scm/git/git.go
@@ -13,6 +13,7 @@ import (
 	"strings"
 
 	"github.com/openshift/source-to-image/pkg/api"
+	"github.com/openshift/source-to-image/pkg/errors"
 	"github.com/openshift/source-to-image/pkg/util"
 	utilglog "github.com/openshift/source-to-image/pkg/util/glog"
 )
@@ -21,7 +22,7 @@ var glog = utilglog.StderrLog
 
 // Git is an interface used by main STI code to extract/checkout git repositories
 type Git interface {
-	ValidCloneSpec(source string) bool
+	ValidCloneSpec(source string) (bool, error)
 	ValidCloneSpecRemoteOnly(source string) bool
 	MungeNoProtocolURL(source string, url *url.URL) error
 	Clone(source, target string, opts api.CloneConfig) error
@@ -95,13 +96,17 @@ func stringInSlice(s string, slice []string) bool {
 
 // ValidCloneSpec determines if the given string reference points to a valid git
 // repository
-func (h *stiGit) ValidCloneSpec(source string) bool {
-	details, _ := ParseFile(source)
-	if details.FileExists && !details.BadRef {
-		return true
+func (h *stiGit) ValidCloneSpec(source string) (bool, error) {
+	details, _, err := ParseFile(source)
+	if err != nil {
+		return false, err
 	}
 
-	return h.ValidCloneSpecRemoteOnly(source)
+	if details.FileExists && !details.BadRef {
+		return true, nil
+	}
+
+	return h.ValidCloneSpecRemoteOnly(source), nil
 }
 
 // ValidCloneSpecRemoteOnly determines if the given string reference points to a valid remote git
@@ -136,7 +141,10 @@ func (h *stiGit) MungeNoProtocolURL(source string, uri *url.URL) error {
 		return nil
 	}
 
-	details, mods := ParseFile(source)
+	details, mods, err := ParseFile(source)
+	if err != nil {
+		return err
+	}
 
 	if details.BadRef {
 		return fmt.Errorf("bad reference following # in %s", source)
@@ -195,11 +203,6 @@ func ParseURL(source string) (*url.URL, error) {
 	return uri, nil
 }
 
-func useCopy(protoSpecified bool, source string) bool {
-	glog.V(4).Infof("useCopy proto spec %v local git repo %v has git bin %v", protoSpecified, isLocalGitRepository(source), hasGitBinary())
-	return !isLocalGitRepository(source) || !hasGitBinary()
-}
-
 // mimic the path munging (make paths absolute to fix file:// with non-absolute paths) done when scm.go:DownloderForSource valided git file urls
 func makePathAbsolute(source string) string {
 	glog.V(4).Infof("makePathAbsolute %s", source)
@@ -217,84 +220,80 @@ func makePathAbsolute(source string) string {
 // expect git clone spec syntax; it also provides details if the file://
 // proto was explicitly specified, if we should use OS copy vs. the git
 // binary, and if a frag/ref has a bad format
-func ParseFile(source string) (details *FileProtoDetails, mods *URLMods) {
+func ParseFile(source string) (*FileProtoDetails, *URLMods, error) {
+	// Checking to see if the user included a "file://" in the call
 	protoSpecified := false
 	if strings.HasPrefix(source, "file://") && len(source) > 7 {
 		protoSpecified = true
 	}
 
+	refSpecified := false
+	path, ref := "", ""
+	if strings.LastIndex(source, "#") != -1 {
+		refSpecified = true
+
+		segments := strings.SplitN(source, "#", 2)
+		path = segments[0]
+		ref = segments[1]
+	} else {
+		path = source
+	}
+
 	// in each valid case, like the prior logic in scm.go did, we'll make the
 	// paths absolute and prepend file:// to the path which callers should
 	// switch to
+	if doesExist(path) {
 
-	// if source, minus potential file:// prefix and #ref suffix,
-	// exists as is, denote and return
-	if doesExist(source) || doesExist(strings.SplitN(source, "#", 2)[0]) {
-		details = &FileProtoDetails{
-			FileExists:     true,
-			UseCopy:        useCopy(protoSpecified, source),
-			ProtoSpecified: protoSpecified,
-			BadRef:         false,
+		// Is there even a valid .git repository?
+		isValidGit, err := isValidGitRepository(path)
+		hasGit := false
+		if isValidGit {
+			hasGit = hasGitBinary()
 		}
-		mods = &URLMods{
-			Scheme: "file",
-			Path:   makePathAbsolute(strings.TrimPrefix(source, "file://")),
-		}
-		return
-	}
 
-	// need to see if this was a file://<valid file>#ref
-	if protoSpecified && strings.LastIndex(source, "#") != -1 {
-		segments := strings.SplitN(source, "#", 2)
-		// given last index check above, segement should
-		// be of len 2
-		path, ref := segments[0], segments[1]
-		// file does not exist, return bad
-		if !doesExist(path) {
-			details = &FileProtoDetails{
-				UseCopy:        false,
-				FileExists:     false,
-				ProtoSpecified: protoSpecified,
-				BadRef:         false,
-			}
-			mods = nil
-			return
-		}
-		// if ref/frag bad, return bad
-		if !gitSSHURLPathRef.MatchString(ref) {
-			details = &FileProtoDetails{
-				UseCopy:        false,
+		if err != nil || !isValidGit || !hasGit {
+			details := &FileProtoDetails{
+				UseCopy:        true,
 				FileExists:     true,
+				BadRef:         false,
 				ProtoSpecified: protoSpecified,
-				BadRef:         true,
 			}
-			mods = nil
-			return
+			mods := &URLMods{
+				Scheme: "file",
+				Path:   makePathAbsolute(strings.TrimPrefix(path, "file://")),
+				Ref:    ref,
+			}
+			return details, mods, err
 		}
 
-		// return good
-		details = &FileProtoDetails{
-			UseCopy:        useCopy(protoSpecified, source),
+		// Check is the #ref is valid
+		badRef := refSpecified && !gitSSHURLPathRef.MatchString(ref)
+
+		details := &FileProtoDetails{
+			BadRef:         badRef,
 			FileExists:     true,
 			ProtoSpecified: protoSpecified,
-			BadRef:         false,
+			// this value doesn't really matter, we should not proceed if the git ref is bad
+			// but let's fallback to "copy" mode if the ref is invalid.
+			UseCopy: badRef,
 		}
-		mods = &URLMods{
+
+		mods := &URLMods{
 			Scheme: "file",
-			Path:   makePathAbsolute(strings.TrimPrefix(source, "file://")),
+			Path:   makePathAbsolute(strings.TrimPrefix(path, "file://")),
 			Ref:    ref,
 		}
-		return
+		return details, mods, nil
 	}
 
-	// general return bad
-	details = &FileProtoDetails{
+	// File does not exist, return bad
+	details := &FileProtoDetails{
 		UseCopy:        false,
 		FileExists:     false,
 		BadRef:         false,
 		ProtoSpecified: protoSpecified,
 	}
-	return
+	return details, nil, nil
 }
 
 // ParseSSH will see if the input string is a valid git clone spec
@@ -366,11 +365,52 @@ func ParseSSH(source string) (*URLMods, error) {
 	return nil, fmt.Errorf("unable to parse ssh git clone specification:  %s", source)
 }
 
-// isLocalGitRepository checks if the specified directory has .git subdirectory (it
-// is a Git repository)
-func isLocalGitRepository(dir string) bool {
-	_, err := os.Stat(fmt.Sprintf("%s/.git", strings.TrimPrefix(dir, "file://")))
-	return !(err != nil && os.IsNotExist(err))
+// isValidGitRepository checks to see if there is a .git repository in the
+// directory and if the repository is valid -- i.e. it has remotes or commits
+func isValidGitRepository(dir string) (bool, error) {
+	gitDir := filepath.Join(strings.TrimPrefix(dir, "file://"), ".git")
+
+	// Check to see if .git directory even exists
+	if !doesExist(gitDir) {
+		// The direcory is not a git repo, no error
+		return false, nil
+	}
+
+	// Search the content of the .git directory for content
+	directories := [2]string{
+		filepath.Join(gitDir, "objects"),
+		filepath.Join(gitDir, "refs"),
+	}
+
+	// For the directories we search, if the git repo has been used, there will
+	// be some file.  We don't just search the base git repository because of the
+	// hook samples that are normally generated with `git init`
+	isEmpty := true
+	for _, dir := range directories {
+		err := filepath.Walk(dir, func(path string, info os.FileInfo, err error) error {
+			// If we find a file, the git directory is "not empty"
+			// We're looking for object blobs, and ref files
+			if info != nil && !info.IsDir() {
+				isEmpty = false
+				return filepath.SkipDir
+			}
+
+			return err
+		})
+
+		if err != nil && err != filepath.SkipDir {
+			// There is a .git, but we've encountered an error
+			return true, err
+		}
+
+		if !isEmpty {
+			return true, nil
+		}
+	}
+
+	// Since we know there's a .git directory, but there is nothing in it, we
+	// throw an error
+	return true, errors.NewEmptyGitRepositoryError(dir)
 }
 
 // doesExist checks if the path exists, removing file:// if needed for OS FS check
--- a/pkg/scm/git/git_test.go
+++ b/pkg/scm/git/git_test.go
@@ -10,27 +10,79 @@ import (
 	"testing"
 
 	"github.com/openshift/source-to-image/pkg/api"
+	"github.com/openshift/source-to-image/pkg/errors"
 	"github.com/openshift/source-to-image/pkg/test"
 )
 
+// Creates a git directory with one unlikely but possible commit hash
 func createLocalGitDirectory(t *testing.T) string {
-	dir, err := ioutil.TempDir(os.TempDir(), "s2i-test")
+	dir, err := ioutil.TempDir(os.TempDir(), "gitdir-s2i-test")
 	if err != nil {
 		t.Error(err)
 	}
-	os.Mkdir(filepath.Join(dir, ".git"), 0600)
+	os.MkdirAll(filepath.Join(dir, ".git/refs/heads"), 0777)
+	os.MkdirAll(filepath.Join(dir, ".git/refs/remotes"), 0777)
+	os.MkdirAll(filepath.Join(dir, ".git/branches"), 0777)
+	os.MkdirAll(filepath.Join(dir, ".git/objects/fo"), 0777)
+	os.Create(filepath.Join(dir, ".git/objects/fo") + "12345678901234567890123456789012345678") // 40 character SHA-1 hash
 	return dir
 }
 
-func TestIsLocalGitRepository(t *testing.T) {
+func createEmptyLocalGitDirectory(t *testing.T) string {
+	dir, err := ioutil.TempDir(os.TempDir(), "gitdir-s2i-test")
+	if err != nil {
+		t.Error(err)
+	}
+	os.MkdirAll(filepath.Join(dir, ".git/refs/heads"), 0777)
+	os.MkdirAll(filepath.Join(dir, ".git/refs/remotes"), 0777)
+	os.MkdirAll(filepath.Join(dir, ".git/branches"), 0777)
+	os.MkdirAll(filepath.Join(dir, ".git/objects"), 0777)
+	return dir
+}
+
+func TestIsValidGitRepository(t *testing.T) {
 	d := createLocalGitDirectory(t)
 	defer os.RemoveAll(d)
-	if isLocalGitRepository(d) == false {
+
+	// We have a .git that is populated
+	// Should return true with no error
+	ok, err := isValidGitRepository(d)
+	if !ok {
 		t.Errorf("The %q directory is git repository", d)
 	}
-	os.RemoveAll(filepath.Join(d, ".git"))
-	if isLocalGitRepository(d) == true {
+
+	if err != nil {
+		t.Errorf("isValidGitRepository returned an unexpected error: %q", err.Error())
+	}
+
+	d = createEmptyLocalGitDirectory(t)
+	defer os.RemoveAll(d)
+
+	// There are no tracking objects in the .git repository
+	// Should return true with an EmptyGitRepositoryError
+	ok, err = isValidGitRepository(d)
+	if !ok {
+		t.Errorf("The %q directory is a git repository, but is empty", d)
+	}
+
+	if err != nil {
+		if e, ok := err.(errors.Error); !ok || e.ErrorCode != errors.EmptyGitRepositoryError {
+			t.Errorf("isValidGitRepository returned an unexpected error: %q, expecting EmptyGitRepositoryError", err.Error())
+		}
+	} else {
+		t.Errorf("isValidGitRepository returned no error, expecting EmptyGitRepositoryError")
+	}
+
+	d = filepath.Join(d, ".git")
+
+	// There is no .git in the provided directory
+	// Should return false with no error
+	if ok, err = isValidGitRepository(d); ok {
 		t.Errorf("The %q directory is not git repository", d)
+	}
+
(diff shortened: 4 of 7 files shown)

### CODE REVIEW

**Comment by oatmealraisin** [comment]
Made the "Warning: " all caps

File: pkg/scm/git/git.go [hunk 1]
```diff
@@ -196,8 +196,19 @@ func ParseURL(source string) (*url.URL, error) {
 }
 
 func useCopy(protoSpecified bool, source string) bool {
-	glog.V(4).Infof("useCopy proto spec %v local git repo %v has git bin %v", protoSpecified, isLocalGitRepository(source), hasGitBinary())
-	return !isLocalGitRepository(source) || !hasGitBinary()
+	glog.V(4).Infof("useCopy proto spec %v local git repo %v has git bin %v empty git repository %v", protoSpecified, isLocalGitRepository(source), hasGitBinary(), isEmptyGitRepository(source))
+
+	if isLocalGitRepository(source) {
+		if !hasGitBinary() {
+			fmt.Println("WARNING: " + source + " is a Git repository, but you do not have git installed.")
```
bparees: glog.Warning ?

pkg/scm/git/git.go
```diff
@@ -196,8 +196,19 @@ func ParseURL(source string) (*url.URL, error) {
...
+		}
+
+		if isEmptyGitRepository(source) {
+			fmt.Println("WARNING: " + source + " is a Git repository, but is empty.")
```
bparees: glog.Warning?

bparees: tell them to use --copy if they want to use the local files instead.

pkg/scm/git/git.go
```diff
@@ -379,6 +390,27 @@ func doesExist(dir string) bool {
 	return !(err != nil && os.IsNotExist(err))
 }
 
+// isEmptyGitRepository checks if there are any commits, remotes, or branches
+// tracked in the repository, returns true if it is empty
+func isEmptyGitRepository(dir string) bool {
+	gitDir := strings.TrimPrefix(dir, "file://") + "/.git/"
+
+	// Search the content of the .git directory for content
+	cmd := "find " +
```
bparees: this isn't going to work on windows, needs a more portable solution.

pkg/scm/git/git.go
bparees: or rather, tell them you *are* using --copy, since that is what is happening.

pkg/scm/git/git.go
```diff
@@ -196,8 +196,19 @@ func ParseURL(source string) (*url.URL, error) {
...
+		if !hasGitBinary() {
+			fmt.Println("WARNING: " + source + " is a Git repository, but you do not have git installed.")
+		}
+
+		if isEmptyGitRepository(source) {
+			fmt.Println("WARNING: " + source + " is a Git repository, but is empty.")
+		}
+	}
+
+	return !isLocalGitRepository(source) || !hasGitBinary() || isEmptyGitRepository(source)
```
php-coder: I see that we're executing `isLocalGitRepository()` twice always and also executing `hasGitBinary()` and `isEmptyGitRepository()` twice sometimes. Because they're doing IO, maybe it's better to execute them only once? I mean, let's extract them to variables.

oatmealraisin: I agree, I'll change that

oatmealraisin:
[test]
@bparees ptal?

pkg/scm/git/git.go
```diff
@@ -379,6 +395,33 @@ func doesExist(dir string) bool {
...
 }
 
+// isEmptyGitRepository checks if there are any commits or remotes tracked in
+// the repository, returns true if it is empty
+func isEmptyGitRepository(dir string) bool {
+	var isEmpty = true // The return value, true for empty
+
+	gitDir := strings.TrimPrefix(dir, "file://") + "/.git"
+	// Search the content of the .git directory for content
+	directories := [2]string{gitDir + "/objects", gitDir + "/refs"}
```
php-coder: Could using of `/` lead to a problem on Windows?

pkg/scm/git/git.go
```diff
@@ -379,6 +395,33 @@ func doesExist(dir string) bool {
...
+
+	for _, dir := range directories {
+		filepath.Walk(dir,
+			func(path string, info os.FileInfo, err error) error {
+				if info != nil {
+					// If we find a file, the git directory is "not empty"
+					// We're looking for object blobs, and ref files
+					if !info.IsDir() {
+						isEmpty = false
```
php-coder: If we found a file do we need to continue or one file is enough? In this case we can put `break` here, right?

pkg/scm/scm_test.go
```diff
@@ -14,7 +14,12 @@ func createLocalGitDirectory(t *testing.T) string {
 	if err != nil {
 		t.Error(err)
 	}
-	os.Mkdir(filepath.Join(dir, ".git"), 0600)
+	os.MkdirAll(filepath.Join(dir, ".git/refs/heads"), 0777)
+	os.MkdirAll(filepath.Join(dir, ".git/refs/remotes"), 0777)
+	os.MkdirAll(filepath.Join(dir, ".git/branches"), 0777)
+	os.MkdirAll(filepath.Join(dir, ".git/objects/foo"), 0777)
+	os.Create(filepath.Join(dir, ".git/objects/foo") + "bar")
+	//os.Create(filepath.Join(dir, ".git/refs/heads") + "master")
```
php-coder: I hope that we need this commented code for something..

pkg/scm/git/git.go
php-coder: BTW, instead of prepending `gitDir` to each element of the array here, we can prepend it later, when we're passing first argument to `filepath.Walk()`. It's up to you, of course, it's just idea.

pkg/scm/git/git.go
php-coder: BTW, one more idea is to don't ignore `err` argument. I'm not sure how to use it in this case, probably, we could write it and at the end if `isEmpty` is `true` then show a warning with this error.

php-coder:
I've reviewed a bit, but the last opinion on @bparees of course, so you can ignore my comments, but I would be happy if some of ideas will help somehow.

pkg/scm/scm_test.go
oatmealraisin: Thanks for the feedback!  I forgot to take this out.  Since we already create a `bar` object, it is a "valid" git repository, because that would represent a commit.  I'll add a comment explaining that

pkg/scm/git/git.go
oatmealraisin: Correct, we only need 1 file to confirm it's a git repo.

As for the error, I don't know if we should do error checking in such a simple function.  Maybe we could combine the isLocalGitRepo() and isEmptyGitRepo() to isValidGitRepo(), with an extra error return value?

pkg/scm/git/git.go
```diff
@@ -196,8 +196,24 @@ func ParseURL(source string) (*url.URL, error) {
 }
 
 func useCopy(protoSpecified bool, source string) bool {
-	glog.V(4).Infof("useCopy proto spec %v local git repo %v has git bin %v", protoSpecified, isLocalGitRepository(source), hasGitBinary())
-	return !isLocalGitRepository(source) || !hasGitBinary()
+	isEmpty := isEmptyGitRepository(source)
+	isLocal := isLocalGitRepository(source)
+	hasGit := hasGitBinary()
```
bparees: you can optimize this further i think by lazily initialized the variables as needed inside the isLocal if block

eg if it's not a local git repo, you don't care if "hasGit" or "isEmpty" are true.

pkg/scm/git/git.go
bparees: yeah this should be using filepath.Join

pkg/scm/git/git.go
bparees: yeah you can just bubble the error up and ultimately fail the whole thing.

oatmealraisin: I'll combine them and do some error handling, this would keep us from assuming --copy as well :)

pkg/scm/git/git.go
oatmealraisin: noted, thanks

pkg/scm/git/git.go
oatmealraisin: Would it be OK if I got rid of that glog, or split it up?  I'll put something up

bparees: yeah you can rework it.

oatmealraisin:
Updated to bubble errors up and stop the program.  

New output: Git installed
```
s2i build file://$PWD openshift/ruby-22-centos7 newapp
ERROR: An error occurred: The git repository "file:///home/rymurphy/documents/projects/go/src/github.com/openshift/source-to-image/asdf" has no tracking information or commits
```
New output: Git not installed
```
 s2i build file://$PWD openshift/ruby-22-centos7 newapp
ERROR: An error occurred: Git was not found in the PATH
```
@bparees  @php-coder ptal?

pkg/scm/git/clone.go
```diff
@@ -23,7 +23,9 @@ func (c *Clone) Download(config *api.Config) (*api.SourceInfo, error) {
 	hasRef := len(config.Ref) > 0
 	cloneConfig := api.CloneConfig{Quiet: true, Recursive: !config.IgnoreSubmodules && !hasRef}
 
-	if c.ValidCloneSpec(config.Source) {
+	if ok, err := c.ValidCloneSpec(config.Source); err != nil {
+		return nil, err
+	} else if ok {
```
php-coder: Small neat: we don't need `else` if we're returning from the first branch.

pkg/scm/git/git.go
```diff
@@ -217,83 +219,95 @@ func makePathAbsolute(source string) string {
...
-	if protoSpecified && strings.LastIndex(source, "#") != -1 {
-		segments := strings.SplitN(source, "#", 2)
-		// given last index check above, segement should
-		// be of len 2
-		path, ref := segments[0], segments[1]
-		// file does not exist, return bad
-		if !doesExist(path) {
+	if doesExist(path) {
+		var isValidGit bool
+		isValidGit, err = isValidGitRepository(path)
```
php-coder: You can write it in a one line, right?
```go
isValidGit, err := isValidGitRepository(path)
```

pkg/scm/git/git.go
```diff
@@ -492,3 +556,27 @@ func pipeToLog(reader io.Reader, log func(...interface{})) {
...
+}
+
+type InvalidReferenceError struct {
+	ref    string // The invalid reference
+	source string // The input Git repo, with the ref
+}
+
+func (err InvalidReferenceError) Error() string {
+	return fmt.Sprintf("Invalid reference \"%s\" in %s", err.ref, err.source)
+}
```
php-coder: I'm not sure, but maybe it will be better to move them to `pkg/errors/errors.go`?

pkg/scm/git/git.go
oatmealraisin: This causes an error with go because it thinks we're shadowing the named return value err :(

pkg/scm/git/git.go
oatmealraisin: I'm ok with either!  Is that where all errors go?  It would make sense either way

pkg/scm/git/clone.go
oatmealraisin: Good point, will change

pkg/scm/git/git.go
php-coder: Ah, ok.

pkg/scm/git/git.go
php-coder: As I said, I'm not sure how it should be. Will wait @bparees response.

pkg/scm/git/git.go
```diff
@@ -217,83 +219,95 @@ func makePathAbsolute(source string) string {
 // expect git clone spec syntax; it also provides details if the file://
 // proto was explicitly specified, if we should use OS copy vs. the git
 // binary, and if a frag/ref has a bad format
-func ParseFile(source string) (details *FileProtoDetails, mods *URLMods) {
+func ParseFile(source string) (details *FileProtoDetails, mods *URLMods, err error) {
+	// NOTE: Why don't we just trim the source here?
```
bparees: not sure what you mean. we're explicitly needing to track the difference between something like:
file://bparees@github.com  (user specified a file with an unusual name)
and
bparees@github.com (user specified an ssh syntax w/ no explicit protocol)

what are you proposing w/ trim?

pkg/scm/git/git.go
```diff
@@ -217,83 +219,95 @@ func makePathAbsolute(source string) string {
...
 			}
 			mods = nil
 			return
 		}
-		// if ref/frag bad, return bad
-		if !gitSSHURLPathRef.MatchString(ref) {
+
+		if refSpecified {
+			// if ref/frag bad, return bad
+			// NOTE: Should we throw an error here?
```
bparees: no, we're already returning BadRef in the struct, assume the caller is responsible for handling that if it cares.

pkg/scm/git/git.go
```diff
@@ -217,83 +219,95 @@ func makePathAbsolute(source string) string {
...
 				BadRef:         false,
+				ProtoSpecified: protoSpecified,
 			}
 			mods = nil
 			return
 		}
-		// if ref/frag bad, return bad
-		if !gitSSHURLPathRef.MatchString(ref) {
+
+		if refSpecified {
```
bparees: i don't see the purpose of this new if statement.  if the ref is "" it will always be valid anyway, so the existing logic should work, no?

pkg/scm/git/git.go
```diff
@@ -217,83 +219,95 @@ func makePathAbsolute(source string) string {
...
+					FileExists:     true,
+					ProtoSpecified: protoSpecified,
+					UseCopy:        false,
+				}
+				mods = nil
+				return
+			}
+
+			// return good
+			isValidGit, _ := isValidGitRepository(path)
```
bparees: why are you checking this again when you already did it on 247?

pkg/scm/git/git.go
bparees: yeah it would make sense to move them there.

pkg/scm/scm.go
```diff
@@ -23,9 +23,13 @@ func DownloaderForSource(s string, forceCopy bool) (build.Downloader, string, er
...
 	}
 
-	details, mods := git.ParseFile(s)
+	details, mods, err := git.ParseFile(s)
 	glog.V(4).Infof("return from ParseFile file exists %v proto specified %v use copy %v", details.FileExists, details.ProtoSpecified, details.UseCopy)
+	if err != nil {
+		return nil, s, err
+	}
 
+	// NOTE: This could be handled by an error
```
bparees: i think it's being handled correctly, it's up to the caller whether it wants to try to press on w/ the other git information returned and ignore the bad ref, or bail out.

pkg/scm/git/git.go
oatmealraisin: Will do

pkg/scm/git/git.go
oatmealraisin: You understood correctly, I was wondering what the use-cases for file:// were.

Within ParseFile, whenever we use `path` we trim the prefix before using it.  If we trimmed it at the beginning, we wouldn't have to make those checks further down.  

Since the function name states we're working with a file, would it make sense to just trim the file:// prefix outright?  We'd still be tracking its use with `protoSpecified`

bparees: six of one, half dozen of the other.

pkg/scm/git/git.go
oatmealraisin: It's more for the good return value.  We could put the error check outside of the if statement, but we'd have to keep the good return inside of it.

pkg/scm/git/git.go
oatmealraisin: I'm not sure what you mean :(

pkg/scm/git/git.go
bparees: what error check?

pkg/scm/git/git.go
bparees: "these two things are basically the same so it doesn't really matter"

you're welcome to change it, but if you update fileExists to not strip file://, you need to make sure no one who is calling it expects it to retain that behavior.

pkg/scm/git/git.go
oatmealraisin: On line 262, ```if !gitSSHURLPathRef.MatchString(ref)```

oatmealraisin: I see what you're talking about now, sorry for the confusion

oatmealraisin:
After adding the errors to our errors package, the new outputs follow the format

```
s2i build file:///$PWD openshift/ruby-22-centos7 newapp
ERROR: An error occurred: The git repository "file:////home/rymurphy/documents/projects/go/src/github.com/openshift/source-to-image/asdf" has no tracking information or commits
ERROR: Suggested solution: Either remove the .git directory from the project, or update its tracking information.
ERROR: If the problem persists consult the docs at https://github.com/openshift/source-to-image/tree/master/docs. Eventually reach us on freenode #openshift or file an issue at https://github.com/openshift/source-to-image/issues providing us with a log from your build using --loglevel=3
```
@bparees What do you think about this output?

bparees:
> ERROR: Suggested solution: Either remove the .git directory from the project, or update its tracking information.


the suggested solutions would be any of:

1) commit the local files to the repository
2) remove the .git folder
3) use --copy instead

oatmealraisin:
New output
```
ERROR: An error occurred: The git repository "file:///home/rymurphy/documents/projects/go/src/github.com/openshift/source-to-image/asdf" has no tracking information or commits
ERROR: Suggested solution: Either commit files to the Git repository, remove the .git directory from the project, or use --copy to ignore the repository.
ERROR: If the problem persists consult the docs at https://github.com/openshift/source-to-image/tree/master/docs. Eventually reach us on freenode #openshift or file an issue at https://github.com/openshift/source-to-image/issues providing us with a log from your build using --loglevel=3
```

Also had to make changes to make sure --copy would work even with a faulty Git
@bparees ptal?

pkg/errors/errors.go
```diff
@@ -232,3 +234,24 @@ func NewUserNotAllowedError(image string, onbuild bool) error {
...
 	}
 }
+
+// NewMissingGitBinaryError returns a new error that indicates that we couldn't
+// find a Git binary in the path.
+func NewMissingGitBinaryError() error {
+	return Error{
+		Message:    "Git was not found in the PATH",
+		ErrorCode:  MissingGitBinaryError,
+		Suggestion: "Install Git from your distribution's vendor",
```
bparees: or use --copy to ignore the git repository

oatmealraisin: @bparees  Good catch, updated

pkg/scm/git/git.go
```diff
@@ -217,65 +220,57 @@ func makePathAbsolute(source string) string {
...
 				BadRef:         false,
+				ProtoSpecified: protoSpecified,
 			}
 			mods = nil
 			return
 		}
+
 		// if ref/frag bad, return bad
-		if !gitSSHURLPathRef.MatchString(ref) {
+		if refSpecified && !gitSSHURLPathRef.MatchString(ref) {
```
bparees: you're only going to get here if err!=nil above, is that intentional?  doesn't seem right.

oatmealraisin: Yes I believe so.  Before we were working with just the path, but now we're checking to see if the user included a valid reference to a branch.  

We could just add `!gitSSHURLPathRef...` in the struct declaration below, 
```
 		details = &FileProtoDetails{
 			FileExists:     true,
			UseCopy:        !isValidGit,
 			ProtoSpecified: protoSpecified,
 			BadRef:         !gitSSHURLPathRef.MatchString(ref),
 		}
```
and let the caller figure out that the source input has a bad ref.

oatmealraisin:
@bparees I've update isValidGit to not check for a Git binary, and moved that check up to ParseFile.  I've also removed the predeclared returns, and updated test cases. ptal?

pkg/scm/git/git.go
```diff
@@ -217,84 +220,75 @@ func makePathAbsolute(source string) string {
...
-			details = &FileProtoDetails{
-				UseCopy:        false,
-				FileExists:     false,
-				ProtoSpecified: protoSpecified,
+	// in each valid case, like the prior logic in scm.go did, we'll make the
+	// paths absolute and prepend file:// to the path which callers should
+	// switch to
+	if doesExist(path) {
+		isValidGit, err := isValidGitRepository(path)
+		hasGit := hasGitBinary()
```
bparees: no reason to check this if isValidGit is false, right?

pkg/scm/git/git.go
```diff
@@ -217,84 +220,75 @@ func makePathAbsolute(source string) string {
...
+		if err != nil {
```
bparees: shouldn't this be:  if err!=nil || !isValidGit || !hasGit {

?

under any of those circumstances we should be returning UseCopy: true (and returning the err, if any, so the caller can assess if it wants to proceed)

pkg/scm/git/git.go
```diff
@@ -217,84 +220,75 @@ func makePathAbsolute(source string) string {
...
-		if !gitSSHURLPathRef.MatchString(ref) {
-			details = &FileProtoDetails{
-				UseCopy:        false,
+		if refSpecified && !gitSSHURLPathRef.MatchString(ref) {
+			details := &FileProtoDetails{
+				BadRef:         true,
 				FileExists:     true,
 				ProtoSpecified: protoSpecified,
-				BadRef:         true,
+				UseCopy:        !(isValidGit && hasGit),
```
bparees: and then this can be UseCopy: false  since you've eliminated the "invalid git repo" and "missing git binary" cases above, if you make the change i suggested.

pkg/scm/git/git.go
```diff
@@ -217,84 +220,75 @@ func makePathAbsolute(source string) string {
...
 			}
-			mods = nil
-			return
+			return details, nil, nil
```
bparees: I think these two blocks where you're setting up details for the case where we do have a valid git repo and git binary should be combined:
```
badRef:=refSpecified && !gitSSHURLPathRef.MatchString(ref)
details:= &FileProtoDetails {
				UseCopy:        false,
				FileExists:     false,
				ProtoSpecified: protoSpecified,
				BadRef:         badRef,
}
mods= &URLMods{
			Scheme: "file",
			Path:   makePathAbsolute(strings.TrimPrefix(source, "file://")),
			Ref:    ref,
		}
return details, mods, nil
```
The old code returned mods as "nil" in the case that there was a BadRef, but it's not obvious to me that it matters whether mods gets returned as nil, or gets returned with the proper value.  I think the assumption the code made was that if BadRef was true, the caller wasn't going to do anything further.  But if that's true, it's fine to actually return a valid mods also.

oatmealraisin:
@bparees I've made the ParseFile function more lazy in it's checks, ptal?

pkg/scm/git/git.go
```diff
@@ -217,84 +220,98 @@ func makePathAbsolute(source string) string {
...
+		path = segments[0]
+		ref = segments[1]
+	} else {
+		path = source
+	}
+
 	// in each valid case, like the prior logic in scm.go did, we'll make the
 	// paths absolute and prepend file:// to the path which callers should
 	// switch to
+	if doesExist(path) {
```
bparees: this is what i was trying to guide you towards.   I think it retains all the logic we've discussed, but is more compact:

```
	// in each valid case, like the prior logic in scm.go did, we'll make the
	// paths absolute and prepend file:// to the path which callers should
	// switch to
	if doesExist(path) {

		// Is there even a valid .git repository?
		isValidGit, err := isValidGitRepository(path)
		hasGit:=false
		if isValidGit {
		  hasGit=hasGitBinary()
		}
		// err indicates empty git repo (should also consider throwing an error for no .git folder
		// so the caller knows specifically why it's not a valid git repo)
		if err != nil || !isValidGit || !hasGit {
			details := &FileProtoDetails{
				UseCopy:        true,
				FileExists:     true,
				BadRef:         false,  
				ProtoSpecified: protoSpecified,
			}
			mods := &URLMods{
				Scheme: "file",
				Path:   makePathAbsolute(strings.TrimPrefix(path, "file://")),
				Ref:    ref,
			}
			return details, mods, err
		}

		// Check is the #ref is valid
		badRef:=refSpecified && !gitSSHURLPathRef.MatchString(ref)

		details := &FileProtoDetails{
			BadRef:         badRef,
			FileExists:     true,
			ProtoSpecified: protoSpecified,
			// this value doesn't really matter, we should not proceed if the git ref is bad
			// but let's fallback to "copy" mode if the ref is invalid.
			UseCopy:        badRef,  
		}

		mods := &URLMods{
			Scheme: "file",
			Path:   makePathAbsolute(strings.TrimPrefix(path, "file://")),
			Ref:    ref,
		}
		return details, mods, nil
	}

	// File does not exist, return bad
	details := &FileProtoDetails{
		UseCopy:        false,
		FileExists:     false,
		BadRef:         false,
		ProtoSpecified: protoSpecified,
	}
	return details, nil, nil
```

oatmealraisin:
@bparees I've put in your suggestion, ptal?

pkg/scm/git/git.go
```diff
@@ -217,84 +220,81 @@ func makePathAbsolute(source string) string {
...
+		hasGit := false
+		if isValidGit {
+			hasGit = hasGitBinary()
 		}
-		// if ref/frag bad, return bad
-		if !gitSSHURLPathRef.MatchString(ref) {
-			details = &FileProtoDetails{
-				UseCopy:        false,
+		// err indicates empty git repo (should also consider throwing an error for no .git folder
+		// so the caller knows specifically why it's not a valid git repo)
```
bparees: this comment was a suggestion to you.  either implement it or remove the comment, but it doesn't belong here (if you're not going to do it now, but still think it's a good idea, it should be to a //TODO on isValidGitRepository

bparees:
> @bparees I've put in your suggestion, ptal?

did it make sense?

oatmealraisin:
@bparees I decided not to implement an error for missing .git

pkg/scm/git/git.go
```diff
@@ -366,11 +365,52 @@ func ParseSSH(source string) (*URLMods, error) {
 	return nil, fmt.Errorf("unable to parse ssh git clone specification:  %s", source)
 }
 
-// isLocalGitRepository checks if the specified directory has .git subdirectory (it
-// is a Git repository)
-func isLocalGitRepository(dir string) bool {
-	_, err := os.Stat(fmt.Sprintf("%s/.git", strings.TrimPrefix(dir, "file://")))
-	return !(err != nil && os.IsNotExist(err))
+// isValidgitRepository checks to see if there is a .git repository in the
```
bparees: isValidGitRepository

pkg/scm/git/git_test.go
```diff
@@ -13,23 +13,56 @@ import (
...
+
+	// We have a .git that is populated
+	// Should return true with no error
+	if ok, err := isValidGitRepository(d); ok == false || err != nil {
 		t.Errorf("The %q directory is git repository", d)
 	}
-	os.RemoveAll(filepath.Join(d, ".git"))
-	if isLocalGitRepository(d) == true {
+
+	os.RemoveAll(d)
```
bparees: why is this one not a defer before you do the test?

pkg/scm/git/git_test.go
```diff
@@ -13,23 +13,56 @@ import (
...
+	d = createEmptyLocalGitDirectory(t)
+	defer os.RemoveAll(d)
+
+	// There are no tracking objects in the .git repository
+	// Should return true with an error
+	if ok, err := isValidGitRepository(d); ok == false || err == nil {
+		t.Errorf("The %q directory is not a git repository", d)
```
bparees: for completeness you should check that the error you got is the type of error (emptygitdir) you expected.

pkg/scm/git/git_test.go
```diff
@@ -13,23 +13,56 @@ import (
...
+	}
+
+	d = filepath.Join(d, ".git")
+
+	// There is no .git in the provided directory
+	// Should return false with no error
+	if ok, _ := isValidGitRepository(d); ok == true {
```
bparees: you should check that err==nil

pkg/scm/scm.go
```diff
@@ -23,8 +23,11 @@ func DownloaderForSource(s string, forceCopy bool) (build.Downloader, string, er
 		return &empty.Noop{}, s, nil
 	}
 
-	details, mods := git.ParseFile(s)
+	details, mods, err := git.ParseFile(s)
 	glog.V(4).Infof("return from ParseFile file exists %v proto specified %v use copy %v", details.FileExists, details.ProtoSpecified, details.UseCopy)
+	if err != nil && !forceCopy {
```
bparees: this should be checking the type of error that came back.  the only error you can ignore here is an "EmptyGitDir" error, right?

pkg/scm/git/git_test.go
oatmealraisin: A bit of confusion about how defer works, will change

pkg/scm/git/git.go
oatmealraisin: Good catch

oatmealraisin:
@bparees I added some checks for error types within the tests, as well as to scm.go, ptal?

pkg/scm/git/git_test.go
```diff
@@ -10,28 +10,80 @@ import (
...
 	}
-	os.RemoveAll(filepath.Join(d, ".git"))
-	if isLocalGitRepository(d) == true {
+
+	if err != nil {
+		t.Errorf("isValidGitRepository returned an unexpected error: %q", err.Error())
+	}
+
+	d = createEmptyLocalGitDirectory(t)
+	defer os.RemoveAll(d)
```
bparees: actually now i'm wondering how this will work also.  is it going to call RemoveAll with the current value of "d" at the time of the return, or is it going to call RemoveAll with the value of "d" at the time the defer was declared?  Obviously we need the latter behavior, or you need to use separate variables for each.

bparees:
one last question on the behavior of defer, and needs commit squashing.

pkg/scm/git/git_test.go
oatmealraisin: That's what I was wondering before :) I looked it up, defer evaluates the arguments when called but delays the actual function until the surrounding function returns.

oatmealraisin:
I'll squash now

bparees:
[merge]

pkg/errors/errors.go
```diff
@@ -232,3 +234,24 @@ func NewUserNotAllowedError(image string, onbuild bool) error {
 		Suggestion: fmt.Sprintf("modify image %q to use a numeric user within the allowed range or build without the --allowed-uids flag", image),
 	}
 }
+
+// NewMissingGitBinaryError returns a new error that indicates that we couldn't
+// find a Git binary in the path.
+func NewMissingGitBinaryError() error {
```
php-coder: @oatmealraisin Ryan, this method looks unused. Should we use it somewhere or just remove?

oatmealraisin: @php-coder We could use it when hasGitBinary() is callled on line 251 of git/git.go, but I think we talked about whether we should error out when the user doesn't have Git installed and decided that we would just use --copy instead.  

Are we tight about removing unused code?  Imo it's a legitimate error, and we may use it if we decide otherwise.  I'm not very opposed to just nixing it, thought.

bparees: since we don't have immediate plans to resume using this, we should remove it.

PR #554, opened 2016-07-25
Move to kube and engine2

gabemontero:
OK, let it begin  ..... 

@bparees @csrwng @jupierce @oatmealraisin @php-coder @PI-Victor @soltysh


csrwng:
@gabemontero since you're already at it... :) I think it would make sense to switch away from the Godeps directory to a vendor directory like origin.

gabemontero:
@csrwng new commit added to move from _workspace to vendor  ... seems to be working for build, the 3 test scripts, and make verify

soltysh:
I've looked into this PR and I'm terrified by the amount of deps needed to get this in :cry: The binary itself grows 3 times (from 11MB to 33MB), it's insane. There are two questions that should be asked: 

1. is it possible to vendor a single/a few subdirs?
2. if 1 is yes, then try to minimize upstream the amount of deps needed to have `k8s.io/kubernetes/pkg/kubelet` 

Without 1, I'm not sure if it pays to have the entire k8s just to use that abstraction layer. If maintaining our own abstraction wouldn't be cheaper?

gabemontero:
I developed your concern as well @soltysh when getting full k8s to compile,
and was curious if others were going to chime in.
My experience with godeps is not extensive, but I've been told that cherry
picking packages from a dependency could get tricky
and to try to avoid it.

The question of talking solely to engine-api vs. k8s/engine-api did come up
in the beginning and @bparees said we wanted to
pursue the later.

Pending comments on those two points, I can start experimenting with a
version on the side that attempts to prune the k8s godep
while others comment on the s2i specific changes.

thanks

gabemontero:
OK, @bparees and I have already gone over this and have a decision, though I'll post the results here to allow others to debate / contest as they see fit.

What I did was:
1) git removed everything except the engine-api and pkg/kubelet/dockertools packages
2) attempted a build and captured the packages that were needed
2) on the captured output parsed out the needed packages and `git reset` and `git checkout` the packages back in
3) got s2i to compile again with a reduced Godeps.

The size savings in the end weren't huge.  The particulars:
1) The new s2i binary is `42984208` (I was seeing around 49M before ..using go 1.6) ... an s2i I built from the current HEAD is `13219968`
2) So still almost 4 times the size
3) I was able to delete 3246 files and about half of the packages I was originally importing.  But one can surmise the beefiest ones are still there.

What changed:
1) not knowing exactly what I needed before, I pulled in all of k8s the first time.

The decision:
1) maintain the use of k8s ... getting its abstraction of the container runtime for free is the point of all this from @bparees 's perspective
2) introduce the reduction I've just done ... I'll update this PR once that is pushed

soltysh:
@gabemontero the only thing that worries me is the reduction impacting the latter k8s rebase process in any way, if not - I'm good with that. Hopefully we'll be able to isolate the upstream package more. I can volunteer to do that one.

gabemontero:
noted @soltysh , and totally, whenever you can start an isolation refactor upstream, that will be great.

gabemontero:
ok, i've updated the PR with the Godeps pruning we have discussed today.

gabemontero:
OK, minor tweaks (script related) to get travis to pass.

This should be ready for folks review when y'all get the chance.  Thanks.

hack/test-stirunimage.sh
```diff
@@ -62,15 +63,6 @@ function test_debug() {
...
 }
 
-set +e
-img_count=$(docker images | grep -c sti_test/sti-fake)
-set -e
-
-if [ "${img_count}" != "10" ]; then
-    echo "You do not have necessary test images, be sure to run 'hack/build-test-images.sh' beforehand."
-    exit 1
-fi
```
bparees: what's the reason for removing this?

gabemontero: those test images are not needed for the actual tests run in this script

bparees: good reason. :)

pkg/api/types.go
```diff
@@ -283,6 +282,25 @@ type DockerConfig struct {
...
+// AuthConfig is our abstraction of the Registry authorization information for whatever
+// docker client we happen to be based on
+type AuthConfig struct {
+	Username      string
+	Password      string
+	Email         string
+	ServerAddress string
+}
+
+// Image is the abstraction of the docker client provider (formerly fsouza, now either
```
bparees: formerly go-dockerclient

pkg/docker/docker.go
```diff
@@ -54,6 +65,9 @@ const (
 	// DefaultDockerTimeout specifies a timeout for Docker API calls. When this
 	// timeout is reached, certain Docker API calls might error out.
 	DefaultDockerTimeout = 20 * time.Second
+
+	// DefaultShmSize  is the default ShmSize to use (in bytes) if not specified.
```
bparees: there's an extra space here and should probably spell our "shared memory size" in the description (at least i'm assuming that's what this value is)

pkg/docker/docker.go
```diff
@@ -106,26 +120,25 @@ type Docker interface {
 // Client contains all methods called on the go Docker
 // client.
```
bparees: comment is wrong now?

pkg/docker/docker.go
```diff
@@ -106,26 +120,25 @@ type Docker interface {
...
-	UploadToContainer(id string, opts docker.UploadToContainerOptions) error
-	DownloadFromContainer(id string, opts docker.DownloadFromContainerOptions) error
-	RemoveContainer(opts docker.RemoveContainerOptions) error
-	CommitContainer(opts docker.CommitContainerOptions) (*docker.Image, error)
-	CopyFromContainer(opts docker.CopyFromContainerOptions) error
-	BuildImage(opts docker.BuildImageOptions) error
-	InspectContainer(id string) (*docker.Container, error)
-	Ping() error
+	ContainerWait(ctx context.Context, containerID string) (int, error)
+	// no CommitContainer in kube_docker_client.go, but engine-api has ContainerCommit
```
bparees: ContainerCommit

pkg/docker/docker.go
```diff
@@ -106,26 +120,25 @@ type Docker interface {
...
+	ContainerCommit(ctx context.Context, container string, options dockertypes.ContainerCommitOptions) (dockertypes.ContainerCommitResponse, error)
+	// CopyToContainer is the UploadToContainer equivalent
+	CopyToContainer(ctx context.Context, container, path string, content io.Reader, opts dockertypes.CopyToContainerOptions) error
+	// no CopyFromContainer in kube_docker_client.go ... exists in engine-api, is the DownloadFromContainer equivalent
+	CopyFromContainer(ctx context.Context, container, srcPath string) (io.ReadCloser, dockertypes.ContainerPathStat, error)
+	// no BuildImage in kube_docker_client.go
```
bparees: ImageBuild?

pkg/docker/docker.go
```diff
@@ -106,26 +120,25 @@ type Docker interface {
...
+	ImageBuild(ctx context.Context, buildContext io.Reader, options dockertypes.ImageBuildOptions) (dockertypes.ImageBuildResponse, error)
+	// no Ping in kube_docker_client.go or engine-api ... going to borrow a small bit from fsouza
```
bparees: this comment seems to imply there should be a Ping method here, is there supposed to be?  or we don't need it anymore?

in general these comments seem like they need to be cleaned up/removed.

pkg/docker/docker.go
```diff
@@ -256,35 +273,58 @@ type BuildImageOptions struct {
...
+			return nil, tlsErr
+		}
+		httpClient = &http.Client{
+			Transport: k8snet.SetTransportDefaults(&http.Transport{
+				TLSClientConfig: tlsc,
+			}),
+		}
 	}
+
+	client, err := dockerapi.NewClient(config.Endpoint, "", httpClient, nil)
```
bparees: isn't httpClient going to be nil if there was no cert/key/CA file?

csrwng: it is, and that's ok ... the docker client will use a default client

pkg/docker/docker.go
```diff
@@ -256,35 +273,58 @@ type BuildImageOptions struct {
...
 	}
+	k8sDocker := dockertools.ConnectToDockerOrDie(config.Endpoint)
 	return &stiDocker{
-		client:   client,
-		pullAuth: auth,
+		kubeDockerClient: k8sDocker,
+		client:           client,
+		httpClient:       httpClient,
+		dialer:           &net.Dialer{},
+		pullAuth: dockertypes.AuthConfig{
```
bparees: this is not abstracting us away from docker as much as i had assumed the kube libs would :(

pkg/docker/docker.go
```diff
@@ -296,13 +336,65 @@ func (d *stiDocker) GetImageWorkdir(name string) (string, error) {
...
 func (d *stiDocker) GetImageEntrypoint(name string) ([]string, error) {
-	image, err := d.client.InspectImage(name)
+	image, err := d.kubeDockerClient.InspectImage(name)
 	if err != nil {
 		return nil, err
 	}
 	return image.Config.Entrypoint, nil
 }
 
+// do is snippets of code borrowed from fsouza and engine-api for basick HTTP Rest flows;
```
bparees: s/basick/basic/

pkg/docker/docker.go
```diff
@@ -296,13 +336,65 @@ func (d *stiDocker) GetImageWorkdir(name string) (string, error) {
...
+// minimally used for the Ping operation, but could be used for POST's as well
+// if ever useful for debug
+func (d *stiDocker) do(method, path string, body io.Reader) (*http.Response, error) {
+	//TODO - for now, we are forgoing the version check and specific version requests that exist in fsouza;
```
bparees: s/fsouza/go-dockerclient/

pkg/docker/docker.go
```diff
@@ -358,48 +450,61 @@ func (d *stiDocker) UploadToContainerWithCallback(src, dest, container string, w
...
+	ctx, cancel := getDefaultContext(DefaultDockerTimeout)
+	defer cancel()
+	return d.client.CopyToContainer(ctx, container, path, r, dockertypes.CopyToContainerOptions{})
 }
 
 // DownloadFromContainer downloads file (or directory) from the container.
 func (d *stiDocker) DownloadFromContainer(containerPath string, w io.Writer, container string) error {
-	opts := docker.DownloadFromContainerOptions{Path: containerPath, OutputStream: w}
-	return d.client.DownloadFromContainer(container, opts)
+	//CopyFromContainer(ctx context.Context, container, srcPath string) (io.ReadCloser, types.ContainerPathStat, error)
```
bparees: looks like dead code/comment.

pkg/docker/docker.go
```diff
@@ -358,48 +450,61 @@ func (d *stiDocker) UploadToContainerWithCallback(src, dest, container string, w
...
-		return false, nil
 	}
-	return false, err
+	if err != nil {
+		if dockerapi.IsErrImageNotFound(err) {
+			return false, nil
+		}
+		return false, errors.NewInspectImageError(name, err)
+	}
+	return false, nil
```
bparees:
```
if err!=nil && !dockerapi.IsErrImageNotFound(err) {
  return false, errors.NewInspectImageError(name,err)
}
return false, nil
```

is equivalent, i think

pkg/docker/docker.go
```diff
@@ -452,58 +558,77 @@ func (d *stiDocker) CheckAndPullImage(name string) (*docker.Image, error) {
...
 	name = getImageName(name)
-	image, err := d.client.InspectImage(name)
+	inspect, err := d.kubeDockerClient.InspectImage(name)
 	if err != nil {
 		glog.V(4).Infof("error inspecting image %s: %v", name, err)
 		return nil, errors.NewInspectImageError(name, err)
 	}
-	return image, nil
+	if inspect != nil {
+		image := api.Image{}
```
bparees: image:=&api.Image{} and then you don't need the &s everywhere else.

pkg/docker/docker.go
```diff
@@ -690,14 +812,13 @@ func (d *stiDocker) RunContainer(opts RunContainerOptions) error {
...
-	var (
-		imageMetadata *docker.Image
-		err           error
-	)
-	if opts.PullImage {
-		imageMetadata, err = d.CheckAndPullImage(image)
-	} else {
-		imageMetadata, err = d.client.InspectImage(image)
+	inspect, err := d.kubeDockerClient.InspectImage(image)
+	var imageMetadata *api.Image = &api.Image{}
```
bparees: imageMetadata:=&api.Image{}

pkg/docker/docker.go
```diff
@@ -452,58 +558,77 @@ func (d *stiDocker) CheckAndPullImage(name string) (*docker.Image, error) {
...
 	}
-	image, err := d.client.InspectImage(name)
+	resp, err := d.kubeDockerClient.InspectImage(name)
 	if err != nil {
-		glog.V(4).Infof("error inspecting image %s: %v", name, err)
-		return nil, errors.NewInspectImageError(name, err)
+		return nil, errors.NewPullImageError(name, err)
+	}
+	if resp != nil {
+		image := api.Image{}
```
bparees: image:=&api.Image{}

pkg/docker/docker.go
```diff
@@ -542,6 +666,9 @@ func (d *stiDocker) GetScriptsURL(image string) (string, error) {
 	if err != nil {
 		return "", err
 	}
+	if imageMetadata == nil {
```
bparees: under what condition is imagemetadata nil, but we didn't get an error on CheckAndPullImage?

pkg/docker/docker.go
```diff
@@ -665,21 +795,13 @@ func determineCommandBaseDir(opts RunContainerOptions, imageMetadata *docker.Ima
...
-			}
-		}
-		liveports = liveports + "\n"
+func dumpContainerInfo(container *dockertypes.ContainerCreateResponse, d *stiDocker, image string) {
+	containerJSON, err := d.kubeDockerClient.InspectContainer(container.ID)
+	if err != nil {
+		return
 	}
+
+	liveports := "\n\nPort Bindings:  \n" + fmt.Sprintf("%+v\n", containerJSON.NetworkSettings.NetworkSettingsBase.Ports)
```
bparees: this prints out as nicely as the for loop did?  seems like it's just going to dump json formatted text?

pkg/docker/docker.go
```diff
@@ -768,19 +892,21 @@ func (d *stiDocker) RunContainer(opts RunContainerOptions) error {
...
-		if _, err = d.client.AttachToContainerNonBlocking(attachOpts); err != nil {
-			glog.V(0).Infof("error: Unable to attach to container %q with options %+v: %v", containerName, attachOpts, err)
-			return err
-		}
+		go func() {
+			attachOpts := opts.asDockerAttachToContainerOptions()
+			if attachError = d.kubeDockerClient.AttachToContainer(container.ID, attachOpts, opts.asDockerAttachToStreamOptions()); attachError != nil {
+				glog.V(0).Infof("error: Unable to attach to container %q with options %+v: %v", containerName, attachOpts, attachError)
+			}
+		}()
```
bparees: since we're doing this in a go-routine, how do we actually know the attach has been registered before we invoke the start?

pkg/docker/docker.go
```diff
@@ -800,15 +926,18 @@ func (d *stiDocker) RunContainer(opts RunContainerOptions) error {
 			// the container exits normally).  dump port/etc information for the user.
 			dumpContainerInfo(container, d, image)
 		}
+
 		// Return an error if the exit code of the container is
 		// non-zero.
 		glog.V(4).Infof("Waiting for container %q to stop ...", containerName)
-		exitCode, err := d.client.WaitContainer(container.ID)
+		ctx, cancel := getDefaultContext(15 * time.Minute)
```
bparees: where did this timeout come from and how does it compare with what we were doing previously?

pkg/docker/docker.go
```diff
@@ -823,6 +952,9 @@ func (d *stiDocker) RunContainer(opts RunContainerOptions) error {
 			c.Close()
 		}
 
+		if attachError != nil {
+			return attachError
+		}
```
bparees: shouldn't we be looking for this earlier?  (and don't we need to ensure the attach has run before checking it?  and since we're setting it in a go routine i'm guessing we need a memory barrier to ensure we're seeing the updated value)

pkg/docker/docker.go
```diff
@@ -864,54 +991,60 @@ func (d *stiDocker) GetImageID(name string) (string, error) {
...
 		dockerOpts.Memory = opts.CGroupLimits.MemoryLimitBytes
-		dockerOpts.Memswap = opts.CGroupLimits.MemorySwap
+		dockerOpts.MemorySwap = opts.CGroupLimits.MemorySwap
 		dockerOpts.CPUShares = opts.CGroupLimits.CPUShares
 		dockerOpts.CPUPeriod = opts.CGroupLimits.CPUPeriod
 		dockerOpts.CPUQuota = opts.CGroupLimits.CPUQuota
 	}
 	glog.V(2).Infof("Building container using config: %+v", dockerOpts)
-	return d.client.BuildImage(dockerOpts)
+	ctx, cancel := getDefaultContext(15 * time.Minute)
```
bparees: same timeout question here..where did it come from and how does it compare to the previous behavior?

pkg/docker/docker.go
```diff
@@ -452,58 +558,77 @@ func (d *stiDocker) CheckAndPullImage(name string) (*docker.Image, error) {
...
+
+func updateImageWithInspect(image *api.Image, inspect *dockertypes.ImageInspect) {
+	image.ID = inspect.ID
+	if inspect.Config != nil {
+		image.ConfigLabels = inspect.Config.Labels
+		image.ConfigEnv = inspect.Config.Env
+	}
+	if inspect.ContainerConfig != nil {
+		image.ContainerLabels = inspect.ContainerConfig.Labels
+		image.ContainerEnv = inspect.ContainerConfig.Env
```
bparees: would it make sense to more closely align the api.Image struct with the docker image struct?

pkg/docker/util.go
```diff
@@ -54,27 +64,88 @@ func GetImageRegistryAuth(auths *client.AuthConfigurations, imageName string) cl
...
+func LoadImageRegistryAuth(dockerCfg io.Reader) *AuthConfigurations {
+	auths, err := NewAuthConfigurations(dockerCfg)
 	if err != nil {
 		glog.V(0).Infof("error: Unable to load docker config")
 		return nil
 	}
 	return auths
 }
 
+// begin borrowed from fsouza
```
bparees: kube and/or docker-engine have nothing for processing auth/config files?

bparees: also s/fsouza/go-dockerclient/ here and everywhere else.

bparees:
@gabemontero first pass complete, just of the main commit and i didn't review the _test files in this pass.

gabemontero:
Sounds good @bparess, thanks for the note of demarcation.  I'll start
consuming / making updates based on these comments.

pkg/api/types.go
gabemontero: part of next push

pkg/docker/docker.go
gabemontero: yes shared memory - part of next push

pkg/docker/docker.go
gabemontero: yes it is.  For Client:

`Client contains all methods used when interacting directly with docker engine-api instead of the k8s abstraction around docker engine-api`

I also updated the comment around the Docker interface:

`Docker is the interface between STI and the k8s abstraction around docker engine-api.`

pkg/docker/docker.go
gabemontero: yep changed

pkg/docker/docker.go
gabemontero: yep

pkg/docker/docker.go
gabemontero: There *used* to be a Ping method in the Client interface.  But yes, overall, these comments were work in progress type reminders for me in the end.  I agree they can be removed.  Doing that for all of them.

pkg/docker/docker.go
gabemontero: @csrwng is correct; in fact, when I put a dummy/empty httpClient when no cert/key/CA file, that caused problems.

pkg/docker/docker.go
gabemontero: Agreed ... same with the pull options, remove container options

pkg/docker/docker.go
gabemontero: part of next push

pkg/docker/docker.go
gabemontero: part of next push

pkg/docker/docker.go
gabemontero: yep

pkg/docker/docker.go
gabemontero: duh yeah

pkg/docker/docker.go
gabemontero: not even sure how I landed with that ... gofmt should catch that :-)

pkg/docker/docker.go
gabemontero: yep

pkg/docker/docker.go
gabemontero: I could have swore I hit this during testing (should have left a comment), and at the time it appeared as if there was a path where the image was not found, but there was a no not found error.

That said, I walked down the code path, and it looks like it is returning a not found error in that case.

I'll remove it (for now), and it it shows up again when I re-run the test suite, I'll update here (and comment the code as to why we need the check).

gabemontero: also, I do remember later on adding a nil check in the private getScriptsURL (for path other than CheckAndPull).  So yeah, this is no longer needed.

pkg/docker/docker.go
gabemontero: yeah there is less available in this area with engine-api than there was with go-dockerclient;  we are in fact only printing a map of string keys and string values, not JSON.

it looked OK to me, but sure, I can cycle through the map and assign labels for the keys/values, and add new lines.

gabemontero: ok ... yeah, upon review, I sliced/diced more than needed ...it should be farily close (minus that missing information I mentioned) to what was there before

soltysh:
You have `github.com/Sirupsen/logrus` in two versions in `Godeps.json`, both should be the same.

gabemontero:
I had caught that with version 1 of the Godeps, but it snuck back in when I
did the pruning ... thanks for the catch.

pkg/docker/docker.go
gabemontero: Yeah, I should have added more comments here.  And I'll preface this by saying that I concede an change may be warranted here.

First, on what occurred with the go-dockerclient:
- remember, the go-dockerclient basically had the same situation since attach was non-blocking 
- go-dockerclient attach returned errors either a) when the container id was empty (should not occur for us since we just created the container we are attaching to) and b) the http post returned an error

So, as your comment below shows, you see that the attachError is defined outside the go-routine.  I started down the path of checking it sooner and blocking until it "changed" somehow, then thought it seemed hacky, so I pulled that and pushed the vanilla check down.

Circling back, where I was headed originally was:
1) set attachError to some dummy error
2) after the go-routine is launched, block until attachError is either changed to nil, or something other than our dummy error
3) then move onto start

That at least would in my mind mirror the go-dockerclient behavior

Too hacky?  Or do you have similar behaving alternative?

pkg/docker/docker.go
gabemontero: see above comment

pkg/docker/docker.go
bparees: channels are the standard way to communicate state between go routines.  Setup a channel and have the go routine send either the error, or nil, through the channel when it's done.

and then in the main code, do a read from the channel to block until you have a result from the go routine.

gabemontero: Thiught of that initially too but the go routine that does the attach will
block and not update the channel the main thread is waiting on so we end up
in the same boat.

So I can't help but circle back to the hacky poll the err for small but
sufficient amonut to insure the post for the attach did not fail
"immediately"

bparees: you can do that w/ a select on the channel(s)  (select on the error channel and a timer channel)

gabemontero: ahhh ... right, I've seen that used elsewhere.  OK, will work on that next.

pkg/docker/docker.go
gabemontero: I wanted to keep this minimal from a field perspective, and I wasn't a fan of a struct within a struct that had just one field.

pkg/docker/docker.go
gabemontero: a combination of things:
1) the k8s client talks about short term ops and long term ops (pull/build image, attach/wait container) ... the default in k8s for short term is 2 minutes
2) the go-dockerclient did not appear to set a timeout on the client side by default, they just allow you to set them (we never did, but used util.TimeoutAfter around some operations, where are default was 20 seconds);  so go-dockerclient's default is 0; following the trail down into go's http support and time support (see go http's `setRequestCancel` in client.go), it was blocking forever (essentially the now time - 0)

That correlates with what I remember from adding the `--run` support.  If the container stays up, you had to ctrl-c to finish the `s2i` command.

And I believe I've heard of builds "hanging" before (correct me if I'm wrong).

Background done.  Rationale:
1) I was uncomfortable waiting forever on both, though perhaps it makes sense for the run container ... I could put the max int/log there 
2) for the build, 15 was based on some experimentation and ad hoc heuristics ... I'm certainly amenable to suggestions

Bottom line, waiting forever

pkg/docker/docker.go
gabemontero: see above comment

pkg/docker/util.go
gabemontero: the string fsouza should no longer exists in any of the repo's go files

pkg/docker/docker.go
bparees: i'm just worried about the pain if it grows in the future

pkg/docker/docker.go
bparees: I think we should retain the existing behavior (wait forever). If we hardcode the build to a 15min timeout, no one is going to be able to build anything that takes longer than 15 mins, which is clearly unacceptable, and while we could add a --timeout argument to s2i, we'd then also have to add that field to openshift which again, seems undesirable.

for run it definitely makes sense to wait forever.

pkg/docker/util.go
gabemontero: on the file processing, there was nothing that could take a reader of streamed output; it always assumed the struct was hydrated at the k8s wrapper / engine-api levels

pkg/docker/docker.go
gabemontero: ok ... I'll make it forever for both

pkg/docker/util.go
bparees: And the code they use to hydrate it isn't public?  The reason i'm worried is that docker has previously changed the format of the config.json file and broken us, so i'd rather depend on an api for reading this file than doing it ourselves.

gabemontero: The go-dockerclient method is public, and the json/object schema from
docker of course is.

But continuing to depend on go-dockerclient is as worrisome a prospect, no?

gabemontero: I'll send an email to the engine-api maintainer we've communicated with ... see if he reveals something I missed.

bparees: definitely depending on go-dockerclient isn't what we want, i guess i'm still struggling to understand how k8s handles reading a docker config.json and finding auths, and why we can't leverage it in some way.

bparees: (or, if not k8s, at least engine-api)

gabemontero: the k8s stuff also pulls from well known env vars ... it was my thinking that those env vars get set when the various openshift components run, but that doesn't apply for s2i

pkg/docker/docker.go
gabemontero: in fact, the util.TimeoutAfter() stuff I mentioned earlier seems to do the exact trick with the channels.... I'm just reusing that.

gabemontero: Actually, what I feared earlier sort of does in fact transpire, even with select on error/timer channels, like is done in both util.TimeoutAfter and in pkg/tar/tar.go, unless we re-swizzle what is done on timeout.

My testing confirms that we always end up timing out because since the attach blocks, and the container never *exits* because it is created, not started, we always end up exiting on the timer channel.

I think the best we can do is say use our default, quick action, default docker timeout of 20 seconds, and if the error hasn't occurred by then, assume the attach is good, and move onto the 
start.

gabemontero:
I've gone ahead and pushed my updates based on @bparees 's comments to date (review of the test files still pending) with the following things to note:
- note the *atypical* interpretation of timeout on the attach call ... see comments and discussion thread off of @bparees 's comment
- I'm still mulling the comment on mimicking structs with structs in the Image abstraction ... simplicity (why only have a struct that encompases as single field ... add the struct later if needed) vs. unforseen future concerns related to growth

pkg/docker/docker.go
gabemontero: oops ... missed this one ... pushing momentarily

gabemontero:
On the attach ... bottom line, the k8s blocking wrapper of the engine-api attach is hurting us more than helping.  I'm bypassing it for now and going directly to engine-api.  We can work with @soltysh and others to make the upstream version more flexible.

gabemontero:
OK, I've pushed the changes for attach.  I call the engine-api attach, and use the error / timer channels to confirm the core attached worked, and we can move onto the start container in the main thread.  The attach go routine then blocks on the IO processing.  Since the k8s wrapper did not make their hijack response processing public, I had to copy from that code.  

Minimally, if the k8s wrapper could expose that method publicly, we could get rid of the duplicate code.  Ideally, if the k8s attach wrapper returned both the error for the core attach call as well as return channel struct or some such that encapsulated the reader/writer streaming, and allow the caller to block on that as desired, that would be ideal.

pkg/docker/docker.go
bparees: yup, sounds good.  I wasn't suggesting we add everything, just that for the fields we are using, we mirror the upstream structure so if we need to pull in more fields later, it is natural.

gabemontero:
OK, I've updated my response on the format of our Image struct abstraction, as well as added some unit tests around access to the engine-api code via kube code (param validation, etc.).

@bparees (and others who are interested) - ptal

pkg/docker/docker.go
gabemontero: On Mon, Aug 1, 2016 at 3:38 PM, Ben Parees <notifications@github.com> wrote:

> In pkg/docker/docker.go
> <https://github.com/openshift/source-to-image/pull/554#discussion_r73039222>
> :
>
> > +		image := api.Image{}
> > +		updateImageWithInspect(&image, resp)
> > +		return &image, nil
> > +	}
> > +	return nil, nil
> > +}
> > +
> > +func updateImageWithInspect(image *api.Image, inspect *dockertypes.ImageInspect) {
> > +	image.ID = inspect.ID
> > +	if inspect.Config != nil {
> > +		image.ConfigLabels = inspect.Config.Labels
> > +		image.ConfigEnv = inspect.Config.Env
> > +	}
> > +	if inspect.ContainerConfig != nil {
> > +		image.ContainerLabels = inspect.ContainerConfig.Labels
> > +		image.ContainerEnv = inspect.ContainerConfig.Env
>
> yup, sounds good. I wasn't suggesting we add everything, just that for the
> fields we are using, we mirror the upstream structure so if we need to pull
> in more fields later, it is natural.
>
Cool deal.

gabemontero:
@bparees bump / what is my position on the @bparees queue ?

bparees:
@gabemontero sorry, has been on my list for the last day or two.  looking now.

gabemontero:
On Wed, Aug 3, 2016 at 2:16 PM, Ben Parees <notifications@github.com> wrote:

> @gabemontero <https://github.com/gabemontero> sorry, has been on my list
> for the last day or two. looking now.
>
no worries @bparees - thanks!

pkg/docker/docker.go
```diff
@@ -439,7 +538,7 @@ func (d *stiDocker) CheckAndPullImage(name string) (*docker.Image, error) {
 	}
 
 	image, err := d.CheckImage(name)
-	if err != nil && err.(errors.Error).Details != docker.ErrNoSuchImage {
+	if err != nil && dockerapi.IsErrImageNotFound(err) {
```
bparees: should be a ! here?

(and worrisome that means we don't have tests to cover this, so maybe we need some more unit tests for methods like this)

pkg/docker/docker.go
```diff
@@ -563,7 +684,10 @@ func (d *stiDocker) GetAssembleInputFiles(image string) (string, error) {
 }
 
 // getScriptsURL finds a scripts url label in the image metadata
-func getScriptsURL(image *docker.Image) string {
+func getScriptsURL(image *api.Image) string {
+	if image == nil {
+		return ""
+	}
```
bparees: why do we need this check here but not in other places where we take an image pointer?

pkg/docker/docker.go
```diff
@@ -665,39 +789,90 @@ func determineCommandBaseDir(opts RunContainerOptions, imageMetadata *docker.Ima
...
-		// Ports is of the following type:  map[docker.Port][]docker.PortBinding
-		for port, bindings := range cont.NetworkSettings.Ports {
-			liveports = liveports + "\n  Container Port:  " + port.Port()
-			liveports = liveports + "\n        Protocol:  " + port.Proto()
-			liveports = liveports + "\n        Public Host / Port Mappings:"
-			for _, binding := range bindings {
-				liveports = liveports + "\n            IP: " + binding.HostIP + " Port: " + binding.HostPort
-			}
+	for port, bindings := range containerJSON.NetworkSettings.NetworkSettingsBase.Ports {
+		liveports = liveports + "\n  Container Port:  " + string(port)
```
bparees: you've dropped the public host/port mappings header?

pkg/docker/docker.go
```diff
@@ -665,39 +789,90 @@ func determineCommandBaseDir(opts RunContainerOptions, imageMetadata *docker.Ima
...
+			liveports = liveports + "\n            IP: " + binding.HostIP + " Port: " + binding.HostPort
 		}
-		liveports = liveports + "\n"
 	}
+	liveports = liveports + "\n"
 	glog.V(0).Infof("\n\n\n\n\nThe image %s has been started in container %s as a result of the --run=true option.  The container's stdout/stderr will be redirected to this command's glog output to help you validate its behavior.  You can also inspect the container with docker commands if you like.  If the container is set up to stay running, you will have to Ctrl-C to exit this command, which should also stop the container %s.  This particular invocation attempts to run with the port mappings %+v \n\n\n\n\n", image, container.ID, container.ID, liveports)
 }
 
+// b copy from kube_docker_client.go ... essentially, if that code could give is a way to inspect the error from engin-api's ContainerAttach call prior
+// to blocking on the IO, we could just leverage the kube_docker_client.go code entirely ... essentially a "non-blocking" attach like go-dockerclient provided
```
bparees: i don't think this comment is supposed to be here.

bparees: or is // b supposed to mean // begin ?  (figured that out based on the // e at the end, i think it could be made more obvious)

pkg/docker/docker.go
```diff
@@ -768,19 +946,52 @@ func (d *stiDocker) RunContainer(opts RunContainerOptions) error {
...
+			defer cancel()
+			resp, err := d.client.ContainerAttach(ctx, container.ID, opts.asDockerAttachToContainerOptions())
+			errorChannel <- err
+			if err != nil {
+				glog.V(0).Infof("error: Unable to attach to container %q: %v", containerName, err)
+				return
+			}
+			defer resp.Close()
+			sopts := opts.asDockerAttachToStreamOptions()
+			attachLoggingError = d.holdHijackedConnection(sopts.RawTerminal, sopts.InputStream, sopts.OutputStream, sopts.ErrorStream, resp)
```
bparees: so this can happen after the container has started.  is that a problem in terms of missing any log output?

pkg/docker/docker.go
```diff
@@ -800,15 +1011,18 @@ func (d *stiDocker) RunContainer(opts RunContainerOptions) error {
 			// the container exits normally).  dump port/etc information for the user.
 			dumpContainerInfo(container, d, image)
 		}
+
 		// Return an error if the exit code of the container is
 		// non-zero.
 		glog.V(4).Infof("Waiting for container %q to stop ...", containerName)
-		exitCode, err := d.client.WaitContainer(container.ID)
+		ctx, cancel := getDefaultContext((1<<63 - 1) * time.Nanosecond) // infinite duration ... go does not expost max duration constant
```
bparees: s/expost/expose/

bparees: not that it likely matters, but why nanoseconds?

bparees: also go does expose max int constants, you should probably use that instead of 1<<63 -1, though it is cute :)

pkg/docker/docker.go
```diff
@@ -864,54 +1073,60 @@ func (d *stiDocker) GetImageID(name string) (string, error) {
...
 }
 
 // BuildImage builds the image according to specified options
 func (d *stiDocker) BuildImage(opts BuildImageOptions) error {
-	dockerOpts := docker.BuildImageOptions{
-		Name:                opts.Name,
-		NoCache:             true,
-		SuppressOutput:      false,
-		RmTmpContainer:      true,
-		ForceRmTmpContainer: true,
```
bparees: since we've lost these two remove tmp container fields, what is the behavior?

pkg/docker/docker.go
```diff
@@ -864,54 +1073,60 @@ func (d *stiDocker) GetImageID(name string) (string, error) {
...
+	ctx, cancel := getDefaultContext(((1<<63 - 1) * time.Nanosecond)) // infinite duration ... go does not expost max duration constant
+	defer cancel()
+	resp, err := d.client.ImageBuild(ctx, opts.Stdin, dockerOpts)
+	if err != nil {
+		return err
+	}
+	defer resp.Body.Close()
+	// since can't pass in output stream to engine-api, need to copy contents of
+	// the output stream they create into our output stream
+	_, err = io.Copy(opts.Stdout, resp.Body)
```
bparees: does this mean instead of streaming output, we're going to dump it all in a single chunk?

bparees: similar question for other places where you are copying io streams

pkg/docker/docker_test.go
```diff
@@ -24,201 +30,336 @@ func TestContainerName(t *testing.T) {
...
 	return &stiDocker{
-		client:   client,
-		pullAuth: docker.AuthConfiguration{},
+		kubeDockerClient: k8sDocker,
+		client:           client,
+		pullAuth:         dockertypes.AuthConfig{},
 	}
 }
 
+//NOTE, neither k8s kube client nor engine api make their error types public, so we can access / instantiate them like we did for the analogous
```
bparees: s/can/can't/?

pkg/docker/docker_test.go
```diff
@@ -24,201 +30,336 @@ func TestContainerName(t *testing.T) {
...
+// ones in go-dockerclient.
+// Hence, we test with errors created in these tests and set on the fake client  vs. those actually defined in k8s/engine-api.
+// Also, the fake client does save the parameters passed in; but instead, you can confirm that it was called via AssertCalls
```
bparees: s/does/doesn't/?

bparees: and the fake docker client is under our control, so why don't we update it to save the parameters?

pkg/docker/docker_test.go
```diff
@@ -24,201 +30,336 @@ func TestContainerName(t *testing.T) {
...
-			t.Errorf("Docker inspect called with unexpected image name: %s",
-				def.docker.InspectImageName)
+		if err != nil && len(def.expectedError) > 0 && !strings.Contains(err.Error(), def.expectedError) {
+			t.Errorf("Test - %s: Expected error: Got: %+v", test, err)
 		}
 	}
 }
-
 func TestCheckAndPullImage(t *testing.T) {
+	// in addition to the deltas mentioned up top, kube fake client can't distinguish between exists and successful pull, and
```
bparees: it could by saving that information in its state, right?

pkg/docker/docker_test.go
```diff
@@ -24,201 +30,336 @@ func TestContainerName(t *testing.T) {
...
+		expectedError: "unable to get test_image:latest",
 	}
 	tests := map[string]testDef{
-		"ImageExists":    imageExistsTest,
-		"ImagePulled":    imagePulledTest,
-		"InspectError":   inspectErrorTest,
-		"PullError":      pullErrorTest,
-		"ErrorAfterPull": errorAfterPullTest,
+		"ImageExists":  imageExistsTest,
+		"InspectError": inspectErrorTest,
```
bparees: i think we need to improve the fake docker client stub so we can retain these test cases.

pkg/docker/docker_test.go
```diff
@@ -24,201 +30,336 @@ func TestContainerName(t *testing.T) {
...
 
 func TestRemoveContainer(t *testing.T) {
 	fakeDocker := &test.FakeDockerClient{}
 	dh := getDocker(fakeDocker)
+	fake := dh.kubeDockerClient.(*dockertools.FakeDockerClient)
 	containerID := "testContainerId"
-	expectedOpts := docker.RemoveContainerOptions{
-		ID:            containerID,
-		RemoveVolumes: true,
-		Force:         true,
```
bparees: so the kube-client's RemoveContainer isn't setting/passing these down to the docker client?

pkg/docker/util.go
```diff
@@ -54,27 +64,88 @@ func GetImageRegistryAuth(auths *client.AuthConfigurations, imageName string) cl
...
+func LoadImageRegistryAuth(dockerCfg io.Reader) *AuthConfigurations {
+	auths, err := NewAuthConfigurations(dockerCfg)
 	if err != nil {
 		glog.V(0).Infof("error: Unable to load docker config")
 		return nil
 	}
 	return auths
 }
 
+// begin borrowed from go-dockerclient
```
bparees: i take it we got no response on whether there was something in k8s we could use for this

bparees:
@gabemontero ok, round 2 complete.  Beyond the comments i've added, at this point my main concerns are:

1) it feels like we've lost some test coverage and i'd like you to either reassure me that's not the case, or tell me why we can't retain the coverage we had before
2) i may need you to walk me through some of the attach + stream handling logic in person

pkg/docker/docker.go
gabemontero: Yeah I mentioned this briefly in scrum and I thought in some of the comments near the top of docker_test.go - the engine-api error types themselves are not public, only these methods which allow you to discern the type of error.  So unlike before with unit tests with go-dockerclient, we can't instantiate the specific error that will pass or fail their helper method in the fashion we would like.  Is there another alternative I'm missing.

Aside from the test question - code will be updated in next push.

gabemontero: Here is the def of the image not found error in engine-api:

```
type imageNotFoundError struct {
	imageID string
}
```

pkg/docker/docker.go
gabemontero: in the other spots, either the getScriptsURL call preceded it, and essentially shield those methods, or the image struct was constructed before hand, and we just passed the pointer to method vs. the struct itself (save space on the stack, etc.)

that said, I can make the nil check for that as an input parameter consistent, but in part I explicitly didn't do that pervasively because I've been dinged for unnecessary nil checks in prior code reviews (like when the caller insures a param is not nil).

Just let me know which way you want me to go - thx.

pkg/docker/docker.go
gabemontero: sure, I'll make it more obvious

pkg/docker/docker.go
gabemontero: simply an oversight after iterating over this too many times ... will add back.

pkg/docker/docker.go
gabemontero: no more so than with go-dockerclient 's non-blocking version of attach; the processing of stdout/stdin occurs in go routines there as well, in a similar fashion, after the "attach http post" has returned.

pkg/docker/docker.go
gabemontero: On the nanosecond thing, I based it off this snippet from golang's time.go:

```
// A Duration represents the elapsed time between two instants
// as an int64 nanosecond count.  The representation limits the
// largest representable duration to approximately 290 years.
type Duration int64

const (
	minDuration Duration = -1 << 63
	maxDuration Duration = 1<<63 - 1
)

But from your prompting, I did find `math.MaxInt64` ... changing the 1<<63-1 to that :-)
```

pkg/docker/docker.go
bparees: can we not construct a test case in which the kube api is going to return an ErrImageNotFoundErr?

gabemontero: duh ... a dummy image name that does not exist ... yep, i'll work on that

gabemontero: wait ... i'm getting mixed up ... unless we connect to a real docker daemon, no, I don't see how ... I'll swing by and discuss.

gabemontero: ok ... @bparees and I talked, on same page ... a k8s pull to either make the k8s version image not found error type public, or allow for a way to inject this private error not found into fake client, is desired.

pkg/docker/docker.go
gabemontero: actually I found them when trying to answer your question ...the field name did not include `TmpContainer` and that threw me off.

Adding as part of next push.

pkg/docker/docker.go
gabemontero: comparing go-dockerclient and engine-api, it is essentially the same code, i.e the call to to a copy on the same thread of execution....however, go-dockerclient, when not raw terminal, has demux version of copy which buffers the processing of the http resp body into stdout/stderr.  Perhaps that gives the illusion of progressive updating of the build logs.

I'll have to run some experiments, comparing to two forms, to confirm the similarity, or uncover the difference.

pkg/docker/docker.go
bparees: ok

pkg/docker/docker.go
bparees: ok i think it's fine as is.

pkg/docker/docker.go
gabemontero: I discovered after running the test suites (and maybe remember) why a ! had been put there (though the code I had there is still not correct) .... best as I can tell, there is a bug with engine-api or docker.  The inspect code (which is the only way we can get the image metadata ... pull does not provide it) will look for a `http.StatusNotFound` on the http get for inspect, and create the engine-api image not found error, however, when we access an image not available locally, we get this for the error message are result from `dockerapi.IsErrImageNotFound`:
```
no such image: "openshift/php-55-centos7:latest" and result from engine-api error check false
```
Looking at the message for the engine-api image not found error, the first n is captialized ... i.e. `No such image`.  So perhaps that is another clue that it is not the error type.

Bottom line, short term, I have to punt on `dockerapi.IsErrImageNotFound` and check the message string instead (yuck).  There is not an error message / type / method in engine-api that has `no such image` (lower case `n`) for the error.

gabemontero: as such, I could add a test case that has `no such image` in the message and check for that ;-)

pkg/docker/docker_test.go
gabemontero: yes, part of next push

pkg/docker/docker_test.go
gabemontero: per our discussion at your cube, I am (re)using the fake docker client from k8s for the operations it supports.  We control the engine-api fake client for the operations like attach or build which k8s does not expose.

pkg/docker/util.go
gabemontero: correct, we did not ... neither k8s or engine-api

pkg/docker/docker_test.go
gabemontero: There is an image field that is publicly accessible that we can manipulate to discern whether the image exists locally:  `Image                *dockertypes.ImageInspect`

And for pull, you can either inject an error, or add a image name to a list of pulled images which is *NOT* publicly accessible, either by itself or from a public method:  `pulled               []string`

pkg/docker/docker_test.go
gabemontero: i'll respond to your summary comment about the fake docker stub about the options to consider...

pkg/docker/docker_test.go
gabemontero: this is another case of the abstraction line moving wrt the docker client ... we set those in our RemoveContainer:
```
func (d *stiDocker) RemoveContainer(id string) error {
	opts := dockertypes.ContainerRemoveOptions{
		RemoveVolumes: true,
		Force:         true,
	}
	return d.kubeDockerClient.RemoveContainer(id, opts)

}
```

gabemontero:
@bparees OK, I've pushed the straight forward updates.  To the overriding concerns:
1) unit test coverage loss:  in my mind this boils down to reusing the "fake k8s docker abstraction" or building our own.  

In theory, some of the tests we lost could/should be covered by the k8s unit tests.  Based on my scan, this is not the case.  Do we want to be in the business of contributing to k8s in this space (either our team or the plat mgmt team I suppose)?  And if we do, how much of this feature do we want to gate on updating k8s?  

Hindsight being 20-20 here ... I'm leaning toward punting on the k8s fake client and adding a "fake k8s abstraction client" to the "fake engine-api client" we currently have.  Maybe a day of additional work, unforeseen hiccups notwithstanding.

Thoughts?

Other notes:
- since we decided not to merge this PR until after QE ran their regression, I'm of the mind to have our QE get started once we are happy with the non-test go files using the `s2i` binary I would provide them, independent of what we do in the unit test space (like if we punt on the current k8s fake client).

Thoughts?

- as to coverage elsewhere on the removed tests (lower level pull and inspect permutations), between the integration/stirun tests, everyday use, and QE's efforts for this card, coverage my guess would be fine, but from a quick, automated perspective, verifying those paths via unit tests can't be beat.

2) on the attach path, I believe there was some back and forth in the comments, but certainly we can further discuss, cross reference code together, etc. however you like.

thanks

pkg/docker/docker.go
bparees: should you be using the kube api's IsErrImageNotFound, not the docker api one?  is that why you're seeing a difference in the error content?

pkg/docker/docker_test.go
bparees: ok so doesn't that solve the problem of how to force an ErrImageNotFound error?

pkg/docker/docker_test.go
bparees: yeah so when i made this comment i assumed we were using *our* fake docker client stub, not something that was part of k8s.

bparees:
> Do we want to be in the business of contributing to k8s in this space (either our team or the plat mgmt team I suppose)? 

since we're beholden to the api now (and will be moreso in origin), it's worth spending a few hours if we can get some upstream PRs that make our lives better.

> And if we do, how much of this feature do we want to gate on updating k8s? 

none.

> Hindsight being 20-20 here ... I'm leaning toward punting on the k8s fake client and adding a "fake k8s abstraction client" to the "fake engine-api client" we currently have. Maybe a day of additional work, unforeseen hiccups notwithstanding.

yeah i'm torn on that.  on the plus side, more control and we'll ensure we're passing what we want to, to the k8s api.  on the downside, it's not going to help us confirm we're actually using the k8s api "correctly".  so overall i think i'd prefer not to go down that path, but i'm definitely open to us discussing it more since you've got the expertise at this point.

> since we decided not to merge this PR until after QE ran their regression, I'm of the mind to have our QE get started once we are happy with the non-test go files using the s2i binary I would provide them, independent of what we do in the unit test space (like if we punt on the current k8s fake client).

i guess i don't see this as urgent enough to justify jumping through that hoop. (and i think we're pretty close anyway).

> as to coverage elsewhere on the removed tests (lower level pull and inspect permutations), between the integration/stirun tests, everyday use, and QE's efforts for this card, coverage my guess would be fine

fair enough.

pkg/docker/docker.go
gabemontero: On Thu, Aug 4, 2016 at 5:18 PM, Ben Parees <notifications@github.com> wrote:

> In pkg/docker/docker.go
> <https://github.com/openshift/source-to-image/pull/554#discussion_r73604610>
> :
>
> > @@ -439,7 +538,7 @@ func (d *stiDocker) CheckAndPullImage(name string) (*docker.Image, error) {
> >  	}
> >
> >  	image, err := d.CheckImage(name)
> > -	if err != nil && err.(errors.Error).Details != docker.ErrNoSuchImage {
> > +	if err != nil && dockerapi.IsErrImageNotFound(err) {
>
> should you be using the kube api's IsErrImageNotFound, not the docker api
> one? is that why you're seeing a difference in the error content?
>
this struct is defined in kube_docker_client.go:

```
// imageNotFoundError is the error returned by InspectImage when image not
found.
type imageNotFoundError struct {
    ID string
}
```
but there is no public method that interrogates it like what exists in
dockerapi (unless go-oracle and grep are deceiving me)

pkg/docker/docker_test.go
gabemontero: All the wrappers of errors at the different layers and private error types
is making it difficult, but I've
only spent a little bit of time... I'll circle back to this, spend some
more time, and report back.

pkg/docker/docker.go
bparees: so you're calling d.CheckImage which in turn is calling kube.InspectImage and kube.InspectImage is eating the dockerapi imageNotFoundError and instead returning a kube.imageNotFoundError.

it's the kube.imageNotFoundError that has "no such image" (lowercase) as its message output.

so as expected, you're getting a kube.imageNotFoundError.

but you're right, i thought i saw a method for checking the error type in kube, but i was wrong, it was something else, i don't see any way to confirm the error type other than, as you said, checking the message text, which is bad.

gabemontero: On Thu, Aug 4, 2016 at 5:41 PM, Ben Parees <notifications@github.com> wrote:

> In pkg/docker/docker.go
> <https://github.com/openshift/source-to-image/pull/554#discussion_r73608220>
> :
>
> > @@ -439,7 +538,7 @@ func (d *stiDocker) CheckAndPullImage(name string) (*docker.Image, error) {
> >  	}
> >
> >  	image, err := d.CheckImage(name)
> > -	if err != nil && err.(errors.Error).Details != docker.ErrNoSuchImage {
> > +	if err != nil && dockerapi.IsErrImageNotFound(err) {
>
> so you're calling d.CheckImage which in turn is calling kube.InspectImage
> and kube.InspectImage is eating the dockerapi imageNotFoundError and
> instead returning a kube.imageNotFoundError.
>
> it's the kube.imageNotFoundError that has "no such image" (lowercase) as
> its message output.
>
Yep, my earlier scan for "no such image" did not include k8s.

> so as expected, you're getting a kube.imageNotFoundError.
>
> but you're right, i thought i saw a method for checking the error type in
> kube, but i was wrong, it was something else, i don't see any way to
> confirm the error type other than, as you said, checking the message text,
> which is bad.
>
agreed (bad) but no choice under current circumstances

pkg/docker/docker_test.go
gabemontero: On Thu, Aug 4, 2016 at 5:34 PM, Gabe Montero <gmontero@redhat.com> wrote:

> All the wrappers of errors at the different layers and private error types
> is making it difficult, but I've
> only spent a little bit of time... I'll circle back to this, spend some
> more time, and report back.
>

OK, so I've got a *form* of err image not found test ... i create an error
with "no such image" in the text and inject it into the fake client, and
then
confirm that the ultimate CheckAndPullImage call does not return an error.

gabemontero:
OK, I've pushed test updates for the (hacked) image not found test I described inline with the review comments, as well as a pull error test I realized we could do.  The confirm the image pulled test is still not possible because the pulled flag in the fake docker client is not public nor accessible from a public method.  The error after pull test is not possible because you can't distinguish which inspect call should receive the first error (other than an additional inspect after the pull, there is nothing that would produce the error)

gabemontero:
@bparees - are we good to say "igtm", but hold off on merging while I provide and `s2i` binary to QE to start regression testing on?

gabemontero:
cool - thanks @bparees

gabemontero:
@dongboyan77 reported back with only 1 error from his regression set, and that error (with `s2i rebuild ..`) exists with the existing `s2i`.

We can post the merge comment, once @oatmealraisin 's in progress s2i bump to origin completes.

gabemontero:
I have asked @dongboyan77 to open a bugzilla for the `s2i rebuild ...` item.

gabemontero:
I saw the `s2i rebuild...` issue on a clean ec2 instance where `s2i version` reports `s2i v1.1.1-18-ge63e078`

oatmealraisin:
> We can post the merge comment, once @oatmealraisin 's in progress s2i bump to origin completes.

@gabemontero I've stopped progress on that, please watch [#10359](https://github.com/openshift/origin/pull/10359) for updates on the bump fyi

gabemontero:
thanks for the clarifycation @oatmealraisin

gabemontero:
hey @bparees with https://github.com/openshift/origin/pull/10359 merged can we post the merge comment on this one?

bparees:
@gabemontero there were a few more commits into s2i i want to bump before we drop the nuke as it were, should be fast: https://github.com/openshift/origin/pull/10425

bparees:
[merge]

gabemontero:
Oooh ... the pkg/docker unit test displayed something I've never seen before:
```
==================
WARNING: DATA RACE
Read by goroutine 27:
  github.com/openshift/source-to-image/pkg/docker.(*stiDocker).RunContainer.func4()
      github.com/openshift/source-to-image/pkg/docker/_test/_obj_test/docker.go:1254 +0x1403
  github.com/openshift/source-to-image/pkg/util/interrupt.(*Handler).Run()
      /data/src/github.com/openshift/source-to-image/_output/local/go/src/github.com/openshift/source-to-image/pkg/util/interrupt/interrupt.go:78 +0x148
  github.com/openshift/source-to-image/pkg/docker.(*stiDocker).RunContainer()
      github.com/openshift/source-to-image/pkg/docker/_test/_obj_test/docker.go:1255 +0x1357
  github.com/openshift/source-to-image/pkg/docker.TestRunContainer()
      /data/src/github.com/openshift/source-to-image/_output/local/go/src/github.com/openshift/source-to-image/pkg/docker/docker_test.go:643 +0x3704
  testing.tRunner()
      /usr/lib/golang/src/testing/testing.go:473 +0xdc

Previous write by goroutine 36:
  github.com/openshift/source-to-image/pkg/docker.(*stiDocker).RunContainer.func4.1()
      github.com/openshift/source-to-image/pkg/docker/_test/_obj_test/docker.go:1162 +0x6ae

Goroutine 27 (running) created at:
  testing.RunTests()
      /usr/lib/golang/src/testing/testing.go:582 +0xae2
  testing.(*M).Run()
      /usr/lib/golang/src/testing/testing.go:515 +0x11d
  main.main()
      github.com/openshift/source-to-image/pkg/docker/_test/_testmain.go:134 +0x385

Goroutine 36 (finished) created at:
  github.com/openshift/source-to-image/pkg/docker.(*stiDocker).RunContainer.func4()
      github.com/openshift/source-to-image/pkg/docker/_test/_obj_test/docker.go:1164 +0x324
  github.com/openshift/source-to-image/pkg/util/interrupt.(*Handler).Run()
      /data/src/github.com/openshift/source-to-image/_output/local/go/src/github.com/openshift/source-to-image/pkg/util/interrupt/interrupt.go:78 +0x148
  github.com/openshift/source-to-image/pkg/docker.(*stiDocker).RunContainer()
      github.com/openshift/source-to-image/pkg/docker/_test/_obj_test/docker.go:1255 +0x1357
  github.com/openshift/source-to-image/pkg/docker.TestRunContainer()
      /data/src/github.com/openshift/source-to-image/_output/local/go/src/github.com/openshift/source-to-image/pkg/docker/docker_test.go:643 +0x3704
  testing.tRunner()
      /usr/lib/golang/src/testing/testing.go:473 +0xdc
==================
```

I'll start researching golang data race, and if I can't repro locally, see if we can infer a solution from the stack trace.

gabemontero:
(yep, doesn't show up locally)

gabemontero:
unit test race fixed, but the integration test with S2I_RACE set also found a similar race in sti.go ... working on fix now for that

gabemontero:
the integration test data race was an interesting one that is unrelated to the docker client conversion ... it may have been there for a while because the integration tests don't seem to run with the env vars set to trigger `-race` on the go test.

I'm going to still correct the error ... it entails closing the error reader passed into utils.go:StreamContainerIO prior to access the errOutput strings ... the close will exit StreamContainerIO and free up the writer access.

gabemontero:
[test]

gabemontero:
OK, the code tweak to have the unit tests pass the pkg/docker package when race detection is enabled starts at https://github.com/openshift/source-to-image/pull/554/commits/7f8cd156d02d3d3aff2ab6195299fa281e61fc69#diff-bf07e8f23e3aaca43157f262317e094cR976 where the change provides protection for `attachLoggingError` variable.

An example of the code change (unrelated to the docker client conversion) which handles the data race condition found with the integration tests when race detection was enabled is at https://github.com/openshift/source-to-image/pull/554/commits/7f8cd156d02d3d3aff2ab6195299fa281e61fc69#diff-77c4df2e21c6a783c0928af25a29ba8dR442 where we have to proactively close out the error reader handled by `StreamContainerIO` so that the active write to `errOutput` is shut down.

bparees:
lgtm [merge]

PR #555, opened 2016-07-25
Fixing permission issue during rm_injections

jupierce:
Post bump, this should fix origin 10009. 
StreamFileAsTar makes an attempt to chmod the rm-injections script to 666 before injecting it with a tar stream, but StreamFileAsTarWithCallback was actually setting these permissions on the primordial source for rm-injections. 

Before this PR, rm-injections ended up as root:root 600 in the build container. Now it should be 666 and source-able by the default user.


jupierce:
@bparees

bparees:
lgtm [merge]

PR #556, opened 2016-07-28
Fix path to assemble-runtime script when it's inside of the runtime image

php-coder:
PTAL @bparees @PI-Victor

### DIFF
--- a/pkg/build/strategies/sti/postexecutorstep.go
+++ b/pkg/build/strategies/sti/postexecutorstep.go
@@ -251,7 +251,7 @@ func (step *startRuntimeImageAndUploadFilesStep) execute(ctx *postExecutorStepCo
 		return fmt.Errorf("Couldn't get working dir of %q image: %v", image, err)
 	}
 
-	commandBaseDir := workDir
+	commandBaseDir := filepath.Join(workDir, "scripts")
 	useExternalAssembleScript := step.builder.externalScripts[api.AssembleRuntime]
 	if !useExternalAssembleScript {
 		// script already inside of the image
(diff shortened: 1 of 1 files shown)


php-coder:
Not fully tested yet.

pkg/build/strategies/sti/postexecutorstep.go
```diff
@@ -251,7 +251,7 @@ func (step *startRuntimeImageAndUploadFilesStep) execute(ctx *postExecutorStepCo
 		return fmt.Errorf("Couldn't get working dir of %q image: %v", image, err)
 	}
 
-	commandBaseDir := workDir
+	commandBaseDir := filepath.Join(workDir, "scripts")
 	useExternalAssembleScript := step.builder.externalScripts[api.AssembleRuntime]
 	if !useExternalAssembleScript {
```
bparees: i don't understand why this if-check is not being triggered for the scenario you described and setting commandBaseDir correctly so we run the script that is already part of the image?

pkg/build/strategies/sti/postexecutorstep.go
```diff
@@ -266,7 +266,7 @@ func (step *startRuntimeImageAndUploadFilesStep) execute(ctx *postExecutorStepCo
 	}
 
 	cmd := fmt.Sprintf(
-		"while [ ! -f %q ]; do sleep 0.5; done; %s/scripts/%s; exit $?",
+		"while [ ! -f %q ]; do sleep 0.5; done; %s/%s; exit $?",
```
bparees: nm, i understand now.

bparees:
lgtm but let me know when you've tested it for both images w/ scripts, and repos that override the scripts.

php-coder:
@bparees Fix is correct. Could you merge it, please?

bparees:
[merge]

PR #557, opened 2016-08-05
log stderr as info messages

bparees:
@php-coder ptal, the current output looks horrible because npm (among other things) uses stderr for its output, so you end up with `ERROR: some normal npm operation` in your normal s2i output.

this seems like the lesser of two evils to me right now.


bparees:
(and yeah i know we discussed this at length, though i don't remember all the suggested alternatives or implications)

php-coder:
Should we do the same change for a container with `assemble-runtime` script? Looks like we should.

bparees:
@php-coder done

php-coder:
LGTM

bparees:
[merge]

Issue #559, opened 2016-08-07
s2i 1.1+ (?) not working with Docker images built with Docker 1.11.


GrahamDumpleton:
Looks like your calculations to work out whether need to override an Entrypoint in S2I builder image, and then run it, mucks up somehow with builder images built with recent Docker version.

Loglevel 8 output right at end.

```
$ docker pull grahamdumpleton/warp0-debian8-python27
Using default tag: latest
latest: Pulling from grahamdumpleton/warp0-debian8-python27
357ea8c3d80b: Already exists
52befadefd24: Already exists
3c0732d5313c: Already exists
ceb711c7e301: Already exists
45c62b477be4: Already exists
4dd63f0a83a0: Already exists
c1128f7427c5: Already exists
c58a76176b96: Already exists
314e244625ac: Pull complete
b061daf4245b: Pull complete
Digest: sha256:ef887938c0e61b0996e0f3afde827517a8fb77f1661218ec6e44b0719ed7bd76
Status: Downloaded newer image for grahamdumpleton/warp0-debian8-python27:latest

$ docker inspect grahamdumpleton/warp0-debian8-python27 | grep DockerVersion
        "DockerVersion": "1.11.2",

$ s2i version
s2i v1.1.1

$ s2i build https://gitlab.com/osevg/python-flask-modwsgi.git grahamdumpleton/warp0-debian8-python27 flask

ERROR: exec: "/bin/env": stat /bin/env: no such file or directory

ERROR: An error occurred: API error (500): Container command not found or does not exist.

$ ~/OpenShift/s2i-1.0.6/s2i build https://gitlab.com/osevg/python-flask-modwsgi.git grahamdumpleton/warp0-debian8-python27 flask
I0807 11:58:52.524189 06038 clone.go:32] Downloading "https://gitlab.com/osevg/python-flask-modwsgi.git" ...
I0807 11:58:54.933371 06038 install.go:251] Using "assemble" installed from "image:///opt/app-root/s2i/bin/assemble"
I0807 11:58:54.933425 06038 install.go:251] Using "run" installed from "image:///opt/app-root/s2i/bin/run"
I0807 11:58:54.933446 06038 install.go:251] Using "save-artifacts" installed from "image:///opt/app-root/s2i/bin/save-artifacts"
---> Installing application source
---> Building application from source
-----> Installing dependencies with pip (requirements.txt)
[...]

PR #560, opened 2016-08-07
Change override entrypoint to /usr/bin/env

bparees:
@grahamdumpleton ptal. 

fixes https://github.com/openshift/source-to-image/issues/559
[test]

### DIFF
--- a/pkg/build/strategies/sti/sti.go
+++ b/pkg/build/strategies/sti/sti.go
@@ -27,7 +27,7 @@ import (
 
 var (
 	// DefaultEntrypoint is the default entry point used when starting containers
-	DefaultEntrypoint = []string{"/bin/env"}
+	DefaultEntrypoint = []string{"/usr/bin/env"}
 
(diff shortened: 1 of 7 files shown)


bparees:
@php-coder ptal.

php-coder:
LGTM.

BTW, maybe it would be good to mention this in the docs (for example, `docs/debugging-s2i.md`). As far I understand this case is rare, but as a user that will find it, I would like to have a bit info about this. For example, we could mention that S2I prefers images with `/usr/bin/env` and if user doesn't have such image, he could create a symlink.

bparees:
[merge]

bparees:
@php-coder it remains to be seen if this is the long term solution, there's still some discussion going on, so i'm not going to bother doc'ing it just yet.

bparees:
@php-coder (discussion is here: https://github.com/openshift/source-to-image/issues/475#issuecomment-238088669)

PR #562, opened 2016-08-08
Move doc.go from pkg/build/ignore to pkg/ignore

php-coder:
Looks like `doc.go` was accidentally placed to the wrong location from its first commit (19442d1ba101f08e5d4cb8d6c44da426130849aa). Let's move this file to others.

PTAL @gabemontero @bparees


bparees:
[merge]

PR #565, opened 2016-08-09
Moved createLocalGitDirectory to test package

oatmealraisin:
This allows multiple unit tests to use these functions, which were
previously copy/pasted between them.

@bparees @php-coder ptal?


bparees:
lgtm
[merge]

PR #563, opened 2016-08-08
hack/rebase-to-origin.sh: exclude pkg/docker/test dir and preserve LICENSE file

php-coder:
PTAL @bparees @PI-Victor @mfojtik

PR #567, opened 2016-08-10
Removed pkg/test from excluded packages in script

oatmealraisin:
This allows us to use Git testing utilities in origin.

Perhaps in the future we should merge the Git utilities that are spread between the projects?

Sorry for the quick one-liners :)

### DIFF
--- a/hack/rebase-to-origin.sh
+++ b/hack/rebase-to-origin.sh
@@ -28,7 +28,6 @@ readonly exclude_pkgs=(
   pkg/create
   pkg/docker/test
   pkg/run
-  pkg/test
   pkg/version
 )
 


bparees:
hm.  this was just added by @php-coder a few days ago to prevent us pushing this code up to origin.

it does seem like it could be useful from within origin so i'm ok w/ merging it, but will let @php-coder weigh in first in case he sees a good reason not to.

oatmealraisin:
It's only for the CreateLocalGitRepository() function, if that changes anything

php-coder:
>hm. this was just added by @php-coder a few days ago to prevent us pushing this code up to origin.

In the PR https://github.com/openshift/source-to-image/pull/563 I excluded `pkg/docker/test` directory. `pkg/test` was in the script since the beginning. Let's merge it and see what happens then :)

bparees:
[merge]

PR #568, opened 2016-08-11
Fixing line number accuracy with glog wrapper

### LINKED ISSUE #558: reported log lines are wrong
run:
s2i build https://github.com/openshift/nodejs-ex centos/nodejs-4-centos7  --loglevel=5 

you will see in the output:
```
E0804 23:28:46.056805   28401 postexecutorstep.go:328] npm info it worked if it 
ends with ok
```

but the output is actually being logged by:
https://github.com/openshift/source-to-image/blob/master/pkg/build/strategies/sti/sti.go#L587

(the postexecutorstep isn't being invoked at all because this isn't an extended build).

bafflingly, if you comment out the postexecutor step line, you'll see the log attributed to yet another incorrect location.  If you then comment out *that* location, the reporting is correct.

Also if you change the postexecutor code to instead log to info instead of err, the reported location changes.

So something very weird is going on w/ the stack frame attribution, possibly due to the presence of go routines, but it makes no sense to me that it's getting attributed to code that's not even being invoked in these flows (i added print statements to confirm that).

jupierce:
Fixes #558 
s/Fixes/Improves

### DIFF
--- a/pkg/build/strategies/sti/postexecutorstep.go
+++ b/pkg/build/strategies/sti/postexecutorstep.go
@@ -322,10 +322,10 @@ func (step *startRuntimeImageAndUploadFilesStep) execute(ctx *postExecutorStepCo
 		return err
 	}
 
-	go dockerpkg.StreamContainerIO(outReader, nil, glog.V(0).Info)
+	go dockerpkg.StreamContainerIO(outReader, nil, func(a ...interface{}) { glog.V(0).Info(a...) })
 
 	errOutput := ""
-	go dockerpkg.StreamContainerIO(errReader, &errOutput, glog.Info)
+	go dockerpkg.StreamContainerIO(errReader, &errOutput, func(a ...interface{}) { glog.Info(a...) })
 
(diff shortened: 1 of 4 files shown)


jupierce:
File/line glog header will now reference the a non-random file.
```
I0811 10:09:51.646242    6534 sti.go:586] ---> Installing application source ...
I0811 10:09:51.647425    6534 sti.go:586] ---> Building your Node application from source
I0811 10:09:51.803139    6534 sti.go:593] npm info it worked if it ends with ok
I0811 10:09:51.803534    6534 sti.go:593] npm info using npm@2.15.1
I0811 10:09:51.803546    6534 sti.go:593] npm info using node@v4.4.2
```
This change should be compatible with glog or the glog wrapper.

bparees:
@jupierce you rock.
[merge]

bparees:
but it looks like you need to do some gofmt'ing :)

jupierce:
@bparees fmt'd :)

bparees:
[merge]

PR #570, opened 2016-08-12
s2i: s2i rebuild broken

gabemontero:
Bug 1366475
Bugzilla link https://bugzilla.redhat.com/show_bug.cgi?id=1366475
Detailed explanation
an attempt to pull the builder image to obtain labels was made; the correct thing was
to pull the locally build image we were rebuilding

@bparees @php-coder PTAL

### DIFF
--- a/pkg/api/describe/describer.go
+++ b/pkg/api/describe/describer.go
@@ -10,6 +10,7 @@ import (
 
 	"github.com/openshift/source-to-image/pkg/api"
 	"github.com/openshift/source-to-image/pkg/build"
+	"github.com/openshift/source-to-image/pkg/docker"
 )
 
 // Config returns the Config object in nice readable, tabbed format.
@@ -97,16 +98,21 @@ func describeBuilderImage(config *api.Config, image string, out io.Writer) {
 		Tag:                config.Tag,
 		IncrementalAuthentication: config.IncrementalAuthentication,
 	}
-	build.GenerateConfigFromLabels(c)
-	if len(c.DisplayName) > 0 {
-		fmt.Fprintf(out, "Builder Name:\t%s\n", c.DisplayName)
-	}
-	fmt.Fprintf(out, "Builder Image:\t%s\n", config.BuilderImage)
-	if len(c.BuilderImageVersion) > 0 {
-		fmt.Fprintf(out, "Builder Image Version:\t%s\n", c.BuilderImageVersion)
-	}
-	if len(c.BuilderBaseImageVersion) > 0 {
-		fmt.Fprintf(out, "Builder Base Version:\t%s\n", c.BuilderBaseImageVersion)
+	pr, err := docker.GetBuilderImage(c)
+	if err == nil {
+		build.GenerateConfigFromLabels(c, pr)
+		if len(c.DisplayName) > 0 {
+			fmt.Fprintf(out, "Builder Name:\t%s\n", c.DisplayName)
+		}
+		fmt.Fprintf(out, "Builder Image:\t%s\n", config.BuilderImage)
+		if len(c.BuilderImageVersion) > 0 {
+			fmt.Fprintf(out, "Builder Image Version:\t%s\n", c.BuilderImageVersion)
+		}
+		if len(c.BuilderBaseImageVersion) > 0 {
+			fmt.Fprintf(out, "Builder Base Version:\t%s\n", c.BuilderBaseImageVersion)
+		}
+	} else {
+		fmt.Fprintf(out, "Error describing image:\t%s\n", err.Error())
 	}
 }
 
--- a/pkg/build/config.go
+++ b/pkg/build/config.go
@@ -9,12 +9,15 @@ import (
 
 // GenerateConfigFromLabels generates the S2I Config struct from the Docker
 // image labels.
-func GenerateConfigFromLabels(config *api.Config) error {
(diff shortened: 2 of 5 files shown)


pkg/docker/util.go
```diff
@@ -286,6 +286,28 @@ func GetBuilderImage(config *api.Config) (*PullResult, error) {
...
+// GetRebuildImage obtains the metadata information for the image
+// specified in a s2i rebuild operation.  Assumptions are made that
+// the build is available locally since it should have been previously built.
+func GetRebuildImage(config *api.Config) (*PullResult, error) {
+	d, err := New(config.DockerConfig, config.PullAuthentication)
+	if err != nil {
+		return nil, err
+	}
+
+	r, err := PullImage(config.Tag, d, api.PullNever, false)
```
bparees: why not pull if not present?  they could have pushed it somewhere.

gabemontero: i was making the assumption that it was always local  ... if we can't make that assumption, do we repurpose the builder pull policy or previous image pull policy, or do we have to add another field in the config for rebuild pull policy ?

gabemontero: and until the deprecated field is remove, i can leverage the ForcePull field as well

pkg/docker/util.go
```diff
@@ -286,6 +286,28 @@ func GetBuilderImage(config *api.Config) (*PullResult, error) {
 	return r, nil
 }
 
+// GetRebuildImage obtains the metadata information for the image
+// specified in a s2i rebuild operation.  Assumptions are made that
+// the build is available locally since it should have been previously built.
+func GetRebuildImage(config *api.Config) (*PullResult, error) {
```
php-coder: Now we have 3 methods that looks very similar, at least, all of them pulling and checking image at the end. I suggest to extract this logic to `pullAndCheck(image string, docker Docker, pullPolicy PullPolicy, forcePull, allowedUIDs boolean) error` method and modify existing methods to invoke it.

pkg/api/describe/describer.go
```diff
@@ -97,7 +97,7 @@ func describeBuilderImage(config *api.Config, image string, out io.Writer) {
 		Tag:                config.Tag,
 		IncrementalAuthentication: config.IncrementalAuthentication,
 	}
-	build.GenerateConfigFromLabels(c)
+	build.GenerateConfigFromLabels(c, false)
```
php-coder: Note that here we're ignoring return value that is a possible error.

pkg/build/config.go
```diff
@@ -9,8 +9,14 @@ import (
 
 // GenerateConfigFromLabels generates the S2I Config struct from the Docker
 // image labels.
-func GenerateConfigFromLabels(config *api.Config) error {
-	result, err := docker.GetBuilderImage(config)
+func GenerateConfigFromLabels(config *api.Config, rebuild bool) error {
```
php-coder: The fact that `GenerateConfigFromLabels` method also pulling the image looks strange and unnatural. Because we're invoking this method only from twice, I suggest to pass `PullResult` to this method as an argument and pull the image from the method that invokes `GenerateConfigFromLabels`. In this case, we can also remove the `rebuild` argument.

php-coder:
@gabemontero Looks like it was broken long time ago or it's not?

gabemontero:
@php-coder yep, as I mentioned in the bugzilla, no recent changes jumped out as something that would have broke `s2i rebuild`; thus, I suspect it has been broken for quite a while

pkg/api/describe/describer.go
gabemontero: yes i took this as "describe as much as you can" .... I think we should leave this as is

pkg/build/config.go
gabemontero: sure, I can make that change ... good idea wrt the separation

pkg/docker/util.go
gabemontero: part of next push

pkg/docker/util.go
gabemontero: thinking about it some more, I'll reuse the builder policy ... the tester for example specified the -p option ... seems natural

gabemontero:
responses to comments pushed

pkg/api/describe/describer.go
```diff
@@ -97,7 +98,8 @@ func describeBuilderImage(config *api.Config, image string, out io.Writer) {
 		Tag:                config.Tag,
 		IncrementalAuthentication: config.IncrementalAuthentication,
 	}
-	build.GenerateConfigFromLabels(c)
+	pr, _ := docker.GetBuilderImage(c)
```
bparees: why aren't you checking this error?

gabemontero: The error was ignored before.  Assumption I took was to describe as much as
possible.  I was inclined to not change that.

php-coder:
LGTM. Thank you!

pkg/api/describe/describer.go
bparees: but now pr could be nil, in which case GenerateConfigFromLabels is going to error anyway, seems better to report the error as early as possible, otherwise it's going to get wrapped/lost (before the error from the pull got bubbled up as is)

gabemontero: OK ... pushed update that does not ignore the err

bparees:
lgtm
[merge]

PR #571, opened 2016-08-16
Document how to invoke embedded scripts from custom scripts

php-coder:
PTAL @bparees 

https://trello.com/c/FJ7ZVLhR/1016-3-document-and-investigate-s2i-exit-hooks

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ ready-to-run images by injecting source code into a Docker container and letting
 
 For a deep dive on S2I you can view [this presentation](https://www.youtube.com/watch?v=flI6zx9wH6M).
 
-Want to try it right now?  Download the [latest release](https://github.com/openshift/source-to-image/releases/tag/v1.1.0) and run:
+Want to try it right now?  Download the [latest release](https://github.com/openshift/source-to-image/releases/latest) and run:
 
 	$ s2i build git://github.com/openshift/django-ex centos/python-35-centos7 hello-python
 	$ docker run -p 8080:8080 hello-python
@@ -39,9 +39,6 @@ For example, to create a reproducible build pipeline for Tomcat (the popular Jav
 4. Invoke source-to-image a second time using the WAR file from the previous step and the initial Tomcat image to create the runtime image
 
 By placing our build logic inside of images, and by combining the images into multiple steps, we can keep our runtime environment close to our build environment (same JDK, same Tomcat JARs) without requiring build tools to be deployed to production.
-
-Learn more about [building your own images](#getting-started) and [how to use a non-builder image for the final application image](https://github.com/openshift/source-to-image/blob/master/docs/runtime_image.md).
-
 
 ## Goals
 
@@ -71,7 +68,7 @@ image:
 Additionally for the best user experience and optimized `s2i` operation we suggest images
 to have `/bin/sh` and `tar` commands available.
 
-See a practical tutorial on how to create a builder image [here](examples/README.md) and read [this](https://github.com/openshift/source-to-image/blob/master/docs/builder_image.md) for a detailed description of the requirements and scripts along with examples of builder images.
+See [a practical tutorial on how to create a builder image](examples/README.md) and read [a detailed description of the requirements and scripts along with examples of builder images](docs/builder_image.md).
 
 ## Build workflow
 
@@ -240,7 +237,10 @@ $ s2i build git://github.com/bparees/openshift-jee-sample openshift/wildfly-100-
 $ docker run --rm -i -p :8080 -t test-jee-app
 ```
 
-Interested in more advanced `s2i` usage? See [here](https://github.com/openshift/source-to-image/blob/master/docs/cli.md)
-for detailed descriptions of the different CLI commands with examples.
+Want to know more? Read the following resources:
 
-Running into some issues and need some advice debugging?  Peruse [here](https://github.com/openshift/source-to-image/blob/master/docs/debugging-s2i.md) for some tips.
+* [Descriptions and examples of the S2I commands](docs/cli.md)
+* [Instructions for using builder images](docs/user_guide.md)
+* [Guidance for S2I builder image creators](docs/builder_image.md)
+* [Using a non-builder image for the base of the application image](docs/runtime_image.md)
+* [Troubleshooting and debugging S2I problems](docs/debugging-s2i.md)
--- a/docs/user_guide.md
+++ b/docs/user_guide.md
@@ -1,0 +1,46 @@
+# Using S2I images
+
+S2I builder images normally include `assemble` and `run` scripts, but the default behavior of those scripts may not be suitable for all users. This document will cover a few approaches for customizing the behavior of an S2I builder that includes default scripts.
+
+## Invoking scripts embedded in an image
+
+Typically builder images provide their own version of the [S2I scripts](builder_image.md#s2i-scripts) that cover the most common use-cases. If these scripts don't fulfill your needs, S2I provides a way of overriding them by adding custom ones in the `.s2i/bin` directory. However, by doing this you are completely replacing the standard scripts. In some cases this is acceptable but in other scenarios you may prefer to execute a few commands before (or after) the scripts while retaining the logic of the script provided in the image. In this case it is possible to create a wrapper script that will execute custom logic and delegate further work to the default script in the image.
+
+To determine the location of the scripts inside of the builder image, we need to look at at the value of `io.openshift.s2i.scripts-url` label. We can use `docker inspect` for that:
+
+```console
+$ docker inspect --format='{{ index .Config.Labels "io.openshift.s2i.scripts-url" }}' openshift/wildfly-100-centos7
+image:///usr/libexec/s2i
+```
+
+Here we are inspected the `openshift/wildfly-100-centos7` builder image and found out that the scripts are in the `/usr/libexec/s2i` directory.
+
+With this knowledge we can invoke any of these scripts from our own by wrapping its invocation.
+
+Example of `.s2i/bin/assemble` script:
+```bash
+#!/bin/bash
+echo "Before assembling"
+
+/usr/libexec/s2i/assemble
+rc=$?
+
+if [ $rc -eq 0 ]; then
+    echo "After successful assembling"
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -240,7 +237,10 @@ $ s2i build git://github.com/bparees/openshift-jee-sample openshift/wildfly-100-
 $ docker run --rm -i -p :8080 -t test-jee-app
 ```
 
-Interested in more advanced `s2i` usage? See [here](https://github.com/openshift/source-to-image/blob/master/docs/cli.md)
-for detailed descriptions of the different CLI commands with examples.
+Want to know more? Read the following resources:
 
-Running into some issues and need some advice debugging?  Peruse [here](https://github.com/openshift/source-to-image/blob/master/docs/debugging-s2i.md) for some tips.
+* [description of the different S2I commands with examples](docs/cli.md)
+* [tips for using builder images](docs/user_guide.md)
```
bparees: i think tips is the wrong word.  it's more like "Instructions for using builder images"  (at least that should be the goal)

README.md
```diff
@@ -240,7 +237,10 @@ $ s2i build git://github.com/bparees/openshift-jee-sample openshift/wildfly-100-
 $ docker run --rm -i -p :8080 -t test-jee-app
 ```
 
-Interested in more advanced `s2i` usage? See [here](https://github.com/openshift/source-to-image/blob/master/docs/cli.md)
-for detailed descriptions of the different CLI commands with examples.
+Want to know more? Read the following resources:
 
-Running into some issues and need some advice debugging?  Peruse [here](https://github.com/openshift/source-to-image/blob/master/docs/debugging-s2i.md) for some tips.
+* [description of the different S2I commands with examples](docs/cli.md)
```
bparees: "Descriptions and examples of the S2I commands"

README.md
```diff
@@ -240,7 +237,10 @@ $ s2i build git://github.com/bparees/openshift-jee-sample openshift/wildfly-100-
...
+* [tips for using builder images](docs/user_guide.md)
+* [S2I builder image requirements](docs/builder_image.md)
```
bparees: "Guidance for S2I builder image creators"  (there really aren't requirements for builder images, that's one of the points of s2i, that you can use any image as a builder image)

README.md
```diff
@@ -240,7 +237,10 @@ $ s2i build git://github.com/bparees/openshift-jee-sample openshift/wildfly-100-
...
+* [how to use a non-builder image for the final application image](docs/runtime_image.md)
```
bparees: "Using a non-builder image for the base of the application image"

docs/user_guide.md
```diff
@@ -0,0 +1,37 @@
+# Using S2I images
+
+## Invoking scripts embedded into an image
+
+When there is a requirement for **adding custom logic** to [S2I scripts](builder_image.md#s2i-scripts), it is possible to execute standartd scripts inside of custom scripts from user’s repository. Note that it only works when you have to add logic (for example, to generate some static assets after executing `assemble` script) and not for the case of modifying this logic. In latter case you have to replace standard scripts.
```
bparees: s/standartd/standard/

docs/user_guide.md
```diff
@@ -0,0 +1,37 @@
+# Using S2I images
```
bparees: i think this needs an intro paragraph.  who's the audience, what am i going to learn.  maybe a table of context too.  jumping right into "invoking scripts embedded into an image" isn't a gentle enough entrypoint to this topic.

docs/user_guide.md
bparees: this first sentence is pretty confusing.  
1) "when there is a requirement for adding custom logic to s2i scripts"?  I think what you mean here is basically "Most S2I builder images include assemble/run scripts in the image.  If there is a need to augment the steps being performed by those scripts, there are two options:  1) replace the script entirely by including your own scripts in your source repo or 2) include a new script in your source repo that performs the additional steps before/after invoking the built in scripts (note that for run script, only invoking before is valid).  Option 2 is only applicable when logic is being added, not changed, and when the logic can be invoked before or after the existing script logic.  To change existing logic or inject additional actions in the middle of the existing logic, the entire script must be replaced/overridden."

something like that, anyway. please don't take it verbatim, i'm typing off the top of my head here :)

docs/user_guide.md
```diff
@@ -0,0 +1,37 @@
...
+When there is a requirement for **adding custom logic** to [S2I scripts](builder_image.md#s2i-scripts), it is possible to execute standartd scripts inside of custom scripts from user’s repository. Note that it only works when you have to add logic (for example, to generate some static assets after executing `assemble` script) and not for the case of modifying this logic. In latter case you have to replace standard scripts.
+
+To determine the location of the scripts inside of the builder image, we need to look at at the value of `io.openshift.s2i.scripts-url` label. We can use `docker inspect` for that:
+
+```console
+$ docker inspect --format=' index .Config.Labels "io.openshift.s2i.scripts-url"' openshift/wildfly-100-centos7 
+image:///usr/libexec/s2i
+```
+
+Here we are inspected `openshift/wildfly-100-centos7` builder image and find out that the scripts lie in the `/usr/libexec/s2i` directory.
```
bparees: s/inspected/inspected the/

s/lie/are/

bparees: s/find/found/

docs/user_guide.md
```diff
@@ -0,0 +1,37 @@
...
+
+With this knowledge we can easily invoke any of these scripts from our own, by wrapping it's invocation.
```
bparees: s/it's/its/
s/easily//
s/own, by/own by/

docs/user_guide.md
```diff
@@ -0,0 +1,37 @@
...
+if [ $rc -eq 0 ]; then
+    echo "After successful assembling"
+else
+    echo "After failed assembling"
+fi
+
+exit $rc
+```
+
+Example shows custom `assemble` script that prints the message, executes standard `assemble` script from the image and prints another message depending on the exit code of the `assemble` script.
```
bparees: s/Example/The example/
s/shows custom/shows a custom/

docs/user_guide.md
```diff
@@ -0,0 +1,37 @@
...
+
+In the same way, it is possible to wrap all other scripts (`save-artifacts`, `assemble-runtime` and `run`).
```
bparees: not exactly the same.  if you wrap a run script, any logic you have after the invocation of the run script will not get executed.

also you need to make sure you "exec" the image's run script i think, otherwise your wrapping script will be pid 1 and get all the signal handling.

docs/user_guide.md
php-coder: I did updates for other comments, but I'm stuck with this: we have 2 variants already but none of them we can use and I cannot come up with another one :(

bparees: sorry what do you mean we have 2 variants but none of them we can use?

php-coder: @bparees I mean that the first is mine, original, and the second is yours, but you said "don't take it verbatim".

php-coder: Ok, I got the idea. I'm working on a new version at this moment..

bparees: yeah i just meant "this still needs some rephrasing/wordsmithing"

php-coder: @bparees I rewrote it completely :-/ PTAL.

docs/user_guide.md
```diff
@@ -0,0 +1,46 @@
+# Using S2I images
+
+When you have tried to use OpenShift builder images and found out that your application requires yet other actions for building/running then you are trying to leverage user scripts from `.s2i/bin`. To make your path smooth here we have collected some approaches and not always obvious receipts.
```
bparees: I thought we had higher aspirations for this document, that it would be a general usage guide, but i guess until we have that we can keep it simple.

that said, this paragraph still needs work.  How about:  "S2I builder images normally include assemble and run scripts, but the default behavior of those scripts may not be suitable for all users.  This document will cover a few approaches for customizing the behavior of an S2I builder that includes default assemble/run scripts."

docs/user_guide.md
```diff
@@ -0,0 +1,46 @@
...
+
+## Invoking scripts embedded into an image
```
bparees: s/into/in/

docs/user_guide.md
```diff
@@ -0,0 +1,46 @@
...
+
+Typically builder images provide its own version of [S2I scripts](builder_image.md#s2i-scripts) that are covering the most common use-cases. When these scripts don't fulfill your needs, S2I provides a way of overriding them by adding custom ones in the `.s2i/bin` directory. But doing this you are completely replacing the standard scripts. In some cases this is acceptable but what if you just need to execute a few commands before (or after) the scripts and then proceed execution flow as before? In this case it is possible to create a wrapper script that will execute custom logic and delegate further work to a standard script.
```
bparees: s/its/their/
s/version of/version of the/
s/that are covering/that cover/
w/When these scripts/If these scripts/
s/But doing this/However, by doing this/
s/In some cases..../In some cases this is acceptable but in other scenarios you may prefer to execute a few commands before(or after) the scripts while retaining the logic of the script provided in the image."
s/to a standard script/to the default script in the image./

bparees:
@php-coder one more round of changes and then i think it's good.

docs/user_guide.md
php-coder: >I thought we had higher aspirations for this document, that it would be a general usage guide, but i guess until we have that we can keep it simple.

A journey of a thousand miles begins with a single step (c)

> that said, this paragraph still needs work. How about: "S2I builder images normally include assemble and run scripts, but the default behavior of those scripts may not be suitable for all users. This document will cover a few approaches for customizing the behavior of an S2I builder that includes default assemble/run scripts."

Ok, I updated this paragraph but I removed last occurrence of "assemble/run" to reduce repetition.

php-coder:
@bparees Thank you for your feedback! PTAL again.

bparees:
@php-coder squash and lgtm

php-coder:
@bparees Squashed and ready to merge.

bparees:
[merge]

PR #572, opened 2016-08-16
Fix Environment and EnvironmentSpec duplication

### LINKED ISSUE #545: Environment and EnvironmentSpec duplication
I've found that S2I has two exactly the same structures for representing environment variable: [`Environment`](https://github.com/openshift/source-to-image/blob/9350cd1ef8afc9c91dcdc96e1d972cbcc6f36181/pkg/scripts/environment.go#L14-L18) and [`EnvironmentSpec`](https://github.com/openshift/source-to-image/blob/9350cd1ef8afc9c91dcdc96e1d972cbcc6f36181/pkg/api/types.go#L238-L242).

I think that we don't need to have both and one of them should be superseded by another.

php-coder:
PTAL @bparees 

Let me know if you think that I also should rename `EnvironmentSpec` to `Environment`.

Fixes #545

### DIFF
--- a/pkg/build/strategies/sti/sti.go
+++ b/pkg/build/strategies/sti/sti.go
@@ -358,7 +358,7 @@ func createBuildEnvironment(config *api.Config) []string {
 		glog.V(3).Infof("No user environment provided (%v)", err)
 	}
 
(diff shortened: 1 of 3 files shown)


bparees:
lgtm but waiting for the docker-client change to drop before merging.

PR #574, opened 2016-08-19
s2i: intermittent hang

gabemontero:
Bug 1368074
Bugzilla link https://bugzilla.redhat.com/show_bug.cgi?id=1368074
Detailed explanation
A timing window existed, seen first with incremental builds and the new engine-api/k8s docker client, where the attach container would never return because of a blocking channel call

@bparees PTAL

### DIFF
--- a/pkg/docker/docker.go
+++ b/pkg/docker/docker.go
@@ -15,7 +15,6 @@ import (
 	"path/filepath"
 	"runtime"
 	"strings"
-	"sync"
 	"syscall"
 	"time"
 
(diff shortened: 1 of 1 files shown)


bparees:
lgtm [merge]

gabemontero:
I only got the known flake (the layered one, right?) when i ran the integration tests but the incremental builds had a couple of couldn't find the save-artifacts invoked markers.  Perhaps rather than a `case default:` in the change we do a timer in case we are exiting too soon.  I'll look into it either later this weekend or on Monday.

gabemontero:
[test]

@bparees I reworked the fix, backing out the change to the borrowed k8s logic.  While that change handled the hang encountered, it did not properly account for stdin closing well before stdout did.

The new fix backs off the mutex.Lock/Unlock usage (too all or nothing in hindsight), and uses the fact that we have already moved past the container wait return (meaning we can assume the container has exited) and simply waits for the attach to have actually exited, or for a time out to expire.

If we don't see any data race errors flagged in the unit test, nor missing incremental output in the integration tests, we should be good to go.

gabemontero:
OK, tests are clean.  @bparees - see if you are good with the new form of the fix, post comments / merge tag at your discretion please.

bparees:
lgtm [merge]

PR #575, opened 2016-08-22
only override entrypoint in image if one is set

bparees:
@php-coder and @gabemontero ptal.  this is a follow up to the ongoing issues described in:
https://github.com/openshift/source-to-image/issues/475#issuecomment-238088669

### DIFF
--- a/pkg/docker/docker.go
+++ b/pkg/docker/docker.go
@@ -878,25 +878,24 @@ func (d *stiDocker) RunContainer(opts RunContainerOptions) error {
 		}
 	}
 
-	// if the original image has no entrypoint, and the run invocation
-	// is trying to set an entrypoint, ignore it.  We only want to
-	// set the entrypoint if we need to override a default entrypoint
-	// in the image.  This allows us to still work with a minimal image
-	// that does not contain "/usr/bin/env" since we don't attempt to override
-	// the entrypoint.
-	if len(opts.Entrypoint) != 0 {
-		entrypoint, err := d.GetImageEntrypoint(image)
-		if err != nil {
-			return err
-		}
-		if len(entrypoint) == 0 {
-			opts.Entrypoint = nil
-		}
-	}
-
 	if err != nil {
 		glog.V(0).Infof("error: Unable to get image metadata for %s: %v", image, err)
 		return err
+	}
+
+	entrypoint, err := d.GetImageEntrypoint(image)
+	if err != nil {
+		return fmt.Errorf("Couldn't get entrypoint of %q image: %v", image, err)
+	}
+
+	// If the image has an entrypoint already defined,
+	// it will be overridden either by DefaultEntrypoint,
+	// or by the value in opts.Entrypoint.
+	// If the image does not have an entrypoint, but
+	// opts.Entrypoint is supplied, opts.Entrypoint will
+	// be respected.
+	if len(entrypoint) != 0 && len(opts.Entrypoint) == 0 {
+		opts.Entrypoint = DefaultEntrypoint
 	}
 
 	// tarDestination will be passed as location to PostExecute function
--- a/pkg/build/strategies/sti/postexecutorstep.go
+++ b/pkg/build/strategies/sti/postexecutorstep.go
@@ -273,7 +273,6 @@ func (step *startRuntimeImageAndUploadFilesStep) execute(ctx *postExecutorStepCo
 
 	opts := dockerpkg.RunContainerOptions{
 		Image:           image,
-		Entrypoint:      DefaultEntrypoint,
 		PullImage:       false, // The PullImage is false because we've already pulled the image
 		CommandExplicit: []string{"/bin/sh", "-c", cmd},
 		Stdout:          outWriter,
--- a/pkg/build/strategies/sti/sti.go
+++ b/pkg/build/strategies/sti/sti.go
@@ -26,9 +26,6 @@ import (
 )
 
 var (
-	// DefaultEntrypoint is the default entry point used when starting containers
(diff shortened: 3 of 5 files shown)


bparees:
[test]

bparees:
(the idea being to minimize the scenarios under which we attempt to override the entrypoint, to avoid the edge cases where our override logic ends up setting the entrypoint to a path that doesn't actually exist)

php-coder:
Maybe we can replace [existing logic inside of `stiDocker`](https://github.com/openshift/source-to-image/blob/master/pkg/docker/docker.go#L881-L895) by this check? `docker` interface already has many S2I specific code so I don't think that it will get dirtier with this change :)

In any case LGTM.

bparees:
@php-coder i don't get your suggestion.

php-coder:
@bparees 

Right now we have an execution flow like this:
1) invoking `docker.GetImageEntrypoint(image)`
2) checking Entrypoint and modify it if needed
3) invoking `docker.RunContainer(opts)`
4) here we're again invoking `docker.GetImageEntrypoint(image)` and checking Entrypoint and modifying it

We are doing 1) and 2) at 3 places and this means that this code is duplicated. Also we're invoking `docker.GetImageEntrypoint()` twice. I see only downsides and don't see any benefits of it. So, I was suggesting to do this inspection at the single place -- inside of `docker.RunContainer()`.

bparees:
@php-coder thanks, reworked based on your suggestion...I had forgotten that other code existed, despite being the one who put it in :(

ptal.

pkg/docker/docker.go
```diff
@@ -878,20 +878,19 @@ func (d *stiDocker) RunContainer(opts RunContainerOptions) error {
...
+	// it will be overridden either by DefaultEntrypoint,
+	// or by the value in opts.Entrypoint.
+	// If the image does not have an entrypoint, but
+	// opts.Entrypoint is supplied, opts.Entrypoint will
+	// be respected.
+	if len(entrypoint) != 0 && len(opts.Entrypoint) == 0 {
+		opts.Entrypoint = DefaultEntrypoint
 	}
 
 	if err != nil {
```
php-coder: Looks like we should move our checks for Entrypoint after this `if`.

bparees: whoops, good catch.  fixed.

php-coder:
LGTM.

The logic around Entrypoint gets bigger and bigger. I hope that you have a full picture of this in you mind because I'm starting to lose it. Or at least we have the tests that stop us to do a wrong step.

I see that we have flow like this:
Original Image -> RunContainer -> CommitContainer

And there is 2 possible cases (correct me where I'm wrong):
1) <none> -> <none> -> <none>
In other words, if there no Entrypoint we leave it as is.
2) </bin/bar> -> </usr/bin/env> -> </usr/bin/env>
Here we have to change custom entrypoint to `env` to allow execution of `assemble`.

I'm sure that I miss something here, right?

bparees:
case (1) you have correct.
case (2) is wrong, it will go /bin/bar -> /usr/bin/env -> /bin/bar

I think the logic pretty straightforward at this point:

If the image came w/ an entrypoint, override it with "/usr/bin/env" for purposes of running the image(for assemble, save artifacts, etc) (actually now i'm wondering why we can't set the entrypoint to "" instead of "/usr/bin/env".. @mfojtik did you try that when you first developed this fix?)

When committing the container(which may be running with an overridden entrypoint), set the entrypoint back to what the image originally came with.


(To answer my other question about why we are setting it to /usr/bin/env instead of just an empty array, Graham noted in this original report that `When running docker run, if it has an existing ENTRYPOINT you can't from memory just supply --entrypoint '' as it will actually try and run the supplied argument (empty string) still.` but i'm not sure if that's true when setting an endpoint via the api... that may be a CLI "optimization".  Would still be interested to hear from @mfojtik though, given all the pain setting it to /usr/bin/env is causing us)

mfojtik:
@bparees if you set the entrypoint to empty string, docker will report that as entrypoint set to "null" in `docker inspect`. But I believe they will default it in that case to `/bin/sh -c` as `docker run --entrypoint="" -it centos bash` works just fine. So I think Graham is not correct (you can try yourself) or they fixed this in Docker. 

I think the original idea was:

1) If your builder image has entrypoint set to something weird, we will change it to "/usr/bin/env" so we can happily run S2I scripts. When we commit the application image, we revert to original entrypoint. This might be the case when you have database image that has entrypoint set to something like "mysqld" and you want to run script inside this container. 
2) If the entrypoint is not set, it is a no-op for us

I'm getting lost in why this is getting this complex and causing pain ;-)

bparees:
> I'm getting lost in why this is getting this complex and causing pain

@mfojtik because there is no universal entrypoint that will work across all images, therefore we need a solution that does not require setting the entrypoint to a particular value in order to override the existing entrypoint the image declares.

Barring that i'm trying to minimize the cases in which we set the entrypoint, hence this PR.

bparees:
[merge]

bparees:
> @bparees if you set the entrypoint to empty string, docker will report that as entrypoint set to "null" in docker inspect.

@mfojtik 

```
$ cat Dockerfile 
FROM centos

ENTRYPOINT echo hello world
```

```
$ docker run --entrypoint="" 07fd1434e8da
hello world
```

```
$ docker run --entrypoint='' 07fd1434e8da
hello world
```

so no, setting entrypoint to emptystring does not appear to override the ENTRYPOINT in the image or cause it to be set to "/bin/sh".  I don't know for sure how the api behaviors if you provide `opts.entrypoint=nil` or `opts.entrypoint=[]string{}` but one would assume it would behave the same way.

PR #576, opened 2016-08-25
bump default docker timeout

gabemontero:
related to Bug 1370265
but refactored pulls to isolate default timeout change from waitgroup change ... separate pull coming for that 

@bparees @csrwng @rhcarvalho PTAL

### DIFF
--- a/pkg/docker/docker.go
+++ b/pkg/docker/docker.go
@@ -67,7 +67,7 @@ const (
 
 	// DefaultDockerTimeout specifies a timeout for Docker API calls. When this
 	// timeout is reached, certain Docker API calls might error out.
-	DefaultDockerTimeout = 20 * time.Second
+	DefaultDockerTimeout = 60 * time.Second
 
 	// DefaultShmSize is the default shared memory size to use (in bytes) if not specified.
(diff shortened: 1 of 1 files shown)


bparees:
@rhcarvalho would really appreciate if you could take a look at this fix, we're trying to resolve a blocker bug for 3.3.  The go stacks for the hang we're trying to address are in the referenced bug.

The net is wg.Done() could get called twice (once for the source upload and once for the timeout error on starting the container)

bparees:
[test]

gabemontero:
changed this pull to just updating the default timeout to facilitate upstream pull in origin targeted for 3.3

will create new pull with wg change that we'll target as origin upstream pull for 3.3.1

bparees:
[merge]

rhcarvalho:
@gabemontero @bparees is this still reflecting the value we want, or shall we revert back and rely on just fixing the WaitGroup usage?

gabemontero:
Yes we still want the longer timeout based on what we are seeing in QE and
online envs wrt docker responsiveness.
Independent of the WaitGroup stuff.

PR #580, opened 2016-09-01
if found = true, add break is better

guangxuli:
Adding a break is a better manner though in a test case, just a nit.


bparees:
[merge]

PR #577, opened 2016-08-26
s2i: build hung, stuck

gabemontero:
Bug 1370265
Bugzilla link https://bugzilla.redhat.com/show_bug.cgi?id=1370265
Detailed explanation
a timeout on container start could result on waitgroup.Done being called twice, the waitgroup going negative, and down a hung/panic path

the refactoring of pull 576 completed with this pull

@bparees @csrwng @rhcarvalho PTAL

PR #581, opened 2016-09-01
Fix WaitGroup usage

rhcarvalho:
Calling wg.Done() does not "cancel" the WaitGroup, it does introduce a
bug in which depending on timing the WaitGroup counter might go
negative, what should never happen.

Would be nice to rewrite the method to be correct concurrent code, but
for now should be enough to avoid the panic condition and note that
there is a potential leak.

---

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1370265

### DIFF
--- a/pkg/build/strategies/sti/sti.go
+++ b/pkg/build/strategies/sti/sti.go
@@ -560,7 +560,6 @@ func (builder *STI) Execute(command string, user string, config *api.Config) err
 		}()
 
 		opts.Stdin = r
-		defer wg.Wait()
 	}
 
 	go func(reader io.Reader) {
@@ -591,15 +590,16 @@ func (builder *STI) Execute(command string, user string, config *api.Config) err
 	go dockerpkg.StreamContainerIO(errReader, &errOutput, func(a ...interface{}) { glog.Info(a...) })
 
 	err := builder.docker.RunContainer(opts)
-	if util.IsTimeoutError(err) {
-		// Cancel waiting for source input if the container timeouts
-		wg.Done()
-	}
 	if e, ok := err.(errors.ContainerError); ok {
 		// even with deferred close above, close errReader now so we avoid data race condition on errOutput;
 		// closing will cause StreamContainerIO to exit, thus releasing the writer in the equation
 		errReader.Close()
 		return errors.NewContainerError(config.BuilderImage, e.ErrorCode, errOutput)
+	}
+	// Do not wait for source input if the container times out.
+	// FIXME: this potentially leaks a goroutine.
+	if !util.IsTimeoutError(err) {
+		wg.Wait()
 	}
 	return err
 }


rhcarvalho:
I propose this instead of #577, @gabemontero @bparees PTAL.
Sorry for coming with a PR, was easier to explain with code than lots of comments.

pkg/build/strategies/sti/sti.go
```diff
@@ -533,7 +533,7 @@ func (builder *STI) Execute(command string, user string, config *api.Config) err
 		close(injectionComplete)
 	}
 
-	wg := sync.WaitGroup{}
+	var wg sync.WaitGroup
```
rhcarvalho: We don't need to initialize WaitGroups, just declare them and they are ready for use.
Idiomatic example: https://golang.org/pkg/sync/#example_WaitGroup

pkg/build/strategies/sti/sti.go
```diff
@@ -560,7 +560,6 @@ func (builder *STI) Execute(command string, user string, config *api.Config) err
 		}()
 
 		opts.Stdin = r
-		defer wg.Wait()
```
rhcarvalho: This is a very tricky defer statement, there is no clean way to "cancel" it.

Calling `wg.Done()` earlier introduces a concurrency problem. As @bparees and @gabemontero noted, `wg.Done()` could end up being called multiple times sending the counter below 0, causing a panic.

Correct code arranges to call `wg.Done()` as many times as `wg.Add(n)` in all cases.

pkg/build/strategies/sti/sti.go
```diff
@@ -591,16 +590,17 @@ func (builder *STI) Execute(command string, user string, config *api.Config) err
...
-	}
 	if e, ok := err.(errors.ContainerError); ok {
 		// even with deferred close above, close errReader now so we avoid data race condition on errOutput;
 		// closing will cause StreamContainerIO to exit, thus releasing the writer in the equation
 		errReader.Close()
 		return errors.NewContainerError(config.BuilderImage, e.ErrorCode, errOutput)
 	}
+	// Do not wait for source input if the container times out.
+	// FIXME: this potentially leaks a goroutine.
+	if !util.IsTimeoutError(err) {
```
gabemontero: removing the  !config.LayeredBuild check of the prior fix seems incorrect to me

pkg/build/strategies/sti/sti.go
```diff
@@ -591,16 +590,17 @@ func (builder *STI) Execute(command string, user string, config *api.Config) err
...
+		wg.Wait()
+	}
```
rhcarvalho: Ideally we'd drop a `wg.Wait()` here, without the `if`, but the `if` here is trying to cover whatever use case it had before, which I am not really sure is the correct thing to do, but we can probably defer to a rewrite. Assign that rewrite to me :-)

pkg/build/strategies/sti/sti.go
```diff
@@ -591,16 +590,17 @@ func (builder *STI) Execute(command string, user string, config *api.Config) err
 	go dockerpkg.StreamContainerIO(errReader, &errOutput, func(a ...interface{}) { glog.Info(a...) })
 
 	err := builder.docker.RunContainer(opts)
-	if util.IsTimeoutError(err) {
-		// Cancel waiting for source input if the container timeouts
-		wg.Done()
```
rhcarvalho: As I said, this is not a correct mechanism for cancellation, instead we will only call `wg.Wait()` if and when we want to block waiting for the goroutine that uploads source code to terminate.

pkg/build/strategies/sti/sti.go
rhcarvalho: See https://play.golang.org/p/kpKJh9Prei

If you never call `wg.Add`, it is perfectly ok and correct to call `wg.Wait()`, so we don't need to add that extra condition, it is irrelevant.

gabemontero: On Thursday, September 1, 2016, Rodolfo Carvalho <notifications@github.com>
wrote:

> In pkg/build/strategies/sti/sti.go
> <https://github.com/openshift/source-to-image/pull/581#discussion_r77222225>
> :
>
> >  	if e, ok := err.(errors.ContainerError); ok {
> >  		// even with deferred close above, close errReader now so we avoid data race condition on errOutput;
> >  		// closing will cause StreamContainerIO to exit, thus releasing the writer in the equation
> >  		errReader.Close()
> >  		return errors.NewContainerError(config.BuilderImage, e.ErrorCode, errOutput)
> >  	}
> > +	// Do not wait for source input if the container times out.
> > +	// FIXME: this potentially leaks a goroutine.
> > +	if !util.IsTimeoutError(err) {
>
> See https://play.golang.org/p/kpKJh9Prei
>
> If you never call wg.Add, it is perfectly ok and correct to call wg.Wait(),
> so we don't need to add that extra condition, it is irrelevant.
>
Ah - ok yeah that makes sense now - the count is still 0.

pkg/build/strategies/sti/sti.go
gabemontero: don't let the PM team see this :-) ... they prefer the `:=`

gabemontero:
LGTM ... I'll delete the other pull when this merges.

@bparees - shall we assign the associated bugzilla and UPSTREAM vendor update to @rhcarvalho :-) ?

pkg/build/strategies/sti/sti.go
bparees: i also prefer := :)

rhcarvalho:
> @bparees - shall we assign the associated bugzilla and UPSTREAM vendor update to @rhcarvalho :-) ?

Would be fun, but I don't want to overcommit :)

I'll be happy to send more s2i PRs when I come back (soon) :smile: , missing you guys

pkg/build/strategies/sti/sti.go
rhcarvalho: This is not a necessary change, so will revert it to minimize the patch ;)

pkg/build/strategies/sti/sti.go
bparees: i think we need the if-check around the wg.Wait() because the bad scenario would be:

1) injections start (wg.Add(1) occurs)
2) call to start container times out (container never starts)
3) injections never finish because the container never started
4) wg.Done() never gets called because the injections never finish
5) wg.Wait() never completes

which is the original hang we were trying to fix.

that's not to say a rewrite couldn't improve it, but at least as it stands now i believe it's necessary.

bparees:
lgtm [merge]
@gabemontero please do the upstream PR to origin, we've already abused @rhcarvalho enough :)

gabemontero:
roger that @bparees

gabemontero:
and thx @rhcarvalho :-)

pkg/build/strategies/sti/sti.go
rhcarvalho: @bparees thanks for challenging my tired mind :)
Really, good to have another pair of eyes and good brain to reason about it.

> 1) injections start (wg.Add(1) occurs)
> 2) call to start container times out (container never starts)

This means that after

```go
err := builder.docker.RunContainer(opts)
```

`util.IsTimeoutError(err) == true`, right?


> 3) injections never finish because the container never started

Yes, and this causes a goroutine leak.

> 4) wg.Done() never gets called because the injections never finish
> 5) wg.Wait() never completes

When `util.IsTimeoutError(err) == true`, `wg.Wait()` is not called. In all other cases, i.e. `!util.IsTimeoutError(err)`, `wg.Wait()` is called.

rhcarvalho: @bparees does this make sense to you after all?

In my mind we're trading a panic for a leak, and based on previous discussions this is a "less critical" leak because this code is running in the builder pod and the process is short-lived.

bparees: we're in agreement, my point was that we *need* the "if !util.IsTimeoutError(err)" check.  

It sounded to me like you were implying we might not need it, but on rereading I guess all you were saying is "ideally we'd refactor this code so this check is not needed, but until we do that, it is needed, so here it is"

so i think we're good (and yeah, goroutine leak. oh well. luckily this code is running in a short lived container in openshift, or a short lived process in s2i)

PR #582, opened 2016-09-05
[test/run] fix unboud variable in the template

mikz:
container_ip is unbound variable, it is a function that has to be called

after fixing this the script correctly executes with `set -u`


bparees:
lgtm [merge]

bparees:
[merge]

PR #584, opened 2016-09-06
better defer waitgroup Done pattern

gabemontero:
@bparees PTAL

stems from recent discussion in https://github.com/openshift/origin/pull/10776


bparees:
lgtm [merge]

PR #585, opened 2016-09-07
Change file exists check to any error returned by os.Stat

csrwng:
When a ssh url is passed to os.Stat in Windows, the error returned is something other than  ErrNotExist, which results in an error when 'oc new-app' is called with a git url in Windows.

With this change, any error returned by os.Stat will result in us assuming the file doesn't exist.


csrwng:
@gabemontero ptal
@bparees fyi

bparees:
lgtm [merge]

gabemontero:
LGTM

csrwng:
ok, now it's fixed

bparees:
[merge]

PR #586, opened 2016-09-15
Fix container_ip function on Mac

luiscoms:
When using docker machine on Mac tests can't access container ip directly, so I'm using docker machine ip.
The host port could be change, so we get it from container.

### DIFF
--- a/pkg/create/templates/test.go
+++ b/pkg/create/templates/test.go
@@ -1,8 +1,7 @@
 package templates
 
 // TestRunScript is a simple test script that verifies the S2I image.
-const TestRunScript = `
-#!/bin/bash
+const TestRunScript = `#!/bin/bash
 #
 # The 'run' performs a simple test that verifies the S2I image.
 # The main focus here is to exercise the S2I scripts.
@@ -18,7 +17,7 @@ IMAGE_NAME=${IMAGE_NAME-{{.ImageName}}-candidate}
 # Determining system utility executables (darwin compatibility check)
 READLINK_EXEC="readlink"
 MKTEMP_EXEC="mktemp"
-if (echo "$OSTYPE" | egrep -qs 'darwin'); then
+if [[ "$OSTYPE" =~ 'darwin' ]]; then
   ! type -a "greadlink" &>"/dev/null" || READLINK_EXEC="greadlink"
   ! type -a "gmktemp" &>"/dev/null" || MKTEMP_EXEC="gmktemp"
 fi
@@ -44,7 +43,19 @@ container_exists() {
 }
 
 container_ip() {
-  docker inspect --format="{{"{{"}} .NetworkSettings.IPAddress {{"}}"}}" $(cat $cid_file)
+  if [ ! -z "$DOCKER_HOST" ] && [[ "$OSTYPE" =~ 'darwin' ]]; then
+    docker-machine ip
+  else
+    docker inspect --format="{{"{{"}} .NetworkSettings.IPAddress {{"}}"}}" $(cat $cid_file)
+  fi
+}
+
+container_port() {
+  if [ ! -z "$DOCKER_HOST" ] && [[ "$OSTYPE" =~ 'darwin' ]]; then
+    docker inspect --format="{{"{{"}}(index .NetworkSettings.Ports \"$test_port/tcp\" 0).HostPort{{"}}"}}" "$(cat "${cid_file}")"
+  else
+    echo $test_port
+  fi
 }
 
 run_s2i_build() {
@@ -108,19 +119,14 @@ test_usage() {
 }
 
 test_connection() {
-  echo "Testing HTTP connection..."
+  echo "Testing HTTP connection (http://$(container_ip):$(container_port))"
   local max_attempts=10
   local sleep_time=1
   local attempt=1
   local result=1
   while [ $attempt -le $max_attempts ]; do
-    echo "Sending GET request to http://$(container_ip):${test_port}/"
-    if (echo "$OSTYPE" | egrep -qs 'darwin'); then
-      echo "Warning for OSX users: if you can't access the container's IP $(container_ip) directly (because you use boot2docker for example)"
-      echo "you should run the curl command in a container, for example using:"
-      echo "docker run --rm -it sequenceiq/alpine-curl curl -s -w %{http_code} -o /dev/null http://$(container_ip):${test_port}/"
-    fi
-    response_code=$(curl -s -w %{http_code} -o /dev/null http://$(container_ip):${test_port}/)
+    echo "Sending GET request to http://$(container_ip):$(container_port)/"
+    response_code=$(curl -s -w %{http_code} -o /dev/null http://$(container_ip):$(container_port)/)
     status=$?
     if [ $status -eq 0 ]; then
       if [ $response_code -eq 200 ]; then


bparees:
@PI-Victor @csrwng need one of you mac guys to review this please :)

PI-Victor:
uhm, sorry for the late reply. i'll have a look at this

pkg/create/templates/test.go
```diff
@@ -43,7 +43,19 @@ container_exists() {
 }
 
 container_ip() {
-  docker inspect --format="{{"{{"}} .NetworkSettings.IPAddress {{"}}"}}" $(cat $cid_file)
+  if (echo "$OSTYPE" | egrep -qs 'darwin'); then
```
php-coder: IMHO we can use `grep` (or even `fgrep`) here instead of `egrep` because we don't use regular expressions here at all. Also I think that brackets could be omitted (`if echo "$OSTYPE" | grep -qs 'darwin'; then`).

pkg/create/templates/test.go
```diff
@@ -43,7 +43,19 @@ container_exists() {
 }
 
 container_ip() {
-  docker inspect --format="{{"{{"}} .NetworkSettings.IPAddress {{"}}"}}" $(cat $cid_file)
+  if (echo "$OSTYPE" | grep -qs 'darwin'); then
```
rhcarvalho: Is it right to assume that everyone on macOS will have the right IP from `docker-machine ip`?

If that's the case, again more Bash nitting, we don't even need `echo` and `grep`:

```bash
case "$OSTYPE" in
  darwin*)  docker-machine ip ;;
  *)        docker inspect --format="{{"{{"}} .NetworkSettings.IPAddress {{"}}"}}" $(cat $cid_file) ;;
esac
```

(inspired by [StackOverflow](http://stackoverflow.com/questions/394230/detect-the-os-from-a-bash-script#394235))

rhcarvalho:
Changing the test runner template to know about `docker-machine` seems like a bad move. Please correct me if I'm wrong as I'm not familiar with the details, folks on Windows and macOS can now run Docker without Docker Machine, and I believe this patch would break running the tests for them.

@luiscoms can't you instead ssh to the Docker Machine VM and run the tests from there?

rhcarvalho:
@luiscoms did you arrive at this by creating your own s2i image from the template, or are you trying to test some specific existing image?

luiscoms:
@rhcarvalho I meld some experiences that I have when creating and editing some images on my own.
Some examples:

[luiscoms/s2i-java](https://github.com/luiscoms/s2i-java)
[luiscoms/springboot-maven3-centos](https://github.com/luiscoms/springboot-maven3-centos)
[luiscoms/s2i-nginx](https://github.com/luiscoms/s2i-nginx)

pkg/create/templates/test.go
csrwng: @luiscoms isn't it sufficient to have the change you made in line 132? (without having to call docker-machine ip). Afaik containers can communicate with each other using the IP that Docker gives them, even if you can't access them from the host Mac.

luiscoms: @csrwng The warning on that line is just informative and could be deleted if these changes works.

csrwng: @luiscom what I meant is that you can get the address of the container in the same way as before:
```
docker inspect --format="{{"{{"}} .NetworkSettings.IPAddress {{"}}"}}" $(cat $cid_file)
```

Simply change the way that it's curled on the Mac. Because you are starting another container to curl, the container should have access to the test container ip, even if your Mac doesn't.

luiscoms: @csrwng You are right about to start another container to run `curl`, my changes are just to avoid this approach and test image on the local machine.

csrwng: ahh got it

csrwng: one approach to avoid depending on docker-machine (and supporting Docker for Mac) could be to use socat to forward traffic for you

luiscoms: But it will replace the `docker-machine` dependency to `socat` dependency

csrwng: True ... the problem I see with docker-machine is that it may be installed and it might even have a valid ip but it could be that I'm configured to point to a different Docker. For example, in my case, I have both docker-machine and Docker for Mac ... docker-machine ip will return a valid ip because I do have a default machine created. However, most of the time my DOCKER_HOST variable is not set, which means that 'docker run' will use Docker for Mac.

pkg/create/templates/test.go
```diff
@@ -43,7 +43,19 @@ container_exists() {
 }
 
 container_ip() {
-  docker inspect --format="{{"{{"}} .NetworkSettings.IPAddress {{"}}"}}" $(cat $cid_file)
+  if [ ! -z "$DOCKER_HOST" ] && (echo "$OSTYPE" | grep -qs 'darwin'); then
```
rhcarvalho: Suggestion:

```bash
[[ "$OSTYPE" =~ 'darwin' ]]
```

(And if you feel like cleanning up Bash, we could also use that in [line 20](https://github.com/openshift/source-to-image/pull/586/files#diff-1a14ec8a07d06e9da1a51072db9046bfR20), and other occurrences if any)

pkg/create/templates/test.go
```diff
@@ -43,7 +43,19 @@ container_exists() {
...
+    docker-machine ip
+  else
+    docker inspect --format="{{"{{"}} .NetworkSettings.IPAddress {{"}}"}}" $(cat $cid_file)
+  fi
+}
+
+container_port() {
+  if [ ! -z "$DOCKER_HOST" ] && (echo "$OSTYPE" | grep -qs 'darwin'); then
+    docker inspect --format='{{(index (index .NetworkSettings.Ports "8080/tcp") 0).HostPort}}' $(cat $cid_file)
```
rhcarvalho: The `{{` is interpreted by the `template/text` package -- this whole block is a string with a Bash script that gets processed as a template.

Just like in the line we are replacing, we need to "escape" the `{{`. The existing solution involves using the template language to print the `{{`:

```
docker inspect --format='{{"{{"}}(index (index .NetworkSettings.Ports "8080/tcp") 0).HostPort{{"{{"}}' $(cat $cid_file)
```

(Note the `{{"{{"}}` part)

An alternative solution I've seen when a Go template includes a Go template is to [specify a custom delimiter](https://golang.org/pkg/text/template/#Template.Delims), what would require changing the code that processes the templates + updating the templates to use the new delimiters.

pkg/create/templates/test.go
```diff
@@ -107,19 +119,14 @@ test_usage() {
...
   while [ $attempt -le $max_attempts ]; do
-    echo "Sending GET request to http://$(container_ip):${test_port}/"
-    if (echo "$OSTYPE" | egrep -qs 'darwin'); then
-      echo "Warning for OSX users: if you can't access the container's IP $(container_ip) directly (because you use boot2docker for example)"
-      echo "you should run the curl command in a container, for example using:"
-      echo "docker run --rm -it sequenceiq/alpine-curl curl -s -w %{http_code} -o /dev/null http://$(container_ip):${test_port}/"
-    fi
-    response_code=$(curl -s -w %{http_code} -o /dev/null http://$(container_ip):${test_port}/)
+    echo "Sending GET request to http://$(container_ip):$(container_port)/"
+    response_code=$(curl -s -w %{http_code} -o /dev/null http://$(container_ip):$(container_port)/)
```
rhcarvalho: There was so long time I didn't look at this that I forgot how macOS support was lacking. @luiscoms thanks for improving it!

pkg/create/templates/test.go
luiscoms: Thanks for correct me!

pkg/create/templates/test.go
luiscoms: I like this suggestion!

pkg/create/templates/test.go
```diff
@@ -43,7 +43,20 @@ container_exists() {
...
+  if [ ! -z "$DOCKER_HOST" ] && [[ "$OSTYPE" =~ 'darwin' ]]; then
+    docker-machine ip
+  else
+    docker inspect --format="{{"{{"}} .NetworkSettings.IPAddress {{"}}"}}" $(cat $cid_file)
+  fi
+}
+
+container_port() {
+  if [ ! -z "$DOCKER_HOST" ] && [[ "$OSTYPE" =~ 'darwin' ]]; then
+    docker inspect --format='{{"{{"}}(index (index .NetworkSettings.Ports "8080/tcp") 0).HostPort{{"{{"}}' $(cat $cid_file)
```
rhcarvalho: Almost! The second part needs to be `{{"}}"}}` :wink: 

And the [`index` function](https://golang.org/pkg/text/template/#hdr-Functions) takes a variable number of arguments, so we can accomplish the same with:

```
docker inspect --format='{{"{{"}}(index .NetworkSettings.Ports "8080/tcp" 0).HostPort{{"}}"}}' $(cat $cid_file)
```

or, suggestion, quoting and bracketing to ensure we're passing a single argument to `docker inspect`:

```
docker inspect --format='{{"{{"}}(index .NetworkSettings.Ports "8080/tcp" 0).HostPort{{"}}"}}' "$(cat "${cid_file}")"
```

pkg/create/templates/test.go
```diff
@@ -43,7 +43,19 @@ container_exists() {
...
+  if [ ! -z "$DOCKER_HOST" ] && [[ "$OSTYPE" =~ 'darwin' ]]; then
+    docker-machine ip
+  else
+    docker inspect --format="{{"{{"}} .NetworkSettings.IPAddress {{"}}"}}" $(cat $cid_file)
+  fi
+}
+
+container_port() {
+  if [ ! -z "$DOCKER_HOST" ] && [[ "$OSTYPE" =~ 'darwin' ]]; then
+    docker inspect --format="{{"{{"}}(index .NetworkSettings.Ports \"$test_port/tcp\" 0).HostPort{{"}}"}}" "$(cat "${cid_file}")"
```
rhcarvalho: Hard to read, but should work!

In hindsight, I regret my note about quoting and bracketing, since we have no consistency in their use in this script...

luiscoms:
@csrwng Can you test it please?

rhcarvalho:
@PI-Victor @csrwng could you please test this on macOS?

luiscoms:
I just rebase this branch

csrwng:
I'm testing it locally, will update shortly

csrwng:
So I haven't been able to make this work for me. Using docker-machine with virtualbox driver, I get:
https://gist.github.com/csrwng/95415fdf07e0c4a8d8c84bbde813a7ef
With Docker for Mac, it works (kind of) but a test won't succeed because the ip of the container is not accessible from the host:
https://gist.github.com/csrwng/b5eba998f46dab4b3e1f0f2ec8464d13

It definitely could be I'm doing something wrong

luiscoms:
@csrwng I guess that is that issue #603 provoking this

jim-minter:
@luiscoms Hi Luis, please could you rebase onto latest s2i making sure you're building on top of a branch which contains bd892aff569b6514925d82ab64299c028dbb5031? I'm hoping that commit should fix #603.

luiscoms:
It is already rebased to latest release, @csrwng could you please test again?

csrwng:
@luiscoms `make test` worked great now using `docker-machine`. Thx! Please squash your commits.

csrwng:
[merge]

PR #593, opened 2016-09-26
Additional build information

PI-Victor:
This PR adds additional build information so that S2I can tell OpenShift a couple of things:
Why did the build fail. (besides fetch source, [which is implemented on the OpenShift side], container commit, builder image pull, runtime image pull, assemble runtime failed, assemble failed, etc. )

It's related to: https://github.com/openshift/origin/pull/10817

### DIFF
--- a/pkg/api/types.go
+++ b/pkg/api/types.go
@@ -325,7 +325,85 @@ type Result struct {
 
 	// ImageID describes resulting image ID.
 	ImageID string
+
+	// BuildInfo holds information about the result of a build.
+	BuildInfo BuildInfo
 }
+
+// BuildInfo holds information about a particular step in the build process.
+type BuildInfo struct {
+	// FailureReason is a camel case reason that is used by the machine to reply
+	// back to the OpenShift builder with information why any of the steps in the
+	// build, failed.
+	FailureReason StepFailureReason
+}
+
+// StepFailureReason holds the type of failure that occured during the build
+// process.
+type StepFailureReason string
+
+const (
+	// ReasonAssembleFailed is the reason associated with the Assemble script
+	// failing.
+	ReasonAssembleFailed StepFailureReason = "AssembleFailed"
+
+	// ReasonPullBuilderImageFailed is the reason associated with failing to pull
+	// the builder image.
+	ReasonPullBuilderImageFailed StepFailureReason = "PullBuilderImageFailed"
+
+	// ReasonPullRuntimeImageFailed is the reason associated with failing to pull
+	// the runtime image.
+	ReasonPullRuntimeImageFailed StepFailureReason = "PullRuntimeImageFailed"
+
+	// ReasonCommitContainerFailed is the reason associated with failing to
+	// commit the container to the final image.
+	ReasonCommitContainerFailed StepFailureReason = "ContainerCommitFailed"
+
+	// ReasonFetchSourceFailed is the reason associated with failing to download
+	// the source of the build.
+	ReasonFetchSourceFailed StepFailureReason = "FetchSourceFailed"
+
+	// ReasonDockerImageBuildFailed is the reasons associated with a failed
+	// Docker image build.
+	ReasonDockerImageBuildFailed StepFailureReason = "DockerImageBuildFailed"
+
+	// ReasonDockerFileCreateFailed is the reason associated with failing to create a
+	// Dockerfile for a build.
+	ReasonDockerFileCreateFailed StepFailureReason = "DockerFileCreationFailed"
+
+	// ReasonInvalidArtifactsMapping is the reason associated with an
+	// invalid artifacts mapping of files that need to be copied.
+	ReasonInvalidArtifactsMapping StepFailureReason = "InvalidArtifactsMapping"
+
+	// ReasonArtifactsFetchFailed is the reason associated with a failure to
+	// download specified scripts in the application image.
+	ReasonArtifactsFetchFailed StepFailureReason = "FetchScriptsFailed"
+
+	// ReasonFSOperationFailed is the reason associated with a failed fs
+	// operation. Create, remove directory, copy file, etc.
+	ReasonFSOperationFailed StepFailureReason = "FileSystemOperationFailed"
+
+	// ReasonInstallScriptsFailed is the reason associated with a failure to
+	// install scripts in the builder image.
+	ReasonInstallScriptsFailed StepFailureReason = "InstallScriptsFailed"
+
+	// ReasonGenericS2IBuildFailed is the reason associated with a broad range of
+	// failure.
+	ReasonGenericS2IBuildFailed StepFailureReason = "GenericS2IBuildFailed"
+
+	// ReasonUnmetS2IDependencies is the failure reason associated with a
+	// builder image that doesn't contain required dependencies for building the
+	// app.
+	ReasonUnmetS2IDependencies StepFailureReason = "UnmetBuilderImageDependencies"
+
+	// ReasonTarSourceFailed is the failurea reason associated with a failure to
+	// tar the current source.
+	ReasonTarSourceFailed StepFailureReason = "TarSourceFailed"
+
+	// ReasonOnBuildForbidden is the failure reason associated with an image that
+	// uses the ONBUILD instruction when it's not allowed.
+	ReasonOnBuildForbidden StepFailureReason = "OnBuildForbidden"
+)
 
 // InstallResult structure describes the result of install operation
 type InstallResult struct {
--- a/pkg/build/strategies/layered/layered.go
+++ b/pkg/build/strategies/layered/layered.go
@@ -138,24 +138,30 @@ func (builder *Layered) SourceTar(config *api.Config) (io.ReadCloser, error) {
 // Build handles the `docker build` equivalent execution, returning the
 // success/failure details.
 func (builder *Layered) Build(config *api.Config) (*api.Result, error) {
+	buildResult := &api.Result{}
+
 	if config.HasOnBuild && config.BlockOnBuild {
-		return nil, fmt.Errorf("builder image uses ONBUILD instructions but ONBUILD is not allowed")
+		buildResult.BuildInfo.FailureReason = api.ReasonOnBuildForbidden
+		return buildResult, fmt.Errorf("builder image uses ONBUILD instructions but ONBUILD is not allowed")
 	}
 
 	if err := builder.CreateDockerfile(config); err != nil {
-		return nil, err
+		buildResult.BuildInfo.FailureReason = api.ReasonDockerFileCreateFailed
+		return buildResult, err
 	}
 
 	glog.V(2).Info("Creating application source code image")
 	tarStream, err := builder.SourceTar(config)
 	if err != nil {
-		return nil, err
+		buildResult.BuildInfo.FailureReason = api.ReasonTarSourceFailed
+		return buildResult, err
 	}
 	defer tarStream.Close()
 
 	namedReference, err := reference.ParseNamed(builder.config.BuilderImage)
 	if err != nil {
-		return nil, err
+		buildResult.BuildInfo.FailureReason = api.ReasonGenericS2IBuildFailed
+		return buildResult, err
 	}
 	newBuilderImage := fmt.Sprintf("%s:s2i-layered-%d", namedReference.Name(), time.Now().UnixNano())
 
@@ -187,7 +193,8 @@ func (builder *Layered) Build(config *api.Config) (*api.Result, error) {
 
 	glog.V(2).Infof("Building new image %s with scripts and sources already inside", newBuilderImage)
 	if err = builder.docker.BuildImage(opts); err != nil {
-		return nil, err
+		buildResult.BuildInfo.FailureReason = api.ReasonDockerImageBuildFailed
+		return buildResult, err
 	}
 
 	// upon successful build we need to modify current config
@@ -202,21 +209,22 @@ func (builder *Layered) Build(config *api.Config) (*api.Result, error) {
 	} else {
 		builder.config.ScriptsURL, err = builder.docker.GetScriptsURL(newBuilderImage)
 		if err != nil {
-			return nil, err
+			buildResult.BuildInfo.FailureReason = api.ReasonGenericS2IBuildFailed
+			return buildResult, err
 		}
 	}
 
 	glog.V(2).Infof("Building %s using sti-enabled image", builder.config.Tag)
 	if err := builder.scripts.Execute(api.Assemble, config.AssembleUser, builder.config); err != nil {
+		buildResult.BuildInfo.FailureReason = api.ReasonAssembleFailed
 		switch e := err.(type) {
 		case errors.ContainerError:
-			return nil, errors.NewAssembleError(builder.config.Tag, e.Output, e)
+			return buildResult, errors.NewAssembleError(builder.config.Tag, e.Output, e)
 		default:
-			return nil, err
+			return buildResult, err
 		}
 	}
+	buildResult.Success = true
 
-	return &api.Result{
-		Success: true,
-	}, nil
+	return buildResult, nil
 }
--- a/pkg/build/strategies/onbuild/onbuild.go
+++ b/pkg/build/strategies/onbuild/onbuild.go
@@ -75,7 +75,8 @@ func New(config *api.Config, overrides build.Overrides) (*OnBuild, error) {
 	return builder, nil
 }
 
-// SourceTar produces a tar archive containing application source and stream it
+// SourceTar produces a tar archive containing application source and streams
+// it
 func (builder *OnBuild) SourceTar(config *api.Config) (io.ReadCloser, error) {
 	uploadDir := filepath.Join(config.WorkingDir, "upload", "src")
 	tarFileName, err := builder.tar.CreateTarFile(config.WorkingDir, uploadDir)
@@ -87,14 +88,17 @@ func (builder *OnBuild) SourceTar(config *api.Config) (io.ReadCloser, error) {
 
 // Build executes the ONBUILD kind of build
 func (builder *OnBuild) Build(config *api.Config) (*api.Result, error) {
+	buildResult := &api.Result{}
+
 	if config.BlockOnBuild {
-		return nil, fmt.Errorf("builder image uses ONBUILD instructions but ONBUILD is not allowed")
+		buildResult.BuildInfo.FailureReason = api.ReasonOnBuildForbidden
+		return buildResult, fmt.Errorf("builder image uses ONBUILD instructions but ONBUILD is not allowed")
 	}
 	glog.V(2).Info("Preparing the source code for build")
 	// Change the installation directory for this config to store scripts inside
 	// the application root directory.
 	if err := builder.source.Prepare(config); err != nil {
-		return nil, err
+		return buildResult, err
 	}
 
 	// If necessary, copy the STI scripts into application root directory
@@ -102,13 +106,15 @@ func (builder *OnBuild) Build(config *api.Config) (*api.Result, error) {
 
 	glog.V(2).Info("Creating application Dockerfile")
 	if err := builder.CreateDockerfile(config); err != nil {
-		return nil, err
+		buildResult.BuildInfo.FailureReason = api.ReasonDockerFileCreateFailed
+		return buildResult, err
 	}
 
 	glog.V(2).Info("Creating application source code image")
 	tarStream, err := builder.SourceTar(config)
 	if err != nil {
-		return nil, err
+		buildResult.BuildInfo.FailureReason = api.ReasonTarSourceFailed
+		return buildResult, err
 	}
 	defer tarStream.Close()
 
@@ -121,7 +127,8 @@ func (builder *OnBuild) Build(config *api.Config) (*api.Result, error) {
 
 	glog.V(2).Info("Building the application source")
 	if err = builder.docker.BuildImage(opts); err != nil {
-		return nil, err
+		buildResult.BuildInfo.FailureReason = api.ReasonDockerImageBuildFailed
+		return buildResult, err
 	}
 
 	glog.V(2).Info("Cleaning up temporary containers")
@@ -131,7 +138,8 @@ func (builder *OnBuild) Build(config *api.Config) (*api.Result, error) {
 
 	if len(opts.Name) > 0 {
 		if imageID, err = builder.docker.GetImageID(opts.Name); err != nil {
-			return nil, err
+			buildResult.BuildInfo.FailureReason = api.ReasonGenericS2IBuildFailed
+			return buildResult, err
 		}
 	}
 
--- a/pkg/build/strategies/sti/postexecutorstep.go
+++ b/pkg/build/strategies/sti/postexecutorstep.go
@@ -55,9 +55,10 @@ func (step *storePreviousImageStep) execute(ctx *postExecutorStepContext) error 
 	if step.builder.incremental && step.builder.config.RemovePreviousImage {
 		glog.V(3).Info("Executing step: store previous image")
 		ctx.previousImageID = step.getPreviousImage()
-	} else {
-		glog.V(3).Info("Skipping step: store previous image")
+		return nil
 	}
+
+	glog.V(3).Info("Skipping step: store previous image")
 	return nil
 }
 
@@ -79,9 +80,10 @@ func (step *removePreviousImageStep) execute(ctx *postExecutorStepContext) error
 	if step.builder.incremental && step.builder.config.RemovePreviousImage {
 		glog.V(3).Info("Executing step: remove previous image")
 		step.removePreviousImage(ctx.previousImageID)
-	} else {
-		glog.V(3).Info("Skipping step: remove previous image")
+		return nil
 	}
+
+	glog.V(3).Info("Skipping step: remove previous image")
 	return nil
 }
 
@@ -128,6 +130,7 @@ func (step *commitImageStep) execute(ctx *postExecutorStepContext) error {
 
 	ctx.imageID, err = commitContainer(step.docker, ctx.containerID, cmd, user, step.builder.config.Tag, step.builder.env, entrypoint, ctx.labels)
 	if err != nil {
+		step.builder.result.BuildInfo.FailureReason = api.ReasonCommitContainerFailed
 		return err
 	}
 
@@ -383,10 +386,10 @@ func (step *invokeCallbackStep) execute(ctx *postExecutorStepContext) error {
 		glog.V(3).Info("Executing step: invoke callback url")
 		step.builder.result.Messages = step.callbackInvoker.ExecuteCallback(step.builder.config.CallbackURL,
 			step.builder.result.Success, ctx.labels, step.builder.result.Messages)
-	} else {
-		glog.V(3).Info("Skipping step: invoke callback url")
+		return nil
 	}
 
+	glog.V(3).Info("Skipping step: invoke callback url")
 	return nil
 }
 
--- a/pkg/build/strategies/sti/sti.go
+++ b/pkg/build/strategies/sti/sti.go
@@ -119,7 +119,7 @@ func New(config *api.Config, overrides build.Overrides) (*STI, error) {
 	if len(config.RuntimeImage) > 0 {
 		builder.runtimeDocker, err = dockerpkg.New(config.DockerConfig, config.RuntimeAuthentication)
 		if err != nil {
-			return builder, err
+			return nil, err
 		}
 
 		builder.runtimeInstaller = scripts.NewInstaller(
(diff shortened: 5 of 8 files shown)


pkg/api/types.go
```diff
@@ -320,8 +320,48 @@ type Result struct {
...
+	// ReasonPullBuilderImageFailed is the reason associated with failing to pull
+	// the builder image.
+	ReasonPullBuilderImageFailed StepFailureReason = "PullBuilderImageFailed"
+	// ReasonRuntimeImageFailed is the reason associated with failing to pull the
+	// runtime image.
+	ReasonRuntimeImageFailed StepFailureReason = "PullRuntimeImageFailed"
+	// CommitContainerFailed is the reason associated with failing to commit the
+	// container to the final image.
+	ReasonCommitContainerFailed StepFailureReason = "ContainerCommitFailed"
+
```
rhcarvalho: This is the only new line in the whole block, looks strange, perhaps remove it?

pkg/build/strategies/onbuild/onbuild.go
```diff
@@ -88,7 +89,7 @@ func (builder *OnBuild) SourceTar(config *api.Config) (io.ReadCloser, error) {
 // Build executes the ONBUILD kind of build
 func (builder *OnBuild) Build(config *api.Config) (*api.Result, error) {
 	if config.BlockOnBuild {
-		return nil, fmt.Errorf("builder image uses ONBUILD instructions but ONBUILD is not allowed")
+		return nil, errors.New("builder image uses ONBUILD instructions but ONBUILD is not allowed")
```
rhcarvalho: The commit message says "be pedantic about errors.New", what does that mean here?

pkg/build/strategies/sti/postexecutorstep.go
```diff
@@ -79,20 +81,18 @@ func (step *removePreviousImageStep) execute(ctx *postExecutorStepContext) error
...
-	}
-
-	glog.V(1).Infof("Removing previously-tagged image %s", previousImageID)
-	if err := step.docker.RemoveImage(previousImageID); err != nil {
-		glog.V(0).Infof("error: Unable to remove previous image: %v", err)
+	if previousImageID != "" {
+		glog.V(1).Infof("Removing previously-tagged image %s", previousImageID)
+		if err := step.docker.RemoveImage(previousImageID); err != nil {
+			glog.V(0).Infof("error: Unable to remove previous image: %v", err)
+		}
```
rhcarvalho: This change looks like in the opposite direction. Normally we remove nesting, not add. The previous code seemed fine, return earlier under some condition, do something else in the general case.

pkg/build/strategies/sti/postexecutorstep.go
```diff
@@ -115,18 +115,18 @@ func (step *commitImageStep) execute(ctx *postExecutorStepContext) error {
 	ctx.labels = createLabelsForResultingImage(step.builder, step.docker, step.image)
 
 	// Set the image entrypoint back to its original value on commit, the running
-	// container has "env" as its entrypoint and we don't want to commit that.
+	// container has "env" as its entrypoint and we don't want to commit that
```
rhcarvalho: Why remove the punctuation? In multiple other places we're adding a missing punctuation, it's inconsistent to remove it here.

pkg/build/strategies/sti/postexecutorstep.go
```diff
@@ -115,18 +115,18 @@ func (step *commitImageStep) execute(ctx *postExecutorStepContext) error {
...
 	entrypoint, err := step.docker.GetImageEntrypoint(step.image)
 	if err != nil {
 		return fmt.Errorf("Couldn't get entrypoint of %q image: %v", step.image, err)
 	}
 	// If the image has no explicit entrypoint, set it to an empty array
-	// so we don't default to leaving the entrypoint as "env" upon commit.
+	// so we don't default to leaving the entrypoint as "env" upon commit
```
rhcarvalho: Same note about punctuation.

pkg/util/duration.go
```diff
@@ -0,0 +1,55 @@
...
+	defer updateBuildInfo(buildInfo, failureReason, start)
+	imageID, err := containerCommitFn(dockerClient, containerID, cmd, user, tag, env, entrypoint, label)
+	if err != nil {
+		failureReason = api.ReasonFetchSourceFailed
+	}
+	return buildInfo, imageID, err
+}
+
+// updateBuildInfo updates a api.BuildStepInfo
+// NOTE: i think []api.BuildStepInfo should be a pointer here!?
```
rhcarvalho: Yes, because you are appending to the slice, which may include an allocation.

pkg/build/strategies/onbuild/onbuild.go
bparees: that are we are being pedantic about not using fmt.Errorf when there is no formatting to be done.

pkg/build/strategies/sti/postexecutorstep.go
PI-Victor: you're right, removed them by mistake

pkg/build/strategies/sti/sti.go
```diff
@@ -273,7 +274,9 @@ func (builder *STI) Prepare(config *api.Config) error {
 
 	// fetch sources, for their .s2i/bin might contain s2i scripts
 	if len(config.Source) > 0 {
-		if builder.sourceInfo, err = builder.source.Download(config); err != nil {
+		// NOTE: needs timing
+		builder.sourceInfo, err = duration.DownloadTimer(builder.source.Download, builder.result.BuildInfo, config)
```
bparees: this is not the right way to do this.  just note the start time, run the download, note the current time, subtract, done.  Creating a bunch of wrapper functions to do that isn't adding any value, you're not getting any reuse out of them.

but i don't want any of this duration related content in this PR.  your card for 3.4 is to get the failure reasons in place and i don't want this PR getting held up by discussion about how to handle the duration work which is for 3.5.

PI-Victor: ok, will add just the failure reasons.
because the card (the 3.5 one) says we need the times for assembly and image pull?
https://trello.com/c/KmmYvStu/996-5-separate-out-build-substatus-and-times

pkg/build/strategies/onbuild/onbuild.go
PI-Victor: no, i meant i was being.

pkg/api/types.go
```diff
@@ -320,8 +320,48 @@ type Result struct {
...
+type BuildStepInfo struct {
+	// FailureReason is a camel case reason that is used by the machine to reply
+	// back to the OpenShift builder with information as why any of the steps in
+	// the build, failed.
+	FailureReason StepFailureReason
+
+	// FuncDuration holds the elapsed time for a specific build step operation
+	// (git  clone, assemble container, etc.) in nanoseconds.
+	FuncDuration int64
```
bparees: i guess openshift does not call each s2i step (clone, assemble, commit) separately, but it does call clone explicitly.  If we need s2i to record and report the assemble+commit time then i guess we'll have to do that, but again let's do it in a separate PR after this card is complete.

PI-Victor: ok

pkg/api/types.go
```diff
@@ -320,8 +320,64 @@ type Result struct {
...
+	// ReasonInvalidArtifactsMappingFailed is the reason associated with an
+	// invalid artifacts mapping of files that need to be copied.
+	ReasonInvalidArtifactsMappingFailed StepFailureReason = "InvalidArtifactsMappingFailed"
+
+	// ReasonFailedFileSystemOperation is the reason associated with a failed fs
+	// operation. create, remove directory, copy file, etc.
+	ReasonFailedFileSystemOperation StepFailureReason = "FileSystemOperationFailed"
+
+	// ReasonInstallScriptsFailed
+	ReasonInstallScriptsFailed StepFailureReason = "InstallScriptsFailed"
```
PI-Victor: @bparees how far the rabbit hole should i go? trying to account for all failure steps i think is a lot. on the other hand returning only some failure cases to the user in openshift and nothing for others doesn't help either.
i thought about returning more generic failures, but telling the user "build failed at Execute step" doesn't seem i'm helping him identify anything.

PI-Victor: ok, on the other hand, not that many failure cases left... but still, maybe this is way too much

bparees: Don't worry about completely exhaustive. We can always add more granularity later. 

Image pull failures, script execution failures, and file injection errors seems like it probably could easily the bulk of the scenarios that will matter to users. I'm ok with a generic "failed in s2i, see build log" for the rest. 

Honestly "failed to clone", "failed to assemble", "failed to push" are the critical ones because those are the three most common errors and one is a code problem (assemble) and the other two point to config /auth issues so it's important to give the user immediate hints about the nature of the failure.

PI-Victor:
let's see what blows up so far. [test]

PI-Victor:
just an update, a nil pointer reference is hogging up this PR (because of the implementation not being great), otherwise i think it's close to being wrapped up

PI-Victor:
@csrwng @gabemontero any of guys mind reviewing this? Ben's still out till next week, but it would be nice to move along meanwhile.

gabemontero:
I'll take a peek @PI-Victor , though note I don't have merge powers for the s2i repo.

pkg/build/strategies/layered/layered.go
```diff
@@ -138,24 +145,33 @@ func (builder *Layered) SourceTar(config *api.Config) (io.ReadCloser, error) {
...
 	if config.HasOnBuild && config.BlockOnBuild {
-		return nil, fmt.Errorf("builder image uses ONBUILD instructions but ONBUILD is not allowed")
+		buildResult.BuildInfo = updater.UpdateFailureReason(buildResult.BuildInfo, api.ReasonGenericS2IBuildFailed)
+		return buildResult, fmt.Errorf("builder image uses ONBUILD instructions but ONBUILD is not allowed")
 	}
 
 	if err := builder.CreateDockerfile(config); err != nil {
-		return nil, err
+		// NOTE: check if CreateDockerfile needs to be added to failure cases.
+		buildResult.BuildInfo = updater.UpdateFailureReason(buildResult.BuildInfo, api.ReasonGenericS2IBuildFailed)
```
gabemontero: I'm only seeing CreateDockerfile called with layered and onbuild.  I think it would help to be more specific than generic s2i build failed.   Perhaps even have a unique reason for layered and another for onbuild.

pkg/build/strategies/layered/layered.go
```diff
@@ -138,24 +145,33 @@ func (builder *Layered) SourceTar(config *api.Config) (io.ReadCloser, error) {
...
+		return buildResult, err
 	}
 
 	glog.V(2).Info("Creating application source code image")
 	tarStream, err := builder.SourceTar(config)
 	if err != nil {
-		return nil, err
+		// NOTE: same as above.
```
gabemontero: Same with SourceTar - only layered and onbuild

pkg/build/strategies/layered/layered.go
```diff
@@ -138,24 +145,33 @@ func (builder *Layered) SourceTar(config *api.Config) (io.ReadCloser, error) {
...
+		buildResult.BuildInfo = updater.UpdateFailureReason(buildResult.BuildInfo, api.ReasonGenericS2IBuildFailed)
+		return buildResult, err
 	}
 	defer tarStream.Close()
 
 	namedReference, err := reference.ParseNamed(builder.config.BuilderImage)
 	if err != nil {
-		return nil, err
+		// NOTE: same comment.
```
gabemontero: ParsedNamed is more generic and used more pervasively.  I could see this one staying generic.

pkg/build/strategies/onbuild/onbuild.go
```diff
@@ -49,17 +53,20 @@ func New(config *api.Config, overrides build.Overrides) (*OnBuild, error) {
 		tar:    tar.New(),
 	}
 	// Use STI Prepare() and download the 'run' script optionally.
-	s, err := sti.New(config, overrides)
+	// NOTE: update build status?
```
gabemontero: Not obvious to me exactly what this entails ... please elaborate.  Which struct/field specifically could be updated?

pkg/build/strategies/onbuild/onbuild.go
```diff
@@ -87,28 +95,33 @@ func (builder *OnBuild) SourceTar(config *api.Config) (io.ReadCloser, error) {
 
 // Build executes the ONBUILD kind of build
 func (builder *OnBuild) Build(config *api.Config) (*api.Result, error) {
+	buildResult := &api.Result{}
+
 	if config.BlockOnBuild {
-		return nil, fmt.Errorf("builder image uses ONBUILD instructions but ONBUILD is not allowed")
+		buildResult.BuildInfo = updater.UpdateFailureReason(buildResult.BuildInfo, api.ReasonGenericS2IBuildFailed)
```
gabemontero: In line with my other comment, but perhaps counter to the "you do not need to be completely exhaustive" guidance @bparees gave, but my initial reaction is that if we don't have a specific error, at least a generic "config mismatch" reason would be good.

pkg/build/strategies/onbuild/onbuild.go
```diff
@@ -87,28 +95,33 @@ func (builder *OnBuild) SourceTar(config *api.Config) (io.ReadCloser, error) {
...
+		return buildResult, err
 	}
 
 	// If necessary, copy the STI scripts into application root directory
 	builder.copySTIScripts(config)
 
 	glog.V(2).Info("Creating application Dockerfile")
 	if err := builder.CreateDockerfile(config); err != nil {
-		return nil, err
+		buildResult.BuildInfo = updater.UpdateFailureReason(buildResult.BuildInfo, api.ReasonGenericS2IBuildFailed)
```
gabemontero: Earlier comment (if we decide to go with it) would apply here too.

pkg/build/strategies/onbuild/onbuild.go
```diff
@@ -131,7 +146,8 @@ func (builder *OnBuild) Build(config *api.Config) (*api.Result, error) {
 
 	if len(opts.Name) > 0 {
 		if imageID, err = builder.docker.GetImageID(opts.Name); err != nil {
-			return nil, err
+			buildResult.BuildInfo = updater.UpdateFailureReason(buildResult.BuildInfo, api.ReasonGenericS2IBuildFailed)
```
gabemontero: Perhaps at least a generic "docker interaction error" vs. generic s2i error (and assuming we don't want a "get image id" error).

pkg/build/strategies/sti/sti.go
```diff
@@ -196,17 +208,24 @@ func (builder *STI) Build(config *api.Config) (*api.Result, error) {
 		glog.V(1).Infof("Running %q in %q", api.Assemble, config.Tag)
 	}
 	if err := builder.scripts.Execute(api.Assemble, config.AssembleUser, config); err != nil {
+		builder.result.BuildInfo = updater.UpdateFailureReason(builder.result.BuildInfo, api.ReasonGenericS2IBuildFailed)
```
gabemontero: an error on executing the assemble script is a generic s2i error?

pkg/build/strategies/strategies.go
```diff
@@ -6,27 +6,40 @@ import (
...
-func Strategy(config *api.Config, overrides build.Overrides) (build.Builder, error) {
+func Strategy(config *api.Config, overrides build.Overrides) (build.Builder, []*api.BuildStepInfo, error) {
+	var builder build.Builder
+	var buildInfo []*api.BuildStepInfo
+	var err error
+
 	image, err := docker.GetBuilderImage(config)
 	if err != nil {
-		return nil, err
+		buildInfo = updater.UpdateFailureReason(buildInfo, api.ReasonGenericS2IBuildFailed)
```
gabemontero: at least a generic "docker interaction" sort of error if "failed getting builder image" is not appropriate

PI-Victor:
@gabemontero you helped me identify some problems i had, i addressed all of your comments except a few about more specific cases, i can do that as a follow up PR, instead of wasting more time figuring out where else i should add them, and what else fits that specific pattern.
thank you very much!

PI-Victor:
@bparees PTAL when you're back.

pkg/build/strategies/layered/layered.go
```diff
@@ -41,10 +42,13 @@ type Layered struct {
 }
 
 // New creates a Layered builder.
-func New(config *api.Config, scripts build.ScriptsHandler, overrides build.Overrides) (*Layered, error) {
+func New(config *api.Config, scripts build.ScriptsHandler, overrides build.Overrides) (*Layered, []*api.BuildStepInfo, error) {
```
rhcarvalho: I find it strange that, in s2i itself, all calls to `New` and `Strategy` and `GetStrategy` ignore the second return value...

IIUC, all that `[]*api.BuildStepInfo` conveys in a list of errors.
If that's the case, wouldn't it be a cleaner design to keep the function signature as it was, and make `[]*api.BuildStepInfo` implement the `error` interface?

If what I'm saying makes sense, sorry if it doesn't, we could store the original error within the "FailureReason", thus this PR would be about wrapping errors allowing for easy identification of what step went wrong and why.

WDYT?

PI-Victor: i understand what you mean. though being clear about it, this PR doesn't add anything useful for s2i itself (or it might be of use in the future, or a follow up PR), but for origin. in openshift, i don't want to compare errors, i want to compare constants when i try to find out what went wrong. Also, whenever we add the timing of functions, it should be easier to transport the information up the stack and into openshift.
openshift uses the Build() interface, that's why the second value is ignored in s2i. the build() returns a *result, that's where i need the info.

gabemontero:
Cool deal!

pkg/api/types.go
```diff
@@ -320,8 +320,81 @@ type Result struct {
...
+	// install scripts in the builder image.
+	ReasonInstallScriptsFailed StepFailureReason = "InstallScriptsFailed"
+
+	// ReasonGenericS2IBuildFailed is the reason associated with a broad range of
+	// failure.
+	ReasonGenericS2IBuildFailed StepFailureReason = "GenericS2IBuildFailed"
+
+	// ReasonUnmentDependenciesFailed is the failure reason associated with a builder
+	// image that doesn't contain required dependencies for building the app.
+	ReasonUnmentDependenciesFailed StepFailureReason = "UnmetBuilderImageDependenciesFailed"
```
bparees: 1) should be "unmet" 
2) it's not dependencies for building the app, it's really s2i dependencies, right?  (tar, sh)

pkg/build/strategies/layered/layered.go
```diff
@@ -70,6 +74,9 @@ func getDestination(config *api.Config) string {
 // accessible and non-empty directory.
 func checkValidDirWithContents(name string) bool {
 	items, err := ioutil.ReadDir(name)
+	// NOTE: obfuscated and unhandled error
+	// what if we don't have read rights to the repo?
+	// this should be replaced with a generic print.
```
bparees: that's the callers problem to decide how to deal with...  even this warning probably doesn't belong in this function.

pkg/build/strategies/onbuild/onbuild.go
bparees: yeah this ONBUILD case seems like one where it actually is worth having a specific reason code (contrary to my previous advice about being generic).

pkg/build/strategies/onbuild/onbuild.go
bparees: this one is ok as is for me.

For the bulk of these errors we can't give them enough information anyway, we need to log the error and the user needs to look in the build logs to understand what went wrong.

the case we're really trying to solve are the "you fat fingered your git repo url" type cases where they shouldn't have to dig through a log to see the git clone operation failed.

pkg/build/strategies/sti/sti.go
```diff
@@ -140,39 +146,45 @@ func New(config *api.Config, overrides build.Overrides) (*STI, error) {
...
 	}
-
 	builder.garbage = build.NewDefaultCleaner(builder.fs, builder.docker)
-	builder.layered, err = layered.New(config, builder, overrides)
+	builder.layered, buildInfo, err = layered.New(config, builder, overrides)
+
+	// NOTE: shouldn't the above error be properly handled? if the layered build
+	// failed, then it means the build failed. the buildInfo is already updated
+	// with failure reasons so there's nothing more to be done here but to return
+	// it up the stack
```
bparees: yes it should.  please handle it.

pkg/build/strategies/sti/sti.go
```diff
@@ -196,17 +208,24 @@ func (builder *STI) Build(config *api.Config) (*api.Result, error) {
...
 		case errors.ContainerError:
 			if !isMissingRequirements(e.Output) {
-				return nil, err
+				builder.result.BuildInfo = updater.UpdateFailureReason(builder.result.BuildInfo, api.ReasonUnmentDependenciesFailed)
+				return builder.result, err
 			}
 			glog.V(1).Info("Image is missing basic requirements (sh or tar), layered build will be performed")
+			// NOTE: separation of concerns, if we fail here, we shouldn't try
+			// another strategy here, but somewhere up the stack. In other words, a
+			// build() shouldn't call another build().
```
bparees: debatable.  what you've got here is a uber-build strategy that understands how to delegate down to other build strategies if it can't do the build itself, and the uber-build understands the overall current state of the system (has already cloned the source, etc)

pkg/build/strategies/sti/sti.go
```diff
@@ -252,35 +275,45 @@ func (builder *STI) Prepare(config *api.Config) error {
...
-				return fmt.Errorf("Invalid runtime artifacts mapping: %q -> %q: destination must be a relative path", volumeSpec.Source, volumeSpec.Destination)
+			var volumeErr error
+
+			switch {
+			case !path.IsAbs(volumeSpec.Source):
+				volumeErr = fmt.Errorf("Invalid runtime artifacts mapping: %q -> %q: source must be an absolute path", volumeSpec.Source, volumeSpec.Destination)
+			case path.IsAbs(volumeSpec.Destination):
+				volumeErr = fmt.Errorf("Invalid runtime artifacts mapping: %q -> %q: destination must be a relative path", volumeSpec.Source, volumeSpec.Destination)
+			case strings.HasPrefix(volumeSpec.Destination, ".."):
+				volumeErr = fmt.Errorf("Invalid runtime artifacts mapping: %q -> %q: destination cannot start with '..'", volumeSpec.Source, volumeSpec.Destination)
```
bparees: IsAbs already covers this possibility, no?  it's not absolute path if it starts with ".."

pkg/build/strategies/sti/sti.go
```diff
@@ -289,6 +322,7 @@ func (builder *STI) Prepare(config *api.Config) error {
 
 	if len(config.RuntimeImage) > 0 && builder.runtimeInstaller != nil {
 		optionalRuntime := builder.runtimeInstaller.InstallOptional(builder.optionalRuntimeScripts, config.WorkingDir)
+		// which err are we checking here, the check is already done above?
```
bparees: seems like this should be removed.

pkg/build/strategies/sti/sti.go
```diff
@@ -304,21 +338,25 @@ func (builder *STI) Prepare(config *api.Config) error {
...
+
+		builder.externalScripts[r.Script] = r.Downloaded
+		builder.installedScripts[r.Script] = r.Installed
+		builder.scriptsURL[r.Script] = r.URL
 	}
 
-	// see if there is a .s2iignore file, and if so, read in the patterns an then search and delete on
+	// see if there is a .s2iignore file, and if so, read in the patterns an then
+	// search and delete on
+	// NOTE: antipattern?
```
bparees: 1) if you want to raise questions about the code unrelated to the changes you're making, this probably isn't the best way to do it since it's creating a distraction from the primary focus of your PR
2) can you be more specific about your concern?

pkg/util/updater/updater.go
```diff
@@ -0,0 +1,24 @@
+package updater
+
+import (
+	"github.com/openshift/source-to-image/pkg/api"
+)
+
+// UpdateFailureReason updates the build status with the failure reason.
+func UpdateFailureReason(buildInfo []*api.BuildStepInfo, reason api.StepFailureReason) []*api.BuildStepInfo {
```
bparees: Given that you're appending, calling this AppendFailureReason makes more sense.

pkg/api/types.go
```diff
@@ -320,8 +320,81 @@ type Result struct {
 
 	// ImageID describes resulting image ID.
 	ImageID string
+
+	// BuildInfo holds information about the result of a build.
+	BuildInfo []*BuildStepInfo
```
bparees: Are we ever actually going to have more than one failure reason returned?

pkg/util/updater/updater.go
```diff
@@ -0,0 +1,24 @@
...
+	return append(buildInfo, &api.BuildStepInfo{
+		FailureReason: reason,
+	})
+}
+
+// InstantiateBuildResult is used to instantiate builder.result to avoid a nil
+// pointer reference while trying to transmit build information up the stack
+// through *builder.result and also to avoid mutating the *result in case it's
+// not empty.
+func InstantiateBuildResult(result *api.Result) *api.Result {
```
bparees: this function name is confusing since it doesn't always instantiate anything.

pkg/api/types.go
PI-Victor: it's not just failure reasons, but a more generic build information. it's one of the reasons why i didn't name it buildfailurereasons or something similar. we could attach different data to it, such as the timing of functions (sorry to bring it up again :)) or whatever else we need.

pkg/build/strategies/sti/sti.go
PI-Victor: not sure what you mean?
the destination path can't be absolute e.g. '/path/absolute/workdir'
or start with '..' e.g.: '../path/absolute/workdir'
this looks fine to me, am i missing something?

PI-Victor: oh, hold on. yeah, it's not an absolute path i see what you mean, but you're not allowed to specify mappings that aren't relative to the current workdir. '../path/workdir' is relative, but it's not a valid mapping because it points to the directory above your workdir.

PI-Victor:
@bparees this is ready for the final review. i've simplified everything and remove the extra package with the helper functions in favor of a bit of code duplication. i had one nit with surfacing failure reasons that i couldn't get to.
i also fixed an existing bug, that we have that might cause s2i to panic in certain cases.

pkg/build/strategies/sti/sti.go
PI-Victor: i've reproduced this because i'm confused. https://play.golang.org/p/XkZR8jcOLW 
a switch statement continues till it hits a true case or a default if it doesn't? 
by the way, the first check is against volumeSpec.Source and the next two are for volumeSpec.Destination in case you think the first is against the same?

bparees: a switch statement (in go) stops on the first match, it does not "fallthrough"

anyway no i was looking at the last 2 cases, but I thought it was trying to report an error for non-absolute paths.  I see now that it's trying to report errors for absolute paths, so it's fine.

bparees:
lgtm, squash and i'll merge.

PI-Victor:
@bparees squashed.

pkg/api/types.go
```diff
@@ -325,8 +325,86 @@ type Result struct {
...
+	// BuildInfo holds information about the result of a build.
+	BuildInfo BuildInfo
+}
+
+// BuildInfo holds information about a particular step in the build process.
+type BuildInfo struct {
+	// FailureReason is a camel case reason that is used by the machine to reply
+	// back to the OpenShift builder with information why any of the steps in the
+	// build, failed.
+	FailureReason StepFailureReason
```
bparees: @rhcarvalho perfect is the enemy of the good. this is fine.

PI-Victor:
@bparees PTAL. i've simplified even further, addressed your comments and also squashed.

pkg/build/strategies/layered/layered.go
```diff
@@ -48,13 +48,14 @@ func New(config *api.Config, scripts build.ScriptsHandler, overrides build.Overr
...
-	return &Layered{
+	builder := &Layered{
 		docker:  d,
 		config:  config,
 		fs:      util.NewFileSystem(),
 		tar:     tarHandler,
 		scripts: scripts,
-	}, nil
+	}
+	return builder, nil
```
bparees: why this change?

PI-Victor: hmmm... a left over. fixed it!

bparees:
[merge]

PR #594, opened 2016-09-27
bz1373281 allow >2 slashes in docker URI

jim-minter:
@bparees ptal

### DIFF
--- a/pkg/build/strategies/layered/layered.go
+++ b/pkg/build/strategies/layered/layered.go
@@ -12,6 +12,7 @@ import (
 	"regexp"
 	"time"
 
+	"github.com/docker/distribution/reference"
 	"github.com/openshift/source-to-image/pkg/api"
 	"github.com/openshift/source-to-image/pkg/build"
 	"github.com/openshift/source-to-image/pkg/docker"
@@ -152,26 +153,11 @@ func (builder *Layered) Build(config *api.Config) (*api.Result, error) {
 	}
 	defer tarStream.Close()
 
-	dockerImageReference, err := docker.ParseImageReference(builder.config.BuilderImage)
+	namedReference, err := reference.ParseNamed(builder.config.BuilderImage)
 	if err != nil {
 		return nil, err
 	}
-	// if we fall down this path via oc new-app, the builder image will be a docker image ref ending
-	// with a @<hex image id> instead of a tag; simply appending the time stamp to the end of a
-	// hex image id ref is not kosher with the docker API; so we remove the ID piece, and then
-	// construct the new image name
-	var newBuilderImage string
-	if len(dockerImageReference.ID) == 0 {
-		newBuilderImage = fmt.Sprintf("%s-%d", builder.config.BuilderImage, time.Now().UnixNano())
-	} else {
-		if len(dockerImageReference.Registry) > 0 {
-			newBuilderImage = fmt.Sprintf("%s/", dockerImageReference.Registry)
-		}
-		if len(dockerImageReference.Namespace) > 0 {
-			newBuilderImage = fmt.Sprintf("%s%s/", newBuilderImage, dockerImageReference.Namespace)
-		}
-		newBuilderImage = fmt.Sprintf("%s%s:s2i-layered-%d", newBuilderImage, dockerImageReference.Name, time.Now().UnixNano())
-	}
+	newBuilderImage := fmt.Sprintf("%s:s2i-layered-%d", namedReference.Name(), time.Now().UnixNano())
 
 	outReader, outWriter := io.Pipe()
 	defer outReader.Close()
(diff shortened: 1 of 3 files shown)


pkg/docker/util.go
```diff
@@ -187,46 +186,27 @@ func ParseImageReference(spec string) (ImageReference, error) {
...
-			return ref, fmt.Errorf("the docker pull spec %q must be two or three segments separated by slashes", spec)
-		}
-		ref.Name = repoParts[0]
-		ref.Tag = tag
-		ref.ID = id
-		return ref, nil
-	default:
-		return ref, fmt.Errorf("the docker pull spec %q must be two or three segments separated by slashes", spec)
+		// registry(/dir)+/name, we suppose
+		ref.Registry = nameParts[0]
```
bparees: how do we safely make this assumption?  should we check it for : and . like above?
(are : or . valid characters in a docker image path segment?)

jim-minter: I don't know how to make the assumption more safe.  ":" is not a valid character in a docker image path segment but is used for hostnames and tags.  Unfortunately "." is a valid character in a docker image path segment; nevertheless this heuristic is used today in the openshift code and it seemed plausible to copy it.

There is an element of guesswork to this.  The previous version of the code made the same assumption here as this version does.

Technically, I still think there's an argument to using "github.com/docker/distribution/reference" as the canonical parser so we match docker behaviour - principle of least surprise

jim-minter:
@bparees I've ended up testing the registry in the 2x'/' case as you suggested above - ptal

pkg/docker/util.go
```diff
@@ -187,46 +186,29 @@ func ParseImageReference(spec string) (ImageReference, error) {
...
-			ref.Registry = repoParts[0]
-			ref.Namespace = "library"
-			ref.Name = repoParts[1]
-			ref.Tag = tag
-			ref.ID = id
-			return ref, nil
+		if strings.ContainsAny(nameParts[0], ":.") || nameParts[0] == "localhost" {
+			// registry/name, we suppose
+			ref.Registry = nameParts[0]
+			ref.Name = "library/" + nameParts[1]
```
bparees: what's the else case?

bparees: ah, i see, ref.name will stay as it came out of parseRepoTag.

bparees: but don't we still need to add "library" in the case where "ref.Name" is something like "mysql"?

pkg/docker/util.go
```diff
@@ -187,46 +186,29 @@ func ParseImageReference(spec string) (ImageReference, error) {
...
-		ref.ID = id
-		return ref, nil
-	case 1:
-		// name
-		if len(repoParts[0]) == 0 {
-			return ref, fmt.Errorf("the docker pull spec %q must be two or three segments separated by slashes", spec)
+		if strings.ContainsAny(nameParts[0], ":.") || nameParts[0] == "localhost" {
+			// registry(/dir)+/name, we suppose
+			ref.Registry = nameParts[0]
+			ref.Name = strings.Join(nameParts[1:], "/")
```
bparees: a 1 segment name that's just a registry can't be valid, can it?

jim-minter:
@bparees ok, here's a different attempt, which is actually to remove the ParseImageReference function entirely and use the docker registry functions...

pkg/build/strategies/layered/layered.go
```diff
@@ -161,16 +162,10 @@ func (builder *Layered) Build(config *api.Config) (*api.Result, error) {
...
+		newBuilderImage = fmt.Sprintf("%s:s2i-layered-%d", namedReference.Name(), time.Now().UnixNano())
 	} else {
-		if len(dockerImageReference.Registry) > 0 {
-			newBuilderImage = fmt.Sprintf("%s/", dockerImageReference.Registry)
-		}
-		if len(dockerImageReference.Namespace) > 0 {
-			newBuilderImage = fmt.Sprintf("%s%s/", newBuilderImage, dockerImageReference.Namespace)
-		}
-		newBuilderImage = fmt.Sprintf("%s%s:s2i-layered-%d", newBuilderImage, dockerImageReference.Name, time.Now().UnixNano())
+		newBuilderImage = fmt.Sprintf("%s-%d", builder.config.BuilderImage, time.Now().UnixNano())
```
bparees: can you see a reason not to include "s2i-layered" in the name here like we do in the other path?

bparees: is there a good reason not to just use namedReference.Name() as the base of the new image name in all cases, regardless of whether it's a "digest" or not?

jim-minter: I don't think so - updated.

bparees:
squash and lgtm
[test]

jim-minter:
Squashed!

bparees:
[merge]

PR #596, opened 2016-09-28
update godeps and use docker-engine for inspect image calls

bparees:


bparees:
[merge]

bparees:
[test]

pkg/docker/docker.go
```diff
@@ -318,7 +331,7 @@ func getDefaultContext(timeout time.Duration) (context.Context, context.CancelFu
 // GetImageWorkdir returns the WORKDIR property for the given image name.
 // When the WORKDIR is not set or empty, return "/" instead.
 func (d *stiDocker) GetImageWorkdir(name string) (string, error) {
-	resp, err := d.kubeDockerClient.InspectImage(name)
+	resp, err := d.InspectImage(name)
```
bparees: the kubeclient inspectimage behavior changed significantly (which is why this pull includes its own implementation of inspectimage now).  as an aside, in talking to @smarterclayton about it, he thinks we probably should not have migrated to the kubeclient interfaces, unfortunately.  So even assuming I can get this to a working state as is, i think in the near future we're going to want to revisit the s2i changes to either:

1) go back to fsouza/go-dockerclient
2) move off kubeclient and only depend on the docker-engine apis

because we're going to be continuously vulnerable to kubeclient changing out from underneath us and being unable to bump the s2i code into origin as a result (which is what happened here).

it's fairly urgent since @jim-minter and @PI-Victor both have pending changes to s2i that we need to get into origin, and that can't happen until we get s2i back to a state where it can be bumped in origin.

gabemontero: Gotcha - for what it is worth, I think 2) is the better long term solution, and at first blush think it is less painful in the short term as well.

gabemontero:
The "use docker-engine for inspect image " commit LGTM

pkg/docker/docker.go
bparees: agreed

gabemontero:
On the godeps stuff, a follow up for a comment I managed to mention in scrum:  the s2i docker version.

Note this godep entry:
```
		{
			"ImportPath": "github.com/docker/docker/pkg/jsonmessage",
			"Comment": "v1.4.1-4831-g0f5c9d3",
			"Rev": "0f5c9d301b9b1cca66b3ea0f9dec3b5317d3686d"
		},
```

1.4.1 is old.  Now, when I did the move off of fsouza, origin happened to be at 1.4.1 as well.

However, I just looked now, and it appears they upgraded with one of the recent rebases (perhaps it is somehow related to drastic changes you saw in the k8s wrapper of the engine-api stuff):

```
		{
			"ImportPath": "github.com/docker/docker/builder/dockerfile/command",
			"Comment": "v1.11.2",
			"Rev": "b9f10c951893f9a00865890a5232e85d770c1087"
		},

```

I certainly would understand if it is too much to take on with this pull, but in addition to probably moving off of k8s and solely to engine-api, a bump in docker version would be a good thing to target (even if not required to move solely to engine-api).

The docker version bump would also alllow us to leverage code like [this](https://github.com/docker/docker/blob/v1.7.0/cliconfig/config.go#L57) to process auth config stored in a file, and throw away the analogous code we copied from fsouza for that.

bparees:
> I certainly would understand if it is too much to take on with this pull, but in addition to probably moving off of k8s and solely to engine-api, a bump in docker version would be a good thing to target (even if not required to move solely to engine-api).

yeah, definitely not something to take on in this pull unfortunately, i'm just going to be happy if i can get this landed as is (i've spent the last 3 days on it).  but agreed that as part of reworking this in the future, we should move up.

bparees:
[merge]

PR #600, opened 2016-09-30
Add gitignore on create

luiscoms:
The build script neds a `git` repository, so the test script run [`git init`](https://github.com/luiscoms/source-to-image/blob/master/pkg/create/templates/test.go#L60)
But when an error occurs on build the script does not clean the `.git/` folder correctly.
To avoid git errors I just add a `.gitignore` file on test folder.


bparees:
[merge]

PR #601, opened 2016-10-03
Add support for custom image labels

mmilata:
Needed for https://trello.com/c/HB9IwGj5/925-5-add-arbitrary-labels-to-docker-images

Work in progress because we probably need to discuss how the API should look like in order to support autogenerated values (e.g. SCM commit hash) in the future.

### DIFF
--- a/pkg/api/describe/describer.go
+++ b/pkg/api/describe/describer.go
@@ -36,6 +36,7 @@ func DescribeConfig(config *api.Config) string {
 		if len(config.EnvironmentFile) > 0 {
 			fmt.Fprintf(out, "Environment File:\t%s\n", config.EnvironmentFile)
 		}
+		printLabels(out, config.Labels)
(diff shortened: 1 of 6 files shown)


bparees:
looks reasonable to me.  i don't think s2i should be concerned w/ the autogeneration aspect, we should handle that in origin and just set explicit label values in the s2i config object when invoking s2i.

mmilata:
@bparees I was thinking that some of the generated labels may have to be generated on the s2i side because openshift doesn't have the cloned repository. Now I see that cloning is done in openshift code so it shouldn't be a problem to access all the information about a repository.

Should I extend this PR with a means to disable the label generation as currently implemented in s2i or is it out of scope of the trello card?

bparees:
Out of scope. We need to leave that in place anyway for compatibility for
now.

Ben Parees | OpenShift

mfojtik:
LGTM, good job with adding more tests :-)

mfojtik:
[merge]

PR #602, opened 2016-10-04
Update nginx example

### LINKED ISSUE #508: nginx-app does not follow the standard directory structure for a s2i application
https://github.com/openshift/source-to-image/tree/master/examples/nginx-app does not follow the standard directory structure for a s2i application i.e. the structure gets created with `s2i create <image name> <destination directory>` 

```
── Dockerfile
├── Makefile
├── .s2i
│   └── bin
│       ├── assemble
│       ├── run
│       ├── save-artifacts
│       └── usage
└── test
    ├── run
    └── test-app
```

luiscoms:
To fix #508 I just update nginx example app to default `s2i create` structure


luiscoms:
It could be run again since I do not change any `Go` file

PI-Victor:
yep, it can. [test]
will give this a review tomorrow.

PI-Victor:
re[test]

examples/nginx-app/Dockerfile
```diff
@@ -1,35 +1,50 @@
 # nginx-centos7
 # Here you can use whatever image base is relevant for your application.
-FROM centos:centos7
+FROM openshift/base-centos7
```
PI-Victor: in line with my comment below, the base image can remain the same.

examples/nginx-app/Dockerfile
```diff
@@ -1,35 +1,50 @@
...
+         s|80|8080|;' /etc/nginx/nginx.conf
+
+# Copy the S2I scripts to /usr/libexec/s2i, since openshift/base-centos7 image sets io.openshift.s2i.scripts-url label that way, or update that label
+COPY ./.s2i/bin/ $STI_SCRIPTS_PATH
+
+# Drop the root user and make the content of /opt/app-root owned by user nginx
+RUN chown -R nginx:nginx /opt/app-root
+
+# Run container as nginx user
+USER nginx
```
PI-Victor: everything that is below this point should actually take into account this comment: https://github.com/openshift/source-to-image/pull/371#issuecomment-160782599

the Makefile, Run as user X, etc these weren't things that we wanted to show the user at first.
The idea of the tutorial was to make it as easy as possible to get up and running with s2i with the bare minimum. I know you spent some time to get this to work, but i don't think it's what we're looking for.
@bparees what do you think, should we accept this customization for the tutorial?

examples/nginx-app/Dockerfile
luiscoms: I changed the image according `s2i create` command

examples/nginx-app/Dockerfile
luiscoms: I follow your recomendation on [Issue 508](https://github.com/openshift/source-to-image/issues/508#issuecomment-225015897)

PI-Victor:
@luiscoms i really appreciate your work and i'm sorry but i don't think this is the intended purpose of the tutorial.
Based on Ben's comment and also on the fact that i've done this from scratch and raised the same concerns while doing it, i'm closing this and also the issue that's related to this, since it's not really an issue.
As stated, the tutorial is fine the way it is, and it's meant to be that way.

Issue #604, opened 2016-10-07
Using a Docker TCP Socket with TLS fails


jasinner:
After upgrading from 1.0.9 to 1.1.2 s2i stopped working with Docker TCP Socket with TLS. I got it working on 1.0.9 by following the instructions here:  https://jasonshepherd.net/2016/09/21/using-openshift-container-platform-s2i-with-a-tcp-docker-daemon/

However after upgrading, the introduction of k8 dockertools library prevents it working again.

bparees:
I suspect this is a dupe of https://github.com/openshift/source-to-image/issues/603, can you tell us more about your environment?

jasinner:
Jasons-MacBook-Pro-3:jboss-dependency-viewer jasonshepherd$ s2i version
s2i v1.1.2
Jasons-MacBook-Pro-3:jboss-dependency-viewer jasonshepherd$ env | grep DOCKER
DOCKER_HOST=tcp://192.168.0.1:2376
DOCKER_TLS_VERIFY=1
DOCKER_CERT_PATH=/Users/jasonshepherd/.docker/
Jasons-MacBook-Pro-3:jboss-dependency-viewer jasonshepherd$ s2i build --loglevel=5 . jboss-eap-7/eap70-openshift jboss-dependecy-viewer
I1010 10:52:06.812003   97743 main.go:68] Running S2I version "v1.1.2"
I1010 10:52:06.816216   97743 docker.go:306] Connecting to docker on tcp://192.168.0.1:2376
I1010 10:52:06.836192   97743 docker.go:559] error inspecting image jboss-eap-7/eap70-openshift:latest: Get http://192.168.0.1:2376/images/jboss-eap-7/eap70-openshift:latest/json?size=1: malformed HTTP response "\x15\x03\x01\x00\x02\x02".
* Are you trying to connect to a TLS-enabled daemon without TLS?
I1010 10:52:06.852941   97743 docker.go:517] error inspecting image jboss-eap-7/eap70-openshift:latest: Get http://192.168.0.1:2376/images/jboss-eap-7/eap70-openshift:latest/json?size=1: malformed HTTP response "\x15\x03\x01\x00\x02\x02".
* Are you trying to connect to a TLS-enabled daemon without TLS?
I1010 10:52:06.853117   97743 main.go:147] 
Error describing image:		unable to get metadata for jboss-eap-7/eap70-openshift:latest
Source:				.
Output Image Tag:		jboss-dependecy-viewer
Environment:			
[...]

PR #605, opened 2016-10-07
Using a Docker TCP Socket with TLS fails

jasinner:
Fixes https://github.com/openshift/source-to-image/issues/604

### DIFF
--- a/cmd/s2i/main.go
+++ b/cmd/s2i/main.go
@@ -146,10 +146,6 @@ $ s2i build . centos/ruby-22-centos7 hello-world-app
 
 			glog.V(2).Infof("\n%s\n", describe.DescribeConfig(cfg))
 
-			if !docker.IsReachable(cfg) {
-				glog.Fatalf("Unable to connect to Docker daemon. Please set the DOCKER_HOST or make sure the Docker socket %q exists", cfg.DockerConfig.Endpoint)
-			}
-
 			builder, err := strategies.GetStrategy(cfg)
 			checkErr(err)
 			result, err := builder.Build(cfg)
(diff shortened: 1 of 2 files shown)


gabemontero:
[test]

cmd/s2i/main.go
```diff
@@ -146,10 +146,6 @@ $ s2i build . centos/ruby-22-centos7 hello-world-app
 
 			glog.V(2).Infof("\n%s\n", describe.DescribeConfig(cfg))
 
-			if !docker.IsReachable(cfg) {
```
gabemontero: @jasinner - thanks for the PR - minimally it is instructive since it lets us know what you did to work around the issue.

As to the precise fix, IMO, given the processing of those environment variables [here](https://github.com/openshift/source-to-image/blob/master/pkg/docker/util.go#L344-L356), I'm curious why the set up [here](https://github.com/openshift/source-to-image/blob/master/pkg/docker/docker.go#L287-L325) did not provide favorable results.

I think the s2i team should use https://github.com/openshift/source-to-image/issues/604 to investigate that and come up with a fix that allows `IsReachable` to succeed with TLS enabled.

@bparees thoughts?

Also, it is possible that the back off of k8s in favor of direct engine-api access could have some bearing here.

pkg/docker/docker.go
```diff
@@ -308,7 +308,7 @@ func New(config *api.DockerConfig, auth api.AuthConfig) (Docker, error) {
 	if err != nil {
 		return nil, err
 	}
-	k8sDocker := dockertools.ConnectToDockerOrDie(config.Endpoint, 0)
```
gabemontero: yeah, again, the backing off of k8s minimally has some bearing, but not passing any endpoint and always relying on the ENV vars does not seem correct.

bparees: yeah i think we need to solve why the default logic isn't working, not remove it entirely.  but @jim-minter fyi.

jim-minter:
Closing in favour of #612

Issue #603, opened 2016-10-05
s2i v1.1.2 reports malformed HTTP response (Docker Toolbox in Windows 10)


rjstone:
On Windows 10 running Docker Toolbox. (uses virtualbox, no Hyper-V available)

```
I1005 02:21:55.886941    8336 docker.go:517] error inspecting image centos/python-35-centos7:latest: Get http://192.168.99.100:2376/images/centos/python-35-centos7:latest/json?size=1: malformed HTTP response "\x15\x03\x01\x00\x02\x02".
```

It's correctly picking up the docker vm from the env as --url="tcp://192.168.99.100:2376"

docker command works just fine and uses this ip:port

**Versions:**  s2i v1.1.2 docker client: 1.12.0 docker server: 1.12.1 API version: 1.24

**Everything works fine if I use s2i v1.1.1** so it seems to be a problem with windows s2i v1.1.2

bparees:
looks like something is attempting an http request against an https endpoint.

rjstone:
I tried variations on the url tcp://192.168.99.100:2376 such as https://192.168.99.100:2376 but it didn't make any difference.

--loglevel=5 revealed it was reading the credential files. Here's the full output:

```
$ s2i-v1-1-2.exe build git://github.com/openshift/ruby-hello-world centos/ruby-22-centos7 hello-world-app --loglevel=5
I1005 12:23:14.699692   19096 main.go:68] Running S2I version "v1.1.2"
I1005 12:23:14.704695   19096 util.go:57] Getting docker credentials for centos/ruby-22-centos7
I1005 12:23:14.704695   19096 util.go:69] Using  credentials for pulling centos/ruby-22-centos7
I1005 12:23:14.716203   19096 docker.go:306] Connecting to docker on tcp://192.168.99.100:2376
I1005 12:23:14.722708   19096 docker.go:559] error inspecting image centos/ruby-22-centos7:latest: Get http://192.168.99.100:2376/images/centos/ruby-22-centos7:latest/json?size=1: malformed HTTP response "\x15\x03\x01\x00\x02\x02".
* Are you trying to connect to a TLS-enabled daemon without TLS?
I1005 12:23:14.748226   19096 docker.go:517] error inspecting image centos/ruby-22-centos7:latest: Get http://192.168.99.100:2376/images/centos/ruby-22-centos7:latest/json?size=1: malformed HTTP response "\x15\x03\x01\x00\x02\x02".
[...]

PR #607, opened 2016-10-10
Remove k8s client

jim-minter:
These patches should remove use of the k8s client from s2i in preparation for work on #603


pkg/docker/docker.go
```diff
@@ -581,20 +518,56 @@ func (d *stiDocker) CheckImage(name string) (*api.Image, error) {
...
+
+	// RegistryAuth is the base64 encoded credentials for the registry
+	base64Auth, err := base64EncodeAuth(d.pullAuth)
 	if err != nil {
 		return nil, errors.NewPullImageError(name, err)
 	}
-	resp, err := d.InspectImage(name)
+	opts := dockertypes.ImagePullOptions{RegistryAuth: base64Auth}
+	ctx := context.Background() // no timeout
+	resp, err := d.client.ImagePull(ctx, name, opts)
```
gabemontero: since opts isn't used elsewhere, just supply `dockertypes.ImagePullOptions(...)` as the param to `d.client.ImagePull`

same with `ctx`

pkg/docker/docker.go
```diff
@@ -618,12 +591,12 @@ func updateImageWithInspect(image *api.Image, inspect *dockertypes.ImageInspect)
...
 // RemoveContainer removes a container and its associated volumes.
 func (d *stiDocker) RemoveContainer(id string) error {
+	ctx := context.Background() // no timeout
 	opts := dockertypes.ContainerRemoveOptions{
 		RemoveVolumes: true,
 		Force:         true,
 	}
-	return d.kubeDockerClient.RemoveContainer(id, opts)
-
+	return d.client.ContainerRemove(ctx, id, opts)
```
gabemontero: similarly, lose `ctx` and `opts` in favor of passing what is assigned to them directly into the method call

pkg/docker/docker.go
```diff
@@ -581,20 +518,56 @@ func (d *stiDocker) CheckImage(name string) (*api.Image, error) {
...
-	if resp != nil {
+	defer resp.Close()
+	reporter := newProgressReporter(name, func() {})
+	reporter.start()
+	defer reporter.stop()
+	decoder := json.NewDecoder(resp)
+	for {
+		var msg dockermessage.JSONMessage
+		err = decoder.Decode(&msg)
+		if err == io.EOF {
```
gabemontero: we've had issues before io.EOF errors get transformed into other types of `EOF` and checks like this don't fire; hence, in this case, we'd end up in a loop that does not exit.

Create a timer (2 minutes per what we do with inspect image) and leverage select on channels so that we either a) timeout after our typical docker interaction timeout or b) break on the EOF

There are examples in other parts of the code of this pattern. (UPDATE: never mind, I see you use the patter in your new progress indicator code  )

Possibly setting the ctx with 2 minutes vs. no timeout might be sufficient as well, but the infinite for loop with no timeout based short circuit still scares me.  

I at least would like to see the protection in both places, but if you disagree certainly bring in others to chime in.

pkg/docker/docker.go
```diff
@@ -581,20 +518,56 @@ func (d *stiDocker) CheckImage(name string) (*api.Image, error) {
...
-	err := d.kubeDockerClient.PullImage(name, d.pullAuth, dockertypes.ImagePullOptions{})
+
+	// RegistryAuth is the base64 encoded credentials for the registry
+	base64Auth, err := base64EncodeAuth(d.pullAuth)
 	if err != nil {
 		return nil, errors.NewPullImageError(name, err)
 	}
-	resp, err := d.InspectImage(name)
+	opts := dockertypes.ImagePullOptions{RegistryAuth: base64Auth}
+	ctx := context.Background() // no timeout
```
gabemontero: i think this should be the 2 minute value like we did on inspect image (which essentially can do a pull image under the covers) vs. no timeout

pkg/docker/docker.go
```diff
@@ -1122,7 +1088,8 @@ func (d *stiDocker) CommitContainer(opts CommitContainerOptions) (string, error)
 
 // RemoveImage removes the image with specified ID
 func (d *stiDocker) RemoveImage(imageID string) error {
-	_, err := d.kubeDockerClient.RemoveImage(imageID, dockertypes.ImageRemoveOptions{})
+	ctx := context.Background() // no timeout
```
gabemontero: let's use the default timeout of 1 minute

pkg/docker/docker.go
```diff
@@ -1143,8 +1110,7 @@ func (d *stiDocker) BuildImage(opts BuildImageOptions) error {
 		dockerOpts.CPUQuota = opts.CGroupLimits.CPUQuota
 	}
 	glog.V(2).Infof("Building container using config: %+v", dockerOpts)
-	ctx, cancel := getDefaultContext(((1<<63 - 1) * time.Nanosecond)) // infinite duration ... go does not expost max duration constant
-	defer cancel()
+	ctx := context.Background() // no timeout
```
gabemontero: again default timeout vs. no timeout

pkg/docker/docker.go
```diff
@@ -931,27 +906,23 @@ func (d *stiDocker) RunContainer(opts RunContainerOptions) error {
...
 
-	containerName := containerNameOrID(container)
-
 	// Container was created, so we defer its removal, and also remove it if we get a SIGINT/SIGTERM/SIGQUIT/SIGHUP.
 	removeContainer := func() {
-		glog.V(4).Infof("Removing container %q ...", containerName)
-		if err := d.RemoveContainer(container.ID); err != nil {
-			glog.V(0).Infof("warning: Failed to remove container %q: %v", containerName, err)
+		glog.V(4).Infof("Removing container %q ...", container.ID)
+		if removeErr := d.RemoveContainer(container.ID); removeErr != nil {
```
gabemontero: i used to get beat up a lot when first joining the team about having differen variables for different errors vs. reusing `err`, with basically it didn't golang stylistic conventions

i won't out and out call for a change, but be prepared others might be more adamant about this

pkg/docker/progressreporter.go
```diff
@@ -0,0 +1,109 @@
...
+
+	dockermessage "github.com/docker/docker/pkg/jsonmessage"
+	"golang.org/x/net/context"
+)
+
+// copy/paste from k8s.io/kubernetes/pkg/kubelet/dockertools/kube_docker_client.go
+
+const (
+	// defaultImagePullingProgressReportInterval is the default interval of image pulling progress reporting.
+	defaultImagePullingProgressReportInterval = 10 * time.Second
```
gabemontero: Start with capital letter

pkg/docker/progressreporter.go
```diff
@@ -0,0 +1,109 @@
...
+
+	// defaultImagePullingStuckTimeout is the default timeout for image pulling stuck. If no progress
+	// is made for defaultImagePullingStuckTimeout, the image pulling will be cancelled.
+	// Docker reports image progress for every 512kB block, so normally there shouldn't be too long interval
+	// between progress updates.
+	defaultImagePullingStuckTimeout = 1 * time.Minute
```
gabemontero: start with capital letter

test/integration/integration_test.go
```diff
@@ -166,7 +162,8 @@ func (i *integrationTest) setup() {
 		FakeScriptsFileURL = "file://" + filepath.Join(testImagesDir, ".s2i", "bin")
 
 		for _, image := range []string{TagCleanBuild, TagCleanBuildUser, TagIncrementalBuild, TagIncrementalBuildUser} {
-			i.dockerClient.RemoveImage(image, dockertypes.ImageRemoveOptions{})
+			ctx := context.Background() // no timeout
```
gabemontero: I would use the default timeout vs. no timeout

pkg/docker/docker.go
```diff
@@ -1032,12 +1003,11 @@ func (d *stiDocker) RunContainer(opts RunContainerOptions) error {
 
 		// Return an error if the exit code of the container is
 		// non-zero.
-		glog.V(4).Infof("Waiting for container %q to stop ...", containerName)
-		ctx, cancel := getDefaultContext(math.MaxInt64 * time.Nanosecond) // infinite duration ... go does not expose max duration constant
-		defer cancel()
+		glog.V(4).Infof("Waiting for container %q to stop ...", container.ID)
+		ctx = context.Background() // no timeout
```
gabemontero: yeah this one is special case - infinite timeout

pkg/docker/docker.go
```diff
@@ -69,8 +66,8 @@ const (
 	// timeout is reached, certain Docker API calls might error out.
 	DefaultDockerTimeout = 60 * time.Second
 
-	// DefaultShmSize is the default shared memory size to use (in bytes) if not specified.
-	DefaultShmSize = int64(1024 * 1024 * 64)
+	// defaultShmSize is the default shared memory size to use (in bytes) if not specified.
+	defaultShmSize = int64(1024 * 1024 * 64)
```
gabemontero: ah ... I now get you decided to mark it private.  I would leave it public.  No harm with it being accessible, and I would not be shocked if something else needs this in the future.  Also I wonder if there is origin specific code that is looking it up.

pkg/docker/docker.go
gabemontero: oops .. meant to delete this comment.  Or maybe I did in another spot.  In any even, the run container paths is special case - leave it infinite.  Apologies for the churn.

pkg/docker/docker.go
jim-minter: This is useful feedback.  With the timeouts generally, I tried to copy them as they stood, but there's an odd mix of no timeout, 1 minute timeout (DefaultDockerTimeout), and one case of 2 minute timeout (InspectImage) in the code.  I was somewhat tempted to rationalise this but wanted to get suggestions first.

jim-minter: TBH, I'd be interested in understanding the justification for any of the timeout smarts here, it's not entirely clear to me how they actually help matters.

pkg/docker/docker.go
gabemontero: ok, i'll defer, but if I catch that we had to make it public in the future, you own me a beer :-)

pkg/docker/docker.go
jim-minter: All of the progress reporter and this code is copy/paste from the kube client, fwiw - nothing new by me.  I would have ditched it but for the fact it monitors progress on the pull, which I do think is better than a stark "no pull may take longer than 2 minutes" rule.

pkg/docker/docker.go
gabemontero: If there were instances where all the k8s docker client / ctx gorp resulted in infinite timeouts under the covers for operations other than run container, that was an oversight that slipped through code inspection and test.

pkg/docker/docker.go
jim-minter: Yup. I think this one's exceptional - I believe it's clearer as removeErr so as not to silently shadow the pre-existing err in context in a different goroutine.

pkg/docker/docker.go
gabemontero: I see - gotcha.  

Long winded aside:  Conversely, FWIW, we were building up a list of issues of similar feel and severity we had found in the k8s client during the original port.  I don't recall this one in particular - probably just slipped through.  But before we got too far down the path of working with the folks upstream to start addressing them, the winds around major rework in that space came, and then ultimately the remove effort you are now undertaking.

Brief summary of the aside: I'm not surprised such shortcomings were not previously addressed :-)

Circling back up:  are you amenable to improving this code borrowed from the k8s client ?

jim-minter: Yup, I've got an idea for this.

gabemontero: cool / thx !

jim-minter:
@gabemontero all of your comments above should be fixed in this new push - please review
[test]

pkg/docker/docker.go
```diff
@@ -199,7 +207,7 @@ func (rco RunContainerOptions) asDockerConfig() dockercontainer.Config {
 		Image:        getImageName(rco.Image),
 		User:         rco.User,
 		Env:          rco.Env,
-		Entrypoint:   dockerstrslice.StrSlice(rco.Entrypoint),
+		Entrypoint:   rco.Entrypoint,
```
gabemontero: sorry if this was here before and I missed earlier - just a question / need for confirmation - was there a type change for this field in engine-api?

pkg/docker/docker.go
gabemontero: did you disagree with my suggestion on just passing 
```
dockertypes.ContainerRemoveOptions {
   RemoveVolumes: true,
   Force:                  true,
}
```
vs. assigning to `opts` and passing that into `ContainerRemove` ?

Upon revisting, I do see that since `getDefaultContext` returns 2 items you have to do what you did.

gabemontero:
Just a couple of minor items in my last pass.

And of course, the commits need to be squashed and ordered appropriately.  The convention we have for these type of changes is that the PR should be left with 2 commits:
- first commit is for all the godep / vendor changes 
- second commit will be for all the changes to the repo's source files under `pkg`

pkg/docker/docker.go
jim-minter: dockerstrslice.StrSlice wraps []string just adding UnmarshalJSON() to unmarshal a string or array of strings in json to a []string.  In all other respects dockerstrslice.StrSlice is a []string.  Since we are not unmarshalling here nor would it make sense to, I removed the reference to dockerstrslice.StrSlice to improve readability and save the import.

jim-minter:
All done.
[test]

gabemontero:
On Thursday, October 13, 2016, Jim Minter <notifications@github.com> wrote:

> *@jim-minter* commented on this pull request.
> ------------------------------
>
> In pkg/docker/docker.go
> <https://github.com/openshift/source-to-image/pull/607>:
>
> > @@ -199,7 +207,7 @@ func (rco RunContainerOptions) asDockerConfig() dockercontainer.Config {
>  		Image:        getImageName(rco.Image),
>  		User:         rco.User,
>  		Env:          rco.Env,
> -		Entrypoint:   dockerstrslice.StrSlice(rco.Entrypoint),
> +		Entrypoint:   rco.Entrypoint,
>
> dockerstrslice.StrSlice wraps []string just adding UnmarshalJSON() to
> unmarshal a string or array of strings in json to a []string. In all other
> respects dockerstrslice.StrSlice is a []string. Since we are not
> unmarshalling here nor would it make sense to, I removed the reference to
> dockerstrslice.StrSlice to improve readability and save the import.
>
Interesting - thx for the details - what spurred my comment was my
recollection that a compile was failing back during the conversion to k8s
cause a field was a StrSlice instead of a []string and i had to find that
import.  I thought it was that field but perhaps not.  And perhaps the type
did change.  In any event we're good thx.

jim-minter:
nit [test]

gabemontero:
@bparess - do we car if the commit for the src code changes comes before the vendoring / godeps commit?  I think we do but want to confrm (in case I'm just being too nit picky)

otherwise I at least am GTM.

jim-minter:
@bparees please see previous comment
@gabemontero I thought it made more sense the other way around too, but I was following what you wrote.  I'll swap the order now.

jim-minter:
@gabemontero done

gabemontero:
OK / cool / apologies for the confusion

IGTM

jim-minter:
@bparees or @gabemontero it'd be great if one of you could set this to merge if you're happy with it.

gabemontero:
On Fri, Oct 14, 2016 at 8:38 AM, Jim Minter <notifications@github.com>
wrote:

> @bparees <https://github.com/bparees> or @gabemontero
> <https://github.com/gabemontero> it'd be great if one of you could set
> this to merge if you're happy with it.
>
I don't have that power, you'll have to wait for @bparees.

pkg/docker/docker.go
```diff
@@ -117,36 +115,37 @@ type Docker interface {
...
-	ctx, _ := context.WithTimeout(context.Background(), 2*time.Minute)
-	resp, _, err := s.client.ImageInspectWithRaw(ctx, name, true)
+func (d stiDocker) InspectImage(name string) (*dockertypes.ImageInspect, error) {
+	ctx, cancel := getDefaultContext()
+	defer cancel()
+	resp, _, err := d.client.ImageInspectWithRaw(ctx, name, true)
 	if err != nil {
-		if dockerapi.IsErrImageNotFound(err) {
-			return nil, fmt.Errorf("no such image :%q", name)
-		}
```
bparees: having this method return an error that contained the string "no such image" was critical to getting a bunch of test flows working in the first place, is the d.client.ImageInspectWithRaw() now handling that for us?  Callers to inspectImage need a way to distinguish between a legitimate error and an image not found error.

pkg/docker/docker.go
```diff
@@ -285,47 +274,47 @@ type BuildImageOptions struct {
...
-		endpoint: config.Endpoint,
 	}, nil
 }
 
-func getDefaultContext(timeout time.Duration) (context.Context, context.CancelFunc) {
-	return context.WithTimeout(context.Background(), timeout)
+func getDefaultContext() (context.Context, context.CancelFunc) {
+	// the intention is: all docker API calls with the exception of known long-
+	// running calls (ContainerWait, ImagePull, ImageBuild) must complete within a
+	// certain timeout otherwise we bale.
```
bparees: s/bale/bail/

pkg/util/timeout.go
```diff
@@ -14,29 +14,38 @@ type TimeoutError struct {
...
+func TimeoutAfter(t time.Duration, errorMsg string, f func(*time.Timer) error) error {
 	c := make(chan error, 1)
+	timer := time.NewTimer(t)
 	go func() {
-		defer close(c)
-		c <- fn()
+		err := f(timer)
+		if !IsTimeoutError(err) {
+			c <- err
+		}
```
bparees: why wouldn't you send the timeout err into the channel too, does it really matter which way the error bubbles up?

bparees:
@jim-minter couple minor nits/questions and then i'll merge.

pkg/docker/docker.go
jim-minter: <hangs head in shame>

pkg/docker/docker.go
```diff
@@ -552,7 +490,7 @@ func (d *stiDocker) CheckAndPullImage(name string) (*api.Image, error) {
 	}
 
 	image, err := d.CheckImage(name)
-	if err != nil && !strings.Contains(err.(errors.Error).Details.Error(), "no such image") {
+	if err != nil && !strings.Contains(err.(errors.Error).Details.Error(), "No such image") {
```
jim-minter: Yes, that's right.  I wanted to write !dockerapi.IsErrImageNotFound here, which works except for in the test harness: it's not possible to synthesise an imageNotFoundError for testing because it's not public within the engine API...

pkg/util/timeout.go
jim-minter: A caller of TimeoutAfter would expect a TimeoutError with its fields populated, and the TimeoutAfter function has the information necessary to do this.  However the TimeoutAfter function does not currently send this information into f().  f() needs a way to signal to TimeoutAfter "I realise I've timed out, I'm giving up and exiting politely".  Within its signature (returning error) the only way of doing this is by return of a special error type.  I've reused TimeoutError for this purpose.  Thus the TimeoutError from f() needs to be found and quietly discarded; the TimeoutError that TimeoutAfter sends will be correctly populated.

Alternatives here would be:
- add another error type for f() to return instead of TimeoutError (don't overload usage of TimeoutError)
- pass errorMsg into f() so any TimeoutError it may return is populated correctly and bubbles up either way
- change the signature of f() so it sends to a channel instead of returning an error, thus f() can decide not to send before it returns

jim-minter:
@bparees repushed with spelling fix

pkg/util/timeout.go
bparees: ok thanks.

bparees:
[merge]

PR #610, opened 2016-10-11
Add ExtractTarStreamFromTarReader

jim-minter:
This PR has two commits:
1) fix unbuffered channel and allow detection of expired timer in goroutine
- resolves an issue whereby a goroutine would hang/leak if a timer expired (and also checks within the goroutine to see if the timer has expired according to reccomended practice)

2) add a function ExtractTarStreamFromTarReader which is just like ExtractTarStreamWithLogging but accepts a tar.Reader-like object.  This is a prereq for cleanup in origin/pkg/bootstrap which should use this code.

### DIFF
--- a/pkg/tar/tar.go
+++ b/pkg/tar/tar.go
@@ -323,13 +323,15 @@ func (t *stiTar) ExtractTarStream(dir string, reader io.Reader) error {
 // exceeds the value of timeout
 func (t *stiTar) ExtractTarStreamWithLogging(dir string, reader io.Reader, logger io.Writer) error {
 	tarReader := tar.NewReader(reader)
-	errorChannel := make(chan error)
-	timeout := t.timeout
-	timeoutTimer := time.NewTimer(timeout)
+	errorChannel := make(chan error, 1)
+	timeoutTimer := time.NewTimer(t.timeout)
 	go func() {
 		for {
 			header, err := tarReader.Next()
-			timeoutTimer.Reset(timeout)
+			if !timeoutTimer.Stop() {
+				break
+			}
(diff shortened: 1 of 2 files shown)


jim-minter:
[test]

pkg/tar/tar.go
```diff
@@ -313,23 +325,31 @@ func (t *stiTar) writeTarHeader(tarWriter *tar.Writer, dir string, path string,
...
-	timeout := t.timeout
-	timeoutTimer := time.NewTimer(timeout)
+	return t.ExtractTarStreamFromTarReader(dir, tarReader, logger)
+}
+
+// ExtractTarStreamFromTarReader extracts files from a given tar stream.
+// Times out if reading from the stream for any given file
+// exceeds the value of timeout
+func (t *stiTar) ExtractTarStreamFromTarReader(dir string, tarReader Reader, logger io.Writer) error {
+	errorChannel := make(chan error, 1)
```
jim-minter: if the timer fires before the goroutine finishes, the parent select will read from the timer channel and the parent function will then exit.  Nothing hangs around to listen to the errorchannel.  If the errorchannel is unbuffered, then when at some point the goroutine writes its result to it, it will block forever.  Whereas, if buffered, the goroutine will then exit out and the channel will be GCed.

(I think this is the only example of this issue in s2i).

bparees: thanks, i learned something. hopefully i'll retain it.

csrwng:
LGTM
[test] failure is a yum mirror issue

csrwng:
[merge]

PR #615, opened 2016-10-17
teach isValidGitRepository about git submodules

### LINKED ISSUE #606: 1.1.2 does not work with git submodules
# Steps to reproduce

s2i version 1.1.2
git version 2.10.0

* git clone https://github.com/3scale/s2i-openresty
* make test

## Result

> ERROR: An error occurred: lstat .../3scale/s2i-openresty/test/test-app/.git/objects: not a directory

## Expected result

successful build

## More info

`test/test-app/.git` is a file with following contents:

> gitdir: ../../.git/modules/test/test-app

It is standard git submodule. 
Version 1.1.1 was working just fine.
git version 2.10.0

## Workaround

```sh
cd test/test-app
rm .git
ln -s ../../.git/modules/test/test-app .git
```

jim-minter:
fixes #606

### DIFF
--- a/pkg/scm/git/git.go
+++ b/pkg/scm/git/git.go
@@ -365,21 +365,67 @@ func ParseSSH(source string) (*URLMods, error) {
 	return nil, fmt.Errorf("unable to parse ssh git clone specification:  %s", source)
 }
 
-// isValidGitRepository checks to see if there is a .git repository in the
+// followGitSubmodule looks at a .git /file/ and tries to retrieve from inside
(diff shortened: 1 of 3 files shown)


jim-minter:
@bparees ptal.  Part of me thinks we're being too clever introspecting .git at all and that we should just stat it and use the result to determine if we think we're dealing with a git repo or not.  Anyway, the history behind why we do more is at #298 and #552.

bparees:
yeah we probably are, but lgtm
[merge]

PR #616, opened 2016-10-17
Use github.com/openshift/origin/pkg/image/reference

### LINKED ISSUE #614: Use common parser for DockerImageReference
Recently `origin` and S2I use their own implementations of `ParseDockerImageReference`. Right now S2I started to use `github.com/docker/distribution/reference` as parser and `origin` started use parser based on same module. I think it would be better to use the same parser for `DockerImageReference` to avoid differences in behavior in the future.

I made a [module](https://github.com/openshift/origin/tree/master/pkg/image/reference) without dependencies on `origin`. It can be used in both repositories. If the parser will be changed, we will not need to synchronize our parsers any more.

https://github.com/openshift/source-to-image/pull/594
https://github.com/openshift/origin/pull/11245

jim-minter:
Dependent on #607 which should be good to go.
Fixes #614.

### DIFF
--- a/pkg/build/strategies/layered/layered.go
+++ b/pkg/build/strategies/layered/layered.go
@@ -12,7 +12,6 @@ import (
 	"regexp"
 	"time"
 
-	"github.com/docker/distribution/reference"
 	"github.com/openshift/source-to-image/pkg/api"
 	"github.com/openshift/source-to-image/pkg/build"
 	"github.com/openshift/source-to-image/pkg/docker"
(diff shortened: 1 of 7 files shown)


pkg/build/strategies/layered/layered.go
```diff
@@ -145,6 +144,10 @@ func (builder *Layered) Build(config *api.Config) (*api.Result, error) {
 		return buildResult, fmt.Errorf("builder image uses ONBUILD instructions but ONBUILD is not allowed")
 	}
 
+	if config.BuilderImage == "" {
+		return nil, fmt.Errorf("repository name must have at least one component")
```
bparees: would this error make more sense as "builder image name cannot be empty"?

jim-minter: yes - I'll change it.  What is there happens to be what was returned in this case before.

bparees:
lgtm 
[merge]

pkg/build/strategies/layered/layered.go
```diff
@@ -146,6 +145,10 @@ func (builder *Layered) Build(config *api.Config) (*api.Result, error) {
 		return buildResult, fmt.Errorf("builder image uses ONBUILD instructions but ONBUILD is not allowed")
 	}
 
+	if config.BuilderImage == "" {
+		return nil, fmt.Errorf("builder image name cannot be empty")
```
PI-Victor: @jim-minter sorry to drop in on your PR, but can you return buildResult here with a generic s2i failure reason instead of nil?
i need openshift to read what kind of error happened in s2i and print it in the output together with the detailed message. something along the lines of:
```go
buildResult.BuildInfo.FailureReason = utilstatus.NewFailureReason(utilstatus.ReasonGenericS2IBuildFailed, utilstatus.ReasonMessageGenericS2iBuildFailed)
return buildResult, fmt.Errorf("builder image name cannot be empty")
```

PI-Victor: i see the merge tag is set, it's ok, i can fix it later on though.

bparees: @PI-Victor @jim-minter i went ahead and canceled the merge.

jim-minter: done, @bparees please merge

bparees:
lgtm, thanks @jim-minter 
[merge]

PR #617, opened 2016-10-17
add failure message

PI-Victor:
@bparees in order to avoid duplication and a really weird long switch statement in OpenShift, i attached the detail messages for the failure reason to s2i, (where they actually belong after all).
PTAL.


bparees:
lgtm [merge]

PR #618, opened 2016-10-18
fix lint warnings

jim-minter:


pkg/api/validation/validation.go
```diff
@@ -8,8 +8,8 @@ import (
 )
 
 // ValidateConfig returns a list of error from validation.
-func ValidateConfig(config *api.Config) []ValidationError {
-	allErrs := []ValidationError{}
+func ValidateConfig(config *api.Config) []Error {
+	allErrs := []Error{}
```
bparees: I particularly hate the lint "stutter" warning, i feel like this name change actually makes the code harder to read (now we have Error and error).

is there a way to explicitly disable stutter checking in lint?

bparees:
annoyance at stutter warnings aside, lgtm.
[merge]

PR #619, opened 2016-10-19
Replace wrong failure reason for assemble failed

PI-Victor:
i've confused !isMissing with isMissing, so the failure reason for this should just be a generic one.
We also don't fail for missing dependencies, but we continue the build with the layered build strategy.


bparees:
squash.

PI-Victor:
@bparees still need to figure something out. i'm testing out the failure of the assemble script and it's not returning the correct error. will request merge a bit later on.

PI-Victor:
@bparees squashed, ptal.

bparees:
[merge]

PR #625, opened 2016-10-28
s2i windows support

### LINKED ISSUE #301: Incremental build error (read/write on closed pipe)
Saving incremental artifacts seems to fail randomly on (perhaps just large) projects. Could be a regression of #171, although ulimit and filesystem hard limits doesn't seem to affect it.
#### I've created a deployable project to replicate the issue, available at:

https://github.com/binarybabel/sti-rails-incremental-demo
_Build config and logs are available in the `extra` directory._
#### Level 5 build log illustrating the problem:

https://raw.githubusercontent.com/binarybabel/sti-rails-incremental-demo/master/extra/incremental-failure.log

### LINKED ISSUE #620: custom s2i scripts loses its executable permission when image built on MS-Windows
# Summary

When building image with s2i on MS-Windows host
And when the code features a custom `.s2i/bin/run` script.
Then that script is injected in the image without executable permission

As a consequence, the image produced by `s2i build` is unbootable.
# Context
- trying to run `s2i build` on MS Windows host with the latest "docker toolbox" (version 1.12.2)
- using s2i version 1.1.1 (as long as version 1.1.2 suffers from #603)
# How to reproduce

Run the following commands on Linux first, then on MS-Windows with the latest "docker toolbox" installed.

``` sh
git clone https://github.com/openshift/ruby-hello-world.git
cd ruby-hello-world

mkdir -p .s2i/bin
echo -e '#!/bin/sh\n\nexec bundle exec ruby app.rb' | tee .s2i/bin/run
chmod +x .s2i/bin/run
git add .s2i/bin/run
git commit -m "Add .s2i/bin/run"

s2i build . centos/ruby-22-centos7 ruby-hello-world          

docker run --rm ruby-hello-world
```

On Linux host, the last `docker run` command boots the image as expected.
On Windows host, the last `docker run` command fails with the following output:

``` sh
Lambda@perfect-dev MINGW64 ~/ruby-hello-world (master)
$ docker run --rm ruby-hello-world
/usr/bin/container-entrypoint: line 2: /tmp/scripts/run: Permission denied
/usr/bin/container-entrypoint: line 2: exec: /tmp/scripts/run: cannot execute: Permission denied
```

It is easy to verify that the problem is that the CMD is not executable:

``` sh
Lambda@perfect-dev MINGW64 ~/ruby-hello-world (master)
$ docker run --rm ruby-hello-world ls -l /tmp/scripts/run
-rw-r--r-- 1 default root 38 Oct 21 14:24 /tmp/scripts/run
```

On Linux host, the CMD is executable, as expected:

``` sh
jp440p:ruby-hello-world(41m|master?) $ docker run --rm -ti ruby-hello-world ls -l /tmp/scripts/
total 12
-rwxr-xr-x 1 default root   40 Oct 21 14:18 run
```

jim-minter:
With this PR, unit and integration tests pass on Windows.

Fixes #620
Fixes #301


jim-minter:
fixes #620

gabemontero:
@jim-minter did you try running `hack/test-stirunimage.sh ` as well ?

bparees:
[test]

bparees:
@jim-minter the fixes comment has to be in the PR summary for it to close the issue.

cmd/s2i/main.go
```diff
@@ -212,7 +212,8 @@ func newCmdRebuild(cfg *api.Config) *cobra.Command {
 				os.Exit(1)
 			}
 
-			if r, err := os.Open(cfg.DockerCfgPath); err == nil {
+			r, err := os.Open(cfg.DockerCfgPath)
+			if err == nil {
```
bparees: i don't strictly care, but what was the motivation for changing this?

hack/test-go.sh
```diff
@@ -26,24 +26,11 @@ find_test_dirs() {
...
-  if [ -z "${S2I_COVER}" ]; then
-    S2I_COVER="-cover -covermode=atomic"
-  fi
-fi
-
 if [ -z ${S2I_RACE+x} ]; then
-  S2I_RACE=""
+  S2I_RACE="-race"
+fi
+if [ -z ${S2I_COVER+x} ]; then
```
bparees: [ -n ${S2I_COVER} ] seems more straight forward?  (assuming i'm correctly understanding the intent as "set S2I_COVER if it's not set to something already")

cmd/s2i/main.go
jim-minter: There were few remaining goshadow errors so I decided to tackle them.  This wasn't a case in question but I changed it to match the other changes in the same function stylistically.

hack/test-go.sh
jim-minter: Yes, indeed, but not when set -o nounset is in force.  I think nounset is overkill myself but it was there before.

jim-minter:
@gabemontero this is now also working against hack/test-stirunimage.sh on Windows.
@bparees I believe this is now good for review.

jim-minter:
FWIW the scripts that I'm using to build a Windows VM for testing are at https://github.com/jim-minter/packer-windows/tree/local .  I'm working on making them more consumable.  This PR works against the Windows 7 and Windows 2012R2 builds (the only two I've tried up to now).

pkg/tar/tar.go
```diff
@@ -95,124 +74,80 @@ type Reader interface {
...
+	return &stiTar{
+		FileSystem: fs,
+		exclude:    DefaultExclusionPattern,
+		timeout:    defaultTimeout,
 	}
-	return t.CreateTarStream(tmpDir, false, writer)
+}
+
+// stiTar is an implementation of the Tar interface
+type stiTar struct {
```
bparees: s2iTar

pkg/build/strategies/sti/sti_test.go
```diff
@@ -549,60 +552,34 @@ func TestSaveArtifactsExtractError(t *testing.T) {
...
-			sourcePath:       "invalid/path",
-			expectedError:    &err,
-		},
-		// 1
-		{
-			validCloneSpec:   false,
-			refSpecified:     false,
-			cloneExpected:    false,
-			checkoutExpected: false,
-			copyExpected:     true,
```
jim-minter: These tests were exercising a code path in pkg/scm/git.(*Clone).Download which I believe was no longer reachable in practice (only by this test) and was somewhat rotten.  I removed the code path and the test.  My logic is: a git.Clone object is only created in scm.go after ValidCloneSpec has returned true.  git.(*Clone).Download had a code path to deal with what happens if ValidCloneSpec later returns false.  I have retained an instrumented call to ValidCloneSpec in git.(*Clone).Download to log an error if this happens, but have otherwise deleted the erronenous code path.

jim-minter:
@bparees all changes made.

bparees:
lgtm, please squash

hack/test-go.sh
rhcarvalho: Even more straightforward, no if:

```bash
S2I_COVER="${S2I_COVER-"-covermode=atomic"}"
```

Or we can just use this to set the value if-not-unset on first usage:

```bash
# ...

# parameter substitution, no assignment:
go test ... "${S2I_COVER-"-covermode=atomic"}"
# ^^^ I think we probably just need this one above, all other occurrences of S2I_COVER may disappear.

# substitution and assignment:
go test ... "${S2I_COVER="-covermode=atomic"}"
```

(Those expansions work with `nounset`)

hack/test-stirunimage.sh
```diff
@@ -14,6 +14,10 @@ function time_now()
 
 mkdir -p /tmp/sti
 WORK_DIR=$(mktemp -d /tmp/sti/test-work.XXXX)
+S2I_WORK_DIR=${WORK_DIR}
+if [[ "$OSTYPE" == "cygwin" ]]; then
```
rhcarvalho: This assumes Windows users use `cygwin`, is that a fair assumption? Hmm well, this is a bash script... already not something very Windows-friendly...

What happens to folks using `cmd.exe` or PowerShell?

pkg/api/types.go
```diff
@@ -21,7 +21,7 @@ const (
 
 // invalidFilenameCharacters contains a list of character we consider malicious
 // when injecting the directories into containers.
-const invalidFilenameCharacters = `\:;*?"<>|%#$!+{}&[],"'` + "`"
+const invalidFilenameCharacters = `\;*?"<>|%#$!+{}&[],"'` + "`"
```
rhcarvalho: Why are we removing `:`?

If we are removing this to accept paths like `C:\Users\name\file.txt`, however, seems that we're then being less restrictive about file names containing `:`?!

And what about `\`, wouldn't it be the same case as `:`?

pkg/api/types.go
```diff
@@ -499,18 +499,16 @@ func (l *VolumeList) Set(value string) error {
...
+	var mount []string
+	pos := strings.LastIndex(value, ":")
+	if pos == -1 {
+		mount = []string{value, "."}
+	} else {
+		mount = []string{value[:pos], value[pos+1:]}
 	}
-	s := VolumeSpec{Source: filepath.Clean(mount[0]), Destination: filepath.Clean(mount[1])}
+	mount[0] = strings.Trim(mount[0], `"'`)
+	mount[1] = strings.Trim(mount[1], `"'`)
```
rhcarvalho: Do we have unit tests for this?

Would be nice to explicitly test this with good and bad input.

rhcarvalho:
For instance, https://github.com/openshift/source-to-image/pull/625/commits/607c6d4e398c8a72f9e3bbec75f9b52a30498ba1 is interesting enough by itself to deserve its own PR and code review, testing... not sure how much that is related to "s2i Windows support".

rhcarvalho:
https://github.com/openshift/source-to-image/pull/625/commits/564ecc4653acfec35aa74570bc3a8343d359b38c is another nice one that would benefit everybody, doesn't need to be blocked by the review of so many other things.

pkg/tar/tar_test.go
```diff
@@ -411,22 +400,10 @@ func TestExtractTarStreamTimeout(t *testing.T) {
...
-		writer.Close()
-	}()
-	extractError := make(chan error, 1)
-	go func() {
-		defer wg.Done()
-		extractError <- th.ExtractTarStream(destDir, reader)
-	}()
-	wg.Wait()
-	err = <-extractError
+	time.AfterFunc(20*time.Millisecond, func() { writer.Close() })
```
rhcarvalho: Hmm, this is catchy. Why do we close the writer after 20ms?!

rhcarvalho:
https://github.com/openshift/source-to-image/pull/625/commits/fdfa4e5fdaa7cbbd557f5bf8e110b97e124466a5 is another good candidate for it's own PR. Very good job :+1: 

I would be sad to see all of those nice changes squashed into a single commit saying "s2i Windows support", when in fact many of the changes have nothing to do with Windows.

test/integration/integration_test.go
```diff
@@ -370,60 +432,6 @@ func TestIncrementalBuildOnBuild(t *testing.T) {
 	integration(t).exerciseIncrementalBuild(TagIncrementalBuildOnBuild, FakeImageOnBuild, false, true, true)
 }
 
-func (i *integrationTest) exerciseInjectionBuild(tag, imageName string, injections []string) {
```
rhcarvalho: Curiosity, why did you move this into `TestInjectionBuild`?

cmd/s2i/main.go
jim-minter: @rhcarvalho yes, correct, line 215 was/is not specifically a case of shadowing.  However, below there was a shadowing case with nearly identical code which I did resolve.  I made the equivalent change from below here as well to maintain the clear similarity...

hack/test-go.sh
jim-minter: @rhcarvalho thanks - applied.

hack/test-stirunimage.sh
jim-minter: Cygwin becomes a prerequisite for using our build and test scaffolding on Windows, but not for running the s2i binary itself, which should work on cygwin and non-cygwin environments (although there is no special testing for the latter currently).

Those who want to build on Windows without cygwin can run go build just as they could before.

I think this is the best and most realistic approach here.

pkg/tar/tar_test.go
jim-minter: Because the test purpose is to prove that the set timeout of 10ms is functional.

test/integration/integration_test.go
jim-minter: Because it was called from one location only.

jim-minter:
@bparees, @rhcarvalho is reluctant to see these commits squashed, I am too FWIW.

bparees:
@jim-minter i agree there is some logical separation to be done, i doubt it's 17 commits though. can you factor it down to 2-3 commits (windows support, tar streaming, govet/shadowing cleanup)?  it will be easier to dig through the history/see what changed/revert things if there are fewer total commits to manage.

rhcarvalho:
More & smaller commits will make my live easier when rebasing some changes I have locally for integration tests... great work here @jim-minter, appreciated.

pkg/tar/tar_test.go
rhcarvalho: Sorry, now I looked at a bigger chunk of context. This is just being moved from line 420. Thanks for the massive clean up. :+1:

cmd/s2i/main.go
```diff
@@ -228,7 +229,8 @@ func newCmdRebuild(cfg *api.Config) *cobra.Command {
 
 			// Attempt to read the .dockercfg and extract the authentication for
 			// docker pull
-			if r, err := os.Open(cfg.DockerCfgPath); err == nil {
+			r, err = os.Open(cfg.DockerCfgPath)
```
rhcarvalho: Hmm, wait... did you notice that we're opening the same file twice? Is that even necessary?! 215 and 231 are opening the same file.. at first sight it doesn't look like the code in between the two calls would actually change the config file, would it?!

hack/test-stirunimage.sh
rhcarvalho: I'm fine with that. Since this is good news for anyone on Windows, I'd suggest adding some documentation in the README or some other document. And a note in the next release /cc @bparees :smile:

pkg/api/types.go
rhcarvalho: I think the change of `invalidFilenameCharacters` and the logical idea that now we want to accept Windows-style paths begs for adding more test cases in `types_test.go` / `TestVolumeListSet`, don't you think?

pkg/tar/tar.go
rhcarvalho: @jim-minter ^^

\<fun\>
will we ever get rid of the sti's in the code base... ;)
\</fun\>

pkg/tar/tar.go
```diff
@@ -307,6 +263,14 @@ func (t *stiTar) writeTarHeader(tarWriter *tar.Writer, dir string, path string,
 	if err != nil {
 		return err
 	}
+	// on Windows, tar.FileInfoHeader incorrectly interprets directory symlinks
+	// as directories.  https://github.com/golang/go/issues/17541
```
rhcarvalho: And I thought Windows had no symlinks... `.lnk` yes, but "directory symlinks", that's new :)

pkg/util/glog/glog.go
```diff
@@ -51,16 +51,47 @@ var (
...
+	return false
+}
+
+// V will returns a logger which will discard output if the specified level is greater than the current logging level.
+func (discarding discard) V(level int32) VerboseLogger {
+	return discarding
+}
+
+// Infof records an info log entry.
+func (discard) Infof(_ string, _ ...interface{}) {
```
rhcarvalho: Since you're passing by, we can drop the `_` everywhere.

It is idiomatic to write this kind of stuff in one line as it was.
Specially the empty functions.

pkg/util/status/build_status.go
```diff
@@ -89,13 +89,6 @@ const (
 	// range of failures.
 	ReasonMessageGenericS2iBuildFailed api.StepFailureMessage = "Generic S2I Build failure - check S2I logs for details"
 
-	// ReasonTarSourceFailed is the failure reason associated with a failure to
-	// tar the current source.
-	ReasonTarSourceFailed api.StepFailureReason = "TarSourceFailed"
-	// ReasonMessageTarSourceFailed is the message associated with a failure to
-	// tar the current source.
-	ReasonMessageTarSourceFailed api.StepFailureMessage = "Failed to tar source files"
-
```
rhcarvalho: Happy to see this go away :-)

@PI-Victor are these values already somehow depended on in Origin code?

pkg/util/fs.go
```diff
@@ -193,7 +273,31 @@ func (h *fs) Open(filename string) (io.ReadCloser, error) {
...
+			info = h.enrichFileInfo(path, info)
+			h.m.Unlock()
+		}
+		return walkFn(path, info, err)
+	}
+	return filepath.Walk(root, wrapper)
+}
+
+func (h *fs) enrichFileInfo(path string, fi os.FileInfo) os.FileInfo {
+	// call me holding the mutex
```
rhcarvalho: Why not lock/unlock here? Make it easier on callers, enforce correctness.

pkg/util/status/build_status.go
PI-Victor: Nay!

pkg/util/fs.go
```diff
@@ -193,7 +273,31 @@ func (h *fs) Open(filename string) (io.ReadCloser, error) {
...
+			h.m.Lock()
+			info = h.enrichFileInfo(path, info)
+			h.m.Unlock()
+		}
+		return walkFn(path, info, err)
+	}
+	return filepath.Walk(root, wrapper)
+}
+
+func (h *fs) enrichFileInfo(path string, fi os.FileInfo) os.FileInfo {
```
rhcarvalho: Can we have a doc comment here? When reading the [call site](https://github.com/openshift/source-to-image/pull/625/files#diff-36a50e367c08aef7a9b2fc62647daf54R126) I wondered what this is meant to do and why. Probably someone in the future will ask the same.

Maybe a better name could help too. "enrich" sounds mystifying to me.

pkg/util/fs.go
rhcarvalho: I see `h.enrichFileInfo(path, info)` is called from multiples places, we should definitely lock/unlock here and not in every call site.

pkg/util/fs.go
```diff
@@ -32,34 +35,111 @@ type FileSystem interface {
...
 }
 
+// ReadDir reads the directory named by dirname and returns a list of directory
+// entries sorted by filename.
 func (h *fs) ReadDir(path string) ([]os.FileInfo, error) {
-	return ioutil.ReadDir(path)
+	fis, err := ioutil.ReadDir(path)
+	if runtime.GOOS == "windows" && err == nil {
+		h.m.Lock()
+		for i := range fis {
```
rhcarvalho: Hmm, maybe the lock outside was aiming to optimizing this here, to avoid acquiring and releasing the lock multiple times in the loop?

One way to get around this and still be able to have lock/unlock within `enrichFileInfo` is to change the signature to take `...os.FileInfo` as a last argument.

```go
// enrichFileInfo does?!
func (h *fs) enrichFileInfo(path string, fis ...os.FileInfo) []os.FileInfo {
	var s []os.FileInfo
	h.m.Lock()
	for fi := range fis {
		if mode, ok := h.fileModes[path]; ok {
			fi = copyFileInfo(fi)
			fi.(*FileInfo).FileMode = mode
		}
		s = append(s, fi)
	}
	h.m.Unlock()
	return fi
}
```

WDYT?

test/integration/integration_test.go
```diff
@@ -71,10 +72,68 @@ func getDefaultContext() (context.Context, context.CancelFunc) {
 
 // TestInjectionBuild tests the build where we inject files to assemble script.
 func TestInjectionBuild(t *testing.T) {
-	integration(t).exerciseInjectionBuild(TagCleanBuild, FakeBuilderImage, []string{
```
rhcarvalho: With this change I noticed you don't call `setup` anymore, as `integration(...)` does. Is there any particular reason for that?

I'm doing some improvements to integration tests that rely on the fact that I can write common initialization to all tests, and the fact that they call `integration` helps with that.

pkg/build/strategies/sti/sti.go
```diff
@@ -568,77 +554,43 @@ func (builder *STI) Execute(command string, user string, config *api.Config) err
...
-		errReader.Close()
-		return errors.NewContainerError(config.BuilderImage, e.ErrorCode, errOutput)
-	}
-	// Do not wait for source input if the container times out.
-	// FIXME: this potentially leaks a goroutine.
-	if !util.IsTimeoutError(err) {
-		wg.Wait()
+		// Must wait for StreamContainerIO goroutine above to exit before reading errOutput.
+		<-c
+		err = errors.NewContainerError(config.BuilderImage, e.ErrorCode, errOutput)
```
csrwng: @rhcarvalho thx, @jim-minter just fyi ... I made a change in this area as well yesterday ... https://github.com/openshift/source-to-image/pull/631 
It looks like your change fixes things as well :) 👍

rhcarvalho:
Adding a compiled C binary `hack/sigintwrap/sigintwrap.exe` to support Windows/cygwin? Sounds scary to me.

jim-minter:
rebased, but not yet ready to be considered for merge

pkg/api/types.go
jim-minter: I don't think this is important enough to be concerned about.  IsInvalidFilename is used in one place and frankly IMO is no more than superstition.  It could be ripped out entirely, although perhaps if it stays as it is, it could contribute protection in an unexpected way one day.  Opening the door to ':' characters could result in a new class of invalid path being passed in, sure, but these will still be picked up eventually as an invalid path just like any other.

test/integration/integration_test.go
jim-minter: ok, will revert.

cmd/s2i/main.go
jim-minter: thanks - will tidy this up on the way

pkg/util/glog/glog.go
jim-minter: @rhcarvalho not clear what you mean by 'we can drop the _ everywhere' ?

pkg/util/fs.go
jim-minter: Thanks for the feedback on this - next push will include documentation and improved locking along the lines of what you suggested.

test/integration/integration_test.go
jim-minter: will revert this to the previous behaviour, as mentioned above.

jim-minter:
> Adding a compiled C binary hack/sigintwrap/sigintwrap.exe to support Windows/cygwin? Sounds scary to me.

@rhcarvalho Yeah.  I think it's minimal and necessary and I don't have a better alternative.  Ping me on IRC if you want to discuss the gory detail.

jim-minter:
@bparees I'm hoping this is ready to merge now.  In conjunction with https://github.com/openshift/vagrant-openshift/pull/520, something like the following commands can be used by the CI system to stand up a Windows instance and test s2i:

```bash
set -ex

export VAGRANT_VAGRANTFILE=hack/windows/Vagrantfile
vagrant up --provider=aws
trap "vagrant destroy" EXIT
vagrant sync-sti -c -s windows
vagrant test-sti --all -d windows
```

pkg/util/glog/glog.go
rhcarvalho: And I meant that one line per empty func + no new lines in between them is common practice in Go, in circumstances like the one we have here.
One common case where it happens is implementing `sort.Interface`:

https://golang.org/pkg/sort/#example_

jim-minter: Thanks - removed the '_'s.  For some reason the first time I tried this I thought I got a compile error as a result, but I guess not!

pkg/scm/git/git.go
```diff
@@ -578,3 +581,60 @@ func pipeToLog(reader io.Reader, log func(...interface{})) {
 		}
 	}
 }
+
+// CreateLocalGitDirectory creates a git directory with a commit
+func (h *stiGit) CreateLocalGitDirectory() (string, error) {
```
rhcarvalho: Shouldn't this be part of some test API/package? Weird to have a method `CreateLocalGitDirectory` that creates a commit with an empty file.

Same for all 3 methods, aren't they better suited for some test package, and not `pkg/scm/git`?

pkg/scm/git/git.go
```diff
@@ -29,6 +29,9 @@ type Git interface {
 	Checkout(repo, ref string) error
 	SubmoduleUpdate(repo string, init, recursive bool) error
 	GetInfo(string) *api.SourceInfo
+	CreateEmptyLocalGitDirectory() (string, error)
+	CreateLocalGitDirectory() (string, error)
+	CreateLocalGitDirectoryWithSubmodule() (string, error)
```
rhcarvalho: -1 for adding these to the `Git` interface. These are used only in tests, why are we making this change?

I see in the commit message:

> - move git functions from test package to git package to protect against import loop

@jim-minter what import loop are you referring to?

rhcarvalho:
@jim-minter this is excellent work I'm looking forward to seeing merged.

WRT to the `sigintwrap`, could we have a `README.md` in `hack/windows/sigintwrap` to explain to whoever will be looking at it:

- Why we need it (Windows support using Cygwin, plus any other relevant information)
- How to build it (including which environment, dependencies, ...)

As a bonus, would be nice if we could detect in CI that the source code in `sigintwrap.c` matches the binary.

pkg/scm/git/git.go
jim-minter: Here are the facts as I see them:
- the 3 routines ideally ought to be grouped together given they're doing similar things
- CreateLocalGitDirectory() is called from pkg/scm/scm_test.go (different package); all the routines are called from pkg/scm/git/*_test.go.
- CreateLocalGitDirectoryWithSubmodule() calls git.ToSlashCygwin() to ensure the pathname sent to `git submodule add` is valid if running on Windows and the `git` binary is a Cygwin binary (which IMO is the likely case when running on Windows).
- git.ToSlashCygwin() is also called by non-test code.

So:
- The routines can't easily go in _test.go files because they're used across multiple packages.
- Unfortunately, putting them in pkg/test is problematic because it creates a import loop pkg/test->pkg/scm/git (pkg/scm/git already imports pkg/test).

Bearing in mind all of the above, and that these routines use the CommandRunner just like other stiGit functions use, and they are more fiddly than they look when you factor in support for testing on Windows (e.g. use of ToSlashCygwin), I opted to put them where you see them currently.  I think I preferred correctness and non-duplication of code over the cleanliness of the Git interface.

jim-minter:
Will do.  Actually, as part of 
https://github.com/openshift/vagrant-openshift/pull/520 , the Windows 
testing AMI will have everything needed to build this (just Cygwin gcc + 
make), so I'll stop checking in the binary and have it built as part of 
the testing.

pkg/scm/git/git.go
rhcarvalho: What I understand:

- There is no import loop as of the commit introducing this change (41011280c32055608f28b113c5e58aa70f341720), `ToSlashCygwin` is only introduced later in 9db4e2056355bcd0ebf6f3d7c215a2a3e5c0e1f8.

- `ToSlashCygwin` is the cause of the import loop

- `ToSlashCygwin` doesn't need to belong in `pkg/scm/git/git.go`, it is something hardly related to "git".

- All calls to `ToSlashCygwin` are a no-op in non-Windows systems.

My proposal:

- Do not move stuff from `pkg/test` into `pkg/scm/git/git.go`
- Move `ToSlashCygwin` (and `isUsingCygwinGit` and `isUsingCygwinGit`) to its own package. Suggestion: `pkg/scm/git/gitpath`?

Bonus:

Use [build constraints](https://golang.org/pkg/go/build/#hdr-Build_Constraints) and get rid of the "Cygwin" part in the function name, and we could have something like:

```go
// pkg/scm/git/path/path_windows.go
// ----------------------------------

package path

func Normalize(path string) (string, error) {
       if usingCygwinGit {
               cmd := exec.Command("cygpath", path)
               out, err := cmd.Output()
               return strings.TrimRight(string(out), "\n"), err
       }
       return path, nil
}
```

```go
// pkg/scm/git/path/path.go
// ----------------------------------

// +build !windows

package path

func Normalize(path string) (string, error) {
       return path, nil
}
```

Then, wherever it gets called (from `pkg/scm/git/*`, `pkg/test/*`, etc), it would be called as:

```go
path.Normalize(...)
```

(Maybe call the package `gitpath` instead of `path`, for the latter is a stdlib package)


---

@jim-minter WDYT?

pkg/scm/git/clone.go
```diff
@@ -19,86 +22,85 @@ type Clone struct {
...
-		// Otherwise the versions will be incorrect.
-		if hasRef && !config.IgnoreSubmodules {
-			glog.V(2).Infof("Cloning sources (deferring submodule init) into %q", targetSourceDir)
-		} else if !hasRef && !config.IgnoreSubmodules {
-			glog.V(2).Infof("Cloning sources and all Git submodules (recursive) into %q", targetSourceDir)
-		} else {
-			glog.V(2).Infof("Cloning sources (ignoring submodules) into %q", targetSourceDir)
-		}
+	if strings.HasPrefix(config.Source, "file://") {
+		s, _ := ToSlashCygwin(strings.TrimPrefix(config.Source, "file://"))
```
rhcarvalho: @jim-minter, missing error handling?

pkg/scm/git/git.go
```diff
@@ -28,18 +31,50 @@ type Git interface {
...
+}
+
+func isUsingCygwinGit() bool {
+	if runtime.GOOS == "windows" {
+		// If we find the cygpath utility (which translates between UNIX-style and
+		// Windows-style paths) in the same directory as git, assume we're using
+		// Cygwin.
+		cygpath, _ := exec.LookPath("cygpath")
```
rhcarvalho: @jim-minter, just making sure I understand what you meant here... we ignore the error because we rely on the `cygpath` variable being the empty string in case of errors and then it would fail the comparison with the `git` path later? Is that how you reasoned here?

pkg/scm/git/git.go
```diff
@@ -28,18 +31,50 @@ type Git interface {
...
+		git, err := exec.LookPath("git")
+		if err == nil && filepath.Dir(cygpath) == filepath.Dir(git) {
```
rhcarvalho: Unlike to happen in practice..., but if I `cd /path/to/git`, then run `s2i`, we might incorrectly return `true` even when `cypath` doesn't exist (because we ignored the error above).

pkg/scm/git/git.go
```diff
@@ -537,7 +570,66 @@ func (h *stiGit) SubmoduleUpdate(repo string, init, recursive bool) error {
...
+	submodules := []string{}
+	rv := []os.FileInfo{}
+	scanner := bufio.NewScanner(r)
+	for scanner.Scan() {
+		text := scanner.Text()
+		m := rx.FindStringSubmatch(text)
+		if m == nil {
+			return nil, fmt.Errorf("unparsable response %q from git ls-files", text)
+		}
+		mode, _ := strconv.ParseInt(m[1], 8, 0)
```
rhcarvalho: @jim-minter, can we safely ignore the error here?

pkg/scm/git/git.go
```diff
@@ -537,7 +570,66 @@ func (h *stiGit) SubmoduleUpdate(repo string, init, recursive bool) error {
...
 	}
-	return h.runner.RunWithOptions(opts, "git", updateArgs...)
+	return h.RunWithOptions(opts, "git", updateArgs...)
+}
+
+// LsTree returns a slice of os.FileInfo objects populated with the paths and
+// file modes of files known to Git.  This is used on Windows systems where the
+// executable mode metadata is lost on git checkout.
+func (h *stiGit) LsTree(repo, ref string, recursive bool) ([]os.FileInfo, error) {
+	rx := regexp.MustCompile("([0-7]{6}) [^ ]+ [0-9a-f]{40}\t(.*)")
```
rhcarvalho: I'd suggest making this regexp global (i.e., in the package level), so that we don't recompile in every call to LsTree (I see that at least when there are submodules we call this func in a loop/recursively).

pkg/scm/git/clone.go
jim-minter: Thanks, will fix.

pkg/scm/git/git.go
jim-minter: Yes.

pkg/scm/git/git.go
jim-minter: Sharp eyes.  Will fix.

pkg/scm/git/git.go
jim-minter: Will fix.

jim-minter:
@rhcarvalho all suggested changes pushed - thanks for the brainstorming on those test routines.  As a result I've put the ToSlashCygwin bits into util, which I think is a big improvement.  Maybe not quite as smart as what you suggested, but definitely much better.

pkg/scm/git/clone.go
```diff
@@ -19,86 +22,94 @@ type Clone struct {
...
-			originalTargetDir := filepath.Join(config.WorkingDir, api.Source)
-			c.RemoveDirectory(originalTargetDir)
-			// we want to copy entire dir contents, thus we need to use dir/. construct
-			path := filepath.Join(targetSourceDir, config.ContextDir) + string(filepath.Separator) + "."
-			err := c.Copy(path, originalTargetDir)
+		if util.UsingCygwinGit {
+			var err error
+			s, err = util.ToSlashCygwin(s)
 			if err != nil {
+				glog.V(1).Infof("ToSlashCygwin failed: %+v", err)
```
rhcarvalho: Why do we have the `+` here? If the error has an underlying struct type, it will print a struct with field names instead of the error message -- I think that's not what we want in log level 1?

rhcarvalho: > 4. If an operand implements the error interface, the Error method will be invoked to convert the object to a string, which will then be formatted as required by the verb (if any).

(https://golang.org/pkg/fmt/)

So actually `%v` and `%+v` seem to do the same for errors: https://play.golang.org/p/OlpINauKU-

rhcarvalho:
To the extent of what I could review, LGTM. Bravo to @jim-minter!

pkg/scm/git/clone.go
rhcarvalho: //cc @bparees 

`glog.V(1)` is for the user-facing messages, isn't it? If that's the case, then `"ToSlashCygwin failed: ..."` seems a bit too cryptic, referring to a name of an internal function.

pkg/scm/git/clone.go
```diff
@@ -19,86 +22,94 @@ type Clone struct {
...
 	}
-	// we want to copy entire dir contents, thus we need to use dir/. construct
-	path := filepath.Join(config.Source, config.ContextDir) + string(filepath.Separator) + "."
-	if !c.Exists(path) {
-		return nil, errors.NewSourcePathError(path)
+
+	cloneConfig := api.CloneConfig{Quiet: true}
+	err = c.Clone(config.Source, targetSourceDir, cloneConfig)
+	if err != nil {
+		glog.V(1).Infof("Git clone failed: %+v", err)
```
rhcarvalho: The `+` is just misleading, as it doesn't apply here IIUC. Suggestion: `%s` or `%v`.

pkg/scm/git/clone.go
bparees: Given that it's an error, it should probably be V(0) and yes, the message should probably be more illuminating.

bparees:
@jim-minter just address @rhcarvalho's remaining comments on how you're logging error messages (level and format string) and then lgtm.

jim-minter:
@bparees all done, thanks

rhcarvalho:
@jim-minter @bparees I tested this branch to see if it fixed #301 (incremental builds hanging for 2min after "Done extracting tar stream"), but in fact it is making incremental builds worse, they hang for ever for me.

My recommendation would be to split the non-Windows-related changes into a separate PR that we can test further, without blocking the Windows support.

Goroutine dump from a hang (see the issue for more details on what code is being run):

<details>

```
I1129 10:40:41.588211    2159 tar.go:379] Creating /tmp/s2i506683693/upload/artifacts/.cabal/share/x86_64-linux-ghc-8.0.1/server-0.0/response
I1129 10:40:41.588280    2159 tar.go:389] Extracting/writing /tmp/s2i506683693/upload/artifacts/.cabal/share/x86_64-linux-ghc-8.0.1/server-0.0/response
I1129 10:40:41.588394    2159 tar.go:325] Creating directory /tmp/s2i506683693/upload/artifacts/.ghc
I1129 10:40:41.588571    2159 tar.go:325] Creating directory /tmp/s2i506683693/upload/artifacts/.ghc/x86_64-linux-8.0.1
I1129 10:40:41.588695    2159 tar.go:325] Creating directory /tmp/s2i506683693/upload/artifacts/.ghc/x86_64-linux-8.0.1/package.conf.d
I1129 10:40:41.588855    2159 tar.go:333] Creating directory /tmp/s2i506683693/upload/artifacts/.ghc/x86_64-linux-8.0.1/package.conf.d
I1129 10:40:41.588883    2159 tar.go:379] Creating /tmp/s2i506683693/upload/artifacts/.ghc/x86_64-linux-8.0.1/package.conf.d/groups-0.4.0.0-53wMJ09LBR64aEJqWveBMG.conf
I1129 10:40:41.588948    2159 tar.go:389] Extracting/writing /tmp/s2i506683693/upload/artifacts/.ghc/x86_64-linux-8.0.1/package.conf.d/groups-0.4.0.0-53wMJ09LBR64aEJqWveBMG.conf
I1129 10:40:41.589106    2159 tar.go:357] Done extracting tar stream





^\I1129 10:53:57.521299    2159 docker.go:926] Removing container "3f978659a9b321357f457edcfda75d4d20d6c73422f3cc57ad8556e1053fb271" ...
I1129 10:53:58.127071    2159 docker.go:930] Removed container "3f978659a9b321357f457edcfda75d4d20d6c73422f3cc57ad8556e1053fb271"
goroutine 44 [running]:
github.com/openshift/source-to-image/pkg/docker.(*stiDocker).RunContainer.func2(0xc15a20, 0xc4203a9b80)
        /home/rodolfo/openshift/src/github.com/openshift/source-to-image/_output/local/go/src/github.com/openshift/source-to-image/pkg/docker/docker.go:936 +0xf4
github.com/openshift/source-to-image/pkg/util/interrupt.(*Handler).signal.func1()
        /home/rodolfo/openshift/src/github.com/openshift/source-to-image/_output/local/go/src/github.com/openshift/source-to-image/pkg/util/interrupt/interrupt.go:102 +0xb5
sync.(*Once).Do(0xc420329820, 0xc4202e7f30)
        /usr/local/go/src/sync/once.go:44 +0xdb
github.com/openshift/source-to-image/pkg/util/interrupt.(*Handler).signal(0xc420329800, 0xc15a20, 0xc4203a9b80)
        /home/rodolfo/openshift/src/github.com/openshift/source-to-image/_output/local/go/src/github.com/openshift/source-to-image/pkg/util/interrupt/interrupt.go:103 +0x60
github.com/openshift/source-to-image/pkg/util/interrupt.(*Handler).Run.func1(0xc4202dd4a0, 0xc420329800)
        /home/rodolfo/openshift/src/github.com/openshift/source-to-image/_output/local/go/src/github.com/openshift/source-to-image/pkg/util/interrupt/interrupt.go:75 +0x82
created by github.com/openshift/source-to-image/pkg/util/interrupt.(*Handler).Run
        /home/rodolfo/openshift/src/github.com/openshift/source-to-image/_output/local/go/src/github.com/openshift/source-to-image/pkg/util/interrupt/interrupt.go:76 +0x103

goroutine 1 [chan receive]:
github.com/openshift/source-to-image/pkg/docker.(*stiDocker).holdHijackedConnection(0xc420302280, 0x870800, 0x0, 0x0, 0xc11d60, 0xc420098098, 0xc11d60, 0xc4200980a8, 0xc19460, 0xc420028558, ...)
        /home/rodolfo/openshift/src/github.com/openshift/source-to-image/_output/local/go/src/github.com/openshift/source-to-image/pkg/docker/docker.go:857 +0x170
github.com/openshift/source-to-image/pkg/docker.(*stiDocker).RunContainer.func3(0x0, 0x0)
        /home/rodolfo/openshift/src/github.com/openshift/source-to-image/_output/local/go/src/github.com/openshift/source-to-image/pkg/docker/docker.go:977 +0x7c4
github.com/openshift/source-to-image/pkg/util/interrupt.(*Handler).Run(0xc420329800, 0xc4201830f0, 0x0, 0x0)
        /home/rodolfo/openshift/src/github.com/openshift/source-to-image/_output/local/go/src/github.com/openshift/source-to-image/pkg/util/interrupt/interrupt.go:78 +0x133
github.com/openshift/source-to-image/pkg/docker.(*stiDocker).RunContainer(0xc420302280, 0x7fff84862674, 0x4, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, ...)
        /home/rodolfo/openshift/src/github.com/openshift/source-to-image/_output/local/go/src/github.com/openshift/source-to-image/pkg/docker/docker.go:1007 +0x904
github.com/openshift/source-to-image/pkg/build/strategies/sti.(*STI).Save(0xc4202ad000, 0xc42029a000, 0x1f, 0xc4202eac80)
        /home/rodolfo/openshift/src/github.com/openshift/source-to-image/_output/local/go/src/github.com/openshift/source-to-image/pkg/build/strategies/sti/sti.go:470 +0xaa7
github.com/openshift/source-to-image/pkg/build/strategies/sti.(*STI).Build(0xc4202ad000, 0xc42029a000, 0x0, 0x0, 0x0)
        /home/rodolfo/openshift/src/github.com/openshift/source-to-image/_output/local/go/src/github.com/openshift/source-to-image/pkg/build/strategies/sti/sti.go:200 +0xd38
main.newCmdBuild.func1(0xc42008b440, 0xc4201dccc0, 0x3, 0x6)
        /home/rodolfo/openshift/src/github.com/openshift/source-to-image/_output/local/go/src/github.com/openshift/source-to-image/cmd/s2i/main.go:156 +0x75c
github.com/openshift/source-to-image/vendor/github.com/spf13/cobra.(*Command).execute(0xc42008b440, 0xc4201dcc00, 0x6, 0x6, 0xc42008b440, 0xc4201dcc00)
        /home/rodolfo/openshift/src/github.com/openshift/source-to-image/_output/local/go/src/github.com/openshift/source-to-image/vendor/github.com/spf13/cobra/command.go:603 +0x439
github.com/openshift/source-to-image/vendor/github.com/spf13/cobra.(*Command).ExecuteC(0xc42008afc0, 0xc420183eb0, 0xc4202592d0, 0xc42008bd40)
        /home/rodolfo/openshift/src/github.com/openshift/source-to-image/_output/local/go/src/github.com/openshift/source-to-image/vendor/github.com/spf13/cobra/command.go:689 +0x367
github.com/openshift/source-to-image/vendor/github.com/spf13/cobra.(*Command).Execute(0xc42008afc0, 0xc420183ea8, 0x1)
        /home/rodolfo/openshift/src/github.com/openshift/source-to-image/_output/local/go/src/github.com/openshift/source-to-image/vendor/github.com/spf13/cobra/command.go:648 +0x2b
main.main()
        /home/rodolfo/openshift/src/github.com/openshift/source-to-image/_output/local/go/src/github.com/openshift/source-to-image/cmd/s2i/main.go:414 +0x660

goroutine 17 [syscall, locked to thread]:
runtime.goexit()
        /usr/local/go/src/runtime/asm_amd64.s:2086 +0x1

goroutine 5 [chan receive]:
github.com/openshift/source-to-image/vendor/github.com/golang/glog.(*loggingT).flushDaemon(0xc2c700)
        /home/rodolfo/openshift/src/github.com/openshift/source-to-image/_output/local/go/src/github.com/openshift/source-to-image/vendor/github.com/golang/glog/glog.go:882 +0x7a
created by github.com/openshift/source-to-image/vendor/github.com/golang/glog.init.1
        /home/rodolfo/openshift/src/github.com/openshift/source-to-image/_output/local/go/src/github.com/openshift/source-to-image/vendor/github.com/golang/glog/glog.go:410 +0x21d

goroutine 6 [syscall]:
os/signal.signal_recv(0xc15a20)
        /usr/local/go/src/runtime/sigqueue.go:116 +0x157
os/signal.loop()
        /usr/local/go/src/os/signal/signal_unix.go:22 +0x22
created by os/signal.init.1
        /usr/local/go/src/os/signal/signal_unix.go:28 +0x41

goroutine 9 [IO wait]:
net.runtime_pollWait(0x7fed21ea02a0, 0x72, 0x3)
        /usr/local/go/src/runtime/netpoll.go:160 +0x59
net.(*pollDesc).wait(0xc420292fb0, 0x72, 0xc4200349d0, 0xc420012068)
        /usr/local/go/src/net/fd_poll_runtime.go:73 +0x38
net.(*pollDesc).waitRead(0xc420292fb0, 0xc13ba0, 0xc420012068)
        /usr/local/go/src/net/fd_poll_runtime.go:78 +0x34
net.(*netFD).Read(0xc420292f50, 0xc4202a2000, 0x1000, 0x1000, 0x0, 0xc13ba0, 0xc420012068)
        /usr/local/go/src/net/fd_unix.go:243 +0x1a1
net.(*conn).Read(0xc4200283f8, 0xc4202a2000, 0x1000, 0x1000, 0x0, 0x0, 0x0)
        /usr/local/go/src/net/net.go:173 +0x70
net/http.(*persistConn).Read(0xc420296400, 0xc4202a2000, 0x1000, 0x1000, 0x638df0, 0xc420034b58, 0x4099fd)
        /usr/local/go/src/net/http/transport.go:1261 +0x154
bufio.(*Reader).fill(0xc4201dd320)
        /usr/local/go/src/bufio/bufio.go:97 +0x10c
bufio.(*Reader).Peek(0xc4201dd320, 0x1, 0xc420034bbd, 0x1, 0x0, 0xc4201dd800, 0x0)
        /usr/local/go/src/bufio/bufio.go:129 +0x62
net/http.(*persistConn).readLoop(0xc420296400)
        /usr/local/go/src/net/http/transport.go:1418 +0x1a1
created by net/http.(*Transport).dialConn
        /usr/local/go/src/net/http/transport.go:1062 +0x4e9

goroutine 10 [select]:
net/http.(*persistConn).writeLoop(0xc420296400)
        /usr/local/go/src/net/http/transport.go:1646 +0x3bd
created by net/http.(*Transport).dialConn
        /usr/local/go/src/net/http/transport.go:1063 +0x50e

goroutine 43 [select, locked to thread]:
runtime.gopark(0x871248, 0x0, 0x833662, 0x6, 0x18, 0x2)
        /usr/local/go/src/runtime/proc.go:259 +0x13a
runtime.selectgoImpl(0xc4202d8f30, 0x0, 0x18)
        /usr/local/go/src/runtime/select.go:423 +0x11d9
runtime.selectgo(0xc4202d8f30)
        /usr/local/go/src/runtime/select.go:238 +0x1c
runtime.ensureSigM.func1()
        /usr/local/go/src/runtime/signal1_unix.go:304 +0x2f3
runtime.goexit()
        /usr/local/go/src/runtime/asm_amd64.s:2086 +0x1

goroutine 16 [IO wait]:
net.runtime_pollWait(0x7fed21ea01e0, 0x72, 0x5)
        /usr/local/go/src/runtime/netpoll.go:160 +0x59
net.(*pollDesc).wait(0xc4202935d0, 0x72, 0xc42003b9d0, 0xc420012068)
        /usr/local/go/src/net/fd_poll_runtime.go:73 +0x38
net.(*pollDesc).waitRead(0xc4202935d0, 0xc13ba0, 0xc420012068)
        /usr/local/go/src/net/fd_poll_runtime.go:78 +0x34
net.(*netFD).Read(0xc420293570, 0xc4202de000, 0x1000, 0x1000, 0x0, 0xc13ba0, 0xc420012068)
        /usr/local/go/src/net/fd_unix.go:243 +0x1a1
net.(*conn).Read(0xc420028450, 0xc4202de000, 0x1000, 0x1000, 0x0, 0x0, 0x0)
        /usr/local/go/src/net/net.go:173 +0x70
net/http.(*persistConn).Read(0xc420296500, 0xc4202de000, 0x1000, 0x1000, 0x638df0, 0xc42003bb58, 0x4099fd)
        /usr/local/go/src/net/http/transport.go:1261 +0x154
bufio.(*Reader).fill(0xc4202dc120)
        /usr/local/go/src/bufio/bufio.go:97 +0x10c
bufio.(*Reader).Peek(0xc4202dc120, 0x1, 0xc42003bbbd, 0x1, 0x0, 0xc4202dc180, 0x0)
        /usr/local/go/src/bufio/bufio.go:129 +0x62
net/http.(*persistConn).readLoop(0xc420296500)
        /usr/local/go/src/net/http/transport.go:1418 +0x1a1
created by net/http.(*Transport).dialConn
        /usr/local/go/src/net/http/transport.go:1062 +0x4e9

goroutine 34 [select]:
net/http.(*persistConn).writeLoop(0xc420296500)
        /usr/local/go/src/net/http/transport.go:1646 +0x3bd
created by net/http.(*Transport).dialConn
        /usr/local/go/src/net/http/transport.go:1063 +0x50e

goroutine 53 [semacquire]:
sync.runtime_notifyListWait(0xc4203061c0, 0x0)
        /usr/local/go/src/runtime/sema.go:267 +0x122
sync.(*Cond).Wait(0xc4203061b0)
        /usr/local/go/src/sync/cond.go:57 +0x80
io.(*pipe).read(0xc420306180, 0xc42037c000, 0x1000, 0x1000, 0x0, 0x0, 0x0)
        /usr/local/go/src/io/pipe.go:47 +0x102
io.(*PipeReader).Read(0xc4200980a0, 0xc42037c000, 0x1000, 0x1000, 0x43f91a, 0xc4202d4728, 0x2)
        /usr/local/go/src/io/pipe.go:129 +0x4c
bufio.(*Reader).fill(0xc4202d46d8)
        /usr/local/go/src/bufio/bufio.go:97 +0x10c
bufio.(*Reader).ReadSlice(0xc4202d46d8, 0xa, 0x2b, 0x7fed21e53320, 0x4547a0, 0x7fed21ef3960, 0x2b)
        /usr/local/go/src/bufio/bufio.go:330 +0xb5
bufio.(*Reader).ReadBytes(0xc4202d46d8, 0x7a370a, 0x1, 0xc42037c000, 0x2, 0x2, 0xc4202d4788)
        /usr/local/go/src/bufio/bufio.go:408 +0x66
bufio.(*Reader).ReadString(0xc4202d46d8, 0x100a, 0x1000, 0xc42037c000, 0x1000, 0x1000)
        /usr/local/go/src/bufio/bufio.go:448 +0x38
github.com/openshift/source-to-image/pkg/docker.StreamContainerIO.func1(0xc11d20, 0xc4200980a0, 0x870828, 0x0, 0xc42035e660)
        /home/rodolfo/openshift/src/github.com/openshift/source-to-image/_output/local/go/src/github.com/openshift/source-to-image/pkg/docker/util.go:167 +0xa8
created by github.com/openshift/source-to-image/pkg/docker.StreamContainerIO
        /home/rodolfo/openshift/src/github.com/openshift/source-to-image/_output/local/go/src/github.com/openshift/source-to-image/pkg/docker/util.go:182 +0x89

goroutine 22 [IO wait]:
net.runtime_pollWait(0x7fed21ea0120, 0x72, 0x6)
        /usr/local/go/src/runtime/netpoll.go:160 +0x59
net.(*pollDesc).wait(0xc4202f0220, 0x72, 0xc4200359d0, 0xc420012068)
        /usr/local/go/src/net/fd_poll_runtime.go:73 +0x38
net.(*pollDesc).waitRead(0xc4202f0220, 0xc13ba0, 0xc420012068)
        /usr/local/go/src/net/fd_poll_runtime.go:78 +0x34
net.(*netFD).Read(0xc4202f01c0, 0xc42030a000, 0x1000, 0x1000, 0x0, 0xc13ba0, 0xc420012068)
        /usr/local/go/src/net/fd_unix.go:243 +0x1a1
net.(*conn).Read(0xc420098038, 0xc42030a000, 0x1000, 0x1000, 0x0, 0x0, 0x0)
        /usr/local/go/src/net/net.go:173 +0x70
net/http.(*persistConn).Read(0xc42006a400, 0xc42030a000, 0x1000, 0x1000, 0x638df0, 0xc420035b58, 0x4099fd)
        /usr/local/go/src/net/http/transport.go:1261 +0x154
bufio.(*Reader).fill(0xc4200926c0)
        /usr/local/go/src/bufio/bufio.go:97 +0x10c
bufio.(*Reader).Peek(0xc4200926c0, 0x1, 0xc420035bbd, 0x1, 0x0, 0xc4202dc660, 0x0)
        /usr/local/go/src/bufio/bufio.go:129 +0x62
net/http.(*persistConn).readLoop(0xc42006a400)
        /usr/local/go/src/net/http/transport.go:1418 +0x1a1
created by net/http.(*Transport).dialConn
        /usr/local/go/src/net/http/transport.go:1062 +0x4e9

goroutine 23 [select]:
net/http.(*persistConn).writeLoop(0xc42006a400)
        /usr/local/go/src/net/http/transport.go:1646 +0x3bd
created by net/http.(*Transport).dialConn
        /usr/local/go/src/net/http/transport.go:1063 +0x50e

goroutine 27 [IO wait]:
net.runtime_pollWait(0x7fed21ea0060, 0x72, 0x7)
        /usr/local/go/src/runtime/netpoll.go:160 +0x59
net.(*pollDesc).wait(0xc4202f0450, 0x72, 0xc4200379d0, 0xc420012068)
        /usr/local/go/src/net/fd_poll_runtime.go:73 +0x38
net.(*pollDesc).waitRead(0xc4202f0450, 0xc13ba0, 0xc420012068)
        /usr/local/go/src/net/fd_poll_runtime.go:78 +0x34
net.(*netFD).Read(0xc4202f03f0, 0xc420318000, 0x1000, 0x1000, 0x0, 0xc13ba0, 0xc420012068)
        /usr/local/go/src/net/fd_unix.go:243 +0x1a1
net.(*conn).Read(0xc420098080, 0xc420318000, 0x1000, 0x1000, 0x0, 0x0, 0x0)
        /usr/local/go/src/net/net.go:173 +0x70
net/http.(*persistConn).Read(0xc42006a500, 0xc420318000, 0x1000, 0x1000, 0x30, 0xc420037b58, 0x43ea0c)
        /usr/local/go/src/net/http/transport.go:1261 +0x154
bufio.(*Reader).fill(0xc4200930e0)
        /usr/local/go/src/bufio/bufio.go:97 +0x10c
bufio.(*Reader).Peek(0xc4200930e0, 0x1, 0x0, 0x1, 0x0, 0xc420417440, 0x0)
        /usr/local/go/src/bufio/bufio.go:129 +0x62
net/http.(*persistConn).readLoop(0xc42006a500)
        /usr/local/go/src/net/http/transport.go:1418 +0x1a1
created by net/http.(*Transport).dialConn
        /usr/local/go/src/net/http/transport.go:1062 +0x4e9

goroutine 28 [select]:
net/http.(*persistConn).writeLoop(0xc42006a500)
        /usr/local/go/src/net/http/transport.go:1646 +0x3bd
created by net/http.(*Transport).dialConn
        /usr/local/go/src/net/http/transport.go:1063 +0x50e

goroutine 40 [IO wait]:
net.runtime_pollWait(0x7fed21e9ffa0, 0x72, 0x8)
        /usr/local/go/src/runtime/netpoll.go:160 +0x59
net.(*pollDesc).wait(0xc4203206f0, 0x72, 0xc4202e59d0, 0xc420012068)
        /usr/local/go/src/net/fd_poll_runtime.go:73 +0x38
net.(*pollDesc).waitRead(0xc4203206f0, 0xc13ba0, 0xc420012068)
        /usr/local/go/src/net/fd_poll_runtime.go:78 +0x34
net.(*netFD).Read(0xc420320690, 0xc42034c000, 0x1000, 0x1000, 0x0, 0xc13ba0, 0xc420012068)
        /usr/local/go/src/net/fd_unix.go:243 +0x1a1
net.(*conn).Read(0xc420028540, 0xc42034c000, 0x1000, 0x1000, 0x0, 0x0, 0x0)
        /usr/local/go/src/net/net.go:173 +0x70
net/http.(*persistConn).Read(0xc420296600, 0xc42034c000, 0x1000, 0x1000, 0x638df0, 0xc4202e5b58, 0x4099fd)
        /usr/local/go/src/net/http/transport.go:1261 +0x154
bufio.(*Reader).fill(0xc4202dd260)
        /usr/local/go/src/bufio/bufio.go:97 +0x10c
bufio.(*Reader).Peek(0xc4202dd260, 0x1, 0xc4202e5bbd, 0x1, 0x0, 0xc42035e360, 0x0)
        /usr/local/go/src/bufio/bufio.go:129 +0x62
net/http.(*persistConn).readLoop(0xc420296600)
        /usr/local/go/src/net/http/transport.go:1418 +0x1a1
created by net/http.(*Transport).dialConn
        /usr/local/go/src/net/http/transport.go:1062 +0x4e9

goroutine 41 [select]:
net/http.(*persistConn).writeLoop(0xc420296600)
        /usr/local/go/src/net/http/transport.go:1646 +0x3bd
created by net/http.(*Transport).dialConn
        /usr/local/go/src/net/http/transport.go:1063 +0x50e

goroutine 59 [semacquire]:
sync.runtime_notifyListWait(0xc420306138, 0xc400000246)
        /usr/local/go/src/runtime/sema.go:267 +0x122
sync.(*Cond).Wait(0xc420306128)
        /usr/local/go/src/sync/cond.go:57 +0x80
io.(*pipe).write(0xc4203060c0, 0xc4203cc008, 0x8000, 0x8001, 0x0, 0x0, 0x0)
        /usr/local/go/src/io/pipe.go:89 +0x191
io.(*PipeWriter).Write(0xc420098098, 0xc4203cc008, 0x8000, 0x8001, 0x1, 0x0, 0x0)
        /usr/local/go/src/io/pipe.go:156 +0x4c
github.com/openshift/source-to-image/vendor/github.com/docker/docker/pkg/stdcopy.StdCopy(0xc11d60, 0xc420098098, 0xc11d60, 0xc4200980a8, 0xc11060, 0xc4202dd860, 0x0, 0x0, 0x0)
        /home/rodolfo/openshift/src/github.com/openshift/source-to-image/_output/local/go/src/github.com/openshift/source-to-image/vendor/github.com/docker/docker/pkg/stdcopy/stdcopy.go:161 +0x429
github.com/openshift/source-to-image/pkg/docker.(*stiDocker).redirectResponseToOutputStream(0xc420302280, 0x0, 0xc11d60, 0xc420098098, 0xc11d60, 0xc4200980a8, 0xc11060, 0xc4202dd860, 0x0, 0x0)
        /home/rodolfo/openshift/src/github.com/openshift/source-to-image/_output/local/go/src/github.com/openshift/source-to-image/pkg/docker/docker.go:826 +0xd2
github.com/openshift/source-to-image/pkg/docker.(*stiDocker).holdHijackedConnection.func1(0xc42035f0e0, 0xc420302280, 0x7cad00, 0xc11d60, 0xc420098098, 0xc11d60, 0xc4200980a8, 0xc42039cd00)
        /home/rodolfo/openshift/src/github.com/openshift/source-to-image/_output/local/go/src/github.com/openshift/source-to-image/pkg/docker/docker.go:839 +0x8c
created by github.com/openshift/source-to-image/pkg/docker.(*stiDocker).holdHijackedConnection
        /home/rodolfo/openshift/src/github.com/openshift/source-to-image/_output/local/go/src/github.com/openshift/source-to-image/pkg/docker/docker.go:840 +0x10c
+ docker run --name test -d -p 8080:8080 -u 1001 test
/usr/bin/docker-current: Error response from daemon: Conflict. The name "/test" is already in use by container b4ccdd34918f05586ff580e4b8684c76a77bca04d625c102aaad0c7419ee1853. You have to remove (or rename) that container to be able to reuse that name..
See '/usr/bin/docker-current run --help'.
+ sleep 0.01
+ curl -sS localhost:8080
+ grep -q Welcome
curl: (7) Failed to connect to localhost port 8080: Connection refused
+ echo '  * create app FAILED'
  * create app FAILED
+ exit 1
```

</details>

(FWIW, I haven't had time to analyze the goroutine dump)

bparees:
> @jim-minter @bparees I tested this branch to see if it fixed #301 (incremental builds hanging for 2min after "Done extracting tar stream"), but in fact it is making incremental builds worse, they hang for ever for me.

are there no incremental build tests in our bucket?  we need to fix that.

rhcarvalho:
> are there no incremental build tests in our bucket? we need to fix that.

IIRC there are, but they don't capture the problem.

jim-minter:
@bparees ready to go

pkg/scm/git/clone.go
```diff
@@ -19,86 +22,94 @@ type Clone struct {
...
 	}
-	// we want to copy entire dir contents, thus we need to use dir/. construct
-	path := filepath.Join(config.Source, config.ContextDir) + string(filepath.Separator) + "."
-	if !c.Exists(path) {
-		return nil, errors.NewSourcePathError(path)
+
+	cloneConfig := api.CloneConfig{Quiet: true}
+	err = c.Clone(config.Source, targetSourceDir, cloneConfig)
+	if err != nil {
+		glog.V(1).Infof("Git clone failed: %v", err)
```
bparees: should follow the pattern of "glog.V(0).Infof("error: Uploading directory to container failed: %v", err)"

jim-minter:
@bparees your 2 additional changes pushed

bparees:
[merge]

PR #626, opened 2016-11-01
remove useless vendored test files

PI-Victor:
we don't run these anywhere, right?
i know the godep tool doesn't vendor test files https://github.com/tools/godep#how-to-use-godep-with-a-new-project

so i'm assuming they were added manually at some point.
[test]


bparees:
godep does vendor them, we didn't add them explicitly.

bparees:
[merge]

PI-Victor:
awesome, my most useful contribution to opensource. 😆

PR #628, opened 2016-11-01
Invoke callback URL when build has failed

### LINKED ISSUE #522: Callback URL doesn't invoked when build fails
[Callback URL section](https://github.com/openshift/source-to-image/blob/master/docs/cli.md#callback-url) a bit outdated:
- `labels` field is missing
- description of `success` says that this field `indicating the result of the build process (true or false)` but this callback never gets executed if build failed. It only executed for a successful builds and hence this field always has value `true` (I'm not sure whether it's a bug in the code or documentation)

php-coder:
Fixes #522

PTAL @bparees

### DIFF
--- a/pkg/build/strategies/sti/postexecutorstep_test.go
+++ b/pkg/build/strategies/sti/postexecutorstep_test.go
@@ -6,7 +6,6 @@ import (
 	"testing"
 
 	"github.com/openshift/source-to-image/pkg/docker"
-	"github.com/openshift/source-to-image/pkg/test"
 )
 
 func TestStorePreviousImageStep(t *testing.T) {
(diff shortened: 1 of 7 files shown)


pkg/util/callback.go
```diff
@@ -34,16 +34,19 @@ func (c *callbackInvoker) ExecuteCallback(callbackURL string, success bool, labe
...
 
-	d := map[string]interface{}{
-		"labels":  labels,
+	data := map[string]interface{}{
 		"payload": buf.String(),
 		"success": success,
 	}
 
+	if len(labels) > 0 {
+		data["labels"] = labels
```
php-coder: With this check when build failed we don't send `labels` field at all (before it was `null`).

pkg/build/strategies/sti/postexecutorstep_test.go
```diff
@@ -235,49 +235,3 @@ func TestReportSuccessStep(t *testing.T) {
...
-	}
-
-	if !reflect.DeepEqual(callbackInvoker.Messages, expectedMessages) {
-		t.Errorf("should invoke ExecuteCallback(Messages=%v) but invoked with %v", expectedMessages, callbackInvoker.Messages)
-	}
-
-	if !reflect.DeepEqual(callbackInvoker.Labels, expectedLabels) {
-		t.Errorf("should invoke ExecuteCallback(Labels=%v) but invoked with %v", expectedLabels, callbackInvoker.Labels)
-	}
-}
```
bparees: is there any testing left for the invokecallback functionality?

php-coder: Yes, and these tests are failing now. I'm going to update them.

php-coder:
@bparees Overall it's OK?

bparees:
> @bparees Overall it's OK?

as long as something is still testing the callback functionality, yes

php-coder:
@bparees PTAL again.

bparees:
[merge]

PR #630, opened 2016-11-01
Remove invalid build constraint

rhcarvalho:
As fas as the git history can tell, the constraint was never referenced
by any script, never used.

```
$ go vet test/integration/integration_test.go
test/integration/integration_test.go:1: invalid non-alphanumeric build
constraint: integration,!no-docker
```

### DIFF
--- a/test/integration/integration_test.go
+++ b/test/integration/integration_test.go
@@ -1,4 +1,4 @@
-// +build integration,!no-docker
+// +build integration
 
 package integration
 


rhcarvalho:
@bparees PTAL

bparees:
i think it was probably a travis thing back when travis couldn't run tests that needed docker?

[merge]

rhcarvalho:
> i think it was probably a travis thing back when travis couldn't run tests that needed docker?

I don't know, maybe. Currently we don't run integration tests in Travis. (and to be sure, are we running it with our bot on merge?)

bparees:
> and to be sure, are we running it with our bot on merge?

yes, the vagrant plugin run_sti_tests command is running them explicitly, though they aren't a part of our Makefile.

PR #631, opened 2016-11-02
Fix hang in s2i strategy by avoiding wait on an error

csrwng:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1390749
If an error occurs in RunContainer that is not a timeout error or a container error, then we wait indefinitely instead of returning an error.

I was able to easily reproduce the hang by simulating errors at different points in RunContainer. I also didn't see a case where a non-nil error should result in us still waiting for the WaitGroup to be done.

### DIFF
--- a/pkg/build/strategies/sti/sti.go
+++ b/pkg/build/strategies/sti/sti.go
@@ -640,12 +640,14 @@ func (builder *STI) Execute(command string, user string, config *api.Config) err
 		errReader.Close()
 		return errors.NewContainerError(config.BuilderImage, e.ErrorCode, errOutput)
 	}
-	// Do not wait for source input if the container times out.
+	// Do not wait for source input if there was an error running the container
 	// FIXME: this potentially leaks a goroutine.
-	if !util.IsTimeoutError(err) {
-		wg.Wait()
+	if err != nil {
+		return err
 	}
-	return err
+
+	wg.Wait()
+	return nil
 }
 
 func (builder *STI) initPostExecutorSteps() {
--- a/pkg/build/strategies/sti/sti_test.go
+++ b/pkg/build/strategies/sti/sti_test.go
@@ -919,6 +919,17 @@ func TestExecuteOK(t *testing.T) {
 	}
 }
 
+func TestExecuteRunContainerError(t *testing.T) {
+	rh := newFakeSTI(&FakeSTI{})
+	fd := rh.docker.(*docker.FakeDocker)
+	runContainerError := fmt.Errorf("an error")
+	fd.RunContainerError = runContainerError
+	err := rh.Execute("test-command", "", rh.config)
+	if err != runContainerError {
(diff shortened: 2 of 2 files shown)


csrwng:
@gabemontero @rhcarvalho ptal
@bparees fyi

csrwng:
[test]

gabemontero:
Yeah I took a pass through RunContainer, and could not see an error that was returned from it that we would still want to wait on.  I'm still a little worried that we are forgetting some history here as to why it was waiting on non-timeout errors.

But pending my paranoia, I at least am GTM.

rhcarvalho:
@csrwng complicated as it may be, do you think you can reproduce the problem with a test case? We'd benefit from having it.

csrwng:
I'll work on the test case ... it also looks like I broke a unit test.

pkg/build/strategies/sti/sti.go
```diff
@@ -642,7 +642,7 @@ func (builder *STI) Execute(command string, user string, config *api.Config) err
 	}
 	// Do not wait for source input if the container times out.
 	// FIXME: this potentially leaks a goroutine.
-	if !util.IsTimeoutError(err) {
+	if err != nil {
```
rhcarvalho: *Disclaimer: I'd need some good time to re-understand all the little nuances here.*

---

Reading this feels really strange "if RunContainer returns a non-nil error, then wait"?!

Where "wait" means "if config.LayeredBuild is false, wait until 'source injection' is complete"...

IIRC the original problem is that 'source injection' misbehaves -- maybe it tended to hang, or maybe we proceeded too early before it was finished. To what I can understand there's a snow-ball effect going on here.

@csrwng @gabemontero are you able to reason about the flow here in a cleaner way?

csrwng: @rhcarvalho that's my bad
It should really read if err == nil then wait... otherwise return the error
Right code now pushed

csrwng:
[test]

pkg/build/strategies/sti/sti.go
```diff
@@ -642,7 +642,7 @@ func (builder *STI) Execute(command string, user string, config *api.Config) err
 	}
 	// Do not wait for source input if the container times out.
 	// FIXME: this potentially leaks a goroutine.
-	if !util.IsTimeoutError(err) {
+	if err == nil {
```
rhcarvalho: Hmm, can I propose the idiom of indenting error flow and keeping the happy path vertical?

```go
if err != nil {
	return err
}

wg.Wait()
return nil
```

(and adjust the comment above the code accordingly)

WDYT?

csrwng: sgtm

csrwng:
Updated code and added unit test

bparees:
lgtm [merge]

PR #636, opened 2016-11-15
Corrected save-artifacts example to prevent tar stream corruption.

clarkhale:
The pushd and popd commands in the example for save-artifacts example emit output to standard out.  This will corrupt the tar stream, and prevent the reuse of the saved artifacts.


bparees:
thanks!
[merge]

PR #637, opened 2016-11-15
[WIP] add build substatus

PI-Victor:
@bparees what do you think about this?

### DIFF
--- a/pkg/api/types.go
+++ b/pkg/api/types.go
@@ -6,6 +6,7 @@ import (
 	"net/url"
 	"path/filepath"
 	"strings"
+	"time"
 
 	utilglog "github.com/openshift/source-to-image/pkg/util/glog"
 
@@ -343,11 +344,55 @@ type Result struct {
 
 // BuildInfo holds information about a particular step in the build process.
 type BuildInfo struct {
+	// StepInfo holds information about build steps.
+	StepInfo []BuildStepInfo
+
 	// FailureReason is a camel case reason that is used by the machine to reply
 	// back to the OpenShift builder with information why any of the steps in the
 	// build, failed.
 	FailureReason FailureReason
 }
+
+// BuildStepInfo holds information about a build step.
+type BuildStepInfo struct {
+	// StepName is the name of the build step.
+	Name StepName
+
+	// StartTime is the start time of the build step in the build process.
+	StartTime time.Time
+	// StopTime is the time the build step has finished (Successfully or not) in
+	// the build process.
+	StopTime time.Time
+}
+
+// StepName is the name of a build step.
+type StepName string
+
+const (
+	// AssembleScriptStep is the build step when the Assemble script is being
+	// run.
+	AssembleScriptStep StepName = "AssembleScriptStep"
+
+	// GatherInputContentStep is the build step when we try to fetch the source for the
+	// build.
+	GatherInputContentStep StepName = "GatherInputContentStep"
+
+	// PullBuilderImageStep is the build step that is associated with pulling the
+	// builder image.
+	PullBuilderImageStep StepName = "PullBuilderImageStep"
+
+	// CommitContainerStep is the build step that is associated with commiting the
+	// container to the builder image.
+	CommitContainerStep StepName = "CommitContainerStep"
+
+	// PullPreviousImageStep is the build step that occurs when we try to pull a
+	// previous image for a incremental build.
+	PullPreviousImageStep StepName = "PullPreviousImageStep"
+
+	// PullRuntimeImageStep is the build step that occurs when we try to pull the
+	// runtime image.
+	PullRuntimeImageStep StepName = "PullRuntimeImageStep"
+)
 
 // StepFailureReason holds the type of failure that occurred during the build
 // process.
--- a/pkg/build/strategies/layered/layered.go
+++ b/pkg/build/strategies/layered/layered.go
@@ -128,7 +128,6 @@ func (builder *Layered) CreateDockerfile(config *api.Config) error {
 // success/failure details.
 func (builder *Layered) Build(config *api.Config) (*api.Result, error) {
 	buildResult := &api.Result{}
-
 	if config.HasOnBuild && config.BlockOnBuild {
 		buildResult.BuildInfo.FailureReason = utilstatus.NewFailureReason(
 			utilstatus.ReasonOnBuildForbidden,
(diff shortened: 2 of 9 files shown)


pkg/api/types.go
```diff
@@ -338,12 +339,41 @@ type Result struct {
...
+	// run.
+	AssembleScriptStep StepName = "AssembleScriptStep"
+
+	// FetchSourceStep is the build step when we try to fetch the source for the
+	// build.
+	FetchSourceStep StepName = "FetchSourceStep"
+
+	// PullBuilderImageStep is the build step that is associated with pulling the
+	// builder image.
+	PullBuilderImageStep StepName = "PullBuilderImageStep"
```
bparees: needs a commitcontainer step

pkg/api/types.go
```diff
@@ -338,12 +339,41 @@ type Result struct {
...
 	FailureReason FailureReason
 }
 
+// BuildStepInfo holds information about a build step.
+type BuildStepInfo struct {
+	// StepName is the name of the build step.
+	StepName StepName
+
+	// StepDuration is the duration of a build step.
+	StepDuration time.Duration
```
bparees: i think we might want the start time and the stop time instead of just the duration.  clients can calculate the duration from the start/stop time if they want, or display the actual start/stop time if desired.

pkg/api/types.go
PI-Victor: i'll add the rest. it was just a sample to see if this is the way to go.

bparees: it looks reasonable to me so far.

PI-Victor:
@bparees i'm not going to do this now, but would it make sense to show, here in s2i after the build has finished successfully, a summary of the build steps? (future PR)

PI-Victor:
p.s. just thought it's worth mentioning. if the step fails e.g.: get builder image, then the timeStart/timeStop will be empty.

bparees:
> i'm not going to do this now, but would it make sense to show, here in s2i after the build has finished successfully, a summary of the build steps?

no.  certainly not in s2i.  *maybe* from within the origin code that wraps the build invocation, but even that i'm not a big fan of.

> p.s. just thought it's worth mentioning. if the step fails e.g.: get builder image, then the timeStart/timeStop will be empty.

why?  a failed step still has a start and finish time...

pkg/build/strategies/layered/layered.go
```diff
@@ -150,18 +151,20 @@ func (builder *Layered) Build(config *api.Config) (*api.Result, error) {
 		buildResult.BuildInfo.FailureReason = utilstatus.NewFailureReason(utilstatus.ReasonGenericS2IBuildFailed, utilstatus.ReasonMessageGenericS2iBuildFailed)
 		return buildResult, errors.New("builder image name cannot be empty")
 	}
-
+	timeStart = time.Now()
 	if err := builder.CreateDockerfile(config); err != nil {
```
PI-Victor: > why? a failed step still has a start and finish time...

i would have to add the function again here and it would double the amount of code.
```
buildResult.BuildInfo.StepInfo = utilstatus.AddStepInfo(buildResult.BuildInfo.StepInfo, api.CreateDockerfileStep, timeStart)
```

a failed step does have a start/stop time, but is it relevant to the user to provide the build timing for a step that failed?

PI-Victor: p.s.: just for the record, i do see a value in that, but is it worth the code duplication?

bparees: why do you have to duplicate code?
```
timeStart=time.Now()
err := builder.CreateDockerfile(config)
buildResult.BuildInfo.StepInfo = utilstatus.AddStepInfo(buildResult.BuildInfo.StepInfo, api.CreateDockerfileStep, timeStart)
if err!=nil {
  buildResult.BuildInfo.FailureReason = utilstatus.NewFailureReason(utilstatus.ReasonDockerfileCreateFailed,   utilstatus.ReasonMessageDockerfileCreateFailed)
  return buildResult, err
}
```

PI-Victor: :| why didn't i think about this. refactoring.

PI-Victor:
[test]

PI-Victor:
@bparees ptal.

PI-Victor:
ah. still need to update the docs for the api const.

pkg/api/types.go
```diff
@@ -339,12 +340,72 @@ type Result struct {
...
+	// PullRuntimeImageStep is the build step that occurs when we try to pull the
+	// runtime image.
+	PullRuntimeImageStep StepName = "PullRuntimeImageStep"
+
+	// CopyRuntimeArtifactsStep is the build step that occurs when we copy the
+	//artifacts from the runtime image.
+	CopyRuntimeArtifactsStep StepName = "CopyRuntimeArtifactsStep"
+
+	// BuildImageStep might not be required.
+	BuildImageStep StepName = "BuildImageStep"
```
bparees: what operations/actions does buildimagestep include?

pkg/api/types.go
```diff
@@ -339,12 +340,72 @@ type Result struct {
...
+
+	// BuildDockerfileStep is the build step that builds an image from a Dockerfile.
+	BuildDockerfileStep StepName = "BuildDockerFileStep"
+
+	// CreateDockerfileStep is the
+	CreateDockerfileStep StepName = "CreateDockerfileStep"
```
bparees: i think you're going too fine-grained here.

pkg/api/types.go
```diff
@@ -339,12 +340,72 @@ type Result struct {
...
+
+	// TarStreamImageStep is the build step that tar streams an image.
+	TarStreamImageStep StepName = "TarStreamImageStep"
```
bparees: same here.

pkg/api/types.go
PI-Victor: https://github.com/openshift/source-to-image/pull/637/files#diff-ba291d0ac25db597b7ad0d29ad65a77fR138 builds the image with the onbuild strategy

bparees: we don't even support onbuild from openshift.

bparees: https://docs.openshift.org/latest/creating_images/s2i.html#using-images-with-onbuild-instructions

we do support "layered" builds (where we do a docker build to inject the source/etc), but that's a pretty rare situation and i'd be content to count it as part of the assembly process just like normal source injection.

PI-Victor:
@bparees remember my question about displaying a summary after the build.

> no. certainly not in s2i. maybe from within the origin code that wraps the build invocation, but even that i'm not a big fan of.

well, as it turns out.
https://github.com/openshift/source-to-image/blob/master/cmd/s2i/main.go#L157-L162
lol.

Issue #640, opened 2016-11-22
panic: runtime error: invalid memory address or nil pointer dereference on Linux x86_64


rafaeltuelho:
using release `v1.1.3`...

`s2i build https://github.com/pmorie/simple-ruby.git openshift/ruby-20-centos7 test-ruby-app`

I get...

```
error: Unable to load docker config
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x4ef3a0]

goroutine 1 [running]:
panic(0x7c26e0, 0xc420012160)
	/usr/local/go/src/runtime/panic.go:500 +0x1a1
github.com/openshift/source-to-image/pkg/docker.GetImageRegistryAuth(0x0, 0x7fff89f4c978, 0x19, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, ...)
	/home/rsoares/Downloads/gocode/src/github.com/openshift/source-to-image/_output/local/go/src/github.com/openshift/source-to-image/pkg/docker/util.go:67 +0x250
main.newCmdBuild.func1(0xc420097200, 0xc4202a41e0, 0x3, 0x3)
	/home/rsoares/Downloads/gocode/src/github.com/openshift/source-to-image/_output/local/go/src/github.com/openshift/source-to-image/cmd/s2i/main.go:118 +0x1003
github.com/openshift/source-to-image/vendor/github.com/spf13/cobra.(*Command).execute(0xc420097200, 0xc4202a4150, 0x3, 0x3, 0xc420097200, 0xc4202a4150)
	/home/rsoares/Downloads/gocode/src/github.com/openshift/source-to-image/_output/local/go/src/github.com/openshift/source-to-image/vendor/github.com/spf13/cobra/command.go:603 +0x439
github.com/openshift/source-to-image/vendor/github.com/spf13/cobra.(*Command).ExecuteC(0xc420096d80, 0xc420183eb0, 0xc4202593d0, 0xc420097b00)
	/home/rsoares/Downloads/gocode/src/github.com/openshift/source-to-image/_output/local/go/src/github.com/openshift/source-to-image/vendor/github.com/spf13/cobra/command.go:689 +0x367
github.com/openshift/source-to-image/vendor/github.com/spf13/cobra.(*Command).Execute(0xc420096d80, 0xc420183ea8, 0x1)
[...]

PR #641, opened 2016-11-22
handle empty docker auths cleanly

bparees:
fixes https://github.com/openshift/source-to-image/issues/640

@jim-minter ptal

### DIFF
--- a/pkg/docker/util.go
+++ b/pkg/docker/util.go
@@ -52,12 +52,14 @@ const (
 // image name and a given set of client authentication objects.
 func GetImageRegistryAuth(auths *AuthConfigurations, imageName string) api.AuthConfig {
 	glog.V(5).Infof("Getting docker credentials for %s", imageName)
+	if auths == nil {
+		return api.AuthConfig{}
+	}
 	ref, err := reference.ParseNamedDockerImageReference(imageName)
 	if err != nil {
 		glog.V(0).Infof("error: Failed to parse docker reference %s", imageName)
 		return api.AuthConfig{}
 	}
-
 	if ref.Registry != "" {
 		if auth, ok := auths.Configs[ref.Registry]; ok {
 			glog.V(5).Infof("Using %s[%s] credentials for pulling %s", auth.Email, ref.Registry, imageName)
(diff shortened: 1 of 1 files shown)


bparees:
[test]

bparees:
(longer term we should probably be removing this custom parser logic and using the docker engine library assuming it has something appropriate.  we're using fsouza/go-dockerclient to handle this in origin)

bparees:
[test]

bparees:
[test]

bparees:
[test]

bparees:
@oatmealraisin ptal

bparees:
@coreydaley ptal

coreydaley:
@bparees lgtm

bparees:
[merge]

jim-minter:
@bparees fyi I never got an e-mail notification for your 'ptal' message and github didn't cc me on the issue, hence the radio silence!

jim-minter:
> (longer term we should probably be removing this custom parser logic and using the docker engine library assuming it has something appropriate. we're using fsouza/go-dockerclient to handle this in origin)

Maybe; actually I had a look at this a while back.  There are two principal problems:
1) upstream, docker has rather moved from .dockercfg files to preferring config.json files, and origin especially has a fair amount of logic centred around .dockercfg files which may be troublesome to adjust.  That bridge may need to be crossed at some point, but for now I thought it added churn for little value.
2) the docker client library code doesn't expose methods for reading in arbitrary .dockercfg files (it looks for $DOCKER_CONFIG/config.json first, then $HOME/.docker/.dockercfg as a fallback). Reading an arbitrary .dockercfg is currently allowed as a s2i command line switch and is also used by Origin through our API, and all of that would need to be deprecated / reworked as well.

jim-minter:
belatedly, fwiw, perhaps this is worth a test case.

bparees:
> fyi I never got an e-mail notification for your 'ptal' message and github didn't cc me on the issue, hence the radio silence!

weird.  github must be censoring me.

Thanks for the background on utility logic for parsing the various dockerconfig formats.  I guess we'll leave it as is for now.

Agreed that it's worthy of a TC.

PR #642, opened 2016-11-28
bump github.com/Microsoft/go-winio

jim-minter:
The minor fixes contained finally get us to a point where we can run things like `go build ./...` or `go test ./...` in the repo, which is nice!


bparees:
[merge]

PR #644, opened 2016-12-01
Goroutine leak, logging and pkg doc

rhcarvalho:
Miscellaneous improvements:

- Send on receiveStdout would block forever in goroutine when
holdHijackedConnection returns earlier.
- We should be updating createOpts.HostConfig.ShmSize before we log
createOpts.HostConfig.
- Pkg doc was outdated.

### DIFF
--- a/pkg/docker/doc.go
+++ b/pkg/docker/doc.go
@@ -1,4 +1,3 @@
-// Contains higher-level Docker operations needed by the S2I builder and executor
-// It uses go-dockerclient to communicate with Docker
-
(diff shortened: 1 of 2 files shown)


rhcarvalho:
@bparees ptal

rhcarvalho:
(Taken these from my local branch when working on #301)

jim-minter:
@bparees all good to merge as far as I'm concerned
thanks @rhcarvalho!

bparees:
[merge]

PR #645, opened 2016-12-01
Godepchecker

jim-minter:
Added a checker to ensure s2i's godeps match origin's.  Currently, warnings are printed but do not cause make verify to bomb out.
Bumped two vendorings to match origin as well.

@soltysh

### DIFF
--- a/tools/godepchecker/godepchecker.go
+++ b/tools/godepchecker/godepchecker.go
@@ -1,0 +1,95 @@
+package main
+
+import (
+	"encoding/json"
+	"fmt"
+	"os"
+	"strings"
+)
+
+// Dependency represents a Golang dependency
+type Dependency struct {
+	ImportPath string
+	Comment    string `json:",omitempty"`
+	Rev        string
+}
+
+// Godeps represents a Godeps/Godeps.json file
+type Godeps struct {
+	ImportPath   string
+	GoVersion    string
+	GodepVersion string
+	Packages     []string
+	Deps         map[string]Dependency
+}
+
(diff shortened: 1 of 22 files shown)


soltysh:
I was rather thinking about verifying the godeps if the are valid, and not if they are the same as in origin. Iow. if `godep restore` works cleanly without any errors.

hack/common.sh
```diff
@@ -293,7 +293,9 @@ s2i::build::place_bins() {
 
       # Move the specified release binaries to the shared S2I_OUTPUT_BINPATH.
       for binary in "${binaries[@]}"; do
-        mv "${full_binpath_src}/${binary}" "${S2I_OUTPUT_BINPATH}/${platform}/"
+        if [ -e "${full_binpath_src}/${binary}" ]; then
+          mv "${full_binpath_src}/${binary}" "${S2I_OUTPUT_BINPATH}/${platform}/"
+        fi
```
bparees: don't we want to error if a binary we expected to build doesn't exist?

tools/godepchecker/godepchecker.go
```diff
@@ -0,0 +1,94 @@
...
+}
+
+func readJSON(filename string) (godeps *Godeps, err error) {
+	f, err := os.Open(filename)
+	if err != nil {
+		return
+	}
+	defer f.Close()
+
+	json.NewDecoder(f).Decode(&godeps)
```
jim-minter: Yes.

bparees:
lgtm
[merge]

PR #646, opened 2016-12-02
advise use http/s or ssh protocol not local protocol

SinghamXiao:
Signed-off-by: yuanxiao <yuan.xiao5@zte.com.cn>

PR #647, opened 2016-12-04
Prefer using https:// instead of git://

rhcarvalho:
Both protocols are "smart" and should have similar performance.
HTTPS is encrypted, while the Git transport is plaintext.
HTTPS verifies the certificate of the server.

The first commit is taken from #646, then I went ahead and updated other examples and tests that clone GitHub repos.

### DIFF
--- a/pkg/create/templates/scripts.go
+++ b/pkg/create/templates/scripts.go
@@ -51,7 +51,7 @@ To use it, install S2I: https://github.com/openshift/source-to-image
 
 Sample invocation:
 
-s2i build git://<source code> {{.ImageName}} <application image>
+s2i build <source code path/URL> {{.ImageName}} <application image>
 
 You can then run the resulting image via:
 docker run <application image>
--- a/pkg/scm/scm.go
+++ b/pkg/scm/scm.go
@@ -45,8 +45,7 @@ func DownloaderForSource(fs util.FileSystem, s string, forceCopy bool) (build.Do
 		return &file.File{FileSystem: fs}, s, nil
 	}
 
-	// If the source is valid  Git protocol (file://, ssh://, git://, git@, etc..) use Git
-	// binary to download the sources
+	// If s is a valid Git clone spec, use git to download the source
 	g := git.New(fs)
 	ok, err := g.ValidCloneSpec(s)
 	if err != nil {
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@ For a deep dive on S2I you can view [this presentation](https://www.youtube.com/
 
 Want to try it right now?  Download the [latest release](https://github.com/openshift/source-to-image/releases/latest) and run:
 
-	$ s2i build git://github.com/openshift/django-ex centos/python-35-centos7 hello-python
+	$ s2i build https://github.com/openshift/django-ex centos/python-35-centos7 hello-python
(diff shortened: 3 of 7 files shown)


rhcarvalho:
[test]

pkg/scm/scm.go
```diff
@@ -45,8 +45,7 @@ func DownloaderForSource(fs util.FileSystem, s string, forceCopy bool) (build.Do
 		return &file.File{FileSystem: fs}, s, nil
 	}
 
-	// If the source is valid  Git protocol (file://, ssh://, git://, git@, etc..) use Git
```
jim-minter: nit, but I personally preferred the comment as it was - the new comment might as well not exist at all as it can be inferred from the code below; the old one at least brought to mind some of the potential options the code might be dealing with.

pkg/create/templates/scripts.go
```diff
@@ -51,7 +51,7 @@ To use it, install S2I: https://github.com/openshift/source-to-image
 
 Sample invocation:
 
-s2i build git://<source code> {{.ImageName}} <application image>
+s2i build <source code path/URL> {{.ImageName}} <application image>
```
jim-minter: nit, I think we prefer URLs to paths - so "URL/path"?

jim-minter:
looks good to me bar the nitty comments

pkg/scm/scm.go
rhcarvalho: > the old one at least brought to mind some of the potential options the code might be dealing with.

IMHO there are better places to tell what are valid clone specs:

- ValidCloneSpec doc comment
- Tests
- Git documentation

I left the comment as a marker of "here is where we decide we'll try a `git clone`", I'm not so concerned with enumerating, in this particular place, all the protocols we might recognize.

Further reasoning, if we add or remove support for a certain protocol, do you think this is an obvious place where we would update the comment, or it would easily drift?

pkg/create/templates/scripts.go
rhcarvalho: I'm okay either way, we can change it.

For context, this is part of the usage script template, used to bootstrap new S2I builder images. Obviously, it won't affect the comments on any existing image.

If the order is really relevant, in my experience I'd use URLs in OpenShift, but tend to use local paths with the `s2i` tool: I often have a local clone of my repo, so I don't need to have `s2i` fetch it from the Internet all the time.

bparees:
@rhcarvalho you're going to need to excise that cherry-picked commit because i ended up squashing and merging that PR, so the commit you've got doesn't exist in our repo.

pkg/create/templates/scripts.go
bparees: i'm with @rhcarvalho, s2i is probably more commonly used with "." as the path, anyway i don't think the ordering is critical.

pkg/scm/scm.go
bparees: @rhcarvalho that is true(that the comment would be better served elsewhere and will get overlooked during updates here), so please move this comment into ValidCloneSpec.

bparees:
@rhcarvalho bump

rhcarvalho:
@bparees sorry I'm just looking at this on my free time, will try again tonight.

bparees:
[test]

PR #648, opened 2016-12-04
Add regression test for incremental builds

rhcarvalho:
Ensure the build takes no more than `DefaultDockerTimeout`, what would
suggest that S2I got stuck waiting for the timeout.

This is a regression test for #301.

I've applied this patch to 12908b75a9d965425ef35a84d729f750f38eeb72 (before @jim-minter's fix) and seen the test fail:

```console
$ hack/test-integration.sh -run=TestIncrementalBuildAndRemovePreviousImage

Running integration tests ...

=== RUN   TestIncrementalBuildAndRemovePreviousImage
[...]
I1204 18:33:58.346799   28390 tar.go:434] Extracting/writing /tmp/s2i614376474/upload/artifacts/save-artifacts-invoked
I1204 18:33:58.346864   28390 tar.go:403] Done extracting tar stream
I1204 18:35:58.879113   28390 docker.go:948] Removing container "48a1c0b42ae492612affc572080b4b619a32e549277f4349ab1f6678bef0a44b" ...
[...]
--- FAIL: TestIncrementalBuildAndRemovePreviousImage (137.40s)
        integration_test.go:501: Test took too long (2m17.156475097s), some operation may have gotten stuck waiting for the DefaultDockerTimeout (2m0s). Inspect the logs to find operations that took long.
FAIL
exit status 1
FAIL    github.com/openshift/source-to-image/test/integration   137.414s

Complete
```

### DIFF
--- a/test/integration/integration_test.go
+++ b/test/integration/integration_test.go
@@ -438,6 +438,7 @@ func (i *integrationTest) exerciseInjectionBuild(tag, imageName string, injectio
 
 func (i *integrationTest) exerciseIncrementalBuild(tag, imageName string, removePreviousImage bool, expectClean bool, checkOnBuild bool) {
 	t := i.t
+	start := time.Now()
 	config := &api.Config{
 		DockerConfig:        docker.GetDefaultDockerConfig(),
 		BuilderImage:        imageName,
@@ -504,6 +505,13 @@ func (i *integrationTest) exerciseIncrementalBuild(tag, imageName string, remove
 
 	if checkOnBuild {
 		i.fileExists(containerID, "/sti-fake/src/onbuild")
+	}
+
+	if took := time.Since(start); took > docker.DefaultDockerTimeout {
+		// https://github.com/openshift/source-to-image/issues/301 is a
+		// case where incremental builds would get stuck until the
+		// timeout.
+		t.Errorf("Test took too long (%v), some operation may have gotten stuck waiting for the DefaultDockerTimeout (%v). Inspect the logs to find operations that took long.", took, docker.DefaultDockerTimeout)
 	}
 }
 
--- a/test/integration/scripts/.s2i/bin/save-artifacts
+++ b/test/integration/scripts/.s2i/bin/save-artifacts
@@ -3,3 +3,7 @@ touch /tmp/src/save-artifacts-invoked
 
 cd /tmp/src
 tar cf - .
+
+# Write some zeros to stdout to mimic the padding that Debian's tar adds
+# https://github.com/openshift/source-to-image/issues/301#issuecomment-263598586
+dd count=1 if=/dev/zero 2> /dev/null


rhcarvalho:
[test]

test/integration/integration_test.go
```diff
@@ -505,6 +506,13 @@ func (i *integrationTest) exerciseIncrementalBuild(tag, imageName string, remove
 	if checkOnBuild {
 		i.fileExists(containerID, "/sti-fake/src/onbuild")
 	}
+
+	if took := time.Since(start); took > docker.DefaultDockerTimeout {
```
jim-minter: I think this approach is worthwhile as long as each expected test run duration is _much_ smaller than DefaultDockerTimeout - otherwise it could lead to flakes.

jim-minter: I wonder whether this approach could/should also be applied to other integration tests - for example (but not limited to) `func (i *integrationTest) exerciseCleanBuild` ?

test/integration/integration_test.go
```diff
@@ -505,6 +506,13 @@ func (i *integrationTest) exerciseIncrementalBuild(tag, imageName string, remove
...
+		// https://github.com/openshift/source-to-image/issues/301 is a
+		// case where incremental builds would get stuck until the
+		// timeout.
+		t.Errorf("Test took too long (%v), some operation may have gotten stuck waiting for the DefaultDockerTimeout (%v). Inspect the logs to find operations that took long.", took, docker.DefaultDockerTimeout)
+	}
```
jim-minter: There's also a sort-of corollary test which would be nice to have: implement the semi-pathological s2i build, which at every step carries out actions which run for longer than DefaultDockerTimeout, which should always succeed.  Don't know whether you consider that within scope for this change, or whether I should open an issue for that?

test/integration/scripts/.s2i/bin/save-artifacts
```diff
@@ -3,3 +3,7 @@ touch /tmp/src/save-artifacts-invoked
 
 cd /tmp/src
 tar cf - .
+
+# Write some zeros to stdout to mimic the padding that Debian's tar adds
+# https://github.com/openshift/source-to-image/issues/301#issuecomment-263598586
+dd count=1 if=/dev/zero 2> /dev/null
```
jim-minter: +1

rhcarvalho: I first tried `echo junk` as you suggested, but for some reason that was not enough to reproduce the problem...

test/integration/integration_test.go
rhcarvalho: Out of scope. Somewhere deep inside I feel `DefaultDockerTimeout` started as just a hack to accommodate synchronization issues. But it can come in handy when operating against a faulty docker daemon.

Worry: such an integration test would be meaningful, but notably slow. Another possibility would be unit tests around the blocks that use the docker client; temporarily decrease `DefaultDockerTimeout`, pass in a fake slow client, then ensure the calls time out.

> at every step carries out actions which run for longer than DefaultDockerTimeout, which should always succeed

You mean should always fail? If they run longer than the timeout, they should be aborted and fail.

jim-minter: > Worry: such an integration test would be meaningful, but notably slow.
The way to implement this would clearly be by decreasing the timeout used for the duration of the test run.

I think the only purpose of the timeout is to protect against potential bad behaviour from the docker daemon (although I would personally prefer us not to do this).  Either way, I think the existence of the timeout must not prevent end users from having builds or hooks or scripts which run for longer than the timeout.  This is one of the things I tried to improve before, which explained why previously in #301 it happened to hang for 2 minutes, and subsequently it hung forever.  We do not use the timeout in attach, pull or wait docker calls for this reason.

So I mean always succeed.

test/integration/integration_test.go
jim-minter: I think it's worth going through all the integration tests and if they clearly run for much less time than the timeout, to add the check in.

bparees:
@rhcarvalho i'm content to merge this as is, or you can add the additional suggestions @jim-minter made, let me know which you plan to do.

rhcarvalho:
@bparees we can improve as a follow up, https://github.com/openshift/source-to-image/issues/652 to track @jim-minter's suggestion

bparees:
[merge]

PR #649, opened 2016-12-04
Make test environment variables work as advertised

rhcarvalho:
`FOO=${FOO:-value}` sets FOO to 'value' if it is null (empty) or unset,
whereas `FOO=${FOO-value}` (without the colon) sets FOO only when FOO is
unset.

`FOO= hack/test-go.sh` is equivalent to `FOO="" hack/test-go.sh`, so
prefer to mention the shorter form.

Let the go tool pick the default cover mode.

Since -coverprofile implies -cover, we don't need to test S2I_COVER when
OUTPUT_COVERAGE is given.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -63,18 +63,17 @@ To run all tests with verbose output:
 
     $ hack/test-go.sh "" -v
 
-To turn off or change the coverage mode, which is `-cover -covermode=atomic` by default, use:
-
-    $ S2I_COVER="" hack/test-go.sh
-
-To run tests without the go race detector, which is on by default, use:
+To run tests without the Go race detector, which is on by default, use:
 
     $ S2I_RACE="" hack/test-go.sh
 
-A line coverage report is run by default when testing a single package.
-To create a coverage report for all packages:
+A line coverage report is printed by default. To turn it off, use:
 
-    $ OUTPUT_COVERAGE=/tmp/reportdir hack/test-go.sh
+    $ S2I_COVER="" hack/test-go.sh
+
+To create an HTML coverage report for all packages:
+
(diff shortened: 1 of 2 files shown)


rhcarvalho:
@bparees @jim-minter seems that 1a0ef1eb7e04e72618c2f9c8abbf644e956d29f4 introduced a regression in how one can use `S2I_RACE` and `S2I_COVER` to control how tests are run, so fixing it here. PTAL.

jim-minter:
yes, looks good to me, thanks @rhcarvalho

bparees:
> FOO= hack/test-go.sh is equivalent to FOO="" hack/test-go.sh, so
prefer to mention the shorter form.

it may be equivalent but i don't think it's as obvious, particularly to people who may know less about bash.. is the two character savings worth the cost in readability?

rhcarvalho:
> it may be equivalent but i don't think it's as obvious, particularly to people who may know less about bash.. is the two character savings worth the cost in readability?

If you find FOO="" better I'm okay with that, this is just a copy-pasteable example, it doesn't have to be the shortest thing, as long as it works.

Having to write `S2I_RACE=' ' hack/test-integration.sh` was the problem for me (regression).

When the value is non-empty, I guess most people write `FOO=bar some-command` in the shell, without the quotes?

bparees:
> When the value is non-empty, I guess most people write FOO=bar some-command in the shell, without the quotes?

that's certainly what i do.

bparees:
> If you find FOO="" better I'm okay with that, this is just a copy-pasteable example, it doesn't have to be the shortest thing, as long as it works.

that's my preference, yeah.  "FOO= " looks like a typo to me.  (maybe just to me though).

rhcarvalho:
@bparees updated

bparees:
thanks @rhcarvalho 
[merge]

Issue #650, opened 2016-12-05
incremental s2i build fails and stuck for nodejs 4 s2i images


sudhirpandey:
we have been using s2i base image for node js 4 from here

https://github.com/sclorg/s2i-nodejs-container

The normal builds work fine but incremental builds starts to fail and would not fall back to normal build

the app is from here
https://github.com/sclorg/s2i-nodejs-container/tree/master/4/test/test-app

reproduced as

have a save-artifacts script has 
```
#!/bin/bash

pushd /opt/app-root/src/
if [ -d .npm ]; then
    # all deps contents to tar stream
    tar cf - .npm
fi
popd
```

First build goes fine without incremental option
```
s2i build 4/test/test-app rhscl/nodejs-4-rhel7 nodejs-sample-app --loglevel=5
```

```
I1205 15:33:00.583796   30065 main.go:68] Running S2I version "v1.1.3"
I1205 15:33:00.584126   30065 util.go:53] Getting docker credentials for rhscl/nodejs-4-rhel7
I1205 15:33:00.589305   30065 docker.go:515] Using locally available image "rhscl/nodejs-4-rhel7:latest"
I1205 15:33:00.593018   30065 main.go:147]
Builder Name:			Node.js 4
Builder Image:			rhscl/nodejs-4-rhel7
Source:				4/test/test-app
Output Image Tag:		nodejs-sample-app
Environment:
Labels:
Incremental Build:		disabled
Remove Old Build:		disabled
Builder Pull Policy:		if-not-present
Previous Image Pull Policy:	if-not-present
Quiet:				disabled
Layered Build:			disabled
Docker Endpoint:		unix:///var/run/docker.sock
Docker Pull Config:		/root/.docker/config.json
Docker Pull User:

I1205 15:33:00.695021   30065 docker.go:515] Using locally available image "rhscl/nodejs-4-rhel7:latest"
I1205 15:33:00.704786   30065 docker.go:515] Using locally available image "rhscl/nodejs-4-rhel7:latest"
I1205 15:33:00.704841   30065 docker.go:726] Image sha256:38c233077e56dc2d6639a646b63772fd486be88438f4fbaca7486792328582a1 contains io.openshift.s2i.scripts-url set to "image:///usr/libexec/s2i"
I1205 15:33:00.704985   30065 scm.go:21] DownloadForSource 4/test/test-app
I1205 15:33:00.705050   30065 git.go:208] makePathAbsolute 4/test/test-app
[...]

PR #651, opened 2016-12-05
clarify the output of save-artifacts cannot include anything except the tar stream

bparees:
fixes https://github.com/openshift/source-to-image/issues/650


bparees:
[merge]

PR #654, opened 2016-12-08
M: Using wildcard * not . in linux shell

SinghamXiao:
Signed-off-by: yuanxiao <yuan.xiao5@zte.com.cn>

### DIFF
--- a/pkg/create/templates/scripts.go
+++ b/pkg/create/templates/scripts.go
@@ -20,11 +20,11 @@ fi
 #
 if [ "$(ls /tmp/artifacts/ 2>/dev/null)" ]; then
   echo "---> Restoring build artifacts..."
-  mv /tmp/artifacts/. ./
+  mv /tmp/artifacts/* ./
 fi
 
 echo "---> Installing application source..."
-cp -Rf /tmp/src/. ./
+cp -Rf /tmp/src/* ./
 
 echo "---> Building application from source..."
 # TODO: Add build steps for your application, eg npm install, bundle install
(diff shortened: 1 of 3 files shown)


pkg/create/templates/scripts.go
```diff
@@ -20,11 +20,11 @@ fi
 #
 if [ "$(ls /tmp/artifacts/ 2>/dev/null)" ]; then
   echo "---> Restoring build artifacts..."
-  mv /tmp/artifacts/. ./
+  mv /tmp/artifacts/* ./
 fi
 
 echo "---> Installing application source..."
-cp -Rf /tmp/src/. ./
+cp -Rf /tmp/src/* ./
```
bparees: this change is going to miss hidden files.  you need to set shopt dotglob if you want to make this change.

pkg/create/templates/scripts.go
```diff
@@ -20,11 +20,11 @@ fi
 #
 if [ "$(ls /tmp/artifacts/ 2>/dev/null)" ]; then
   echo "---> Restoring build artifacts..."
-  mv /tmp/artifacts/. ./
+  mv /tmp/artifacts/* ./
```
bparees: same here.

pkg/create/templates/scripts.go
SinghamXiao: @bparees You mean the folder .s2i. I misunderstand the script. I wander to konw why the code generates the hidden .s2i,? Once used s2i create ... command,, still need to modify files (script, run etc.) in folder .s2i Why not make it visible?

bparees: I mean any hidden folder that is part of the source repository being built/operated on by the assemble script.  The existing logic will include those dirs, your logic will not.

s2i create is just a helper to give you some skeleton/example files when you are starting development of a new s2i builder image.  As you say, it's not a critical feature if you already know how to write the scripts.

As for why s2i create creates the .s2i folder as a hidden folder, probably not for any good reason, it could well be created as non-hidden instead and probably should be.

PR #655, opened 2016-12-08
refactor func checkErr

SinghamXiao:
Signed-off-by: yuanxiao <yuan.xiao5@zte.com.cn>

### DIFF
--- a/pkg/errors/errors.go
+++ b/pkg/errors/errors.go
@@ -2,9 +2,12 @@ package errors
 
 import (
 	"fmt"
+	"os"
+
+	utilglog "github.com/openshift/source-to-image/pkg/util/glog"
 )
 
-// Common STI errors
+// Common S2I errors
 const (
 	InspectImageError int = 1 + iota
 	PullImageError
@@ -24,7 +27,7 @@ const (
 	EmptyGitRepositoryError
 )
 
-// Error represents an error thrown during STI execution
+// Error represents an error thrown during S2I execution
 type Error struct {
 	Message    string
 	Details    error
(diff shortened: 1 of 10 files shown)


bparees:
Given that main.go is the only thing using this function currently, what is the motivation for this refactor?

SinghamXiao:
@bparees Looks very disordered two many functions in file main.go Why not refactor this file？ In addition，it's better to move the funtion CheckErr to a common place. Maybe oneday it's invoked by fuctions in other files。

pkg/errors/errors.go
```diff
@@ -244,3 +247,25 @@ func NewEmptyGitRepositoryError(source string) error {
 		Suggestion: "Either commit files to the Git repository, remove the .git directory from the project, or use --copy to ignore the repository.",
 	}
 }
+
+var glog = utilglog.StderrLog
+
+func CheckErr(err error) {
```
bparees: needs godoc.

bparees:
The question asked when making code changes should be "why?", not "why not?" and "there are too many functions in this file and someone might want to reuse it in the future" doesn't strike me as a very good answer to "why" (if someone does want to reuse it in the future, that would be a logical time to move it into a utility package)

Having said that, if you godoc the function, I will merge this change.

SinghamXiao:
Sorry for my poor English. I will keep learning. 
godoc errors CheckError
func CheckError(err error)
    CheckError checks input error. 1. if the input error is nil, the
    function do nothing but return. 2. if the input error is a kind of Error
    which is thrown during STI execution, the function handle it with
    Suggestion and Details. 3. if the input error is a kind of system Error
    which is unknown, the function exit with 1.

bparees:
two minor issues and please squash your commits and I will merge.
[test]

PR #657, opened 2016-12-12
output error when err != nil

guangxuli:
if ` err == nil`, it is useless to output `err` info.

### DIFF
--- a/pkg/scm/git/git.go
+++ b/pkg/scm/git/git.go
@@ -208,14 +208,16 @@ func ParseURL(source string) (*url.URL, error) {
 	return uri, nil
 }
 
-// mimic the path munging (make paths absolute to fix file:// with non-absolute paths) done when scm.go:DownloderForSource valided git file urls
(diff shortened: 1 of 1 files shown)


pkg/scm/git/git.go
```diff
@@ -208,14 +208,16 @@ func ParseURL(source string) (*url.URL, error) {
...
 	glog.V(4).Infof("makePathAbsolute %s", source)
 	if !strings.HasPrefix(source, "/") {
-		if absolutePath, err := filepath.Abs(source); err == nil {
-			glog.V(4).Infof("makePathAbsolute new path %s err %v", absolutePath, err)
+		absolutePath, err := filepath.Abs(source)
+		if err == nil {
+			glog.V(4).Infof("makePathAbsolute new path %s success", absolutePath)
 			return absolutePath
 		}
+		glog.V(4).Infof("makePathAbsolute new path err %v", err)
```
bparees: should print the source path, it might give us a clue as to why we got an error.

guangxuli: ok

guangxuli:
@bparees fixed.

bparees:
[merge]

PR #663, opened 2016-12-20
use rhel7:inst AMI for Windows CI, to match RHEL CI

jim-minter:
@bparees, turns out I was using the wrong AMI for the RHEL part of the Windows CI.  I haven't got to the bottom of what the difference is between the two, but it repeatedly works with inst and fails with deps.  Please merge and I'll re-enable the CI.


bparees:
[merge]

PR #664, opened 2016-12-21
add zsh completions for s2i command

shiywang:
fixes https://github.com/openshift/source-to-image/issues/665

### DIFF
--- a/cmd/s2i/main.go
+++ b/cmd/s2i/main.go
@@ -31,6 +31,7 @@ import (
 	"github.com/openshift/source-to-image/pkg/tar"
 	"github.com/openshift/source-to-image/pkg/util"
 	"github.com/openshift/source-to-image/pkg/version"
+	"io"
 )
 
 // glog is a placeholder until the builders pass an output stream down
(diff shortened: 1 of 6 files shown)


cmd/s2i/main.go
```diff
@@ -273,22 +274,241 @@ func newCmdCreate() *cobra.Command {
...
+
+	return run(out, root)
+}
+
+func runCompletionBash(out io.Writer, root *cobra.Command) error {
+	return root.GenBashCompletion(out)
+}
+
+func runCompletionZsh(out io.Writer, root *cobra.Command) error {
+	zshInitialilzation := `# Copyright 2016 The Kubernetes Authors.
```
bparees: zshInitialization

pkg/errors/errors.go
```diff
@@ -277,3 +278,9 @@ func CheckError(err error) {
 		os.Exit(1)
 	}
 }
+
+// UsageError checks command usage error.
+func UsageError(cmd *cobra.Command, format string, args ...interface{}) error {
+	msg := fmt.Sprintf(format, args...)
```
bparees: it would seem reasonable to me for UsageError to just take the msg string, and expect callers to have already invoked fmt.Sprintf before passing the message in.

bparees:
the travis failure is this:
```
./pkg/errors/errors.go:285:20: error strings should not be capitalized or end with punctuation or a newline
```

shiywang:
@bparees updated, PTAL, thanks

bparees:
lgtm [merge]

PR #669, opened 2016-12-31
Build with latest available Go minor versions & tip

jawnsy:
This changes the build to:

* Always build with the latest minor version of the last two (supported upstream) major releases of Go (currently the latest minor version in the 1.6 and 1.7 series)
  * At present, this is building with 1.6.3, which seems to be a bug in Travis somewhere (1.6.4 is available)
  * It does correctly build with 1.7.4
* Build using the latest major version (1.8 pre-release) from the Golang git repository, but allow failures, so that we have early warning in case there are any problems due to changes in Go itself

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,8 @@
 language: go
 
 go:
-  - 1.7.4
+  - 1.7.x
+  - master
 
 install:
   - make install-travis
@@ -14,3 +15,8 @@ notifications:
   irc: "chat.freenode.net#openshift-dev"
 
 sudo: false
+
+matrix:
+  allow_failures:
+  - go: master
+  fast_finish: true
--- a/pkg/scm/git/git_munge_test.go
+++ b/pkg/scm/git/git_munge_test.go
@@ -109,11 +109,13 @@ func TestMungeNoProtocolURL(t *testing.T) {
 		uri, err := url.Parse(scenario)
 		if err != nil {
 			t.Errorf("Could not parse url %q", scenario)
+			continue
 		}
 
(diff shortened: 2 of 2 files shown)


jawnsy:
@bparees PTAL

jawnsy:
This currently fails with tip, not sure if it's a flake or a bug: https://travis-ci.org/openshift/source-to-image/builds/187975336

jim-minter:
I like this, and request the following:
- remove the 1.6.x from .travis.yml - we can't support MacOS Sierra users on 1.6 and #675 will take us to using 1.7 (as https://github.com/openshift/origin/pull/12215 did for origin).
- squash the commits.

The failure with golang master is an issue raised by improvements in Golang's URL parser in 1.8: https://github.com/golang/go/commit/c5ccbdd22bdbdc43d541b7e7d4ed66ceb559030e .  This has knock-on effects for both origin and s2i and a review will be needed in due course.  I'll open an issue in origin for that and assign it to me.

If you could also put continue statements in git_munge_test.go directly under `t.Errorf("Could not parse url %q", scenario)` and `t.Errorf("MungeNoProtocolURL returned err: %v", err)`, that should prevent the panic, and the test should still fail (but politely).

jim-minter:
Opened https://github.com/openshift/origin/issues/12403.

jawnsy:
@jim-minter Updated the PR with your requested changes. PTAL. I haven't tested locally, so let's see what Travis says :)

bparees:
[merge]

jim-minter:
Thanks @jawnsy!

Issue #668, opened 2016-12-31
Bump Go version dependency


jawnsy:
In the installation instructions, we specify that we have a dependency on Go 1.4 or newer. However, according to the [upstream release policy](https://golang.org/doc/devel/release.html), only the latest version (currently 1.7) and its immediate predecessor (1.6) receive security support.

> Each major Go release obsoletes and ends support for the previous one. For example, if Go 1.5 has been released, then it is the current release and Go 1.4 and earlier are no longer supported. We fix critical problems in the current release as needed by issuing minor revisions (for example, Go 1.5.1, Go 1.5.2, and so on).
>
> As a special case, we issue minor revisions for critical security problems in both the current release and the previous release. For example, if Go 1.5 is the current release then we will issue minor revisions to fix critical security problems in both Go 1.4 and Go 1.5 as they arise. See the security policy for more details.

As a result, it may be a good idea to increase the dependency to 1.6.

bparees:
i don't think we even build properly on 1.4 anymore, it's just outdated doc, thanks for catching it.

PR #671, opened 2017-01-01
Update required Go version in README

jawnsy:
Closes: #668

Signed-off-by: Jonathan Yu <jawnsy@redhat.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -169,10 +169,9 @@ If a `save-artifacts` script exists, a prior image already exists, and the `--in
 
 ## Dependencies
 
-1. [Docker](http://www.docker.io) >= 1.6
-1. [Go](http://golang.org/) >= 1.4
+1. [docker](https://docker.com) >= 1.6
+1. [Go](https://golang.org/dl/) >= 1.7.1
 1. (optional) [Git](https://git-scm.com/)
-
 
 ## Installation
 


README.md
```diff
@@ -169,11 +169,10 @@ If a `save-artifacts` script exists, a prior image already exists, and the `--in
 
 ## Dependencies
 
-1. [Docker](http://www.docker.io) >= 1.6
-1. [Go](http://golang.org/) >= 1.4
+1. [docker](https://github.com/docker/docker) >= 1.6
```
jawnsy: Now referring to the GitHub repository instead of Docker Engine. Should we bump the minimum docker version as well?

bparees: why change it at all?  this is not a dependency for building, it's a dependency for using the tool.  pointing them to a docker install makes more sense than a source repo.

jawnsy: Okay. I was updating it because `docker.io` is outdated. How about: https://www.docker.com/products/docker-engine ? Or should we just link to the main docker site, https://docker.com ?

bparees: docker.com seems safest/most generic.

jawnsy: Updated :)

jim-minter:
I believe we should be requiring golang 1.7.1 minimum.

bparees:
@jim-minter based on your findings re: macos, i agree. @jawnsy let's make this 1.7.

jawnsy:
@jim-minter @bparees Updated, PTAL

bparees:
[merge]

PR #673, opened 2017-01-04
[DO_NOT_MERGE] minor improvements

PI-Victor:
removes reason and status failure for onbuild strat - since it's not supported in openshift, then it's not needed here.
add runtime artifacts reason and message for failures. should fix https://bugzilla.redhat.com/show_bug.cgi?id=1403712 still need to test it out
rename s2i specific errors package to `s2ierr` to avoid collision with core `errors` package.
make error messages consistent through out the app.
[test]


pkg/build/strategies/sti/sti.go
```diff
@@ -202,6 +208,7 @@ func (builder *STI) Build(config *api.Config) (*api.Result, error) {
 	} else {
 		glog.V(1).Info("Clean build will be performed")
 	}
+	glog.V(0).Infof("Builder result %#v", builder.result.BuildInfo)
```
bparees: You sure this should be V(0)?  what does this output look like in the log?

pkg/build/strategies/sti/sti.go
```diff
@@ -448,6 +455,7 @@ func (builder *STI) Exists(config *api.Config) bool {
 	tag := firstNonEmpty(config.IncrementalFromTag, config.Tag)
 
 	result, err := dockerpkg.PullImage(tag, builder.incrementalDocker, policy, false)
+	glog.V(0).Infof("Test... I GOT HERE err: %s, result: %#v", err, result)
```
bparees: leftover debugging?

pkg/util/status/build_status.go
```diff
@@ -70,11 +70,20 @@ const (
...
 	// ReasonArtifactsFetchFailed is the reason associated with a failure to
-	// download specified scripts in the application image.
-	ReasonArtifactsFetchFailed api.StepFailureReason = "FetchScriptsFailed"
+	// download specified artifacts in the application image.
+	ReasonArtifactsFetchFailed api.StepFailureReason = "FetchArtifactsFailed"
 	// ReasonMessageArtifactsFetchFailed is the message associated with a failure to
-	// download specified scripts in the application image.
-	ReasonMessageArtifactsFetchFailed api.StepFailureMessage = "Failed to fetch scripts specified scripts."
+	// download specified artifacts in the application image.
+	ReasonMessageArtifactsFetchFailed api.StepFailureMessage = "Failed to fetch specified artifacts."
```
bparees: what constitutes "artifacts" here?  what failure causes this to be reported?

pkg/build/strategies/onbuild/onbuild.go
```diff
@@ -90,19 +90,15 @@ func (builder *OnBuild) Build(config *api.Config) (*api.Result, error) {
...
 	// If necessary, copy the STI scripts into application root directory
 	builder.copySTIScripts(config)
 
 	glog.V(2).Info("Creating application Dockerfile")
 	if err := builder.CreateDockerfile(config); err != nil {
-		buildResult.BuildInfo.FailureReason = utilstatus.NewFailureReason(
-			utilstatus.ReasonDockerfileCreateFailed,
-			utilstatus.ReasonMessageDockerfileCreateFailed,
-		)
-		return buildResult, err
```
bparees: though we don't support on-build in openshift today, i don't see that as a good reason to remove this code.  what is the harm in having it?

pkg/util/status/build_status.go
bparees: things artifacts could be referring to, in my mind:  

1) remote assemble/run/save-artifacts scripts we were trying to download from an http server
2) running save-artifacts to extract dependencies from the previous image, failed

pkg/build/strategies/sti/sti.go
PI-Victor: yeah, nah. ignore my debugging

pkg/build/strategies/onbuild/onbuild.go
PI-Victor: it seems like i'm introducing yet another thing that doesn't work. like the result.success field for example and i'm bloating the code for no particular reason.
if you think that we're going to use onbuild at some point in the near future or really want it to be there, then i revert the change.

bparees: not sure why you say it "doesn't work".  Whether the caller currently respects the return value or not, the buildResult object is correct/valid, no?

I'd rather be returning it so that if we ever turn on-build back on for openshift, we don't forget to reintroduce this code.

pkg/util/status/build_status.go
PI-Victor: [this is what constitutes artifacts - scripts](once https://github.com/openshift/source-to-image/blob/28109cfc72cffa557659cf2ea249c03bf14a812a/pkg/build/strategies/sti/sti.go#L374)
[but "artifacts is also used for any files the user specified to be copied"](https://github.com/openshift/source-to-image/blob/28109cfc72cffa557659cf2ea249c03bf14a812a/pkg/build/strategies/sti/sti.go#L275)

PI-Victor: do you think this should be called something else so we could differentiate better?

pkg/build/strategies/onbuild/onbuild.go
PI-Victor: ok, bad phrasing. it works, it's just unused :) i understand your concern i'll add it back.

pkg/util/status/build_status.go
bparees: i think it should stay called scripts since that's what we call it everywhere else (including the field the user would have populated).

bparees: artifacts and scripts should have distinct errors.

PI-Victor: scripts it is, then.

PI-Victor:
@bparees PTAL. tested the bump with openshift 
```
NAME                         TYPE      FROM          STATUS                                 STARTED              DURATION
extended-build-from-repo-1   Source    Git@7323f1a   Failed (FetchRuntimeArtifactsFailed)   About a minute ago   19s
extended-build-from-repo-2   Source    Git@7323f1a   Complete                               55 seconds ago       26s
```

pkg/util/status/build_status.go
```diff
@@ -71,10 +71,18 @@ const (
 
 	// ReasonArtifactsFetchFailed is the reason associated with a failure to
 	// download specified scripts in the application image.
-	ReasonArtifactsFetchFailed api.StepFailureReason = "FetchScriptsFailed"
+	ReasonArtifactsFetchFailed api.StepFailureReason = "FetchArtifactsFailed"
 	// ReasonMessageArtifactsFetchFailed is the message associated with a failure to
 	// download specified scripts in the application image.
-	ReasonMessageArtifactsFetchFailed api.StepFailureMessage = "Failed to fetch scripts specified scripts."
+	ReasonMessageArtifactsFetchFailed api.StepFailureMessage = "Failed to fetch specified scripts."
```
bparees: Seems like this should be renamed to ReasonMessageScriptsFetchFailed (and ReasonScriptsFetchFailed above)

PI-Victor:
addressed nits

bparees:
[merge]

PR #676, opened 2017-01-13
decided return nil end the New() func

xgwang-zte:
already check the err value,decided return nil end the New() func


bparees:
[merge]

PR #678, opened 2017-01-20
optimize execute() func

xgwang-zte:
optimize the execute() func to keep the code  logic clear


bparees:
this contravenes our style practices of using the earliest possible return.

Issue #661, opened 2016-12-14
S2I test/run issues 


marc-despland:
Hello 

I used ```s2i create``` to generate the skeleton of a S2I image project. But when i run the make test i have several issues :

A small one : as the cid_file is empty at the beginning, ```docker inspect``` display usage warning, so i have add some quotes around $(cat $cid_file) like in other functions

A "big" one: the tests failed because after stopping the container, the script try to remove the image ... but the container was not removed yet, so i have add a wait loop like in an other function

[The pact of the test/run script s2i-test-run-patch.txt](https://github.com/openshift/source-to-image/files/651790/s2i-test-run-patch.txt)

[A link to my S2I image](https://github.com/marc-despland/s2i-angular2)

bparees:
Thanks, we'll take a look at it.

coreydaley:
I am working on revamping the skeleton into a working application, and updating the information in the docs folder, but it will take me a few.

coreydaley:
Linked to #688

coreydaley:
Fixed via #688

Issue #546, opened 2016-07-18
Using external repos in integration tests and examples


php-coder:
S2I uses one personal repo in integration tests:

```
test/integration/integration_test.go:   TestSource          = "git://github.com/pmorie/simple-html"
```

Also we're using personal repositories in a different examples:

```
docs/cli.md:$ s2i build git://github.com/mfojtik/sinatra-app-example openshift/ruby-20-centos7 ruby-app
docs/cli.md:$ s2i build --ref=my-branch git://github.com/mfojtik/sinatra-app-example openshift/ruby-20-centos7 ruby-app
docs/cli.md:$ s2i build --scripts-url=file://s2iscripts --destination=/opt git://github.com/mfojtik/sinatra-app-example openshift/ruby-20-centos7 ruby-app
README.md:$ s2i build git://github.com/pmorie/simple-ruby openshift/ruby-20-centos7 test-ruby-app
README.md:$ s2i build git://github.com/bparees/openshift-jee-sample openshift/wildfly-100-centos7 test-jee-app
README.md:$ s2i build git://github.com/pmorie/simple-ruby openshift/ruby-20-centos7 test-ruby-app
```

Should we create a separate repo for all of them to reduce dependency on personal repositories? We can't be sure that all these repos are up-to-date with newest versions. Also from the user's perspective I trust more to non-personal repositories.

See also: https://github.com/openshift/origin/issues/9899

bparees:
the pmorie repo should bre placed with this:
https://github.com/openshift/ruby-hello-world

we should do the same for all the other ruby examples in the doc. (will need to test, obviously).

for the JEE example, we can probably use one of the middleware quickstarts:
https://github.com/jboss-developer/jboss-eap-quickstarts/tree/7.1.x-develop/kitchensink

coreydaley:
Linked to #688

coreydaley:
Fixed via #688

Issue #470, opened 2016-04-18
s2i tool does not check if run user 1001 (for CDK and OpenShift)


alberttwong:
## I was a bit surprised that the s2i tool doesn't warn you.

Most of these images have "USER 1001" (which is the default user) for this reason. I don't know why this is only checked by OpenShift; perhaps it's worth an issue against S2I to at least warn that a privileged security context will be needed on OpenShift.

bparees:
it's an openshift requirement and not an s2i requirement, so it's debatable whether the s2i tooling should be reporting about openshift requirements, given that s2i is intended to be usable w/o openshift.

alberttwong:
maybe we should extend the example since many of use the s2i tool to build for OSE.

bparees:
@alberttwong that's fair.

alberttwong:
Here's a working s2i that is running non-root.  We also tested it against the CDK. 

https://github.com/openshift-s2i/s2i-aspnet/pull/5

PI-Victor:
i'd like to take care of this, but i'm unsure where this conversation is going.
which example should i extend? or should this be, after all, in s2i?

bparees:
s2i create (which bootstraps an example) already uses non-root, this is probably talking about the sample here:  https://github.com/openshift/source-to-image/blob/master/examples/nginx-app

so @PI-Victor the fix would be to update that sample to run as non-root.

alberttwong:
I think generally speaking, we should try to create the best example as possible for people to base their s2i.   That was my intent when building asp-net s2i.

bparees:
@alberttwong i don't explicitly disagree, but there's also a fine line between trying to teach them the minimum set of concepts related to s2i, and trying to teach them all the best practices particularly around things related to openshift.

bparees:
@coreydaley please add this to the work you're doing refactoring the s2i example/doc.

coreydaley:
Linked to #688

coreydaley:
Fixed via #688

PR #688, opened 2017-02-06
Overhaul of Source-to-Image readme(s), templates, and examples

coreydaley:
Updating README(s), templates, and examples

Related issues: #470, #546, #661


coreydaley:
@openshift/devex ptal

coreydaley:
[test]

README.md
```diff
@@ -233,12 +233,12 @@ You can start using `s2i` right away (see [releases](https://github.com/openshif
 with the following test sources and publicly available images:
 
 ```
-$ s2i build https://github.com/pmorie/simple-ruby openshift/ruby-20-centos7 test-ruby-app
+$ s2i build https://github.com/openshift/ruby-hello-world openshift/ruby-20-centos7 test-ruby-app
```
bparees: @pmorie your legacy is dead.

coreydaley: Long live @pmorie !

hack/test-stirunimage.sh
```diff
@@ -130,9 +130,9 @@ check_result $? "${WORK_DIR}/s2i-git-proto.log"
 test_debug "s2i build with --run==true option"
 if [[ "$OSTYPE" == "cygwin" ]]; then
   ( cd hack/windows/sigintwrap && make )
-  hack/windows/sigintwrap/sigintwrap 's2i build https://github.com/bparees/openshift-jee-sample openshift/wildfly-90-centos7 test-jee-app --run=true --loglevel=5' &> "${WORK_DIR}/s2i-run.log" &
+  hack/windows/sigintwrap/sigintwrap 's2i build --ref=7.0.x --context-dir=helloworld https://github.com/jboss-developer/jboss-eap-quickstarts openshift/wildfly-90-centos7 test-jee-app --run=true --loglevel=5' &> "${WORK_DIR}/s2i-run.log" &
```
bparees: wf 101

hack/test-stirunimage.sh
```diff
@@ -130,9 +130,9 @@ check_result $? "${WORK_DIR}/s2i-git-proto.log"
...
 else
-  s2i build https://github.com/bparees/openshift-jee-sample openshift/wildfly-90-centos7 test-jee-app --run=true --loglevel=5 &> "${WORK_DIR}/s2i-run.log" &
+  s2i build --ref=7.0.x --context-dir=helloworld https://github.com/jboss-developer/jboss-eap-quickstarts openshift/wildfly-90-centos7 test-jee-app --run=true --loglevel=5 &> "${WORK_DIR}/s2i-run.log" &
```
bparees: wf 101

pkg/create/templates/scripts.go
```diff
@@ -17,30 +17,35 @@ if [[ "$1" == "-h" ]]; then
...
-cp -Rf /tmp/src/. ./
+# Override the default nginx index.html file.
+# This is what we consider in this example 'installing the application'
+# here you can go ahead an replace this with the actual building of python modules,
+# bundle install, and anything else you need.
 
-echo "---> Building application from source..."
-# TODO: Add build steps for your application, eg npm install, bundle install
+echo "---> Building and installing application from source..."
+cp -Rf /tmp/src/* /usr/share/nginx/html/
```
bparees: mv

docs/cli.md
```diff
@@ -167,19 +167,19 @@ Use this method only for development or local testing.
...
+Build a Java application from a Git source, using the official `openshift/wildfly-101-centos7`
 builder image but overriding the scripts URL from local directory.  The resulting
 image will be named `java-app`:
 
 ```
-$ s2i build --scripts-url=file://s2iscripts https://github.com/bparees/openshift-jee-sample openshift/wildfly-100-centos7 java-app
+$ s2i build --scripts-url=file://s2iscripts --ref=7.1.x --context-dir=kitchensink https://github.com/jboss-developer/jboss-eap-quickstarts openshift/wildfly-101-centos7 java-app
 ```
 
 Build a Ruby application from a Git source, specifying `ref`, and using the official
```
jim-minter: s/specifying `ref`, and // ?

docs/cli.md
```diff
@@ -167,19 +167,19 @@ Use this method only for development or local testing.
...
-`ruby-20-centos7` builder image.  The resulting image will be named `ruby-app`:
+`ruby-23-centos7` builder image.  The resulting image will be named `ruby-app`:
 
 ```
-$ s2i build --ref=my-branch https://github.com/mfojtik/sinatra-app-example openshift/ruby-20-centos7 ruby-app
+$ s2i build --ref=master https://github.com/openshift/ruby-hello-world openshift/ruby-23-centos7 ruby-app
```
jim-minter: s/--ref=master // ?

examples/nginx-centos7/Dockerfile
```diff
@@ -0,0 +1,54 @@
+
+# nginx-centos7
+# Here you can use whatever base image is relevant for your application.
+FROM centos:centos7
+
+# Here you can specify the maintainer for the image that you're building
+MAINTAINER Code Guru <codeguru@example.com>
```
jim-minter: Who?  I'd prefer something closer to `MAINTAINER Your Name <your.name@example.com>`.

examples/nginx-centos7/Dockerfile
```diff
@@ -0,0 +1,54 @@
...
+RUN sed -i 's/80/8080/' /etc/nginx/nginx.conf
+RUN sed -i 's/user nginx;//' /etc/nginx/nginx.conf
+
+# Copy the S2I scripts to /usr/libexec/s2i since we set the label that way
+COPY ./s2i/bin/ /usr/libexec/s2i
+
+# TODO (optional): Copy the builder files into /opt/app-root
+COPY ./test/test-app/* /opt/app-root/
+
+# TODO: Drop the root user and make the needed folders owned by user 1001
```
jim-minter: I think this comment isn't required.

examples/nginx-centos7/Dockerfile
```diff
@@ -0,0 +1,54 @@
...
+LABEL io.k8s.description="Nginx Webserver" \
+    io.k8s.display-name="Nginx 1.6.3" \
+    io.openshift.expose-services="8080:http" \
+    io.openshift.tags="builder,webserver,html,nginx" \
+    # this label tells s2i where to find its mandatory scripts
+    # (run, assemble, save-artifacts)
+    io.openshift.s2i.scripts-url="image:///usr/libexec/s2i"
+
+# Install the nginx web server package and clean the yum cache
+RUN yum install -y epel-release  && \
```
jim-minter: nit: double space

examples/nginx-centos7/Dockerfile
```diff
@@ -0,0 +1,54 @@
...
+
+# TODO: Drop the root user and make the needed folders owned by user 1001
+RUN chown -R 1001:1001 /opt/app-root
+RUN chown -R 1001:1001 /usr/share/nginx
+RUN chown -R 1001:1001 /var/log/nginx
+RUN chown -R 1001:1001 /var/lib/nginx
+RUN touch /run/nginx.pid
+RUN chown -R 1001:1001 /run/nginx.pid
+
+# This default user is created in the openshift/base-centos7 image
```
jim-minter: Out of date comment.

examples/nginx-centos7/Makefile
```diff
@@ -0,0 +1,10 @@
+
+IMAGE_NAME = nginx-centos7
+
+build:
```
jim-minter: there should also be a `.PHONY: build`, FWIW

examples/nginx-centos7/s2i/bin/save-artifacts
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh -e
+#
+# S2I save-artifacts script for the 'nginx-centos7' image.
+# The save-artifacts script streams a tar archive to standard output.
+# The archive contains the files and folders you want to re-use in the next build.
+#
+# For more information see the documentation:
+#	https://github.com/openshift/source-to-image/blob/master/docs/builder_image.md
+#
+tar cf - /usr/share/nginx/html/
```
jim-minter: I don't think this makes sense - doesn't this make it impossible to remove files from /usr/share/nginx/html that were present in previous builds, and also make the actual build result dependent on the existence of previous images?

I think the file and comments should be kept, but the tar line removed.

pkg/create/templates/docker.go
```diff
@@ -3,38 +3,56 @@ package templates
...
-#      io.openshift.expose-services="8080:http" \
-#      io.openshift.tags="builder,x.y.z,etc."
-
-# TODO: Install required packages here:
-# RUN yum install -y ... && yum clean all -y
+# Here you can use whatever base image is relevant for your application.
+FROM centos:centos7
+
+# Here you can specify the maintainer for the image that you're building
+MAINTAINER Code Guru <codeguru@example.com>
```
jim-minter: Who? I'd prefer something closer to `MAINTAINER Your Name <your.name@example.com>`.

pkg/create/templates/docker.go
```diff
@@ -3,38 +3,56 @@ package templates
...
+LABEL io.k8s.description="Nginx Webserver" \
+    io.k8s.display-name="Nginx 1.6.3" \
+    io.openshift.expose-services="8080:http" \
+    io.openshift.tags="builder,webserver,html,nginx" \
+    # this label tells s2i where to find its mandatory scripts
+    # (run, assemble, save-artifacts)
+    io.openshift.s2i.scripts-url="image:///usr/libexec/s2i"
+
+# Install the nginx web server package and clean the yum cache
+RUN yum install -y epel-release  && \
```
jim-minter: nit: double space

examples/nginx-centos7/Dockerfile
```diff
@@ -0,0 +1,54 @@
...
+# Change the default port for nginx 
+# Required if you plan on running images as a non-root user).
+RUN sed -i 's/80/8080/' /etc/nginx/nginx.conf
+RUN sed -i 's/user nginx;//' /etc/nginx/nginx.conf
+
+# Copy the S2I scripts to /usr/libexec/s2i since we set the label that way
+COPY ./s2i/bin/ /usr/libexec/s2i
+
+# TODO (optional): Copy the builder files into /opt/app-root
+COPY ./test/test-app/* /opt/app-root/
```
jim-minter: Question: why is this being done?

examples/nginx-centos7/README.md
```diff
@@ -0,0 +1,82 @@
...
+- bundle install ruby gems
+- setup application specific configuration
+
+The script can also specify a way to restore any saved artifacts from the previous image.   
+
+##### run
+Create a *run* script that will start the application. 
+
+##### save-artifacts (optional)
+- Create a *save-artifacts* script which allows a new build to reuse content from a previous version of the application image.
```
jim-minter: s/^- //

examples/nginx-centos7/test/run
```diff
@@ -0,0 +1,160 @@
+#!/bin/bash
+#
+# The 'run' performs a simple test that verifies the S2I image.
```
jim-minter: it's 160 lines of simple ;-)

hack/test-stirunimage.sh
```diff
@@ -130,9 +130,9 @@ check_result $? "${WORK_DIR}/s2i-git-proto.log"
 test_debug "s2i build with --run==true option"
 if [[ "$OSTYPE" == "cygwin" ]]; then
   ( cd hack/windows/sigintwrap && make )
-  hack/windows/sigintwrap/sigintwrap 's2i build https://github.com/bparees/openshift-jee-sample openshift/wildfly-90-centos7 test-jee-app --run=true --loglevel=5' &> "${WORK_DIR}/s2i-run.log" &
+  hack/windows/sigintwrap/sigintwrap 's2i build --ref=7.0.x --context-dir=helloworld https://github.com/jboss-developer/jboss-eap-quickstarts openshift/wildfly-101-centos7 test-jee-app --run=true --loglevel=5' &> "${WORK_DIR}/s2i-run.log" &
```
jim-minter: --ref=7.1.x everywhere else?

pkg/create/templates/readme.go
```diff
@@ -0,0 +1,91 @@
...
+- bundle install ruby gems
+- setup application specific configuration
+
+The script can also specify a way to restore any saved artifacts from the previous image.   
+
+##### run
+Create a *run* script that will start the application. 
+
+##### save-artifacts (optional)
+- Create a *save-artifacts* script which allows a new build to reuse content from a previous version of the application image.
```
jim-minter: s/^- //

pkg/create/templates/readme.go
```diff
@@ -0,0 +1,91 @@
...
+##### usage (optional) 
+Create a *usage* script that will print out instructions on how to use the image.
+
+##### Make the scripts executable 
+Make sure that all of the scripts are executable by running *chmod +x s2i/bin/**
+
+#### Create the builder image
+The following command will create a builder image named nginx-centos7 based on the Dockerfile that was created previously.
+` +
+	"```\n" +
```
jim-minter: Would be nice for this text and the equivalent README.md to be as closely in sync as possible (bar the {{.ImageName}}s)

examples/nginx-centos7/s2i/bin/assemble
```diff
@@ -16,13 +16,16 @@ fi
...
-# This is what we consider in this example `installing the application`
+
+# Override the default nginx index.html file.
+# This is what we consider in this example 'installing the application'
 # here you can go ahead an replace this with the actual building of python modules,
 # bundle install, and anything else you need.
+
 echo "---> Building and installing application from source..."
-cp -Rf /tmp/src/index.html /usr/share/nginx/html/index.html
+mv -Rf /tmp/src/* /usr/share/nginx/html/
```
jim-minter: does mv have a -R flag?

pkg/create/templates/scripts.go
```diff
@@ -17,30 +17,35 @@ if [[ "$1" == "-h" ]]; then
...
-cp -Rf /tmp/src/. ./
+# Override the default nginx index.html file.
+# This is what we consider in this example 'installing the application'
+# here you can go ahead an replace this with the actual building of python modules,
+# bundle install, and anything else you need.
 
-echo "---> Building application from source..."
-# TODO: Add build steps for your application, eg npm install, bundle install
+echo "---> Building and installing application from source..."
+mv -Rf /tmp/src/* /usr/share/nginx/html/
```
jim-minter: does mv have a -R flag?

pkg/create/templates/docker.go
```diff
@@ -3,38 +3,56 @@ package templates
 // Dockerfile is a default Dockerfile laid down by s2i create
 const Dockerfile = `
 # {{.ImageName}}
-FROM openshift/base-centos7
```
jim-minter: I'm not sure it makes more sense for bootstrap to output a working nginx example compared to its previous behaviour of outputting all the required boilerplate and TODOs marked.  If people want a working nginx example, that's in the examples directory.  Others - thoughts?

examples/nginx-centos7/s2i/bin/save-artifacts
bparees: I considered making a similar comment but i'm not actually sure what s2i will do w/ a save-artifacts script that doesn't output a tar stream to stdout, which is what would happen w/ your suggestion.  I suspect not good things.

pkg/create/templates/docker.go
bparees: +1

coreydaley: I do understand your guys thoughts on this, but, I think that it is more useful for the user to get a fully working application when they use create, something where the make, and make test both work correctly and pass.  I had also considered removing the example that is provided in the repo, but instead provided a copy of what s2i create provides that users can browse through.

examples/nginx-centos7/s2i/bin/assemble
coreydaley: Good catch!

pkg/create/templates/docker.go
jim-minter: If there's a middle ground where a developer starting from a bootstrap template just has to change lines like:
```
# TODO you need to change this
ENTRYPOINT nginx --arg=foo
```
Then I think it's a good plan.  The thing I'm worried about with this PR is that they would be supposed to rip out additional nginx-specific boilerplate which shouldn't be in a generic bootstrap template (but fits OK in an example).

pkg/create/templates/readme.go
coreydaley: They should be, the readme in the example is just a copy of what this produces.

pkg/create/templates/docker.go
bparees: > I do understand your guys thoughts on this, but, I think that it is more useful for the user to get a fully working application when they use create, something where the make, and make test both work correctly and pass.

is it?  why not point them to the example if that's what they want?

coreydaley:
Updated created skeleton, ptal

examples/nginx-centos7/test/run
coreydaley: In my defense, that comment was already there ...

coreydaley:
@bparees updated to reflect your requested changes.

bparees:
lgtm, will let @jim-minter have one last pass if he wants it.

examples/nginx-centos7/s2i/bin/save-artifacts
jim-minter: @coreydaley this doesn't make sense, please change it.  Either no command, or `echo artifact >/tmp/artifact; tar -C /tmp -c artifact` would be preferable.

@bparees fwiw (and fortunately), s2i appears to be perfectly happy with a zero length tar stream.

jim-minter:
after fixing the save-artifacts script, lgtm, thanks!

examples/nginx-centos7/s2i/bin/save-artifacts
coreydaley: @jim-minter According to this (https://github.com/openshift/source-to-image/blob/master/docs/builder_image.md#save-artifacts) that script is correct, there is an index.html file in the directory /usr/share/nginx/html

bparees: it's correct functionally, it's not correct logically because you'd never save/restore your html files in this way.  you save/restore content you pulled down in your assemble script.  the html files are coming out of the git repo (presumably).

jim-minter: My concerns are:

1) when s2i build --incremental is specified, I believe the behaviour implemented will tar/untar all of /usr/share/nginx/html from the old container into /usr/share/nginx/html, then copy the new "source" over that.  So there's no straightforward way for a user to *remove* a file from /usr/share/nginx/html once it gets into a container, and the builds become impure - hardly exemplary behaviour.

2) partially because of 1), the example as it stands confounds more than illuminates.  If I were trying to learn how to persist C object files or a Maven cache across incremental builds, I think this example would do the opposite of helping me to understand the process.

coreydaley:
@jim-minter updated

jim-minter:
@bparees lgtm

bparees:
[merge]

PR #690, opened 2017-02-15
update ONBUILD docs link to non-breaking link

lholmquist:
the current link to the ONBUILD docs on dockerhub, https://docs.docker.com/reference/builder/#onbuild , is no longer working,  looks like they changed the location of those docs


bparees:
lgtm, thanks!
[merge]

PR #693, opened 2017-02-27
use indefinite timeout for docker commit

bparees:
fixes bug https://bugzilla.redhat.com/show_bug.cgi?id=1391665

### DIFF
--- a/pkg/docker/docker.go
+++ b/pkg/docker/docker.go
@@ -328,7 +328,7 @@ func New(config *api.DockerConfig, auth api.AuthConfig) (Docker, error) {
 
 func getDefaultContext() (context.Context, context.CancelFunc) {
 	// the intention is: all docker API calls with the exception of known long-
-	// running calls (ContainerWait, ImagePull, ImageBuild) must complete within a
+	// running calls (ContainerWait, ImagePull, ImageBuild, ImageCommit) must complete within a
 	// certain timeout otherwise we bail.
 	return context.WithTimeout(context.Background(), DefaultDockerTimeout)
 }
@@ -1029,8 +1029,6 @@ func (d *stiDocker) GetImageID(name string) (string, error) {
 // CommitContainer commits a container to an image with a specific tag.
 // The new image ID is returned
 func (d *stiDocker) CommitContainer(opts CommitContainerOptions) (string, error) {
-	ctx, cancel := getDefaultContext()
-	defer cancel()
 	dockerOpts := dockertypes.ContainerCommitOptions{
 		Reference: opts.Repository,
 	}
@@ -1046,7 +1044,7 @@ func (d *stiDocker) CommitContainer(opts CommitContainerOptions) (string, error)
 		glog.V(2).Infof("Committing container with dockerOpts: %+v, config: %+v", dockerOpts, config)
 	}
 
-	resp, err := d.client.ContainerCommit(ctx, opts.ContainerID, dockerOpts)
+	resp, err := d.client.ContainerCommit(context.Background(), opts.ContainerID, dockerOpts)
 	if err == nil {
 		return resp.ID, nil
 	}


bparees:
@openshift/devex ptal.

we are seeing cases where the commit takes longer than 2 minutes and rather than bump it up to another not-long-enough value, i'm in favor of not having a timeout for this long-running operation, like how we handle image pulls and running containers.

bparees:
[test]

pkg/docker/docker.go
```diff
@@ -333,6 +333,10 @@ func getDefaultContext() (context.Context, context.CancelFunc) {
 	return context.WithTimeout(context.Background(), DefaultDockerTimeout)
 }
 
+func getContext(timeout time.Duration) (context.Context, context.CancelFunc) {
```
jim-minter: This function is added but not used - suggest removing it from the patch.

pkg/docker/docker.go
```diff
@@ -333,6 +333,10 @@ func getDefaultContext() (context.Context, context.CancelFunc) {
 	return context.WithTimeout(context.Background(), DefaultDockerTimeout)
```
jim-minter: I'd consider adding 'ImageCommit, ' to the comment on line 331

bparees: done

bparees:
updated and also switched remove container to not be a force per recommendation of the docker team.

jim-minter:
> updated and also switched remove container to not be a force per recommendation of the docker team.

Hmm. I think that previously if you CTRL+C'd s2i while it was doing it's thing, the fact that force was set to true meant that it would clean up its container if it was still running.  With force unset, I think that won't happen any more.  Is that what we want?

bparees:
> Hmm. I think that previously if you CTRL+C'd s2i while it was doing its thing, the fact that force was set to true meant that it would clean up its container if it was still running. 

@rhvgoyal any opinions on this?  I think @jim-minter is correct that this is why we wanted force.  Can you elaborate on why you do not think we should be using force?  You mentioned some cleanup logic that would later remove containers that someone had attempted to cleanup, in what environments would that cleanup logic be running?

rhvgoyal:
force can do lot of bad things. Its like destructive removal of container and can leave lot of internal state uncleaned. To me its more like a debug thing where you really want container to disappear from your view and leave whatever mess behind.

If you are worried about container running, you can first try to stop container yourself and then remove it. Using force is not the solution, IMO.

bparees:
> force can do lot of bad things. Its like destructive removal of container and can leave lot of internal state uncleaned.

can you elaborate on this?  I don't understand why the docker apis would not be clean enough to deal with this properly.

bparees:
since it's not the primary concern, i'm leaving the force removal in place.  we can continue the discussion about whether it should be used or not but for now i want to get the increased timeout fix into origin.

bparees:
[merge]

rhvgoyal:
@bparees if a resource is not free, how can docker api make it free just because you have asked for force removal.

For example, say a process has kept a mount point busy and hence the device busy. Now container can't be removed till that mount point becomes free. 

If you specify force, docker api will clean whatever it can clean and leave rest of it behind. Very fact you have to use force, something is not right. Its more of a debugging option.

bparees:
> @bparees if a resource is not free, how can docker api make it free just because you have asked for force removal.

@rhvgoyal by terminating the container.

rhvgoyal:
@bparees but mount points can leak in other mount namespace where pid does not belong to container.

Also for example, this case of "docker commit" in progress. There is no container process running. One of the docker threads is performing commit operation and device is in use can container can't be removed.

rhvgoyal:
Users should use non force removal of container. And if there is a problem, then either it is a bug fix and should be fixed or it is wrong sequence of removal and more synchronization is needed either in docker api or on the caller side.

For example, in this case, s2i need to wait for commit to finish and then call container removal (non-force version).

bparees:
@rhvgoyal we do wait for the commit to finish before removing the container.  the case the bug hit is when it took over 2 minutes to commit the container (which i think still needs to be investigated, we never see times that long) and we gave up on waiting for the commit, assuming docker had hung, and did our best to clean up what we could, hence force removing the container.

rhvgoyal:
If docker api does not give notion of timeout, then at best you can timeout and try to remove container normally. And if removal fails, try again later. Instead of forcing the removal. 

Container rootfs is busy now and force remval is not going to help with that. 

So I don't think that force removal can be justified in anyway for regular operations.

rhvgoyal:
If docker had hung, how force removal will help?

bparees:
> If docker had hung, how force removal will help?

dockering hanging committing the container doesn't mean the entire docker daemon is necessarily hung or that it's not worth attempting to clean up something.

rhvgoyal:
And who came up with a time limit of 2 minutes. How did you determine that. It is so arbitrary. A big container can easily take more than 2 minutes. I have seen that cloud storage can easily get throttled and that can take more time. VM can just be overloaded in terms of cpu time, and then commit can take a long time. 

So coming with an arbitrary small limit and then assuming docker is hung, is not right.

rhvgoyal:
I think this notion of assuming that if commit does not succeed in 2 mins, container is hung wrong. Should callers of source to image assume same. If source to image can't finish certain image build in 2 mins, s2i is hung and kill it.

bparees:
> I think this notion of assuming that if commit does not succeed in 2 mins, container is hung wrong.

hence why this PR removes that timeout.  However the bug in question is the *only* time we have ever seen 2 minutes be insufficient for a commit.  We have not had a single other issue report where we hit the timeout.  Which makes me think it being an issue with that environment more likely than a particularly unusual container size.

rhvgoyal:
if you think this is environment problem, sure debug that and see where the delay occurred.

bparees:
[merge]

bparees:
[merge]

bparees:
[merge]

rhvgoyal:
@bparees force removal will also leak resources and soon will fill up your thin pool and there is no easy way to clean it up. So another reason why force removal is a bad idea.

https://github.com/docker/docker/issues/31628

jim-minter:
@rhvgoyal thanks - agreed - https://github.com/openshift/source-to-image/pull/706 submitted to cover this.

PR #694, opened 2017-02-28
Gather build step timing information

coreydaley:
Gathers build step timing information for significant build steps and records it, making it available to the 'oc describe' command

Dependency for https://github.com/openshift/origin/pull/13307

### DIFF
--- a/pkg/api/helpers.go
+++ b/pkg/api/helpers.go
@@ -1,0 +1,42 @@
+package api
+
+import "time"
+
+// RecordStageAndStepInfo records details about each build stage and step
+func RecordStageAndStepInfo(stages Stages, stageName StageName, stepName StepName, startTime time.Time, endTime time.Time) Stages {
+	// Make sure that the stages slice is initialized
+	if len(stages) == 0 {
+		stages = make([]StageInfo, 0)
+	}
+
+	// If the stage already exists  update the endTime and Duration, and append the new step.
+	for stageKey, stageVal := range stages {
+		if stageVal.StageName == stageName {
+			stages[stageKey].Duration = endTime.Sub(stages[stageKey].StartTime)
+			if len(stages[stageKey].Steps) == 0 {
+				stages[stageKey].Steps = make([]StepInfo, 0)
+			}
+			stages[stageKey].Steps = append(stages[stageKey].Steps, StepInfo{
+				StepName:  stepName,
+				StartTime: startTime,
+				Duration:  endTime.Sub(startTime),
+			})
+			return stages
+		}
+	}
+
+	// If the stageName does not exist, add it to the slice along with the new step.
+	steps := make([]StepInfo, 0)
+	steps = append(steps, StepInfo{
+		StepName:  stepName,
+		StartTime: startTime,
+		Duration:  endTime.Sub(startTime),
+	})
+	stages = append(stages, StageInfo{
+		StageName: stageName,
+		StartTime: startTime,
+		Duration:  endTime.Sub(startTime),
+		Steps:     steps,
+	})
+	return stages
+}
--- a/pkg/api/types.go
+++ b/pkg/api/types.go
@@ -6,6 +6,7 @@ import (
 	"net/url"
 	"path/filepath"
 	"strings"
+	"time"
 
 	utilglog "github.com/openshift/source-to-image/pkg/util/glog"
 
@@ -341,13 +342,97 @@ type Result struct {
 	BuildInfo BuildInfo
 }
 
-// BuildInfo holds information about a particular step in the build process.
+// BuildInfo contains information about the build process.
 type BuildInfo struct {
+	// Stages contains details about each build stage.
+	Stages Stages
+
 	// FailureReason is a camel case reason that is used by the machine to reply
 	// back to the OpenShift builder with information why any of the steps in the
-	// build, failed.
+	// build failed.
 	FailureReason FailureReason
 }
+
+// Stages is a slice of build stages that have been recorded.
+type Stages []StageInfo
+
+// StageInfo contains details about a build stage.
+type StageInfo struct {
+	// StageName is the identifier for each build stage.
+	StageName StageName
+
+	// StartTime identifies when this stage started.
+	StartTime time.Time
+
+	// Duration identifies how long this stage ran.
+	Duration time.Duration
+
+	// Steps contains details about each build step within a build stage.
+	Steps Steps
+}
+
+// StageName is the identifier for each build stage.
+type StageName string
+
+// Valid StageNames
+const (
+	// StagePullImages pulls the docker images.
+	StagePullImages StageName = "PullImages"
+
+	//StageAssemble runs the assemble steps.
+	StageAssemble StageName = "Assemble"
+
+	// StageBuild builds the source.
+	StageBuild StageName = "Build"
+
+	// StageCommit commits the container.
+	StageCommit StageName = "CommitContainer"
+
+	// StageRetrieve retrieves artifacts.
+	StageRetrieve StageName = "RetrieveArtifacts"
+)
+
+// Steps is a slice of build steps that have been recorded within each build stage.
+type Steps []StepInfo
+
+// StepInfo contains details about a build step.
+type StepInfo struct {
+	// StepName is the identifier for each build step.
+	StepName StepName
+
+	// StartTime identifies when this step started.
+	StartTime time.Time
+
+	// Duration identifies how long this step ran.
+	Duration time.Duration
+}
+
+// StepName is the identifier for each build step.
+type StepName string
+
+// Valid StepNames
+const (
+	// StepPullBuilderImage pulls the builder image.
+	StepPullBuilderImage StepName = "PullBuilderImage"
+
+	// StepPullPreviousImage pulls the previous image for an incremental build.
+	StepPullPreviousImage StepName = "PullPreviousImage"
+
+	// StepPullRuntimeImage pull the runtime image.
+	StepPullRuntimeImage StepName = "PullRuntimeImage"
+
+	// StepAssembleBuildScripts runs the assemble scripts.
+	StepAssembleBuildScripts StepName = "AssembleBuildScripts"
+
+	// StepBuildDockerImage builds the Docker image for layered builds.
+	StepBuildDockerImage StepName = "BuildDockerImage"
+
+	// StepCommitContainer commits the container to the builder image.
+	StepCommitContainer StepName = "CommitContainer"
+
+	// StepRetrievePreviousArtifacts restores archived artifacts from the previous build.
+	StepRetrievePreviousArtifacts StepName = "RetrievePreviousArtifacts"
+)
 
 // StepFailureReason holds the type of failure that occurred during the build
 // process.
--- a/pkg/build/strategies/layered/layered.go
+++ b/pkg/build/strategies/layered/layered.go
@@ -176,7 +176,10 @@ func (builder *Layered) Build(config *api.Config) (*api.Result, error) {
 	docker.StreamContainerIO(outReader, nil, func(s string) { glog.V(2).Info(s) })
 
 	glog.V(2).Infof("Building new image %s with scripts and sources already inside", newBuilderImage)
-	if err := builder.docker.BuildImage(opts); err != nil {
+	startTime := time.Now()
+	err := builder.docker.BuildImage(opts)
+	buildResult.BuildInfo.Stages = api.RecordStageAndStepInfo(buildResult.BuildInfo.Stages, api.StageBuild, api.StepBuildDockerImage, startTime, time.Now())
+	if err != nil {
 		buildResult.BuildInfo.FailureReason = utilstatus.NewFailureReason(
 			utilstatus.ReasonDockerImageBuildFailed,
 			utilstatus.ReasonMessageDockerImageBuildFailed,
@@ -206,7 +209,10 @@ func (builder *Layered) Build(config *api.Config) (*api.Result, error) {
 	}
 
 	glog.V(2).Infof("Building %s using sti-enabled image", builder.config.Tag)
-	if err := builder.scripts.Execute(api.Assemble, config.AssembleUser, builder.config); err != nil {
+	startTime = time.Now()
+	err = builder.scripts.Execute(api.Assemble, config.AssembleUser, builder.config)
+	buildResult.BuildInfo.Stages = api.RecordStageAndStepInfo(buildResult.BuildInfo.Stages, api.StageAssemble, api.StepAssembleBuildScripts, startTime, time.Now())
+	if err != nil {
 		buildResult.BuildInfo.FailureReason = utilstatus.NewFailureReason(
 			utilstatus.ReasonAssembleFailed,
 			utilstatus.ReasonMessageAssembleFailed,
--- a/pkg/build/strategies/sti/postexecutorstep.go
+++ b/pkg/build/strategies/sti/postexecutorstep.go
@@ -9,6 +9,7 @@ import (
 	"path"
 	"path/filepath"
 	"strings"
+	"time"
 
 	"github.com/openshift/source-to-image/pkg/api"
 	dockerpkg "github.com/openshift/source-to-image/pkg/docker"
@@ -128,7 +129,7 @@ func (step *commitImageStep) execute(ctx *postExecutorStepContext) error {
 	if entrypoint == nil {
 		entrypoint = []string{}
 	}
-
+	startTime := time.Now()
 	ctx.imageID, err = commitContainer(
 		step.docker,
 		ctx.containerID,
@@ -139,7 +140,7 @@ func (step *commitImageStep) execute(ctx *postExecutorStepContext) error {
 		entrypoint,
 		ctx.labels,
 	)
-
+	step.builder.result.BuildInfo.Stages = api.RecordStageAndStepInfo(step.builder.result.BuildInfo.Stages, api.StageCommit, api.StepCommitContainer, startTime, time.Now())
 	if err != nil {
 		step.builder.result.BuildInfo.FailureReason = utilstatus.NewFailureReason(
 			utilstatus.ReasonCommitContainerFailed,
@@ -226,6 +227,10 @@ func (step *downloadFilesFromBuilderImageStep) downloadAndExtractFile(artifactPa
 	}()
 
 	if err := step.docker.DownloadFromContainer(artifactPath, fd, containerID); err != nil {
+		step.builder.result.BuildInfo.FailureReason = utilstatus.NewFailureReason(
+			utilstatus.ReasonGenericS2IBuildFailed,
+			utilstatus.ReasonMessageGenericS2iBuildFailed,
+		)
 		return fmt.Errorf("could not download file (%q -> %q) from container %s: %v", artifactPath, fd.Name(), containerID, err)
 	}
 
@@ -260,6 +265,10 @@ func (step *startRuntimeImageAndUploadFilesStep) execute(ctx *postExecutorStepCo
 
 	fd, err := ioutil.TempFile("", "s2i-upload-done")
 	if err != nil {
+		step.builder.result.BuildInfo.FailureReason = utilstatus.NewFailureReason(
+			utilstatus.ReasonGenericS2IBuildFailed,
+			utilstatus.ReasonMessageGenericS2iBuildFailed,
+		)
 		return err
 	}
 	fd.Close()
@@ -281,6 +290,10 @@ func (step *startRuntimeImageAndUploadFilesStep) execute(ctx *postExecutorStepCo
 		destinationDir := filepath.Join(artifactsDir, "scripts")
 		err = step.copyScriptIfNeeded(script, destinationDir)
 		if err != nil {
+			step.builder.result.BuildInfo.FailureReason = utilstatus.NewFailureReason(
+				utilstatus.ReasonGenericS2IBuildFailed,
+				utilstatus.ReasonMessageGenericS2iBuildFailed,
+			)
 			return err
 		}
 	}
@@ -288,6 +301,10 @@ func (step *startRuntimeImageAndUploadFilesStep) execute(ctx *postExecutorStepCo
 	image := step.builder.config.RuntimeImage
 	workDir, err := step.docker.GetImageWorkdir(image)
 	if err != nil {
+		step.builder.result.BuildInfo.FailureReason = utilstatus.NewFailureReason(
+			utilstatus.ReasonGenericS2IBuildFailed,
+			utilstatus.ReasonMessageGenericS2iBuildFailed,
+		)
 		return fmt.Errorf("could not get working dir of %q image: %v", image, err)
 	}
 
@@ -298,9 +315,17 @@ func (step *startRuntimeImageAndUploadFilesStep) execute(ctx *postExecutorStepCo
 		var scriptsURL string
 		scriptsURL, err = step.docker.GetScriptsURL(image)
 		if err != nil {
+			step.builder.result.BuildInfo.FailureReason = utilstatus.NewFailureReason(
+				utilstatus.ReasonGenericS2IBuildFailed,
+				utilstatus.ReasonMessageGenericS2iBuildFailed,
+			)
 			return err
 		}
 		if len(scriptsURL) == 0 {
+			step.builder.result.BuildInfo.FailureReason = utilstatus.NewFailureReason(
+				utilstatus.ReasonGenericS2IBuildFailed,
+				utilstatus.ReasonMessageGenericS2iBuildFailed,
+			)
 			return fmt.Errorf("could not determine scripts URL for image %q", image)
 		}
 		commandBaseDir = strings.TrimPrefix(scriptsURL, "image://")
--- a/pkg/build/strategies/sti/sti.go
+++ b/pkg/build/strategies/sti/sti.go
@@ -10,6 +10,7 @@ import (
 	"path/filepath"
 	"regexp"
 	"strings"
+	"time"
 
 	"github.com/openshift/source-to-image/pkg/api"
 	"github.com/openshift/source-to-image/pkg/build"
@@ -203,7 +204,6 @@ func (builder *STI) Build(config *api.Config) (*api.Result, error) {
 	defer builder.garbage.Cleanup(config)
 
 	glog.V(1).Infof("Preparing to build %s", config.Tag)
-	// The failure reason is updated inside the Prepare function.
 	if err := builder.preparer.Prepare(config); err != nil {
 		return builder.result, err
 	}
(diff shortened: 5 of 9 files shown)


coreydaley:
[test]

pkg/api/types.go
```diff
@@ -343,12 +344,56 @@ type Result struct {
...
+type StepName string
+
+const (
+	// AssembleScriptStep is the build step when the Assemble script is being
+	// run.
+	AssembleScriptStep StepName = "AssembleScriptStep"
+
+	// GatherInputContentStep is the build step when we try to fetch the source for the
+	// build.
+	GatherInputContentStep StepName = "GatherInputContentStep"
```
bparees: this is done by openshift outside of s2i at this point, so whatever number you are collecting here is probably not relevant/accurate.

pkg/api/types.go
```diff
@@ -343,12 +344,56 @@ type Result struct {
...
+	// container to the builder image.
+	CommitContainerStep StepName = "CommitContainerStep"
+
+	// PullPreviousImageStep is the build step that occurs when we try to pull a
+	// previous image for a incremental build.
+	PullPreviousImageStep StepName = "PullPreviousImageStep"
+
+	// PullRuntimeImageStep is the build step that occurs when we try to pull the
+	// runtime image.
+	PullRuntimeImageStep StepName = "PullRuntimeImageStep"
```
bparees: we can leave this out, seeing as we're deprecating this feature from an openshift perspective.

pkg/build/strategies/sti/postexecutorstep.go
```diff
@@ -218,6 +224,11 @@ func (step *downloadFilesFromBuilderImageStep) downloadAndExtractFile(artifactPa
 			utilstatus.ReasonFSOperationFailed,
 			utilstatus.ReasonMessageFSOperationFailed,
 		)
+		step.builder.result.BuildInfo.StepInfo = utilstatus.AddStepInfo(
+			step.builder.result.BuildInfo.StepInfo,
+			api.GatherInputContentStep,
+			timeStart,
+		)
```
bparees: this is, i think, part of the extended-build support so we don't need to worry about timing for it.

pkg/build/strategies/sti/postexecutorstep.go
```diff
@@ -226,6 +237,15 @@ func (step *downloadFilesFromBuilderImageStep) downloadAndExtractFile(artifactPa
...
 	if err := step.docker.DownloadFromContainer(artifactPath, fd, containerID); err != nil {
+		step.builder.result.BuildInfo.FailureReason = utilstatus.NewFailureReason(
+			utilstatus.ReasonGenericS2IBuildFailed,
+			utilstatus.ReasonMessageGenericS2iBuildFailed,
+		)
+		step.builder.result.BuildInfo.StepInfo = utilstatus.AddStepInfo(
+			step.builder.result.BuildInfo.StepInfo,
+			api.GatherInputContentStep,
+			timeStart,
+		)
```
bparees: same here.

pkg/build/strategies/sti/postexecutorstep.go
```diff
@@ -235,6 +255,11 @@ func (step *downloadFilesFromBuilderImageStep) downloadAndExtractFile(artifactPa
 			utilstatus.ReasonGenericS2IBuildFailed,
 			utilstatus.ReasonMessageGenericS2iBuildFailed,
 		)
+		step.builder.result.BuildInfo.StepInfo = utilstatus.AddStepInfo(
+			step.builder.result.BuildInfo.StepInfo,
+			api.GatherInputContentStep,
+			timeStart,
+		)
```
bparees: and here.

pkg/build/strategies/sti/postexecutorstep.go
```diff
@@ -243,8 +268,18 @@ func (step *downloadFilesFromBuilderImageStep) downloadAndExtractFile(artifactPa
...
```
bparees: and here

pkg/build/strategies/sti/postexecutorstep.go
```diff
@@ -243,8 +268,18 @@ func (step *downloadFilesFromBuilderImageStep) downloadAndExtractFile(artifactPa
...
 		return fmt.Errorf("could not extract runtime artifact %q into the directory %q: %v", artifactPath, artifactsDir, err)
 	}
+	step.builder.result.BuildInfo.StepInfo = utilstatus.AddStepInfo(
+		step.builder.result.BuildInfo.StepInfo,
+		api.GatherInputContentStep,
+		timeStart,
+	)
```
bparees: can be removed.

pkg/build/strategies/sti/postexecutorstep.go
```diff
@@ -256,10 +291,20 @@ type startRuntimeImageAndUploadFilesStep struct {
...
 	if err != nil {
+		step.builder.result.BuildInfo.StepInfo = utilstatus.AddStepInfo(
+			step.builder.result.BuildInfo.StepInfo,
+			api.GatherInputContentStep,
+			timeStart,
+		)
+		step.builder.result.BuildInfo.FailureReason = utilstatus.NewFailureReason(
+			utilstatus.ReasonGenericS2IBuildFailed,
+			utilstatus.ReasonMessageGenericS2iBuildFailed,
+		)
```
bparees: can be removed (extended builds)

pkg/build/strategies/sti/postexecutorstep.go
```diff
@@ -281,13 +326,32 @@ func (step *startRuntimeImageAndUploadFilesStep) execute(ctx *postExecutorStepCo
...
 		if err != nil {
+			step.builder.result.BuildInfo.StepInfo = utilstatus.AddStepInfo(
+				step.builder.result.BuildInfo.StepInfo,
+				api.GatherInputContentStep,
+				timeStart,
+			)
+			step.builder.result.BuildInfo.FailureReason = utilstatus.NewFailureReason(
+				utilstatus.ReasonGenericS2IBuildFailed,
+				utilstatus.ReasonMessageGenericS2iBuildFailed,
+			)
```
bparees: remove

pkg/build/strategies/sti/postexecutorstep.go
```diff
@@ -281,13 +326,32 @@ func (step *startRuntimeImageAndUploadFilesStep) execute(ctx *postExecutorStepCo
...
+
+		step.builder.result.BuildInfo.StepInfo = utilstatus.AddStepInfo(
+			step.builder.result.BuildInfo.StepInfo,
+			api.GatherInputContentStep,
+			timeStart,
+		)
+		step.builder.result.BuildInfo.FailureReason = utilstatus.NewFailureReason(
+			utilstatus.ReasonGenericS2IBuildFailed,
+			utilstatus.ReasonMessageGenericS2iBuildFailed,
+		)
```
bparees: remove

pkg/build/strategies/sti/sti.go
```diff
@@ -269,15 +287,21 @@ func (builder *STI) Prepare(config *api.Config) error {
...
 
 	if len(config.RuntimeImage) > 0 {
-		if err = dockerpkg.GetRuntimeImage(config, builder.runtimeDocker); err != nil {
+		timeStart := time.Now()
+		err = dockerpkg.GetRuntimeImage(config, builder.runtimeDocker)
+		builder.result.BuildInfo.StepInfo = utilstatus.AddStepInfo(
+			builder.result.BuildInfo.StepInfo,
+			api.PullRuntimeImageStep,
+			timeStart,
+		)
```
bparees: not needed.

pkg/build/strategies/sti/sti.go
```diff
@@ -345,7 +369,14 @@ func (builder *STI) Prepare(config *api.Config) error {
...
 	// fetch sources, for their .s2i/bin might contain s2i scripts
 	if len(config.Source) > 0 {
-		if builder.sourceInfo, err = builder.source.Download(config); err != nil {
+		timeStart := time.Now()
+		builder.sourceInfo, err = builder.source.Download(config)
+		builder.result.BuildInfo.StepInfo = utilstatus.AddStepInfo(
+			builder.result.BuildInfo.StepInfo,
+			api.GatherInputContentStep,
+			timeStart,
+		)
```
bparees: not needed.  the "source.Download" that is performed when invoking s2i from openshift is just a file copy:
https://github.com/openshift/source-to-image/blob/master/pkg/scm/file/download.go#L21

because openshift has already done the clone operation.

and it's probably not worth calculating the filecopy time.

pkg/build/strategies/sti/sti.go
```diff
@@ -496,6 +533,7 @@ func (builder *STI) Save(config *api.Config) (err error) {
 	extractFunc := func(string) error {
 		extractErr := builder.tar.ExtractTarStream(artifactTmpDir, outReader)
 		io.Copy(ioutil.Discard, outReader) // must ensure reader from container is drained
+
```
bparees: this is a step that i would expect to have its own name and timing (RetrievePreviousArtifacts or something)

pkg/util/status/build_status.go
```diff
@@ -114,10 +116,36 @@ const (
...
-// reason and a message to be displayed
+// reason and a message to be displayed.
 func NewFailureReason(reason api.StepFailureReason, message api.StepFailureMessage) api.FailureReason {
 	return api.FailureReason{
 		Reason:  reason,
 		Message: message,
 	}
 }
+
+// AddStepInfo appends timing information about specific build stepts to the
```
bparees: steps

pkg/util/status/build_status.go
```diff
@@ -114,10 +116,36 @@ const (
...
+func AddStepInfo(stepInfo []api.BuildStepInfo, stepName api.StepName, timeStart time.Time) []api.BuildStepInfo {
+	timeStop := time.Now()
+
+	for i, step := range stepInfo {
+		if step.Name != stepName {
+			continue
+		}
+		stepInfo[i] = api.BuildStepInfo{
+			Name:      step.Name,
+			StartTime: timeStart,
```
bparees: shouldn't you be taking the startTime from the existing Step?

pkg/util/status/build_status.go
```diff
@@ -114,10 +116,36 @@ const (
...
 func NewFailureReason(reason api.StepFailureReason, message api.StepFailureMessage) api.FailureReason {
 	return api.FailureReason{
 		Reason:  reason,
 		Message: message,
 	}
 }
+
+// AddStepInfo appends timing information about specific build stepts to the
+// result object. In case the step was already added, it will update the stop
+// time, this allows timing for substeps to be aggregated under a generic step.
```
bparees: i'm not sure how i feel about this.  If i run 6 steps and label them all "GatherInput" but they are intermingled with other actions, you're going to report that GatherInput took way longer than it really did.

I think we should just aim to structure the code+definition of "steps" better such that we can explicitly start/stop the timing for a given step exactly once.  If there are cases where you think we are going a little bit of input gathering, then go off and do something else, then come back and do some more input gathering, we should discuss them and decide if maybe we just need more fine grained "step" definitions rather than lumping all those things as "input gathering".  And again i think a lot of this goes away anyway once you stop worrying about timings for flows that are either

1) specific to extended builds
2) specific to standalone s2i usage because openshift handles that step externally

coreydaley:
@bparees ptal
I still need to update the godoc for the StepName(s)
I also plan on removing the "GatherInputContent" generic one once I update the tests.

I thought that a map made more sense for the 'Steps' so that they can be accessed directly instead of having to iterate over a slice and see if the step already existed etc.  

Let me know if there is anywhere else you think that we should be including timing info

pkg/build/strategies/layered/layered.go
```diff
@@ -206,16 +208,19 @@ func (builder *Layered) Build(config *api.Config) (*api.Result, error) {
 	}
 
 	glog.V(2).Infof("Building %s using sti-enabled image", builder.config.Tag)
-	if err := builder.scripts.Execute(api.Assemble, config.AssembleUser, builder.config); err != nil {
+	startTime = time.Now()
+	err2 := builder.scripts.Execute(api.Assemble, config.AssembleUser, builder.config)
```
bparees: why not just err =?

pkg/build/strategies/sti/postexecutorstep.go
```diff
@@ -171,7 +172,10 @@ func (step *downloadFilesFromBuilderImageStep) execute(ctx *postExecutorStepCont
 	}
 
 	for _, artifact := range step.builder.config.RuntimeArtifacts {
-		if err := step.downloadAndExtractFile(artifact.Source, artifactsDir, ctx.containerID); err != nil {
+		startTime := time.Now()
+		err := step.downloadAndExtractFile(artifact.Source, artifactsDir, ctx.containerID)
+		step.builder.result.BuildInfo.Steps = utilstatus.RecordStepInfo(step.builder.result.BuildInfo.Steps, api.StepDownloadAndExtractArtifacts, startTime)
+		if err != nil {
```
bparees: i believe this is part of the extended build flow, you can ignore it.

pkg/build/strategies/sti/sti.go
```diff
@@ -269,15 +280,17 @@ func (builder *STI) Prepare(config *api.Config) error {
 	builder.result.WorkingDir = config.WorkingDir
 
 	if len(config.RuntimeImage) > 0 {
-		if err = dockerpkg.GetRuntimeImage(config, builder.runtimeDocker); err != nil {
+		startTime := time.Now()
+		err = dockerpkg.GetRuntimeImage(config, builder.runtimeDocker)
+		builder.result.BuildInfo.Steps = utilstatus.RecordStepInfo(builder.result.BuildInfo.Steps, api.StepPullRuntimeImage, startTime)
+		if err != nil {
```
bparees: irrelevant step for openshift.

pkg/build/strategies/sti/sti.go
```diff
@@ -345,7 +358,10 @@ func (builder *STI) Prepare(config *api.Config) error {
 
 	// fetch sources, for their .s2i/bin might contain s2i scripts
 	if len(config.Source) > 0 {
-		if builder.sourceInfo, err = builder.source.Download(config); err != nil {
+		startTime := time.Now()
+		builder.sourceInfo, err = builder.source.Download(config)
+		builder.result.BuildInfo.Steps = utilstatus.RecordStepInfo(builder.result.BuildInfo.Steps, api.StepDownloadSource, startTime)
```
bparees: per my comment on your original version of this, when running under openshift, "Download" is really just a filesystem copy and isn't worth timing. (and shouldn't be called Download if we did decide to time it).

pkg/util/status/build_status.go
```diff
@@ -114,10 +116,29 @@ const (
...
+
+// RecordStepInfo populates timing information about specific build steps into the
+// result object.
+func RecordStepInfo(steps map[api.StepName]api.StepInfo, stepName api.StepName, startTime time.Time) map[api.StepName]api.StepInfo {
+	// If the step already exists, only update the StopTime
+	if info, ok := steps[stepName]; ok {
+		steps[stepName] = api.StepInfo{
+			StartTime: info.StartTime,
+			StopTime:  time.Now(),
+		}
```
bparees: why create a new object?  just update the existing object's StopTime.

pkg/build/strategies/sti/sti.go
```diff
@@ -228,16 +228,27 @@ func (builder *STI) Build(config *api.Config) (*api.Result, error) {
...
-			)
+		if e, ok := err.(s2ierr.ContainerError); ok {
+			if !isMissingRequirements(e.Output) {
+				builder.result.BuildInfo.FailureReason = utilstatus.NewFailureReason(
+					utilstatus.ReasonAssembleFailed,
+					utilstatus.ReasonMessageAssembleFailed,
+				)
+				return builder.result, err
+			}
+			glog.V(1).Info("Image is missing basic requirements (sh or tar), layered build will be performed")
```
bparees: So for a layered build you're going to do the following:
1) assemble runs and fails, you record an assemble duration (duration=1->1)
2) you start recording a LayeredBuild duration
3) you start recording a build docker image
4) you finish recording a build docker image (duration=3->4)
5) you start recording another assemble step
6) you finish recording that assemble step (will update the total duration of the assemble step from (1) to reflect the time from 1->6)
7) you finish recording the LayeredBuild step (duration=2->7)

I think we can agree that's going to be messy in terms of the data reported and how a user can interpret it...  some durations are being double counted, there is some overlap, etc.

I don't think we should ever nest step durations, and there should be no overlap between them..  
start stepA
stop stepA
start stepB
stop stepB

never:
startA
startB
stopB
stopA

and never:
startA
startB
stopA
stopB

In an ideal world, the total duration of all the steps recorded should == the total duration of the build.

In reality that won't be the case because we're not going to measure the times for every single step of the build process, so the total duration of the steps we record will be less than the total duration of the build.  but it definitely should not be more.

For the layered build case here, I think if you get an error executing assemble and conclude you're going to do a layered build instead, you should just throw away the original start time you noted, don't record the step time at all (the failed assemble invocation will be mostly a no-op).

And then within the layered build logic, record the DockerBuild duration and the Assemble duration.

pkg/api/types.go
```diff
@@ -341,14 +342,67 @@ type Result struct {
...
+
+	// StepPullPreviousImage is the build step that occurs when we try to pull a
+	// previous image for a incremental build.
+	StepPullPreviousImage StepName = "PullPreviousImage"
+
+	// StepRetrievePreviousArtifacts is ...
+	StepRetrievePreviousArtifacts StepName = "RetrievePreviousArtifacts"
+
+	// StepGatherInputContent is ...
+	StepGatherInputContent StepName = "GatherInputContent"
```
coreydaley: This was a leftover for the tests to compile previously, will be removed in the next push

coreydaley:
[test]

coreydaley:
re[test]

coreydaley:
@bparees Updated per your suggestions

coreydaley:
[(5) Separate out build substatus and times [builds]](https://trello.com/c/KmmYvStu/996-5-separate-out-build-substatus-and-times-builds)

pkg/build/strategies/sti/sti.go
jeremyeder: Question:  why not strive to instrument all steps so that "total duration of all the steps recorded should == the total duration of the build" ?

bparees: @jeremyeder because attaining that level of instrumentation is sisyphean and going to be really hard to maintain going forward any time we add/restructure code.

there are a million things we do during builds and trying to name/define steps for all of them is going to end up as a mess, or we're going to be counting things as a single duration that aren't really related.

In practice I expect the total duration will roughly match the duration of the steps we are reporting, because we're adding timing information for any step that we expect will have a meaningful duration.  But in practice we might have a few seconds here and there that are not accounted for where we're doing some internal stuff.  If it turns out they are significantly off, that will be an indication that there is additional work we should be reporting durations for.

pkg/api/types.go
```diff
@@ -341,14 +342,56 @@ type Result struct {
...
+
+const (
+	// StepAssembleBuildScripts is the build step when the Assemble script is being
+	StepAssembleBuildScripts StepName = "AssembleBuildScripts"
+
+	// StepPullBuilderImage is the build step that is associated with pulling the
+	// builder image.
+	StepPullBuilderImage StepName = "PullBuilderImage"
+
+	// StepBuildDockerImage is the step where the Docker image is built.
```
bparees: is the step where the Docker image is build for layered builds.

pkg/build/strategies/sti/postexecutorstep.go
```diff
@@ -171,7 +172,8 @@ func (step *downloadFilesFromBuilderImageStep) execute(ctx *postExecutorStepCont
 	}
 
 	for _, artifact := range step.builder.config.RuntimeArtifacts {
-		if err := step.downloadAndExtractFile(artifact.Source, artifactsDir, ctx.containerID); err != nil {
+		err := step.downloadAndExtractFile(artifact.Source, artifactsDir, ctx.containerID)
+		if err != nil {
```
bparees: why change this?

pkg/build/strategies/sti/sti.go
```diff
@@ -269,15 +279,15 @@ func (builder *STI) Prepare(config *api.Config) error {
 	builder.result.WorkingDir = config.WorkingDir
 
 	if len(config.RuntimeImage) > 0 {
-		if err = dockerpkg.GetRuntimeImage(config, builder.runtimeDocker); err != nil {
+		err = dockerpkg.GetRuntimeImage(config, builder.runtimeDocker)
+		if err != nil {
```
bparees: why change this?

pkg/build/strategies/sti/sti.go
```diff
@@ -345,7 +355,8 @@ func (builder *STI) Prepare(config *api.Config) error {
 
 	// fetch sources, for their .s2i/bin might contain s2i scripts
 	if len(config.Source) > 0 {
-		if builder.sourceInfo, err = builder.source.Download(config); err != nil {
+		builder.sourceInfo, err = builder.source.Download(config)
+		if err != nil {
```
bparees: why change this?

pkg/util/status/build_status.go
```diff
@@ -114,10 +116,31 @@ const (
...
 func NewFailureReason(reason api.StepFailureReason, message api.StepFailureMessage) api.FailureReason {
 	return api.FailureReason{
 		Reason:  reason,
 		Message: message,
 	}
 }
+
+// RecordStepInfo populates timing information about specific build steps into the
+// result object.
+func RecordStepInfo(steps api.Steps, stepName api.StepName, startTime time.Time) api.Steps {
```
bparees: make this a method on the api.Steps struct and just let it mutate itself.

coreydaley:
@bparees updates made, ptal

pkg/build/strategies/sti/sti.go
```diff
@@ -228,20 +228,30 @@ func (builder *STI) Build(config *api.Config) (*api.Result, error) {
 	} else {
 		glog.V(1).Infof("Running %q in %q", api.Assemble, config.Tag)
 	}
-	if err := builder.scripts.Execute(api.Assemble, config.AssembleUser, config); err != nil {
+	startTime := time.Now()
+	err := builder.scripts.Execute(api.Assemble, config.AssembleUser, config)
+
+	if err != nil {
```
bparees: can still use the if err:=foo; err!=nil { syntax here.

coreydaley:
Added k8s.io/kubernetes/pkg/api/unversioned (along with it's dependencies) to the vendor directory as origin will be using unversioned.Time for the api type for the StepInfo.StartTime and StepInfo.StopTime

bparees:
> Added k8s.io/kubernetes/pkg/api/unversioned (along with it's dependencies) to the vendor directory as origin will be using unversioned.Time for the api type for the StepInfo.StartTime and StepInfo.StopTime

adding k8s deps is likely to cause us trouble later.  unversioned.Time is just a wrapper around time.Time, so there's no reason the origin code can't wrap the time object it gets back from s2i, as needed.

```
type Time struct {
	time.Time `protobuf:"-"`
}
```

pkg/build/strategies/sti/sti.go
smarterclayton: Is this internal?  Why aren't we using a tracing library vs creating our own?

jim-minter:
@coreydaley random idea, don't know if it'd be useful: perhaps an elapsed function and a closure would help tie together the measuring better, e.g.:

```golang
package main

import (
	"fmt"
	"net/http"
	"time"
)

func elapsed(f func()) time.Duration {
	start := time.Now()

	f()

	return time.Now().Sub(start)
}

func main() {
	var resp *http.Response
	var err error

	duration := elapsed(func() {
		resp, err = http.Get("https://www.google.com")
	})

	fmt.Printf("request took %s\n", duration)
	if err == nil {
		fmt.Printf("status code was %d\n", resp.StatusCode)
	}
}
```

pkg/api/types.go
```diff
@@ -341,14 +342,75 @@ type Result struct {
...
+
 	// FailureReason is a camel case reason that is used by the machine to reply
 	// back to the OpenShift builder with information why any of the steps in the
-	// build, failed.
+	// build failed.
 	FailureReason FailureReason
 }
 
+// Steps is a map of build steps that have been recorded
+type Steps map[StepName]StepInfo
```
jim-minter: I wonder whether (a) it makes sense for the keys to be ordered to help end users interpret the results, and (b) whether you can use time.Duration instead of storing start and stop times. Example:
```golang
type Step struct {
	Name string
	Duration time.Duration
}

type Steps []Step
```

pkg/build/strategies/sti/sti.go
coreydaley: @smarterclayton Could you provide more insight into what you are thinking?  I found go-metrics (https://github.com/armon/go-metrics), but it seems to do something very similar to what we are doing here, but doesn't really make it easier.  We are just trying to provide running time information for specific steps, but not always entire functions.  Did you have a specific library in mind that I can research further?

pkg/api/types.go
coreydaley: @jim-minter Thanks, I'll take a look at that once I have clarified what @smarterclayton is suggesting with the tracing library.

bparees: ordering makes sense... i'm less convinced on storing the duration because in general i prefer to store the raw values so we can present the data in any form we want.  Among other things, it might prove useful if we have "gaps" in the duration (since we aren't measuring absolutely everything, we could have a build that takes 60s, but steps that only account for 40s.  Being able to see when each step started/stopped could help us find the "missing" 20s).

Obviously we'd present duration values to the user, but the api object would store the actual date/time information.

bparees:
> @coreydaley random idea, don't know if it'd be useful: perhaps an elapsed function and a closure would help tie together the measuring better, e.g.:

i have mixed feelings on this, forcing us to wrapper chunks of code into closures seems like it's just going to make the code more complex/nested, though it has the advantage of providing better certainty about how the measurement is started/stopped.

there's also the challenge of cases where we want to "abort" the measurement (if we go to run the assemble script and can't because "tar" is missing, indicating we need to do a layered build, we want to throw away that initial "assemble" duration because it wasn't real, and we're going to re-measure it when we can actually run it after the layered build).  Maybe this can be sufficiently handled by adding an error return to the elapsed() function you propose.

Overall, though, it feels like maybe we're over-engineering this feature.  And of course it depends on what @coreydaley's timing library investigation turns up as well.

bparees:
lgtm pending the resolution to @smarterclayton's concern that we're not using an existing timing measurement library.

coreydaley:
@openshift/devex ptal

examples/nginx-centos7/Dockerfile
```diff
@@ -32,7 +32,6 @@ RUN sed -i 's/user nginx;//' /etc/nginx/nginx.conf
 # Copy the S2I scripts to /usr/libexec/s2i since we set the label that way
 COPY ./s2i/bin/ /usr/libexec/s2i
 
-RUN chown -R 1001:1001 /opt/app-root
```
bparees: ?

pkg/api/types.go
```diff
@@ -341,14 +342,101 @@ type Result struct {
...
+	StageName StageName
+
+	// StartTime identifies when this stage started.
+	StartTime time.Time
+
+	// EndTime identifies when this stage ended.
+	EndTime time.Time
+
+	// Duration identifies how long this stage ran.
+	Duration time.Duration
```
bparees: we shouldn't need both duration and endtime (if you have one, you can get the other), since ultimately we're targeting an api of "starttime" and "duration", just store those two things.

pkg/api/helpers.go
```diff
@@ -0,0 +1,46 @@
...
+	// Make sure that the stages slice is initialized
+	if len(stages) == 0 {
+		stages = make([]StageInfo, 0)
+	}
+
+	// If the stage already exists  update the endTime and Duration, and append the new step.
+	for stageKey, stageVal := range stages {
+		if stageVal.StageName == stageName {
+			stages[stageKey].EndTime = endTime
+			stages[stageKey].Duration = stages[stageKey].Duration + endTime.Sub(startTime)
```
bparees: just record stage duration as:  startTime of the (endtime of the last step)-(starttime of the first step).

(and get rid of the endtime field, per my comment below)

coreydaley:
@bparees 
Updated per your suggestions on the corresponding origin pull request, ptal.

pkg/api/helpers.go
```diff
@@ -0,0 +1,42 @@
...
+func RecordStageAndStepInfo(stages Stages, stageName StageName, stepName StepName, startTime time.Time, endTime time.Time) Stages {
+	// Make sure that the stages slice is initialized
+	if len(stages) == 0 {
+		stages = make([]StageInfo, 0)
+	}
+
+	// If the stage already exists  update the endTime and Duration, and append the new step.
+	for stageKey, stageVal := range stages {
+		if stageVal.StageName == stageName {
+			stages[stageKey].Duration = stages[stageKey].Duration + endTime.Sub(startTime)
```
bparees: stage duration should also just be endtime-starttime.

coreydaley: Updated

coreydaley:
@bparees updated and added a test

pkg/api/helpers.go
```diff
@@ -0,0 +1,42 @@
...
+func RecordStageAndStepInfo(stages Stages, stageName StageName, stepName StepName, startTime time.Time, endTime time.Time) Stages {
+	// Make sure that the stages slice is initialized
+	if len(stages) == 0 {
+		stages = make([]StageInfo, 0)
+	}
+
+	// If the stage already exists  update the endTime and Duration, and append the new step.
+	for stageKey, stageVal := range stages {
+		if stageVal.StageName == stageName {
+			stages[stageKey].Duration = endTime.Sub(startTime)
```
bparees: endTime.Sub(stages[stageKey].startTime)

coreydaley: Fixed

bparees:
[merge]

coreydaley:
Just fixed the travis failure

bparees:
[merge]

PR #695, opened 2017-03-03
Ruby 2.3 image has moved

znmeb:
The Ruby example doesn't work - there's no image at `openshift/ruby-23-centos7`. It's now at `centos/ruby-23-centos7`.


bparees:
thanks

bparees:
[merge]

Issue #701, opened 2017-03-13
Default s2i rebuild image pull policy not work correctly


guangxuli:
**sti rebuild command**
the help description of flag -p:
`-p, --pull-policy string               Specify when to pull the builder image (always, never or if-not-present) (default "if-not-present")`

After running `s2i rebuild hello-world-app:1.2`
outputs : `ERROR: An error occurred: the policy for pull image must be set`

it seems that the default pull policy not works correctly.

PR #702, opened 2017-03-13
fix Default s2i rebuild image pull policy not work correctly

guangxuli:
fix https://github.com/openshift/source-to-image/issues/701


bparees:
lgtm [merge]

PR #711, opened 2017-03-15
only use image user for running containers

bparees:


smarterclayton:
Did not see anything else either.

bparees:
[merge]

PR #712, opened 2017-03-22
Set each default pull policy at begining

guangxuli:


bparees:
[merge]

php-coder:
AFAIR there are 2 ways of invoking S2I -- one from the binary and another one as part of OpenShift. Looks like after this change there is no defaulting logic when S2I is invoking from the OpenShift. I hope that we didn't break something that was working before.

bparees:
> AFAIR there are 2 ways of invoking S2I -- one from the binary and another one as part of OpenShift. Looks like after this change there is no defaulting logic when S2I is invoking from the OpenShift. I hope that we didn't break something that was working before.

@php-coder i checked for this before merging the PR.  the openshift logic always sets the policy:  
https://github.com/openshift/origin/blob/master/pkg/build/builder/sti.go#L214-L223

PR #713, opened 2017-03-22
fix invalid test session

guangxuli:
fix these:
1. make test invalid
2. example pkg/build no test files

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,7 +3,7 @@
 # Targets (see each target for more information):
 #   all: Build code.
 #   build: Build code.
-#   check: Run tests.
+#   check: Run build, verify and tests.
 #   test: Run tests.
 #   clean: Clean up.
 #   release: Build release.
@@ -33,6 +33,7 @@ verify: build
 	hack/verify-golint.sh
 	hack/verify-govet.sh
 	hack/verify-godeps.sh || true # remove this to make godepchecker's warnings actionable
+	hack/verify-bash-completion.sh
 .PHONY: verify
 
 # Install travis dependencies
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -55,11 +55,11 @@ install-travis:
 # Example:
 #   make check
 #   make test
-#   make check WHAT=pkg/build TESTFLAGS=-v
-check:
+#   make check WHAT=pkg/docker TESTFLAGS=-v
+test check:
```
bparees: the "check" target should run the build, verify, and test targets.
the "test" target should run the hack/test-go logic.

guangxuli: oh, i added this according to the comments 
```
#   check: Run tests.
#   test: Run tests.
```

agree, fixed. origin also did like what your specified. thanks.

Makefile
```diff
@@ -55,12 +55,21 @@ install-travis:
 # Example:
 #   make check
 #   make test
-#   make check WHAT=pkg/build TESTFLAGS=-v
-check:
+#   make check WHAT=pkg/docker TESTFLAGS=-v
+check:	verify
```
bparees: why not just `check: verify test` ?

and move verify-bash-completion to the verify target above.

guangxuli: because verify have include build.

bparees: but what is the difference between calling $(MAKE) test and having test as a dependency of the check target?

bparees: i don't think there is one, since you're not passing any arguments, thus there's no reason to do it that way.

guangxuli: yes, I should use test target instead of $(MAKE) test after moving verify-bash-completion script.

guangxuli:
@bparees any more comments? thanks!

bparees:
lgtm
[merge]

PR #716, opened 2017-03-27
add runtime pull policy

guangxuli:
`cfg.RuntimeImagePullPolicy` is used in our code, but the configuration was not add to the command flags. 
/cc @bparees


bparees:
[merge]

Issue #718, opened 2017-03-27
Failed to kill container


DMevada:
Hello,

I am trying to generate a docker container from an image that is available at 
https://github.com/sclorg/s2i-nodejs-container/tree/master/4

Here are the steps that I took to generate the Docker container.

1. make build TARGET=centos7 VERSION=4 (as mentioned in the above repo's instructions)
2. ../s2i build https://github.com/sclorg/s2i-nodejs-container.git --context-dir=4/test/test-app/ centos/nodejs-44-centos7 nodejs-sample-app --loglevel 5

The error that I am facing is that S2I is unable to kill the container. I have not made any local code changes in  the s2i-nodejs-container repo. In other words, I simply cloned the repo, and ran it.

Since the entire logs are 2000 lines, I have uploaded a separate file.

[s2i error.txt](https://github.com/openshift/source-to-image/files/873991/s2i.error.txt)

Most of the logs are caused by npm install, but you will find the error on line 2188.
This test was run on CentOS 7.2, with Docker version: Docker version 17.03.0-ce, build 60ccb22

bparees:
1) since you didn't tag the image you built w/ your make build step, you're actually just building using the nodejs image from docker hub.  which is fine, but just an fyi if you were trying to customize the nodejs image in any way.

2) What makes you think it failed?  the message you're getting is a warning.  When i run your steps i get a return code 0 (success) and a new nodejs-sample-app image exists locally.

DMevada:
1. Yes, I deliberately didn't tag the image because I first wanted it try it without customizing it.

2. Oh, I thought this warning message was a cause of failure because it meant that the docker container that is being built cannot be removed. I also thought that this step must be successful (no warnings or errors), before I can run the container that was generated.

However, you're right, I just tried running the container and it is working as intended. 

Please close this issue, and thanks for your help!

bparees:
it does seem like we've lost our success message somewhere along the way, i'll leave this issue open to make sure it gets added back in.

bparees:
@guangxuli would you like to take a shot at this one?  We just need to ensure that s2i prints a final "build succeeded" or "build failed" message at the end, today apparently it is not doing so, at least for success.

guangxuli:
@bparees no problem :)

helio-frota:
Hi, 
Maybe this is related ?
https://github.com/bucharest-gold/origin-s2i-nodejs/issues/58

This happens after `s2i build`
` warning: Failed to kill container...`

```
[...]

PR #720, opened 2017-03-29
fix kill container failed warning

guangxuli:
fix https://github.com/openshift/source-to-image/issues/718


bparees:
per my comments in the issue, we don't want to change this.

Also this doesn't address the fundamental problem I described in the issue which is that we're not giving the user any summary feedback at the end to tell them whether their build succeeded or failed.

guangxuli:
@bparees  How about this change? Whether the modification address you comment? PTAL.

cmd/s2i/main.go
```diff
@@ -155,7 +155,12 @@ $ s2i build . centos/ruby-22-centos7 hello-world-app
 			builder, _, err := strategies.GetStrategy(cfg)
 			s2ierr.CheckError(err)
 			result, err := builder.Build(cfg)
-			s2ierr.CheckError(err)
+			if err != nil {
+				fmt.Println("Build failed")
+				s2ierr.CheckError(err)
+			} else {
+				fmt.Println("Build success")
```
bparees: should be using glog, not fmt.

success text should be "Build completed successfully"

guangxuli: done

guangxuli:
@bparees  the retrying output :  `retrying in 5ss ...`  so, i remove the char s in this PR.

bparees:
thanks @guangxuli 

lgtm [merge]

PR #723, opened 2017-04-13
rework proxy stripping logic

bparees:


bparees:
[test]

bparees:
@coreydaley ptal

coreydaley:
@bparees I think it would be worth including an example or two in the test that include port numbers, otherwise it lgtm

bparees:
TCs added.
[merge]

PR #728, opened 2017-04-19
Always close pipes in RunContainer(), in order to avoid hang in error cases

jim-minter:
https://bugzilla.redhat.com/show_bug.cgi?id=1442875

@openshift/devex

@bparees ptal - running this against origin tests + extended tests at https://github.com/openshift/origin/pull/13817 then intend to repurpose that PR for the s2i vendoring bump.

### DIFF
--- a/pkg/docker/docker.go
+++ b/pkg/docker/docker.go
@@ -132,6 +132,7 @@ type Docker interface {
 	UploadToContainerWithTarWriter(fs util.FileSystem, srcPath, destPath, container string, makeTarWriter func(io.Writer) s2itar.Writer) error
 	DownloadFromContainer(containerPath string, w io.Writer, container string) error
 	Version() (dockertypes.Version, error)
+	CheckReachable() error
 }
 
 // Client contains all methods used when interacting directly with docker engine-api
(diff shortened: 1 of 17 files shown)


pkg/docker/docker.go
```diff
@@ -900,8 +894,21 @@ func (d *stiDocker) holdHijackedConnection(tty bool, inputStream io.Reader, outp
 }
 
 // RunContainer creates and starts a container using the image specified in opts
-// with the ability to stream input and/or output.
+// with the ability to stream input and/or output.  Any non-nil
+// opts.Std{in,out,err} will be closed upon return.
 func (d *stiDocker) RunContainer(opts RunContainerOptions) error {
+	// Ensure that Std{in,out,err} are closed upon return, including under error
+	// circumstances.
```
bparees: i thought we wanted to leave stderr open in some cases so the caller than read the container's stderr, and only close it if we failed before starting the container?

(as in what sti.go:685 is doing)

jim-minter: Once RunContainer returns, by definition nothing more can be written to std{out,err}, therefore it closes them.  It also closes stdin so that any goroutine still trying to write to the container will error out.

sti.go:685 uses a goroutine + (unbuffered) pipe + *string (should actually be a bytes.Buffer) to keep the stderr data for reading, so we should be good.

bparees: ok

bparees:
[merge]

bparees:
merge failures look likely to have been introduced by this change.

jim-minter:
[test]

bparees:
squash a couple of these commits and lgtm.

jim-minter:
@bparees squashed

bparees:
[merge]

PR #733, opened 2017-04-20
Fix test template to match created directories

soltysh:
@bparees found this while working on my app for PyCon, ptal

### DIFF
--- a/pkg/create/templates/test.go
+++ b/pkg/create/templates/test.go
@@ -24,7 +24,6 @@ fi
 
 test_dir="$($READLINK_EXEC -zf $(dirname "${BASH_SOURCE[0]}"))"
 image_dir=$($READLINK_EXEC -zf ${test_dir}/..)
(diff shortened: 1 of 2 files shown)


soltysh:
I found one another in docker template as well.

pkg/create/templates/test.go
```diff
@@ -24,7 +24,7 @@ fi
 
 test_dir="$($READLINK_EXEC -zf $(dirname "${BASH_SOURCE[0]}"))"
 image_dir=$($READLINK_EXEC -zf ${test_dir}/..)
-scripts_url="file://${image_dir}/.s2i/bin"
+scripts_url="file://${image_dir}/s2i/bin"
```
bparees: looks like this arg isn't even used in this file?

soltysh:
Updated.

bparees:
[merge]

PR #734, opened 2017-04-27
de-boiler-plate credential stripping logic

bparees:


bparees:
[test]

bparees:
@jim-minter ptal

pkg/util/util.go
```diff
@@ -0,0 +1,19 @@
...
+	"github.com/docker/engine-api/types/container"
+)
+
+// SafeForLoggingContainerConfig returns a string version of the container.Config object
+// with sensitive information (proxy environment variables containing credentials)
+// redacted.
+func SafeForLoggingContainerConfig(config *container.Config) string {
+	origEnv := config.Env
+	strippedEnv := StripProxyCredentials(origEnv)
+	config.Env = strippedEnv
```
jim-minter: operating on the original config object and not a shallow copy could in principle expose us to a race in future - suggest copying the config object up front

pkg/util/util.go
```diff
@@ -0,0 +1,19 @@
...
+	s := fmt.Sprintf("%+v", config)
```
jim-minter: FWIW I think I'd envisaged returning a temporary copy and having the caller decide what they'd Sprintf, but this works too :)

jim-minter:
I would be inclined to fix the comment around the potential race, otherwise lgtm.

pkg/util/util.go
bparees: I figured there might be cases where it's easier to embed the logic of "what to not log" into the function itself and saves us the trouble of deep copying the object.

bparees:
> I would be inclined to fix the comment around the potential race, otherwise lgtm.

@jim-minter updated, ptal.

jim-minter:
lgtm

bparees:
[merge]

PR #743, opened 2017-05-09
Pull image before executing S2I script

csrwng:
Checks and pulls an image before executing a script. This should already be happening when the build strategy gets created. However, there may be some time between the time the strategy is created and the time an assemble script is actually run. In that time the image may be garbage collected.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1446925


csrwng:
@openshift/devex ptal

bparees:
[merge]

PR #748, opened 2017-05-12
strip debug symbols on release binary

### LINKED ISSUE #730: Strip debug symbols from release build
Adding the '-s -w' flags to `-ldflags` reduces binary size to about 6M (from about 11M now). The symbols are useful when debugging, but not so much in released versions IMO.

Note that stripping these will still produce useful stacktraces on panics.

bparees:
fixes #730


bparees:
@openshift/devex ptal

coreydaley:
lgtm

bparees:
[merge]

PR #750, opened 2017-05-24
fix anchor for testrun

mmckinst:
I'm not sure if its markdown or an HTML problem but "test/run" with a slash in the anchor isn't working.


bparees:
thanks!
[merge]

PR #753, opened 2017-06-05
Update and/or add docker labels with a JSON file

johnkim76:
Checks if a JSON file with labels exist. If so, update and/or add the labels before the output images is committed.

### DIFF
--- a/docs/new_labels.md
+++ b/docs/new_labels.md
@@ -1,0 +1,27 @@
+# Adding New Labels
+New Docker Labels may be created and/or updated for the output image via the image metadata file.
+
+If a new label is specified in the metadata file, the label will be added in the output image.  However, any label previously defined in the base builder image will be ***overwritten*** in the output image, if the same label name is specified in the image metadata file.
+
+## Image Metadata File Name and Path
+The name and path of the file ***must*** be the following:
+```bash
+/tmp/.s2i/image_metadata.json
+```
+
+## Example 
+The file may have one or more label/value pairs.  Below is the JSON format of the labels, in the image metadata file:
+```bash
+{
+  "labels": [
+    {"labelkey1":"value1"},
+    {"labelkey2":"value2"},
+    .........
+  ]
+}
+
+```
+Note: If the JSON format is different than shown above, it will cause an error.
+
+## Creating the File
+The file should be created during the `assemble` step. 
--- a/pkg/build/strategies/sti/postexecutorstep.go
+++ b/pkg/build/strategies/sti/postexecutorstep.go
@@ -2,6 +2,7 @@ package sti
 
 import (
 	"archive/tar"
+	"encoding/json"
 	"fmt"
 	"io"
 	"io/ioutil"
@@ -18,6 +19,8 @@ import (
 	"github.com/openshift/source-to-image/pkg/util"
 	utilstatus "github.com/openshift/source-to-image/pkg/util/status"
 )
+
+const maximumLabelSize = 10240
 
 type postExecutorStepContext struct {
 	// id of the previous image that we're holding because after committing the image, we'll lose it.
@@ -104,6 +107,8 @@ type commitImageStep struct {
 	image   string
 	builder *STI
 	docker  dockerpkg.Docker
+	fs      util.FileSystem
+	tar     s2itar.Tar
 }
 
 func (step *commitImageStep) execute(ctx *postExecutorStepContext) error {
@@ -116,7 +121,15 @@ func (step *commitImageStep) execute(ctx *postExecutorStepContext) error {
 
 	cmd := createCommandForExecutingRunScript(step.builder.scriptsURL, ctx.destination)
 
+	if err = checkAndGetNewLabels(step.builder, step.docker, step.tar, ctx.containerID); err != nil {
+		return fmt.Errorf("could not check for new labels for %q image: %v", step.image, err)
+	}
+
 	ctx.labels = createLabelsForResultingImage(step.builder, step.docker, step.image)
+
+	if err = checkLabelSize(ctx.labels); err != nil {
+		return fmt.Errorf("label validation failed for %q image: %v", step.image, err)
+	}
 
 	// Set the image entrypoint back to its original value on commit, the running
 	// container has "env" as its entrypoint and we don't want to commit that.
@@ -211,46 +224,10 @@ func (step *downloadFilesFromBuilderImageStep) execute(ctx *postExecutorStepCont
 }
 
 func (step *downloadFilesFromBuilderImageStep) downloadAndExtractFile(artifactPath, artifactsDir, containerID string) error {
-	glog.V(5).Infof("Downloading file %q", artifactPath)
-
-	fd, err := ioutil.TempFile(artifactsDir, "s2i-runtime-artifact")
-	if err != nil {
-		step.builder.result.BuildInfo.FailureReason = utilstatus.NewFailureReason(
-			utilstatus.ReasonFSOperationFailed,
-			utilstatus.ReasonMessageFSOperationFailed,
-		)
-		return fmt.Errorf("could not create temporary file for runtime artifact: %v", err)
+	if res, err := downloadAndExtractFileFromContainer(step.docker, step.tar, artifactPath, artifactsDir, containerID); err != nil {
+		step.builder.result.BuildInfo.FailureReason = res
+		return err
 	}
-	defer func() {
-		fd.Close()
-		os.Remove(fd.Name())
-	}()
-
-	if err := step.docker.DownloadFromContainer(artifactPath, fd, containerID); err != nil {
-		step.builder.result.BuildInfo.FailureReason = utilstatus.NewFailureReason(
-			utilstatus.ReasonGenericS2IBuildFailed,
-			utilstatus.ReasonMessageGenericS2iBuildFailed,
-		)
-		return fmt.Errorf("could not download file (%q -> %q) from container %s: %v", artifactPath, fd.Name(), containerID, err)
-	}
-
-	// after writing to the file descriptor we need to rewind pointer to the beginning of the file before next reading
-	if _, err := fd.Seek(0, os.SEEK_SET); err != nil {
-		step.builder.result.BuildInfo.FailureReason = utilstatus.NewFailureReason(
-			utilstatus.ReasonGenericS2IBuildFailed,
-			utilstatus.ReasonMessageGenericS2iBuildFailed,
-		)
-		return fmt.Errorf("could not seek to the beginning of the file %q: %v", fd.Name(), err)
-	}
-
-	if err := step.tar.ExtractTarStream(artifactsDir, fd); err != nil {
-		step.builder.result.BuildInfo.FailureReason = utilstatus.NewFailureReason(
-			utilstatus.ReasonGenericS2IBuildFailed,
-			utilstatus.ReasonMessageGenericS2iBuildFailed,
-		)
-		return fmt.Errorf("could not extract runtime artifact %q into the directory %q: %v", artifactPath, artifactsDir, err)
-	}
-
 	return nil
 }
 
@@ -450,22 +427,20 @@ func createLabelsForResultingImage(builder *STI, docker dockerpkg.Docker, baseIm
 	}
 
 	configLabels := builder.config.Labels
+	newLabels := builder.newLabels
 
-	return mergeLabels(configLabels, generatedLabels, existingLabels)
+	return mergeLabels(configLabels, generatedLabels, existingLabels, newLabels)
 }
 
-func mergeLabels(configLabels, generatedLabels, existingLabels map[string]string) map[string]string {
-	result := map[string]string{}
-	for k, v := range existingLabels {
-		result[k] = v
+func mergeLabels(labels ...map[string]string) map[string]string {
+	mergedLabels := map[string]string{}
+
+	for _, labelMap := range labels {
+		for k, v := range labelMap {
+			mergedLabels[k] = v
+		}
 	}
-	for k, v := range generatedLabels {
-		result[k] = v
-	}
-	for k, v := range configLabels {
-		result[k] = v
-	}
-	return result
+	return mergedLabels
 }
 
 func createCommandForExecutingRunScript(scriptsURL map[string]string, location string) string {
@@ -481,4 +456,116 @@ func createCommandForExecutingRunScript(scriptsURL map[string]string, location s
 		cmd = path.Join(location, "scripts", api.Run)
 	}
 	return cmd
+}
+
+func downloadAndExtractFileFromContainer(docker dockerpkg.Docker, tar s2itar.Tar, sourcePath, destinationPath, containerID string) (api.FailureReason, error) {
+	glog.V(5).Infof("Downloading file %q", sourcePath)
+
+	fd, err := ioutil.TempFile(destinationPath, "s2i-runtime-artifact")
+	if err != nil {
+		res := utilstatus.NewFailureReason(
+			utilstatus.ReasonFSOperationFailed,
+			utilstatus.ReasonMessageFSOperationFailed,
+		)
+		return res, fmt.Errorf("could not create temporary file for runtime artifact: %v", err)
+	}
+	defer func() {
+		fd.Close()
+		os.Remove(fd.Name())
+	}()
+
+	if err := docker.DownloadFromContainer(sourcePath, fd, containerID); err != nil {
+		res := utilstatus.NewFailureReason(
+			utilstatus.ReasonGenericS2IBuildFailed,
+			utilstatus.ReasonMessageGenericS2iBuildFailed,
+		)
+		return res, fmt.Errorf("could not download file (%q -> %q) from container %s: %v", sourcePath, fd.Name(), containerID, err)
+	}
+
+	// after writing to the file descriptor we need to rewind pointer to the beginning of the file before next reading
+	if _, err := fd.Seek(0, os.SEEK_SET); err != nil {
+		res := utilstatus.NewFailureReason(
+			utilstatus.ReasonGenericS2IBuildFailed,
+			utilstatus.ReasonMessageGenericS2iBuildFailed,
+		)
+		return res, fmt.Errorf("could not seek to the beginning of the file %q: %v", fd.Name(), err)
+	}
+
+	if err := tar.ExtractTarStream(destinationPath, fd); err != nil {
+		res := utilstatus.NewFailureReason(
+			utilstatus.ReasonGenericS2IBuildFailed,
+			utilstatus.ReasonMessageGenericS2iBuildFailed,
+		)
+		return res, fmt.Errorf("could not extract artifact %q into the directory %q: %v", sourcePath, destinationPath, err)
+	}
+
+	return utilstatus.NewFailureReason("", ""), nil
+}
+
+func checkLabelSize(labels map[string]string) error {
+	var sum = 0
+	for k, v := range labels {
+		sum += len(k) + len(v)
+	}
+
+	if sum > maximumLabelSize {
+		return fmt.Errorf("label size '%d' exceeds the maximum limit '%d'", sum, maximumLabelSize)
+	}
+
+	return nil
+}
+
+// check for new labels and apply to the output image.
+func checkAndGetNewLabels(builder *STI, docker dockerpkg.Docker, tar s2itar.Tar, containerID string) error {
+	glog.V(3).Infof("Checking for new Labels to apply... ")
+
+	// metadata filename and its path inside the container
+	metadataFilename := "image_metadata.json"
+	sourceFilepath := filepath.Join("/tmp/.s2i", metadataFilename)
+
+	// create the 'downloadPath' folder if it doesn't exist
+	downloadPath := filepath.Join(builder.config.WorkingDir, "metadata")
+	glog.V(3).Infof("Creating the download path '%s'", downloadPath)
+	if err := os.MkdirAll(downloadPath, 0700); err != nil {
+		glog.Errorf("Error creating dir %q for '%s': %v", downloadPath, metadataFilename, err)
+		return err
+	}
+
+	// download & extract the file from container
+	if _, err := downloadAndExtractFileFromContainer(docker, tar, sourceFilepath, downloadPath, containerID); err != nil {
+		glog.V(3).Infof("unable to download and extract '%s' ... continuing", metadataFilename)
+		return nil
+	}
+
+	// open the file
+	filePath := filepath.Join(downloadPath, metadataFilename)
+	fd, err := os.Open(filePath)
+	if fd == nil || err != nil {
+		return fmt.Errorf("unable to open file '%s' : %v", downloadPath, err)
+	}
+	defer fd.Close()
+
+	// read the file to a string
+	str, err := ioutil.ReadAll(fd)
+	if err != nil {
+		return fmt.Errorf("error reading file '%s' in to a string: %v", filePath, err)
+	}
+	glog.V(3).Infof("new Labels File contents : \n%s\n", str)
+
+	// string into a map
+	var data map[string]interface{}
+
+	if err = json.Unmarshal([]byte(str), &data); err != nil {
+		return fmt.Errorf("JSON Unmarshal Error with '%s' file : %v", metadataFilename, err)
+	}
+
+	// update newLabels[]
+	labels := data["labels"]
+	for _, l := range labels.([]interface{}) {
+		for k, v := range l.(map[string]interface{}) {
+			builder.newLabels[k] = v.(string)
+		}
+	}
+
+	return nil
 }
--- a/pkg/build/strategies/sti/sti.go
+++ b/pkg/build/strategies/sti/sti.go
@@ -65,6 +65,7 @@ type STI struct {
 	incremental            bool
 	sourceInfo             *api.SourceInfo
 	env                    []string
+	newLabels              map[string]string
 
 	// Interfaces
 	preparer  build.Preparer
@@ -124,6 +125,7 @@ func New(client dockerpkg.Client, config *api.Config, fs util.FileSystem, overri
 		externalScripts:        map[string]bool{},
 		installedScripts:       map[string]bool{},
 		scriptsURL:             map[string]string{},
+		newLabels:              map[string]string{},
 	}
 
 	if len(config.RuntimeImage) > 0 {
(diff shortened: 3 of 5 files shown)


pkg/build/strategies/sti/postexecutorstep.go
```diff
@@ -482,3 +455,90 @@ func createCommandForExecutingRunScript(scriptsURL map[string]string, location s
...
+
+	return utilstatus.NewFailureReason("", ""), nil
+}
+
+// check for new labels and apply to the output image.
+func checkAndGetNewLabels(builder *STI, docker dockerpkg.Docker, tar s2itar.Tar, containerID string) error {
+	glog.V(3).Infof("Checking for new Labels to apply... ")
+	newLabelsFileName := "new_labels.json"
+	sourceFilePath := "/opt/docker/" + newLabelsFileName
+	destinationDir := builder.config.WorkingDir + "/upload/"
```
bparees: "upload" doesn't seem like the right destination for this file.  you can create a new dir under the workingdir.

pkg/build/strategies/sti/postexecutorstep.go
```diff
@@ -482,3 +455,90 @@ func createCommandForExecutingRunScript(scriptsURL map[string]string, location s
...
+		return res, fmt.Errorf("could not extract runtime artifact %q into the directory %q: %v", sourcePath, destinationPath, err)
+	}
+
+	return utilstatus.NewFailureReason("", ""), nil
+}
+
+// check for new labels and apply to the output image.
+func checkAndGetNewLabels(builder *STI, docker dockerpkg.Docker, tar s2itar.Tar, containerID string) error {
+	glog.V(3).Infof("Checking for new Labels to apply... ")
+	newLabelsFileName := "new_labels.json"
```
bparees: we're going to want to have the possibility of using this for more than labels.  let's call it "image_metadata.json"

pkg/build/strategies/sti/postexecutorstep.go
```diff
@@ -482,3 +455,90 @@ func createCommandForExecutingRunScript(scriptsURL map[string]string, location s
...
+	sourceFilePath := "/opt/docker/" + newLabelsFileName
+	destinationDir := builder.config.WorkingDir + "/upload/"
+	res, err := downloadAndExtractFileFromContainer(docker, tar, sourceFilePath, destinationDir, containerID)
+	if err != nil {
+		builder.result.BuildInfo.FailureReason = res
+	} else {
+		// check and read the 'newLabelsFileName'.
+		newLablesFilepath := destinationDir + newLabelsFileName
```
bparees: s/Lables/Labels/ but it's moot since this will change once you make the file generic. :)

pkg/build/strategies/sti/postexecutorstep.go
```diff
@@ -482,3 +455,90 @@ func createCommandForExecutingRunScript(scriptsURL map[string]string, location s
...
+		fd, ferr := os.Open(newLablesFilepath)
+		if ferr != nil {
+			glog.V(3).Infof("New LABELs file '%s' could not be opened... continuing...", newLablesFilepath)
```
bparees: this should be a failure.  if they provided a file and we extracted it but then couldn't read it locally, we don't want to commit an image that doesn't have the metadata they supplied.

pkg/build/strategies/sti/postexecutorstep.go
```diff
@@ -482,3 +455,90 @@ func createCommandForExecutingRunScript(scriptsURL map[string]string, location s
...
+			return nil
+		} else if fd != nil {
+			// read contents of the file to a string
+			labelStr, err := ioutil.ReadAll(fd)
+			if err != nil {
+				return fmt.Errorf("error reading file '%s' in to a string: %v", newLablesFilepath, err)
+			}
+			glog.V(3).Infof(" New Labels File contents : \n%s\n", labelStr)
+
+			// unmarshal JSON into a map
```
bparees: again to make this extensible we're going to want a more sophisticated structure.

```
{
  "Labels": [
    {"labelkey1":"value1"},
    {"labelkey2":"value2"},
    .........
  ]
}
```

pkg/build/strategies/sti/postexecutorstep.go
```diff
@@ -482,3 +455,90 @@ func createCommandForExecutingRunScript(scriptsURL map[string]string, location s
...
+			var newLabels map[string]interface{}
+			jerr := json.Unmarshal([]byte(labelStr), &newLabels)
+			if err != nil {
+				glog.V(3).Infof("JSON Unmarshal Error ... Ignoreing '%s' Labels File... Continuing... : %v", newLabelsFileName, jerr)
```
bparees: again should fail the build if we can't parse their metadata.

pkg/build/strategies/sti/postexecutorstep.go
```diff
@@ -482,3 +455,90 @@ func createCommandForExecutingRunScript(scriptsURL map[string]string, location s
...
+
+// check for new labels and apply to the output image.
+func checkAndGetNewLabels(builder *STI, docker dockerpkg.Docker, tar s2itar.Tar, containerID string) error {
+	glog.V(3).Infof("Checking for new Labels to apply... ")
+	newLabelsFileName := "new_labels.json"
+	sourceFilePath := "/opt/docker/" + newLabelsFileName
+	destinationDir := builder.config.WorkingDir + "/upload/"
+	res, err := downloadAndExtractFileFromContainer(docker, tar, sourceFilePath, destinationDir, containerID)
+	if err != nil {
+		builder.result.BuildInfo.FailureReason = res
```
bparees: isn't this going to fail (set a failure reason) for all builds that don't have a metadata file in the container?

pkg/build/strategies/sti/postexecutorstep.go
johnkim76: yes.. will update

pkg/build/strategies/sti/postexecutorstep.go
johnkim76: agreed.  would "metadata" folder work?

bparees: sure

pkg/build/strategies/sti/postexecutorstep.go
johnkim76: I will update

pkg/build/strategies/sti/postexecutorstep.go
johnkim76: I apologize.. will update

pkg/build/strategies/sti/postexecutorstep.go
johnkim76: will update.

pkg/build/strategies/sti/postexecutorstep.go
johnkim76: understood.. will update

pkg/build/strategies/sti/postexecutorstep.go
johnkim76: will update

docs/new_labels.md
```diff
@@ -0,0 +1,25 @@
...
+/opt/docker/new_labels.json
+```
+
+## Example 
+The file may have 1 or more label/value pair.  Below is the JSON format of the new labels file:
+```bash
+{
+    "label_name1": "label1-value",
+    "label_name2": "label2-value",    
+    ...
```
coreydaley: nit
label1_name
label2_name

docs/new_labels.md
```diff
@@ -0,0 +1,25 @@
...
+    "last_label": "last-label-value"
+}
+```
+Note: If the JSON format is different have shown above, the contents of the file will be ignored.
```
coreydaley: different than shown above

pkg/build/strategies/sti/postexecutorstep.go
```diff
@@ -482,3 +455,90 @@ func createCommandForExecutingRunScript(scriptsURL map[string]string, location s
...
+	} else {
+		// check and read the 'newLabelsFileName'.
+		newLablesFilepath := destinationDir + newLabelsFileName
+		fd, ferr := os.Open(newLablesFilepath)
+		if ferr != nil {
```
coreydaley: err instead of ferr

pkg/build/strategies/sti/postexecutorstep.go
```diff
@@ -482,3 +455,90 @@ func createCommandForExecutingRunScript(scriptsURL map[string]string, location s
...
+			if err != nil {
+				glog.V(3).Infof("JSON Unmarshal Error ... Ignoreing '%s' Labels File... Continuing... : %v", newLabelsFileName, jerr)
+			} else {
+				// update newLabels[]
+				for k, v := range newLabels {
+					builder.newLabels[k] = v.(string)
+				}
+			}
+		}
+		defer func() {
```
coreydaley: defer fd.Close()

no reason to wrap it

pkg/build/strategies/sti/postexecutorstep.go
```diff
@@ -482,3 +455,90 @@ func createCommandForExecutingRunScript(scriptsURL map[string]string, location s
...
+			// read contents of the file to a string
+			labelStr, err := ioutil.ReadAll(fd)
+			if err != nil {
+				return fmt.Errorf("error reading file '%s' in to a string: %v", newLablesFilepath, err)
+			}
+			glog.V(3).Infof(" New Labels File contents : \n%s\n", labelStr)
+
+			// unmarshal JSON into a map
+			var newLabels map[string]interface{}
+			jerr := json.Unmarshal([]byte(labelStr), &newLabels)
```
coreydaley: err =

pkg/build/strategies/sti/postexecutorstep.go
```diff
@@ -482,3 +455,90 @@ func createCommandForExecutingRunScript(scriptsURL map[string]string, location s
...
+
+	fd, err := ioutil.TempFile(destinationPath, "s2i-runtime-artifact")
+	if err != nil {
+		res := utilstatus.NewFailureReason(
+			utilstatus.ReasonFSOperationFailed,
+			utilstatus.ReasonMessageFSOperationFailed,
+		)
+		return res, fmt.Errorf("could not create temporary file for runtime artifact: %v", err)
+	}
+	defer func() {
```
coreydaley: Shouldn't the defer be before the return statement?  Probably right after the ioutil.TempFile line

pkg/build/strategies/sti/postexecutorstep.go
```diff
@@ -482,3 +455,90 @@ func createCommandForExecutingRunScript(scriptsURL map[string]string, location s
...
+			utilstatus.ReasonMessageGenericS2iBuildFailed,
+		)
+		return res, fmt.Errorf("could not extract runtime artifact %q into the directory %q: %v", sourcePath, destinationPath, err)
+	}
+
+	return utilstatus.NewFailureReason("", ""), nil
+}
+
+// check for new labels and apply to the output image.
+func checkAndGetNewLabels(builder *STI, docker dockerpkg.Docker, tar s2itar.Tar, containerID string) error {
```
coreydaley: I feel like there is some funkiness going on in this function, but will wait until you re-factor it to review.

docs/new_labels.md
```diff
@@ -0,0 +1,25 @@
...
+If a new label is specified in the file, the label will simply be added in the output image.  However, any label previosly defined in the base builder image will be ***overwritten*** in the output image, if the same label name is specified in the labels file.
+
+## File Name and Path
+The name and path of the labels file ***must*** be the following:
+```bash
+/opt/docker/new_labels.json
+```
+
+## Example 
+The file may have 1 or more label/value pair.  Below is the JSON format of the new labels file:
```
coreydaley: The file may have one or more label/value pairs.

docs/new_labels.md
johnkim76: I will update.

pkg/build/strategies/sti/postexecutorstep.go
johnkim76: thank you, will update

pkg/build/strategies/sti/postexecutorstep.go
johnkim76: understood thank you.

docs/new_labels.md
johnkim76: thank you.  I will need to almost redo this doc since I've only focused on adding labels.

pkg/build/strategies/sti/postexecutorstep.go
johnkim76: thank you, will update.

pkg/build/strategies/sti/postexecutorstep.go
johnkim76: thank you.  will update

pkg/build/strategies/sti/postexecutorstep.go
johnkim76: I will move it. thank you.

docs/new_labels.md
johnkim76: will change it. thank you.

pkg/build/strategies/sti/postexecutorstep.go
johnkim76: @coreydaley when I reviewed the source before my edits, it appears that I did not modify the location of the defer...  how should I proceed?

bparees: The point of the defer is to only set it up if we successfully opened the file.  it's correct where it is.  you can't call fd.Close() if you failed to open/create the file.

johnkim76: ok.. thank you for the clarification. I will leave it alone

pkg/build/strategies/sti/postexecutorstep.go
```diff
@@ -450,11 +422,12 @@ func createLabelsForResultingImage(builder *STI, docker dockerpkg.Docker, baseIm
...
 
 	configLabels := builder.config.Labels
+	newLabels := builder.newLabels
 
-	return mergeLabels(configLabels, generatedLabels, existingLabels)
+	return mergeLabels(configLabels, generatedLabels, existingLabels, newLabels)
 }
 
-func mergeLabels(configLabels, generatedLabels, existingLabels map[string]string) map[string]string {
+func mergeLabels(configLabels, generatedLabels, existingLabels map[string]string, newLabels map[string]string) map[string]string {
```
coreydaley: You should be able to re-factor this to use a variadic function.
https://gobyexample.com/variadic-functions

```
func mergeLabels(labels... map[string]string) map[string]string {
 	mergedLabels := map[string]string{}
 	
	for _, labelMap := range labels {
		for k, v := range labelMap {
			mergedLabels[k] = v
		}
	}
 	return mergedLabels
 }
```

pkg/build/strategies/sti/postexecutorstep.go
```diff
@@ -211,46 +219,10 @@ func (step *downloadFilesFromBuilderImageStep) execute(ctx *postExecutorStepCont
 }
 
 func (step *downloadFilesFromBuilderImageStep) downloadAndExtractFile(artifactPath, artifactsDir, containerID string) error {
```
coreydaley: I don't understand the point of this function.  Did you mean for it to return the error if the err != nil?

docs/new_labels.md
```diff
@@ -0,0 +1,27 @@
...
+If a new label is specified in the metadata file, the label will be added in the output image.  However, any label previosly defined in the base builder image will be ***overwritten*** in the output image, if the same label name is specified in the image metadata file.
+
+## Image Metadata File Name and Path
+The name and path of the file ***must*** be the following:
+```bash
+/opt/docker/image_metadata.json
+```
+
+## Example 
+The file may have one or more label/value pair.  Below is the JSON format of the labels, in the image metadata file:
```
coreydaley: pairs

pkg/build/strategies/sti/postexecutorstep.go
```diff
@@ -116,6 +119,11 @@ func (step *commitImageStep) execute(ctx *postExecutorStepContext) error {
 
 	cmd := createCommandForExecutingRunScript(step.builder.scriptsURL, ctx.destination)
 
+	err = checkAndGetNewLabels(step.builder, step.docker, step.tar, ctx.containerID)
```
coreydaley: if err = checkAndGetNewLabels(step.builder, step.docker, step.tar, ctx.containerID); err != nil {

pkg/build/strategies/sti/postexecutorstep.go
```diff
@@ -482,3 +458,101 @@ func createCommandForExecutingRunScript(scriptsURL map[string]string, location s
...
+		// read the file to a string
+		str, err := ioutil.ReadAll(fd)
+		if err != nil {
+			return fmt.Errorf("error reading file '%s' in to a string: %v", filePath, err)
+		}
+		glog.V(3).Infof("new Labels File contents : \n%s\n", str)
+
+		// string into a map
+		var data map[string]interface{}
+		err = json.Unmarshal([]byte(str), &data)
```
coreydaley: if err = .....; err != nil {

pkg/build/strategies/sti/postexecutorstep.go
```diff
@@ -482,3 +458,101 @@ func createCommandForExecutingRunScript(scriptsURL map[string]string, location s
...
+	// create the 'downloadPath' folder if it doesn't exist
+	downloadPath := filepath.Join(builder.config.WorkingDir, "metadata")
+	glog.V(3).Infof("Creating the download path '%s'", downloadPath)
+	if err := os.MkdirAll(downloadPath, 0700); err != nil {
+		glog.Errorf("Error creating dir %q for '%s': %v", downloadPath, metadataFilename, err)
+		return err
+	}
+
+	// download & extract the file from container
+	_, err := downloadAndExtractFileFromContainer(docker, tar, sourceFilepath, downloadPath, containerID)
```
coreydaley: I think that you could go ahead and wrap this in the if _, err := ......); err != nil and then return nil inside of the if, then get rid of the else { and return nil at the end.

pkg/build/strategies/sti/postexecutorstep.go
johnkim76: thank you, I will update.

pkg/build/strategies/sti/postexecutorstep.go
johnkim76: thank you, this should return err.  I will update.

docs/new_labels.md
johnkim76: I will update.

pkg/build/strategies/sti/postexecutorstep.go
johnkim76: will update.

pkg/build/strategies/sti/postexecutorstep.go
johnkim76: yes, thank you, that looks much cleaner.  will update.

pkg/build/strategies/sti/postexecutorstep.go
johnkim76: will update.

pkg/build/strategies/sti/postexecutorstep.go
```diff
@@ -211,47 +218,11 @@ func (step *downloadFilesFromBuilderImageStep) execute(ctx *postExecutorStepCont
...
-		step.builder.result.BuildInfo.FailureReason = utilstatus.NewFailureReason(
-			utilstatus.ReasonGenericS2IBuildFailed,
-			utilstatus.ReasonMessageGenericS2iBuildFailed,
-		)
-		return fmt.Errorf("could not extract runtime artifact %q into the directory %q: %v", artifactPath, artifactsDir, err)
+		step.builder.result.BuildInfo.FailureReason = res
 	}
-
-	return nil
+	return err
```
coreydaley: This should be inside the if statement, then a return nil at the bottom of the function.

johnkim76: understood. Will update

docs/new_labels.md
```diff
@@ -0,0 +1,27 @@
+# Adding New Labels
+New Docker Labels may be created and/or updated for the output image via the image metadata file.
+
+If a new label is specified in the metadata file, the label will be added in the output image.  However, any label previosly defined in the base builder image will be ***overwritten*** in the output image, if the same label name is specified in the image metadata file.
```
coreydaley: s/previosly/previously/

pkg/build/strategies/sti/postexecutorstep.go
coreydaley:
```
func (step *downloadFilesFromBuilderImageStep) downloadAndExtractFile(artifactPath, artifactsDir, containerID string) error {
	if res, err := downloadAndExtractFileFromContainer(step.docker, step.tar, artifactPath, artifactsDir, containerID); err != nil {
		step.builder.result.BuildInfo.FailureReason = res
		return err
	}
	return nil
}
```

docs/new_labels.md
johnkim76: will update

pkg/build/strategies/sti/postexecutorstep.go
johnkim76: will update.

johnkim76:
I believe I've addressed all of the suggestions thus far.  Please let me know if I've missed anything, thank you.

docs/new_labels.md
```diff
@@ -0,0 +1,27 @@
+# Adding New Labels
+New Docker Labels may be created and/or updated for the output image via the image metadata file.
+
+If a new label is specified in the metadata file, the label will be added in the output image.  However, any label previously defined in the base builder image will be ***overwritten*** in the output image, if the same label name is specified in the image metadata file.
+
+## Image Metadata File Name and Path
+The name and path of the file ***must*** be the following:
+```bash
+/opt/docker/image_metadata.json
```
bparees: let's move this path to /tmp/docker/image_metadata.json.  that's the only path that's pretty much guaranteed to exist in any image and be writable by the assemble script.

pkg/build/strategies/sti/postexecutorstep.go
```diff
@@ -482,3 +451,102 @@ func createCommandForExecutingRunScript(scriptsURL map[string]string, location s
...
+		)
+		return res, fmt.Errorf("could not seek to the beginning of the file %q: %v", fd.Name(), err)
+	}
+
+	if err := tar.ExtractTarStream(destinationPath, fd); err != nil {
+		res := utilstatus.NewFailureReason(
+			utilstatus.ReasonGenericS2IBuildFailed,
+			utilstatus.ReasonMessageGenericS2iBuildFailed,
+		)
+		return res, fmt.Errorf("could not extract runtime artifact %q into the directory %q: %v", sourcePath, destinationPath, err)
```
bparees: s/runtime// since this function is not strictly being used to extract runtime artifacts anymore, it's used for arbitrary content being pulled from a container.

docs/new_labels.md
johnkim76: I will update.

pkg/build/strategies/sti/postexecutorstep.go
johnkim76: I will update.

johnkim76:
@bparees thank you for your suggestions.  I've added a test as you described.

bparees:
@smarterclayton this is about ready to merge from my perspective, so if you have any remaining concerns please let me know.  as it stands, the expected filename is `/tmp/docker/image_metadata.json` (I think you suggested you'd prefer a dash instead of the underscore) and the format is json:

```
{
  "Labels": [
    {"testLabel": "testLabel_value"}
  ]
}
```

bparees:
@johnkim76 please go ahead and squash your commits.  thanks!

bparees:
[test]

docs/new_labels.md
```diff
@@ -0,0 +1,27 @@
+# Adding New Labels
+New Docker Labels may be created and/or updated for the output image via the image metadata file.
+
+If a new label is specified in the metadata file, the label will be added in the output image.  However, any label previously defined in the base builder image will be ***overwritten*** in the output image, if the same label name is specified in the image metadata file.
+
+## Image Metadata File Name and Path
+The name and path of the file ***must*** be the following:
+```bash
+/tmp/docker/image_metadata.json
```
smarterclayton: I'd expect:

```
/tmp/.s2i/image_metadata.json
```

docs/new_labels.md
```diff
@@ -0,0 +1,27 @@
...
+{
+  "Labels": [
+    {"labelkey1":"value1"},
+    {"labelkey2":"value2"},
+    .........
+  ]
+}
+
+```
+Note: If the JSON format is different than shown above, it will cause an error.
```
smarterclayton: Are you doing any maximum size validation of labels?

docs/new_labels.md
johnkim76: ok. I will update

docs/new_labels.md
bparees: it doesn't at present (we don't anywhere else either, for that matter).  

worst case isn't it just going to cause the image commit to fail and then your build fails?

vbatts: Unless you want to add your own max-size, there is not one on upstream LABEL's
https://github.com/moby/moby/blob/v1.12.6/builder/dockerfile/dispatchers.go#L106
it's just a `map[string]string{}`. Granted that means you could blow out the memory on the host loading the data :-)

johnkim76: @smarterclayton do you want any further updates?

smarterclayton: Let's set a reasonable max of 10k for all labels put together here.

smarterclayton:
Please use the same API conventions that we use elsewhere - `Labels` should be `labels`.

docs/new_labels.md
johnkim76: I will update.

johnkim76:
@smarterclayton thank you for the suggestions. I've made the updates.  Please verify and I will squash if all looks ok.

pkg/build/strategies/sti/postexecutorstep.go
```diff
@@ -496,6 +502,15 @@ func downloadAndExtractFileFromContainer(docker dockerpkg.Docker, tar s2itar.Tar
 	return utilstatus.NewFailureReason("", ""), nil
 }
 
+func checkLabelSize(labels map[string]string) error {
+	for k, v := range labels {
+		if len(v) > maximumLabelSize {
```
bparees: i think this is supposed to be checking the total size of all the labels, not the size of each individual label.

johnkim76: Ah, understood, I will update.

bparees:
lgtm, squash and i'll merge.

bparees:
[merge]

PR #759, opened 2017-06-23
Documentation: Installation using go get

### LINKED ISSUE #667: Document installation using "go get"
The current installation instructions discuss downloading and extracting a release tarball.  Many applications deliver themselves using `go get`, which is convenient as it provides a built-in mechanism for updating (`go get -u`) and works across many platforms (since it compiles on the target system).

These instructions work today:

1. Set `GOPATH=/home/user/go` (this happens by default if no GOPATH is set)
1. Add `GOPATH/bin` to `PATH`
1. Run `go get github.com/openshift/source-to-image/cmd/s2i`

`s2i` is then downloaded, built, and installed in `GOPATH/bin`, and can subsequently be updated using `go get -u`.  Other projects can be installed like this; for example, ginkgo is installed using: `go get github.com/onsi/ginkgo/ginkgo` and caddy is installed using: `go get github.com/mholt/caddy/caddy`

It should also work cross-platform, as long as the right version of `go` is available (our build scripts already require it anyways).

coreydaley:
Closes #667

### DIFF
--- a/README.md
+++ b/README.md
@@ -175,6 +175,12 @@ If a `save-artifacts` script exists, a prior image already exists, and the `--in
 
 ## Installation
 
+##### Using `go get`
+
+You can install the s2i binary using `go get` which will download the source-to-image code into your `$GOPATH`, build the s2i binary, and install it into your `$GOPATH/bin`.
+
+```$ go get github.com/openshift/source-to-image/cmd/s2i```
+
 ##### For Mac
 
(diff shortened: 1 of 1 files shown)


coreydaley:
@bparees @jawnsy ptal

jawnsy:
LGTM. Nice!

bparees:
[merge]

PR #760, opened 2017-06-23
Refactoring the S2I command line interface

### LINKED ISSUE #443: Refactor CLI code and usage of os.Exit
I found usages of `os.Exit` in `cmd/s2i/main.go`.
That's a bad idea because it interrupts the runtime without running any deferred calls, no cleanup is done...
Most offenders call `os.Exit` indirectly through [`checkErr`](https://github.com/openshift/source-to-image/blob/cf412b2e88b510c5117dd28965ade70fbcf23b3d/cmd/s2i/main.go#L321).

Also, the project is stuffing all CLI code in a single file, not following the standards we use in Origin.

coreydaley:
Refactoring the S2I cli to fall in line with the openshift client code organization
Closes #443


coreydaley:
[test]

coreydaley:
@openshift/devex ptal
Keep in mind that this was mostly just re-organizing the command line code to fall in line with the openshift client, and to get rid of some of the os.Exit calls in the cli.

pkg/cmd/cmd.go
```diff
@@ -0,0 +1 @@
+package cmd
```
bparees: this must be a hideous golang requirement?  (can't have empty directories that just contain sub-packages?)

coreydaley: correct :(  
Basically gives a "no files to compile error", I believe there is one or two in origin also

pkg/cmd/cli/cli.go
```diff
@@ -0,0 +1,81 @@
...
+	cmdutil "github.com/openshift/source-to-image/pkg/cmd/cli/util"
+	"github.com/openshift/source-to-image/pkg/docker"
+)
+
+// glog is a placeholder until the builders pass an output stream down
+// client facing libraries should not be using glog
+var glog = utilglog.StderrLog
+
+// NewCmdCLI implements the S2I command line functionality.
+func NewCmdCLI(cfg *api.Config) *cobra.Command {
```
bparees: what's the point of this arg?  seems like we only pass nil when we call NewCmdCLI....

cmd/s2i/s2i.go
```diff
@@ -0,0 +1,24 @@
...
+func main() {
+	rand.Seed(time.Now().UTC().UnixNano())
+	if len(os.Getenv("GOMAXPROCS")) == 0 {
+		runtime.GOMAXPROCS(runtime.NumCPU())
+	}
+
+	basename := filepath.Base(os.Args[0])
+	command := cli.CommandFor(basename)
+	if err := command.Execute(); err != nil {
+		os.Exit(1)
```
bparees: shouldn't we dump this error before exiting?

pkg/cmd/cli/cli.go
coreydaley: fixed in next push

cmd/s2i/s2i.go
coreydaley: fixed in next push

pkg/cmd/cli/cli.go
```diff
@@ -0,0 +1,81 @@
...
+	"github.com/openshift/source-to-image/pkg/docker"
+)
+
+// glog is a placeholder until the builders pass an output stream down
+// client facing libraries should not be using glog
+var glog = utilglog.StderrLog
+
+// NewCmdCLI implements the S2I command line functionality.
+func NewCmdCLI() *cobra.Command {
+	rand.Seed(time.Now().UnixNano())
```
jim-minter: can be removed - already done in main()

pkg/cmd/cli/cli.go
```diff
@@ -0,0 +1,81 @@
...
+	return s2iCmd
+}
+
+// CommandFor returns the appropriate command for this base name,
+// or the OpenShift CLI command.
+func CommandFor(basename string) *cobra.Command {
+	// Make case-insensitive and strip executable suffix if present
+	if runtime.GOOS == "windows" {
+		basename = strings.ToLower(basename)
+		basename = strings.TrimSuffix(basename, ".exe")
```
jim-minter: afaics all this code does nothing as basename is not read after write?

pkg/cmd/cli/cli.go
coreydaley: fixed in next push

pkg/cmd/cli/cli.go
coreydaley: fixed in next push

coreydaley:
@bparees @jim-minter updates made, ptal

bparees:
lgtm, will let @jim-minter have the final word.

cmd/s2i/s2i.go
```diff
@@ -0,0 +1,26 @@
...
+	"path/filepath"
+	"runtime"
+	"time"
+
+	"github.com/openshift/source-to-image/pkg/cmd/cli"
+)
+
+func main() {
+	rand.Seed(time.Now().UTC().UnixNano())
+	if len(os.Getenv("GOMAXPROCS")) == 0 {
```
jim-minter: I don't believe this is necessary - is there a specific reason for adding it out of interest?

jim-minter:
nice refactor!  lgtm - question about GOMAXPROCS is curiosity on my part

cmd/s2i/s2i.go
bparees: looks like it was cargo-culted from origin:
https://github.com/openshift/origin/blob/master/cmd/openshift/openshift.go#L30

i think it's a harmless (if probably unnecessary) optimization

bparees:
[merge]

cmd/s2i/s2i.go
rhcarvalho: @jim-minter indeed since Go 1.5 ([release notes](https://golang.org/doc/go1.5)):
> By default, Go programs run with GOMAXPROCS set to the number of cores available; in prior releases it defaulted to 1.

This effectively does nothing, and, as Ben suggests, it must come as inheritance from the early v3 days ;)

coreydaley:
@jim-minter Thanks!

Issue #533, opened 2016-07-08
TestLayeredBuildNoTarNoName is flaking on windows


bparees:
```
I0708 12:22:39.758507   10220 tar.go:308] Adding to tar: /tmp/s2i263248273/upload/src/README.md as src/README.md
I0708 12:22:39.758560   10220 tar.go:308] Adding to tar: /tmp/s2i263248273/upload/src/index.html as src/index.html
I0708 12:22:39.858016   10220 docker.go:722] Removing container "s2i_sti_test_sti_fake_no_tar_83e4f98d" ...
I0708 12:22:39.868368   10220 docker.go:726] Removed container "s2i_sti_test_sti_fake_no_tar_83e4f98d"
I0708 12:22:39.868465   10220 cleanup.go:33] Removing temporary directory /tmp/s2i263248273
I0708 12:22:39.868477   10220 fs.go:158] Removing directory '/tmp/s2i263248273'
I0708 12:22:39.869708   10220 sti.go:578] 
--- FAIL: TestLayeredBuildNoTarNoName (0.71s)
    integration_test.go:316: An error occurred during the build: non-zero (13) exit code from sti_test/sti-fake-no-tar
```

https://ci.openshift.redhat.com/jenkins/job/test_pull_requests_sti/513/consoleFull

no useful message about what failed unfortunately.

gabemontero:
fyi - i see this intermittently with the new k8s/engine-api docker client as well, for whatever that might be worth

bparees:
haven't seen this lately.

bparees:
reopening as we've seen it again, most recently it happened in windows testing:

bparees:
@jim-minter actually now i'm wondering if the new occurrences are specific to the windows setup, looks like maybe we lost connection to the linux docker machine?

```
--- FAIL: TestLayeredBuildNoTarNoName (1.57s)
	integration_test.go:335: An error occurred during the build: write tcp 172.18.11.159:49643->172.18.10.126:2376: wsasend: An established connection was aborted by the software in your host machine.
```

jim-minter:
@bparees agreed.  I'm wondering how on earth to get the information needed to debug whatever is causing this.

bparees:
just failed again here too:
https://ci.openshift.redhat.com/jenkins/job/test_pull_requests_s2i_windows/62/console

jim-minter:
[...]

PR #762, opened 2017-06-26
.s2ifile restore multiple environment variables not working

### LINKED ISSUE #761: Setting multiple environment variables in .s2ifile does not work
Executing the following command work as expected:
s2i build . ubuntu:latest --scripts-url file://.s2i/exec -e "TEST_VAR1=testvar1" -e "TEST_VAR2=testvar2"

If I try to create an .s2ifile with multiple env vars by executing the same command with --use-config added it generates an .s2ifile with "flags":{"env":"TEST_VAR1=testvar1,TEST_VAR2=testvar2" ......}

After this executing "s2i build --use-config" creates a single TEST_VAR1 environment variable with value "testvar1,TEST_VAR2=testvar2" instead of two environment variables.

Also a warning message appears:
WARNING: DEPRECATED: Use multiple -e flags to specify multiple environment variables instead of comma ("TEST_VAR1=testvar1,TEST_VAR2=testvar2")

Environment attached.
Workaround (kindda): using environment file instead

[reproduce.zip](https://github.com/openshift/source-to-image/files/1102327/reproduce.zip)

coreydaley:
Restoring multiple environment variables from an .s2ifile
does not work correctly since removing the ability to supply
a comma separated list of environment variables on the cli.

This pull request restores the environment variables back into
a list before running the command when building the command from
a saved .s2ifile

Fixes #761

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -1,9 +1,11 @@
 package config
 
 import (
-	"io/ioutil"
-
 	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"regexp"
+	"strings"
 
 	"github.com/openshift/source-to-image/pkg/api"
 	utilglog "github.com/openshift/source-to-image/pkg/util/glog"
@@ -12,6 +14,7 @@ import (
 )
 
 var glog = utilglog.StderrLog
+var savedEnvMatcher = regexp.MustCompile("env-[0-9]+")
 
 // DefaultConfigPath specifies the default location of the S2I config file
 const DefaultConfigPath = ".s2ifile"
@@ -24,7 +27,7 @@ type Config struct {
 	Flags        map[string]string `json:"flags,omitempty" yaml:"flags,omitempty"`
 }
 
-// Save persists the S2I command line arguments into disk.
+// Save persists the S2I command line arguments to disk.
 func Save(config *api.Config, cmd *cobra.Command) {
 	c := Config{
 		BuilderImage: config.BuilderImage,
@@ -32,9 +35,14 @@ func Save(config *api.Config, cmd *cobra.Command) {
 		Tag:          config.Tag,
 		Flags:        make(map[string]string),
 	}
-	// Store only flags that have changed
 	cmd.Flags().Visit(func(f *pflag.Flag) {
-		c.Flags[f.Name] = f.Value.String()
+		if f.Name == "env" {
+			for i, env := range config.Environment {
+				c.Flags[fmt.Sprintf("%s-%d", f.Name, i)] = fmt.Sprintf("%s=%s", env.Name, env.Value)
+			}
+		} else {
+			c.Flags[f.Name] = f.Value.String()
+		}
 	})
 	data, err := json.Marshal(c)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


coreydaley:
[test]

coreydaley:
@openshift/devex ptal

bparees:
@coreydaley we shouldn't be writing it out to the file in comma format either, because we lose fidelity if one of their env vars contained a comma.

coreydaley:
@bparees ptal

coreydaley:
@bparees ptal

pkg/config/config.go
```diff
@@ -63,14 +74,31 @@ func Restore(config *api.Config, cmd *cobra.Command) {
...
 	config.BuilderImage = c.BuilderImage
 	config.Source = c.Source
 	config.Tag = c.Tag
+	cmd.Flags().Visit(func(f *pflag.Flag) {
+		if f.Name == "env" {
+			envOverride = true
+		}
+	})
 	for name, value := range c.Flags {
-		// Do not change flags that user sets. Allow overriding of stored flags.
```
jim-minter: suggest preserving the comment

pkg/config/config.go
```diff
@@ -63,14 +74,31 @@ func Restore(config *api.Config, cmd *cobra.Command) {
...
-		if cmd.Flag(name).Changed {
-			continue
+		if name == "env" {
+			for _, v := range strings.Split(value, ",") {
```
jim-minter: cmd.Flag(name).Changed test needed here?

jim-minter:
That moment when you realise your serialisation format is utterly inadequate :cry:
I think this is workable as a quick fix, but an unambiguously new Config struct (version field?) with something like a `Flags []struct { key, value string }` + migration for old versions might be better?

pkg/config/config.go
```diff
@@ -26,6 +29,7 @@ type Config struct {
 
 // Save persists the S2I command line arguments into disk.
 func Save(config *api.Config, cmd *cobra.Command) {
+	glog.V(0).Infof("Saving")
```
bparees: debug?

pkg/config/config.go
```diff
@@ -34,7 +38,13 @@ func Save(config *api.Config, cmd *cobra.Command) {
 	}
 	// Store only flags that have changed
```
bparees: this comment doesn't make sense.  isn't this saving all the flags?  (which seems fine)

pkg/config/config.go
```diff
@@ -49,6 +59,7 @@ func Save(config *api.Config, cmd *cobra.Command) {
 
 // Restore loads the arguments from disk and prefills the Request
 func Restore(config *api.Config, cmd *cobra.Command) {
+	glog.V(0).Infof("Restoring")
```
bparees: debug?

pkg/config/config.go
```diff
@@ -63,14 +74,31 @@ func Restore(config *api.Config, cmd *cobra.Command) {
...
+
+	envOverride := false
 	config.BuilderImage = c.BuilderImage
 	config.Source = c.Source
 	config.Tag = c.Tag
+	cmd.Flags().Visit(func(f *pflag.Flag) {
+		if f.Name == "env" {
+			envOverride = true
+		}
+	})
```
bparees: can't just use a cmd.Flag("env").Changed check for this?  because it's an array flag?

bparees:
> That moment when you realise your pre-existing serialisation format is utterly inadequate 😢

yeah @coreydaley and i were discussing this.  If i had my time machine, this feature would not exist.  I'm somewhat inclined to fix it quickly here, but deprecate it going forward.  We obviously did not think through the version compatibility issues w/ storing/restoring the config.

The new struct is a good idea.

coreydaley:
@bparees @jim-minter updated, ptal

pkg/config/config.go
```diff
@@ -58,19 +66,40 @@ func Restore(config *api.Config, cmd *cobra.Command) {
...
+	if cmd.Flag("env").Changed {
+		envOverride = true
+	}
+
 	for name, value := range c.Flags {
-		// Do not change flags that user sets. Allow overriding of stored flags.
-		if cmd.Flag(name).Changed {
+		if envOverride {
 			continue
 		}
```
bparees: this is in the wrong place, isn't it?

pkg/config/config.go
```diff
@@ -58,19 +66,40 @@ func Restore(config *api.Config, cmd *cobra.Command) {
...
 	config.Source = c.Source
 	config.Tag = c.Tag
+
+	envOverride := false
+	if cmd.Flag("env").Changed {
+		envOverride = true
+	}
+
 	for name, value := range c.Flags {
-		// Do not change flags that user sets. Allow overriding of stored flags.
```
bparees: jim asked to preserve this comment (and I agree)

coreydaley:
@bparees updated, ptal

bparees:
lgtm [merge]

bparees:
flake https://github.com/openshift/source-to-image/issues/533
[merge]

PR #764, opened 2017-06-30
backport of including empty dirs in tar files

bparees:


bparees:
@jim-minter fyi
[merge]

PR #768, opened 2017-07-05
ensure created directories are readable/executable

bparees:
bug 1467819

https://bugzilla.redhat.com/show_bug.cgi?id=1467819

### DIFF
--- a/pkg/test/fs.go
+++ b/pkg/test/fs.go
@@ -106,6 +106,12 @@ func (f *FakeFileSystem) MkdirAll(dirname string) error {
 	return f.MkdirAllError
 }
 
+// MkdirAllWithPermissions creates a new directories on the fake filesystem
+func (f *FakeFileSystem) MkdirAllWithPermissions(dirname string, perm os.FileMode) error {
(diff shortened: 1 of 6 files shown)


bparees:
@csrwng ptal, after recent changes to how we manage directories in tar, our own directories were no longer group/other read/executable meaning we couldn't run the /tmp/script/run script on container start with a random uid.

pkg/test/fs.go
```diff
@@ -106,6 +106,12 @@ func (f *FakeFileSystem) MkdirAll(dirname string) error {
 	return f.MkdirAllError
 }
 
+// MkdirAllWithPermissions creates a new directories on the fake filesystem
```
coreydaley: Update comment with correct function name

bparees:
[test]

csrwng:
LGTM

bparees:
@csrwng added a TC for running overridden assemble/run scripts.

csrwng:
still lgtm

bparees:
tests won't pass until this commit merges, so just manually merging.

Issue #766, opened 2017-07-01
Pass `--quiet` to `git checkout`


yuvipanda:
Currently every time you run s2i with a non-master ref you get:

```
You are in 'detached HEAD' state. You can look around, make experimental
changes and commit them, and you can discard any commits you make in this
state without impacting any branches by performing another checkout.
If you want to create a new branch to retain commits you create, you may
do so (now or later) by using -b with the checkout command again. Example:

  git checkout -b <new-branch-name>

```

If we pass `--quiet` flag to git checkout this won't appear!

Alternatively, we can also use `git reset --hard <ref>` instead.

bparees:
sgtm, but we'll want to decide whether to pass the flag based on the logging verbosity level i think.

PR #774, opened 2017-07-06
make git quiet unless verbose logging is enabled

bparees:
fixes https://github.com/openshift/source-to-image/issues/766

### DIFF
--- a/pkg/scm/git/git.go
+++ b/pkg/scm/git/git.go
@@ -13,6 +13,8 @@ import (
 	"strconv"
 	"strings"
 
+	log "github.com/golang/glog"
+
 	"github.com/openshift/source-to-image/pkg/api"
 	s2ierr "github.com/openshift/source-to-image/pkg/errors"
 	"github.com/openshift/source-to-image/pkg/util"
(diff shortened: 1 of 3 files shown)


bparees:
[test]

bparees:
@openshift/devex ptal

pkg/scm/git/git.go
```diff
@@ -13,6 +13,8 @@ import (
 	"strconv"
 	"strings"
 
+	log "github.com/golang/glog"
```
coreydaley: Why log instead of glog as used elsewhere?

bparees: glog is a wrapper in this file.  if you want to know the actual loglevel that's been set (as I do in the new logic), you need to use the real glog package to check.

bparees: and if you just mean why didn't i import it as "glog" it's because the wrapper variable is named glog.

jim-minter:
lgtm

bparees:
[merge][severity:lowrisk]

PR #775, opened 2017-07-07
Removing Payload field

### LINKED ISSUE #629: Payload field that is sending to the remote server is always empty
[Callback URL documentation](https://github.com/openshift/source-to-image/blob/master/docs/cli.md#callback-url) says that `payload` field contains "a string containing all build messages". During during fixing #522 I found that this field is always empty:
```
POST / HTTP/1.1
Host: 127.0.0.1:8888
User-Agent: Go-http-client/1.1
Content-Length: 216
Content-Type: application/json
Accept-Encoding: gzip

{"labels":{"io.k8s.display-name":"my-test-app","io.openshift.s2i.build.image":"java:8","io.openshift.s2i.build.source-location":"/data/src/github.com/openshift/source-to-image/s2i-test"},"payload":"","success":true}
```

coreydaley:
Closes #629

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -128,13 +128,11 @@ Upon completion (or failure) of a build, `s2i` can execute a HTTP POST to a URL 
 about the build:
 
 * `success` - flag indicating the result of the build process (`true` or `false`)
-* `payload` - list of messages from the build process
 * `labels`  - labels of the resulting image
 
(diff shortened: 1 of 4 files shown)


coreydaley:
[test]

bparees:
[merge]

PR #777, opened 2017-07-13
Adds support for network flag while building image

### LINKED ISSUE #776: Specify which network to use while constructing the image
We're using s2i outside openshift, for some of our local projects.

What happens is that we have a different Network that can access internet (other than the default bridge network), and when we run the s2i command, it creates a Docker Container for this build inside this default network. This doesn't allow our s2i process (that makes some pip/python installs from internet) happen.

How can we specify a docker network other than the default Bridge?

Thanks

rikatz:
This adds support for the --network flag while building a new image.

This does not changes the build behaviour, and is only used if user configures the --network directive. Otherwise, the default known behaviour will be used.

If user specifies an unknown network name, the following error will be returned by Docker API:

``ERROR: An error occurred: Error response from daemon: {"message":"network bla not found"}``

This is necessary due to issue #776 

Thank you very much!

fixes #776


pkg/cmd/cli/cmd/build.go
```diff
@@ -181,6 +181,6 @@ $ s2i build . centos/ruby-22-centos7 hello-world-app
 	buildCmd.Flags().BoolVarP(&(cfg.ForceCopy), "copy", "c", false, "Use local file system copy instead of git cloning the source url")
 	buildCmd.Flags().StringVar(&(cfg.RuntimeImage), "runtime-image", "", "Image that will be used as the base for the runtime image")
 	buildCmd.Flags().VarP(&(cfg.RuntimeArtifacts), "runtime-artifact", "a", "Specify a file or directory to be copied from the builder to the runtime image")
-
+	buildCmd.Flags().StringVarP(&(cfg.Network), "network", "", "", "Specify the default Docker Network name to be used in build process")
```
bparees: unless i'm missing something, i think you can just set cfg.DockerNetworkMode w/ this flag.  The field name isn't ideal, but then you don't need the rest of the code changes, right?

pkg/docker/docker.go
```diff
@@ -246,6 +246,7 @@ func (rco RunContainerOptions) asDockerHostConfig() dockercontainer.HostConfig {
 		hostConfig.Resources.MemorySwap = rco.CGroupLimits.MemorySwap
 		hostConfig.Resources.CgroupParent = rco.CGroupLimits.Parent
 	}
+
```
bparees: delete this newline (less deltas in the commit)

rikatz: OK, did as asked :)

bparees:
[test]

pkg/cmd/cli/cmd/build.go
rikatz: Hum, I don't think so. cfg.DockerNetworkMode points to the default 'bridge' config. There are also some references in there to --net=container:id in the code. I couldn't understand what's that about (maybe something from Openshift build process), but the case here is that the cfg.DockerNetworkMode is used somewhere that cannot be configured :(

So what I did here is set a new --network flag that would be used only if that's explicitly called by the build process, otherwise the default behaviour would be keept.

rikatz: Actually, also I was afraid of breaking something that was working, so what I did was to create a new flag that could be used to exit the regular flow of creating a build container with the default network :)

bparees: when invoked via openshift, we set up a config struct explicitly with DockerNetworkMode set and invoke the s2i code directly (not via cli)

when invoked via the CLI today, s2i does not appear to set DockerNetworkMode to any value (it is an empty string) which I guess is handled as a default by the docker apis when we pass it in.

So the net is when s2i is being invoked via CLI, you should be free to set DockerNetworkMode (or leave it as empty string).

bparees: > Actually, also I was afraid of breaking something that was working, so what I did was to create a new flag that could be used to exit the regular flow of creating a build container with the default network :)

yeah, i can understand why you did that :)  I'm just letting you know i think your fears are unwarranted.  The openshift build codepath doesn't go through the CLI config path.

bparees:
[test]

pkg/cmd/cli/cmd/build.go
rikatz: Thanks :)

I'm facing some issue here, as cfg.DockerNetworkMode implements a DockerNetworkMode constructor (and not a string).

I've seen that other 'specific' structs like 'Injection' that uses a VarP method, but this needs a .Set method. I'm thinking about setting this in 'types.go', is that ok or there's some better way to do this?

Thanks!

bparees: define a scoped variable within NewCmdBuild

have the flag set the scoped variable

have the Run function copy the scoped variable value into the cfg.DockerNetworkMode field.

rikatz:
@bparees Did as asked :) Thank you very much for your patience also ;)

bparees:
looks good, does it work for your environment?

bparees:
[test]

rikatz:
Yes, it does.

@jcmoraisjr take a look and see if this solves our problem :)

bparees:
[merge]

PR #784, opened 2017-08-07
Update godeps to match origin

jim-minter:


jim-minter:
@bparees also I've fixed up the s2i_verify_godeps job so hopefully when this merges we'll be back to green.

bparees:
[merge]

PR #791, opened 2017-08-10
git describe tags in rebase-to-origin.sh

jim-minter:
didn't realise this script existed, but since it does...


bparees:
@jim-minter can you give me an example of what the new content will look like?

jim-minter:
```
                {
                        "ImportPath": "github.com/openshift/source-to-image/pkg/
                        "Comment": "v1.1.7-20-g7da3d3e",
                        "Rev": "7da3d3e97565e652a59fd81286f3956fd43e85a8"
                },
```

It's a nit, but the Comment is supposedly of the form "\<latest relevant tag>-<# commits since that tag>-g\<as many letters from Rev as are required to make the SHA unique in the current repo>".
Previously the Rev was being updated but not the Comment.

bparees:
lgtm
[merge]

PR #794, opened 2017-08-15
remove ec2 instance suffixes

### LINKED ISSUE #793: Windows s2i instance naming breaks pruning script
https://github.com/openshift/source-to-image/blob/112f0d4fd5163e57b1d9414dabc81f905f38a212/hack/windows/Vagrantfile#L19

This line yields instances named like "`origin_test_pull_requests_s2i_windows_258-windows`". The suffix "`windows`" is redundant, and it breaks the verifier instance matching regexes in our pruning scripts - specifically this one: https://github.com/openshift/li/blob/9618207bcf5014071354ce591c4e90b04056b93a/build/dev_tools_constants.rb#L55 - which expect the build number to be at the very end of the instance name.

bparees:
fixes #793

### DIFF
--- a/hack/windows/Vagrantfile
+++ b/hack/windows/Vagrantfile
@@ -8,7 +8,7 @@ Vagrant.configure("2") do |config|
 
   config.vm.define "rhel" do |rhel|
     rhel.vm.provider "aws" do |aws|
-      aws.tags = {"Name" => "#{ENV['USER']}-rhel"}
+      aws.tags = {"Name" => "#{ENV['USER']}"}
       aws.ami = "rhel7:inst"
     end
     rhel.vm.provision "configure-docker-server-linux"
@@ -16,7 +16,7 @@ Vagrant.configure("2") do |config|
 
   config.vm.define "windows" do |windows|
     windows.vm.provider "aws" do |aws|
-      aws.tags = {"Name" => "#{ENV['USER']}-windows"}
+      aws.tags = {"Name" => "#{ENV['USER']}"}
       aws.ami = "windows2012r2:deps"
     end
     windows.vm.provision "configure-docker-client-windows"


bparees:
@dobbymoodge @jim-minter ptal
[test]

jim-minter:
Won't this bring up two different instances with the same name?  Doesn't sound like a good idea to me.  `"Name" => "rhel_#{ENV['USER']}"` is hacky but would it do instead?

dobbymoodge:
huh, looks like the info isn't exactly redundant - this creates two VMs, blahblah-buildnum-rhel and blahblah-buildnum-windows. It looks like the code will need to be a little more complex; it might be best to put the additional info between the image name and the buildnum, which means splitting and reconstructing that string...

dobbymoodge:
@jim-minter That also breaks the prune script, which is looking for something like `/origin_.*(\d+)/`

dobbymoodge:
`s/^(.+)_(\d+)$/\1_rhel_\2/`

jim-minter:
When running locally, it's improbable that the name will be of the form blahblah-buildnum - suggest that if it ends _\d+, then insert the text before the final _\d+, otherwise append as is currently.  Plus comment/link to this issue.

dobbymoodge:
Maybe change the job definition to build `$USER` without the buildnum, and then do something like:
```
"Name" => "#{ENV['USER']}_rhel#{ENV['BUILDNUM'] ? "_#{ENV['BUILDNUM']}" : ""}"
```
or maybe construct that string beforehand so it doesn't look like code vomit.

dobbymoodge:
```
rhel_user = "#{ENV['USER']}_rhel"
rhel_user += "_#{ENV['BUILDNUM']}" if ENV['BUILDNUM']
...
```

bparees:
I was going off @dobbymoodge's issue report which indicated that the ec2 instance name already includes the job name anyway:

> This line yields instances named like "origin_test_pull_requests_s2i_windows_258-windows". 

(and presumably anyone running locally would run rhel OR windows as separate invocations, again passing a unique name to each)

dobbymoodge:
@bparees looks like `vagrant up` with this vagrantfile will always bring up both kinds of VMs.

bparees:
yeah i see that now.  Not really convinced it's a problem, though.  You get two instances w/ the same name, but that doesn't bother ec2 and once is clearly a rhel machine and the other is clearly a windows machine. [/shrug]

bparees:
i guess it might cause the job some confusion when it goes to terminate things.

PR #795, opened 2017-08-15
Parse $USER and set VM name accordingly

### LINKED ISSUE #793: Windows s2i instance naming breaks pruning script

dobbymoodge:
Fixes #793

### DIFF
--- a/hack/windows/Vagrantfile
+++ b/hack/windows/Vagrantfile
@@ -1,4 +1,10 @@
 Vagrant.configure("2") do |config|
+  rhel_vm_name = "#{ENV['USER']}-rhel"
+  windows_vm_name = "#{ENV['USER']}-windows"
+  if ENV['USER'] =~ /^(.*)[_-](\d+)$/
+    rhel_vm_name = "#{$1}-rhel-#{$2}"
+    windows_vm_name = "#{$1}-windows-#{$2}"
+  end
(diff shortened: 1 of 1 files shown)


dobbymoodge:
[test]

dobbymoodge:
HT @bparees, @jim-minter PTAL

dobbymoodge:
Looking better: 
image

dobbymoodge:
Also the regex handles the new names correctly:
```
[46] pry(main)> 'origin_test_pull_requests_s2i_windows-rhel-270' =~ /^(origin).*[_-](\d+)$/
=> 0
[47] pry(main)> puts $1, $2
origin
270
=> nil
[48] pry(main)> 
```

bparees:
lgtm, thanks @dobbymoodge 
[merge]

PR #796, opened 2017-08-16
Add support for assemble-user label

rikatz:
This PR adds support for the assemble-user label in Dockerfile.

Using this label enables the user to determine a different user to assemble the image than the runtime user.

This is needed in images such as standalone PHP, that the artefacts (PHP pages) should not be writable or owned  by the same user running the daemon.

This directive is already supported by s2i cli, and this only adds the support to define this through a label called `io.openshift.s2i.assemble-user`.

The order here is: first the defined assemble-user in cli. If this is not defined, then searches for the label in the image. 

If nothing is defined, the common workflow (with no AssembleUser) will follow.

Thanks!!

### DIFF
--- a/docs/builder_image.md
+++ b/docs/builder_image.md
@@ -76,6 +76,13 @@ and placing them into the appropriate directories inside the image. The workflow
 1. Build any application artifacts.
 1. Install the artifacts into locations appropriate for running.
 
+In the case you need to assemble the Image using a different user than the runtime user defined 
+in ``USER`` directive of Dockerfile, you can achive this by the following ways:
+
+1. use the `--assemble-user` in cmd line
+1. use the label `io.openshift.s2i.assemble-user`
+
+
 #### Example `assemble` script:
 
 **NOTE**: All the examples are written in [Bash](http://www.gnu.org/software/bash/)
--- a/pkg/build/strategies/strategies.go
+++ b/pkg/build/strategies/strategies.go
@@ -23,10 +23,20 @@ func GetStrategy(client docker.Client, config *api.Config) (build.Builder, api.B
 func Strategy(client docker.Client, config *api.Config, overrides build.Overrides) (build.Builder, api.BuildInfo, error) {
 	var builder build.Builder
 	var buildInfo api.BuildInfo
+	var err error
 
 	fs := fs.NewFileSystem()
 
 	startTime := time.Now()
+
+	if config.AssembleUser, err = docker.GetAssembleUser(client, config); err != nil {
+		buildInfo.FailureReason = utilstatus.NewFailureReason(
+			utilstatus.ReasonPullBuilderImageFailed,
+			utilstatus.ReasonMessagePullBuilderImageFailed,
+		)
+		return nil, buildInfo, err
+	}
+
 	image, err := docker.GetBuilderImage(client, config)
 	buildInfo.Stages = api.RecordStageAndStepInfo(buildInfo.Stages, api.StagePullImages, api.StepPullBuilderImage, startTime, time.Now())
 	if err != nil {
--- a/pkg/docker/util.go
+++ b/pkg/docker/util.go
@@ -363,3 +363,19 @@ func GetDefaultDockerConfig() *api.DockerConfig {
 
 	return cfg
 }
+
+// GetAssembleUser finds an assemble user on the given image.
+// This functions receives the config to check if the AssembleUser was defined in command line
+// If the cmd is blank, it tries to fetch the value from the Builder Image defined Label (assemble-user)
+// Otherwise it follows the common flow, using the USER defined in Dockerfile
+func GetAssembleUser(client Client, config *api.Config) (string, error) {
+	if len(config.AssembleUser) > 0 {
+		return config.AssembleUser, nil
+	}
+	d := New(client, config.PullAuthentication)
+	imageData, err := d.GetLabels(config.BuilderImage)
+	if err != nil {
+		return "", err
+	}
+	return imageData[AssembleUserLabel], nil
+}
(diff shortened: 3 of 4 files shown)


coreydaley:
[test]

pkg/build/strategies/sti/sti.go
```diff
@@ -215,13 +215,18 @@ func (builder *STI) Build(config *api.Config) (*api.Result, error) {
...
+	}
+
+	if len(assembleUser) > 0 {
+		glog.V(1).Infof("Running %q in %q as %q user", api.Assemble, config.Tag, assembleUser)
 	} else {
 		glog.V(1).Infof("Running %q in %q", api.Assemble, config.Tag)
 	}
 	startTime := time.Now()
-	if err := builder.scripts.Execute(api.Assemble, config.AssembleUser, config); err != nil {
+	if err := builder.scripts.Execute(api.Assemble, assembleUser, config); err != nil {
```
bparees: I don't think you can change the assembleUser this late in the flow.

strategies.go:Strategy() does a "GetBuilderImage()" call and part of GetBuilderImage(via pullAndCheck() confirms that the config.AssembleUser is part of the AllowedUIDs.  So config.AssembleUser must be set to the final assemble user value before pullAndCheck is invoked, to ensure the assemble user is a legal value.

rikatz: @bparees Thanks!

I'm moving this to strategies.go, but while looking to this check I couldn't find where, in pullAndCheck function and also in CheckAllowedUser function the config.AssembleUser is verified.

Am I missing something here?

Thanks!!!

bparees: https://github.com/openshift/source-to-image/blob/master/pkg/docker/util.go#L304

which calls https://github.com/openshift/source-to-image/blob/master/pkg/docker/util.go#L231-L240

rikatz:
@bparees Thanks for your help.

Did as asked. creating a new function in utils.go and using it in Strategy before building the image.

Hope I don't mess with anything.

Thanks!

pkg/build/strategies/strategies.go
```diff
@@ -27,6 +27,19 @@ func Strategy(client docker.Client, config *api.Config, overrides build.Override
 	fs := fs.NewFileSystem()
 
 	startTime := time.Now()
+
+	assembleUser, err := docker.GetAssembleUser(config.AssembleUser, config.BuilderImage, client, config)
```
coreydaley: Would this work here?
config.AssembleUser, err := docker.GetAssembleUser(config.AssembleUser, config.BuilderImage, client, config)

pkg/build/strategies/strategies.go
```diff
@@ -27,6 +27,19 @@ func Strategy(client docker.Client, config *api.Config, overrides build.Override
...
+
+	if err != nil {
+		buildInfo.FailureReason = utilstatus.NewFailureReason(
+			utilstatus.ReasonPullBuilderImageFailed,
+			utilstatus.ReasonMessagePullBuilderImageFailed,
+		)
+		return nil, buildInfo, err
+	}
+
+	config.AssembleUser = assembleUser
```
coreydaley: Then you could remove this line

pkg/docker/util.go
```diff
@@ -363,3 +363,16 @@ func GetDefaultDockerConfig() *api.DockerConfig {
 
 	return cfg
 }
+
+// GetAssembleUser finds an assemble user on the given image.
```
coreydaley: Add more doc explaining what is happening here, if cfgAssemble user is not blank ...

pkg/docker/util.go
```diff
@@ -363,3 +363,16 @@ func GetDefaultDockerConfig() *api.DockerConfig {
...
+func GetAssembleUser(cfgAssembleUser string, image string, client Client, config *api.Config) (string, error) {
```
coreydaley: Update this function to use the config object to pull all of the information instead of passing two strings (cfgAssembleUser, image) that are contained in config separately.

docs/builder_image.md
```diff
@@ -64,6 +64,11 @@ using `--scripts-url` flag or the `io.openshift.s2i.scripts-url` with the format
 `image:///path/in/image`), then the `--destination` flag or the `io.openshift.s2i.destination`
 label applies only to sources and artifacts.
 
+**NOTE2**: In the case you need to assemble the Image using a different user than the
```
coreydaley: This should not be a note, it should be formatted similar to lines above, and maybe located in the assemble section?

pkg/build/strategies/strategies.go
rikatz: Yeah, my bad. Trying to be a better developer :)

bparees:
lgtm, please squash your commits and i will merge.

pkg/docker/util.go
```diff
@@ -363,3 +363,19 @@ func GetDefaultDockerConfig() *api.DockerConfig {
...
+// If the cmd is blank, it tries to fetch the value from the Builder Image defined Label (assemble-user)
+// Otherwise it follows the common flow, using the USER defined in Dockerfile
+func GetAssembleUser(client Client, config *api.Config) (string, error) {
+	if config.AssembleUser == "" {
+		d := New(client, config.PullAuthentication)
+		imageData, err := d.GetLabels(config.BuilderImage)
+		if err != nil {
+			return "", err
+		}
+		config.AssembleUser = imageData[AssembleUserLabel]
```
bparees: @coreydaley noted that you're mutating the config object here.

it would be better to just return imageData[AssembleUserLabel].

coreydaley: I was thinking maybe a refactor to something along these lines
https://gist.github.com/coreydaley/ca5620d081669d46c9cb37a8eca2fc44

coreydaley:
if tests pass it lgtm

bparees:
@rikatz commits still need squashing please.

rikatz:
@bparees @coreydaley squashed as asked, waiting for the tests to pass :)

bparees:
[merge]

rikatz:
@bparees @coreydaley thanks for your help and your patience :)

coreydaley:
@rikatz No problem.  Thank you for contributing !

PR #812, opened 2017-10-05
allow user labels to override existing labels and generated labels

bparees:


bparees:
@openshift/devex ptal.

@GrahamDumpleton fyi

[test]

bparees:
[merge]

PR #813, opened 2017-10-05
Make it possible to configure example nginx server

bkabrda:
Since me and my team have been using the s2i functionality to configure services in the containers, I think it'd be great to have an example of this in the main repo, so that others could benefit from seeing this approach. Please let me know what you think!


bparees:
thanks @bkabrda 
[merge]

PR #820, opened 2017-10-25
implement functionality to copy symlinks and not their contents

jim-minter:
associated with bugzilla 1506173

### DIFF
--- a/pkg/test/fs/fs.go
+++ b/pkg/test/fs/fs.go
@@ -2,7 +2,6 @@ package fs
 
 import (
 	"bytes"
-	"errors"
 	"io"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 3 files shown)


jim-minter:
[test]

bparees:
So fundamentally this means that if someone previously was building a dir that had a symlink that pointed to a location outside their build dir, it would have worked (we'd have included the actual content).

Now that build will fail because we'll include the symlink which will now be broken since it points to a location outside the build dir.

Correct?

bparees:
(If so, I wonder if the less breaking change would be to retain the old behavior, unless we can't follow the symlink.  If we can't follow the symlink (destination does not exist), then and only then create a broken symlink.

(I have strongly mixed feelings, I wish we never copied the symlinked content, but since we've established that behavior, i'm wary of changing it)

bparees:
[merge]

PR #827, opened 2017-11-10
Minor fixes to create template

asottile:
- Remove spurious newlines at the beginning of files
- README.md and index.html are no longer executable.


bparees:
[merge]

PR #828, opened 2017-11-10
Move from deprecated MAINTAINER to LABEL maintainer=...

asottile:
See https://docs.docker.com/engine/reference/builder/#maintainer-deprecated


bparees:
[merge]

PR #834, opened 2017-11-21
Drop engine api

### LINKED ISSUE #833: Drop github.com/docker/engine-api packages
This is a continuation of https://github.com/openshift/source-to-image/pull/818 and a post-[1.8 k8s rebase](https://github.com/openshift/origin/pull/17115/) task. 
The problem is that `github.com/docker/docker` already contains the types from `github.com/docker/engine-api`, the latter is just an old version of the former. This in the end should clean our vendor dir slightly and eventually lead to a cleaner vendor inside origin, as well.

soltysh:
Fixes #833.

@openshift/sig-master fyi, this should in the end help to limit the pkg/builder vendored dirs as well


pkg/docker/docker.go
```diff
@@ -1053,10 +1053,12 @@ func (d *stiDocker) RunContainer(opts RunContainerOptions) error {
 		// Return an error if the exit code of the container is
 		// non-zero.
 		glog.V(4).Infof("Waiting for container %q to stop ...", container.ID)
-		exitCode, err := d.client.ContainerWait(context.Background(), container.ID)
-		if err != nil {
-			return fmt.Errorf("waiting for container %q to stop: %v", container.ID, err)
-		}
+		// FIXME
+		exitCode := 0
+		// exitCode, err := d.client.ContainerWait(context.Background(), container.ID)
```
soltysh: This is the only bit that is left for me to finish it. It looks like they've changed it significantly and worst thing is that the current conditions (`WaitConditionNotRunning`, `WaitConditionNextExit`, `WaitConditionRemoved`) don't fit us. I was hoping the `WiatConditionNextExit` will, but if the container exits in between this will block until next exit, which is not quite right. I'll dig a bit more into it, and will see what's possible and what not.

soltysh: @bparees ^

bparees:
lgtm pending resolution of the FIXME bits.

soltysh:
@bparees I've fixed the appropriate bits, but I've noticed we're not parsing the docker image spec properly, it was silently falling on `reference.ParseNamed` which requires canonical name, iow. `registry/namespace/image`, I changed that to `reference.WithName` similarly to what we did in origin during rebase. ptal

soltysh:
[test]

soltysh:
/test
[test]

soltysh:
@bparees all green, ptal

jim-minter:
with the caveat that I don't understand the ParseNormalizedNamed change, all this looks good to me.  On this branch, `hack/verify-godeps.sh` indicates:
```
warning: differing godep github.com/opencontainers/runc/libcontainer/user: origin "4d6e6720a7c885c37b4cb083c0d372dda3425120" vs s2i "d223e2adae83f62d58448a799a5da05730228089"
warning: differing godep github.com/docker/go-units: origin "9e638d38cf6977a37a8ea0078f3ee75a7cdb2dd1" vs s2i "e30f1e79f3cd72542f2026ceec18d3bd67ab859c"
warning: differing godep github.com/Microsoft/go-winio: origin "78439966b38d69bf38227fbf57ac8a6fee70f69a" vs s2i "f533f7a102197536779ea3a8cb881d639e21ec5a"
```
I can repurpose #836 to deal with that if appropriate?

soltysh:
> with the caveat that I don't understand the ParseNormalizedNamed change, all this looks good to me. 

Docker changed their API, significantly. This works 😉 

> On this branch, hack/verify-godeps.sh indicates:

Uhm... I thought I've picked the exact same level of deps, I'll fix that right away.

soltysh:
All fixed and green, with 2 lgtm, I'm merging.

[merge]

PR #837, opened 2017-12-07
update incorrect artifacts path in builder_image doc

wgordon17:


docs/builder_image.md
```diff
@@ -92,8 +92,8 @@ and it is assumed that the tar contents unpack into the `/tmp/s2i` directory.
 #!/bin/bash
 
 # restore build artifacts
-if [ "$(ls /tmp/s2i/artifacts/ 2>/dev/null)" ]; then
-    mv /tmp/s2i/artifacts/* $HOME/.
+if [ "$(ls /tmp/artifacts/ 2>/dev/null)" ]; then
+    mv /tmp/artifacts/* $HOME/.
```
bparees: need to fix line 89 too, if you're going to change this.  It's just an example (s2i images are free to set their destination to something other than /tmp).

wgordon17: Absolutely! Thanks. I understand it's just an example, but if the standard s2i images are using `/tmp/artifacts` it helps to use that in the example, because I spent an inordinate amount of time trying to figure out why my assemble script wasn't restoring my artifacts.

bparees:
[merge]

PR #841, opened 2018-01-17
Remove the out-of-dated docker engine-api related hack

louyihua:
Since the code does not rely on docker's engine-api now, this hack should also be removed.


bparees:
lgtm
[merge]

bparees:
[merge]

Issue #847, opened 2018-01-30
 panic: runtime error: invalid memory address or nil pointer dereference


helio-frota:
**Fedora 27**
**s2i v1.1.8**

How to reproduce:

1. clone https://github.com/bucharest-gold/centos7-s2i-nodejs
2. pip install docker-squash --user
3. make all

```console
BUILDER=bucharestgold/centos7-s2i-nodejs:latest NODE_VERSION=9.4.0 ./test/run.sh
Running s2i build --pull-policy never  file://./test/test-app bucharestgold/centos7-s2i-nodejs:latest bucharestgold/centos7-s2i-nodejs-testapp
ERROR: couldn't parse "file://./test/test-app": url "file://./test/test-app" has non-empty host
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x78 pc=0x4edf87]

goroutine 1 [running]:
panic(0x7ce260, 0xc420012140)
	/usr/local/go/src/runtime/panic.go:500 +0x1a1
github.com/openshift/source-to-image/pkg/cmd/cli/cmd.NewCmdBuild.func1(0xc4200a1440, 0xc4201f8d70, 0x3, 0x5)
	/go/src/github.com/openshift/source-to-image/_output/local/go/src/github.com/openshift/source-to-image/pkg/cmd/cli/cmd/build.go:64 +0x19a7
github.com/openshift/source-to-image/vendor/github.com/spf13/cobra.(*Command).execute(0xc4200a1440, 0xc4201f8d20, 0x5, 0x5, 0xc4200a1440, 0xc4201f8d20)
	/go/src/github.com/openshift/source-to-image/_output/local/go/src/github.com/openshift/source-to-image/vendor/github.com/spf13/cobra/command.go:603 +0x439
github.com/openshift/source-to-image/vendor/github.com/spf13/cobra.(*Command).ExecuteC(0xc4200a0fc0, 0xc4201a7e70, 0x48e9d2, 0xc4200a0fc0)
	/go/src/github.com/openshift/source-to-image/_output/local/go/src/github.com/openshift/source-to-image/vendor/github.com/spf13/cobra/command.go:689 +0x367
github.com/openshift/source-to-image/vendor/github.com/spf13/cobra.(*Command).Execute(0xc4200a0fc0, 0x3, 0xc4200a0fc0)
[...]

PR #848, opened 2018-01-30
error out on unparseable urls

bparees:
fixes https://github.com/openshift/source-to-image/issues/847


bparees:
@jim-minter ptal and tell me if you're surprised that `file://./path` is no longer properly parseable in golang (the "." gets treated as a hostname now).

jim-minter:
lgtm: relative paths cannot be expressed in file:// URLs.

bparees:
[merge]

helio-frota:
good to know.
I'll test it [maybe tomorrow]
thanks!

PR #852, opened 2018-02-28
remove golang version check

### LINKED ISSUE #851: Go 1.10 failure due to version comparison bug
https://github.com/openshift/source-to-image/blob/master/hack/common.sh#L204

1.10 is not actually less than 1.4 but the install fails because 1<4

```
==> hack/build-go.sh

Detected go version: go version go1.10 darwin/amd64.
S2I requires go version 1.4 or greater.
Please install Go version 1.4 or later.
```

workaround: string replace 1.4 -> 1.0

bparees:
fixes #851

just straight up removing it, i doubt anyone is running go 1.3 or earlier at this point anyway.

### DIFF
--- a/hack/common.sh
+++ b/hack/common.sh
@@ -195,24 +195,6 @@ EOF
     exit 2
   fi
 
-  # Travis continuous build uses a head go release that doesn't report
-  # a version number, so we skip this check on Travis.  It's unnecessary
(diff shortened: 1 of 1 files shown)


bparees:
[merge]

ilovezfs:
Thank you @bparees!

PR #855, opened 2018-03-07
revert to using docker-engine

bparees:


bparees:
@derekwaynecarr @mrunalp so something in this reversion appears to resolve the issue.  from what @jim-minter and I can see, the meaningful change is the switch from using the docker-engine api to docker, not actual code changes in the s2i code.

bparees:
the issue in the docker client was tracked down and resolved.  don't need this.

PR #866, opened 2018-04-25
Keep Option for Injected Files

adambkaplan:
Adding a `Keep` option to `VolumeSpec` so that files can be kept in s2i builds.

This will be used to support the injection of `ConfigMap` build sources, which should not be truncated.

Trello card: https://trello.com/c/RMKJxJUm/1020-5-allow-using-a-configmap-as-an-input-to-a-build-builds

### DIFF
--- a/test/integration/integration_test.go
+++ b/test/integration/integration_test.go
@@ -403,6 +403,12 @@ func (i *integrationTest) exerciseInjectionBuild(tag, imageName string, injectio
 			t.Errorf("injectionList.Set() failed with error %s\n", err)
 		}
 	}
+	// For test purposes, keep at least one injected source
+	var keptVolume *api.VolumeSpec
+	if len(injectionList) > 0 {
+		injectionList[0].Keep = true
+		keptVolume = &injectionList[0]
+	}
 	config := &api.Config{
 		DockerConfig:      docker.GetDefaultDockerConfig(),
 		BuilderImage:      imageName,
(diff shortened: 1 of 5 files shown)


test/integration/integration_test.go
```diff
@@ -402,6 +402,10 @@ func (i *integrationTest) exerciseInjectionBuild(tag, imageName string, injectio
 			t.Errorf("injectionList.Set() failed with error %s\n", err)
 		}
 	}
+	// Don't truncate the first injected item.
```
bparees: why?

adambkaplan: I decided to keep the behavior of `Set` in place, because I couldn't add a third option to the injection string without breaking Windows directory paths.

Setting `Truncated = false` ensures that we always exercise the "keep" option during the integration test.

test/integration/integration_test.go
```diff
@@ -432,15 +438,35 @@ func (i *integrationTest) exerciseInjectionBuild(tag, imageName string, injectio
...
+				t.Errorf("The file %q must not be empty", f)
+			}
+		}
+	}
+}
+
+func (i *integrationTest) testFile(tag, path string) error {
+	exitCode := i.runInImage(tag, "test -s "+path)
+	if exitCode != 0 {
+		return fmt.Errorf("file %s does not exist in the container %s", path, tag)
```
bparees: not strictly true, since `test -s` also returns "1" if the file exists w/ size zero.  should probably fix this message so it doesn't confuse someone.

adambkaplan:
@bparees updated test messages.

bparees:
[merge]

adambkaplan:
[(5) Allow using a ConfigMap as an input to a build [builds]](https://trello.com/c/RMKJxJUm/1020-5-allow-using-a-configmap-as-an-input-to-a-build-builds)

PR #870, opened 2018-04-27
prevent save-artifact tar extraction from overwriting files outside the working dir

bparees:


bparees:
@gabemontero ptal

gabemontero:
Perhaps unit tests to cover this are possible, but I'm good if you or somebody else does that as a follow up (assuming this needs to get merged before you head out of town).

bparees:
[merge]

Issue #872, opened 2018-04-27
Multiple runtime artifacts provision is not working


ingcsmoreno:
## What happen?
The **How to use a non-builder image for the final application image** document specifies that it's possible to set [multiple runtime artifacts](https://github.com/openshift/source-to-image/blob/master/docs/runtime_image.md#runtime-image-requirements) to be copied from _build_ container to _runtime_ container by passing the `runtime-artifacts` option a semicolon separated list of paths.

I executed a build with the following command:
```
s2i build /Users/Escanor/Git/AwesomeApp s2i/php:builder_7.1-1.8.2 service/awesomeapp:live --runtime-image s2i/php:run_7.1-1.8.2 --runtime-artifact "/var/www:./var;/etc/tokens" -v /tmp/tokens:/etc/tokens
```
## Expected Result
The build works, and both paths are copied into the runtime container.

## Actual Result
The following error occurred:
```
Error: invalid argument "/var/www:./var;/etc/tokens" for --runtime-artifact: invalid characters in filename: "/var/www:./var;/etc/tokens"
```

## Where?
I've (tried to) run the build on the following environments:

- Alpine Linux 3.4 and S2I v1.1.0
- MacOS High Sierra and S2I v1.1.5

Got the same result and error message in both cases.

## Extra info
After a quick inspection to the source code, I found the is a [path validation](https://github.com/openshift/source-to-image/blob/9a1bd4dcd19bca938c3f8aa09cf1748a1560554d/pkg/api/types.go#L26) that explicitly declares the semicolon as an invalid character.  

Shouldn't a paths split take place before this check?

adambkaplan:
@bparees closing this issue (feature merged in origin)

bparees:
I think it could have been closed when the fix was merged into the s2i repo actually.  We don't even support the runtime image/artifacts flow on openshift anymore and the issue was reported against the s2i cli invocation.  But either way, thanks for fixing it :)

PR #873, opened 2018-05-01
Parse multiple volumes via semicolon delimeter.

adambkaplan:
Fixes issue #872

### DIFF
--- a/hack/test-stirunimage.sh
+++ b/hack/test-stirunimage.sh
@@ -140,6 +140,10 @@ test_debug "s2i build with remote git repo"
 s2i build https://github.com/openshift/cakephp-ex docker.io/centos/php-70-centos7 test --loglevel=5 &> "${WORK_DIR}/s2i-git-proto.log"
 check_result $? "${WORK_DIR}/s2i-git-proto.log"
 
+test_debug "s2i build with runtime image"
+s2i build --ref=10.x --context-dir=helloworld https://github.com/wildfly/quickstart docker.io/openshift/wildfly-101-centos7 test-jee-app-thin --runtime-image=docker.io/openshift/wildfly-101-centos7 &> "${WORK_DIR}/s2i-runtime-image.log"
+check_result $? "${WORK_DIR}/s2i-runtime-image.log"
+
 test_debug "s2i build with --run==true option"
 if [[ "$OSTYPE" == "cygwin" ]]; then
   ( cd hack/windows/sigintwrap && make )
--- a/pkg/api/types.go
+++ b/pkg/api/types.go
@@ -527,8 +527,22 @@ func IsInvalidFilename(name string) bool {
 // When the destination is not specified, the source get copied into current
 // working directory in container.
 func (l *VolumeList) Set(value string) error {
+	volumes := strings.Split(value, ";")
+	newVols := make([]VolumeSpec, len(volumes))
+	for i, v := range volumes {
+		spec, err := l.parseSpec(v)
+		if err != nil {
+			return err
+		}
+		newVols[i] = *spec
+	}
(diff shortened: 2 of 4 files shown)


hack/test-stirunimage.sh
```diff
@@ -140,6 +140,10 @@ test_debug "s2i build with remote git repo"
 s2i build https://github.com/openshift/cakephp-ex docker.io/centos/php-70-centos7 test --loglevel=5 &> "${WORK_DIR}/s2i-git-proto.log"
 check_result $? "${WORK_DIR}/s2i-git-proto.log"
 
+test_debug "s2i build with runtime image"
+s2i build --ref=10.x --context-dir=helloworld https://github.com/wildfly/quickstart docker.io/openshift/wildfly-101-centos7 test-jee-app-thin --runtime-image=docker.io/openshift/wildfly-101-centos7 &> "${WORK_DIR}/s2i-runtime-image.log"
```
adambkaplan: Note - the wildfly s2i images were recently updated to support runtime artifacts/builds.

pkg/api/types.go
```diff
@@ -527,8 +527,22 @@ func IsInvalidFilename(name string) bool {
...
+		newVols[i] = spec
+	}
+	*l = append(*l, newVols...)
+	return nil
+}
+
+func (l *VolumeList) parseSpec(value string) (VolumeSpec, error) {
 	if len(value) == 0 {
-		return errors.New("invalid format, must be source:destination")
+		return VolumeSpec{}, errors.New("invalid format, must be source:destination")
```
gabemontero: should return nil vs. empty

pkg/api/types.go
```diff
@@ -541,10 +555,9 @@ func (l *VolumeList) Set(value string) error {
 	mount[1] = strings.Trim(mount[1], `"'`)
 	s := VolumeSpec{Source: filepath.Clean(mount[0]), Destination: filepath.ToSlash(filepath.Clean(mount[1]))}
 	if IsInvalidFilename(s.Source) || IsInvalidFilename(s.Destination) {
-		return fmt.Errorf("invalid characters in filename: %q", value)
+		return VolumeSpec{}, fmt.Errorf("invalid characters in filename: %q", value)
```
gabemontero: should return nil vs. empty

gabemontero:
[test]

gabemontero:
I think s2i still leverages rosie ... let's see

gabemontero:
/test

maybe not rosie

gabemontero:
No I was right ... still rosie ... there she goes

pkg/api/types.go
adambkaplan: Can't do that for value return :/

gabemontero: sorry, the implication was to change the return type to a pointer

gabemontero: unless there are errors that can be selectively ignored by the caller, why return an empty or partial VolumeSpec ... less space on the stack as well

adambkaplan:
[test]

bparees:
this lgtm but one concern:

what's the behavior in origin today if someone specifies a path uses a ;?  Are we erroring out, or are we tolerating the filename/path as being "some;path"?

Because if we're tolerating it as a valid filename today, this PR will change that behavior in origin and that would be a breaking change.

adambkaplan:
Based on the prior state, s2i rejects `some;path` as an illegal file name - `;` is an invalid file name character. See [here](https://github.com/openshift/source-to-image/blob/master/pkg/api/types.go#L26)

bparees:
ok, seems like this should be safe then, thanks.

[merge]

PR #874, opened 2018-05-01
Improve URL redaction

adambkaplan:
Addresses issue brought up in BZ 1571349

### DIFF
--- a/pkg/util/env_test.go
+++ b/pkg/util/env_test.go
@@ -4,60 +4,54 @@ import (
 	"testing"
 )
 
+type envTestCase struct {
+	Input  string
+	Output string
+}
+
 func TestStripProxyCredentials(t *testing.T) {
 
-	inputs := []string{
-		// values w/o protocols are untouched
-		"http_proxy=user:password@hostname.com",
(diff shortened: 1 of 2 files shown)


pkg/util/env_test.go
```diff
@@ -4,60 +4,54 @@ import (
...
+	expected := make([]string, len(cases))
+	for i, c := range cases {
+		inputs[i] = c.Input
+		expected[i] = c.Output
 	}
 	result := SafeForLoggingEnv(inputs)
 	for i := range result {
 		if result[i] != expected[i] {
-			t.Errorf("expected %s to be stripped to %s, but got %s", inputs[i], expected[i], result[i])
+			t.Errorf("expected %s to be changed to %s, but got %s", inputs[i], expected[i], result[i])
```
bparees: `expected %s to be %s`  (this test no longer expects the string to be changed in all cases, there are some cases where it should be untouched)

adambkaplan:
@bparees @gabemontero updated unit test message.

bparees:
[merge]

PR #875, opened 2018-05-01
Fixing tar unit test on macOS

adambkaplan:
Unclear why this is occurring, but this unit test failure was consistent on macOS.


bparees:
[merge]

PR #878, opened 2018-05-14
Option to produce a buildable Dockerfile instead of an image when building

bparees:


bparees:
@openshift/sig-developer-experience ptal

needs tests

not totally happy w/ how cfg.ScriptsURL is being handled today, may tweak that to introduce an ImageScriptsPath option (the challenge is that since we can't inspect the builder image, we can't look up the default assemble/run scripts path from the image metadata, so it needs to be passed in somewhere... right now i'm reusing the existing --scripts-url arg, but that precludes having actual remote assemble/run scripts hosted somewhere else)

pkg/build/strategies/dockerfile/dockerfile.go
```diff
@@ -0,0 +1,330 @@
...
+	for _, injection := range config.Injections {
+		//src := strings.TrimPrefix(injection.Source, builder.config.WorkingDir)
+		src := filepath.Join(api.Injections, injection.Source)
+		buffer.WriteString(fmt.Sprintf("COPY --chown=%s:0 %s %s\n", sanitize(imageUser), sanitize(src), sanitize(injection.Destination)))
+	}
+
+	if assembleProvided {
+		buffer.WriteString(fmt.Sprintf("RUN %s\n", sanitize(filepath.Join(scriptsDestDir, "assemble"))))
+	} else {
+		buffer.WriteString(fmt.Sprintf("RUN %s\n", sanitize(filepath.Join(inImageScriptsDir, "assemble"))))
```
gabemontero: So as part of "the challenge is that since we can't inspect the builder image", we have to just  assume the assemble script is in the builder image .... and if it is not, the failure will be such that the user can sort that out

Could it be beneficial to add a comment to the Dockerfile explaining we've made an assumption the file is there, you need to make sure it is there, etc.

pkg/build/strategies/dockerfile/dockerfile.go
```diff
@@ -0,0 +1,330 @@
...
+	}
+
+	if runProvided {
+		buffer.WriteString(fmt.Sprintf("CMD %s\n", sanitize(filepath.Join(scriptsDestDir, "run"))))
+	} else {
+		buffer.WriteString(fmt.Sprintf("CMD %s\n", sanitize(filepath.Join(inImageScriptsDir, "run"))))
+	}
+
+	//uploadDir := filepath.Join(builder.config.WorkingDir, "upload")
+	//if err := builder.fs.WriteFile(filepath.Join(uploadDir, "Dockerfile"), buffer.Bytes()); err != nil {
```
gabemontero: clean up commented out lines

pkg/build/strategies/dockerfile/dockerfile.go
```diff
@@ -0,0 +1,330 @@
...
+	if err := builder.fs.WriteFile(filepath.Join(builder.config.AsDockerfile), buffer.Bytes()); err != nil {
+		return err
+	}
+	//glog.V(2).Infof("Writing custom Dockerfile to %s", uploadDir)
```
gabemontero: clean up commented out lines

pkg/build/strategies/dockerfile/dockerfile.go
```diff
@@ -0,0 +1,330 @@
...
+		buffer.WriteString(fmt.Sprintf("rm -rf %s", sanitize(injection.Destination)))
+	}
+	if wroteRun {
+		buffer.WriteString("\n")
+	}
+
+	if runProvided {
+		buffer.WriteString(fmt.Sprintf("CMD %s\n", sanitize(filepath.Join(scriptsDestDir, "run"))))
+	} else {
+		buffer.WriteString(fmt.Sprintf("CMD %s\n", sanitize(filepath.Join(inImageScriptsDir, "run"))))
```
gabemontero: same notion re: challenge -> assumption -> possible comment warning

pkg/build/strategies/dockerfile/dockerfile.go
```diff
@@ -0,0 +1,330 @@
...
+		}
+		if config.SourceInfo != nil {
+			builder.sourceInfo = builder.config.SourceInfo
+		}
+	}
+
+	// we are only installing scripts from the source repo,
+	// and if we don't find one there, we'll fall back to the in-image scripts.
+	scriptInstaller := scripts.NewInstaller(
+		"", //builder.config.BuilderImage,
```
gabemontero: clean up

pkg/build/strategies/dockerfile/dockerfile.go
```diff
@@ -0,0 +1,330 @@
...
+	}
+
+	buffer.WriteString(fmt.Sprintf("COPY --chown=%s:0 %s %s\n", sanitize(imageUser), sanitize(builder.uploadSrcDir), sanitize(filepath.ToSlash(sourcesDestDir))))
+
+	glog.V(2).Infof("Processing secret inputs: %#v", config.Injections)
+	config.Injections = util.FixInjectionsWithRelativePath(config.ImageWorkDir, config.Injections)
+	glog.V(2).Infof("Processed secret inputs: %#v", config.Injections)
+
+	for _, injection := range config.Injections {
+		//src := strings.TrimPrefix(injection.Source, builder.config.WorkingDir)
```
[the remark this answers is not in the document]
bparees: will fix

pkg/build/strategies/dockerfile/dockerfile.go
```diff
@@ -0,0 +1,330 @@
...
+		for k, v := range generatedLabels {
+			buffer.WriteString(fmt.Sprintf(" %s=%s", k, v))
+		}
+		buffer.WriteString("\n")
+	}
+
+	// only COPY scripts dir if required scripts are present, i.e. the dir is not empty;
+	// even if the "scripts" dir exists, the COPY would fail if it was empty
+	scriptsProvided, fileNames := checkValidDirWithContents(filepath.Join(builder.config.WorkingDir, builder.uploadScriptsDir))
+	//glog.V(2).Infof("found %v files in script dir %s", fileNames, builder.uploadScriptsDir)
```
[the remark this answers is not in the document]
bparees: will fix

pkg/build/strategies/dockerfile/dockerfile.go
bparees: sure, will add.

pkg/build/strategies/dockerfile/dockerfile.go
bparees: yup

bparees:
@gabemontero comments addressed in `gabe 1` but i think i'm also going to go back and add in an "--image-scripts-url" arg or something along those lines so we can continue to support the remote scripts url when doing this.  (s2i will pull down the remote scripts into the working directory)

pkg/api/types.go
```diff
@@ -249,6 +249,10 @@ type Config struct {
 
 	// ImageWorkDir is the default working directory for the builder image.
 	ImageWorkDir string
+
+	// ImageScriptsPath is the path to the directory containing the assemble/run scripts located within the builder image.
+	// It should be specified as an absolute path (e.g. "/usr/libexec/s2i")
+	ImageScriptsPath string
```
gabemontero: How about `ImageScriptsDir` to be more consistent with `ImageWorkDir` above

Also, `path` could imply either dir or file

pkg/build/strategies/dockerfile/dockerfile.go
```diff
@@ -180,9 +176,9 @@ func (builder *Dockerfile) CreateDockerfile(config *api.Config) error {
 	if assembleProvided {
 		buffer.WriteString(fmt.Sprintf("RUN %s\n", sanitize(filepath.Join(scriptsDestDir, "assemble"))))
 	} else {
-		buffer.WriteString(fmt.Sprintf("# Assemble script sourced from builder image based on user input or image metadata."))
-		buffer.WriteString(fmt.Sprintf("# If this file does not exist in the image, the build will fail."))
-		buffer.WriteString(fmt.Sprintf("RUN %s\n", sanitize(filepath.Join(inImageScriptsDir, "assemble"))))
+		buffer.WriteString(fmt.Sprintf("# Assemble script sourced from builder image based on user input or image metadata.\n"))
+		buffer.WriteString(fmt.Sprintf("# If this file does not exist in the image, the build will fail.\n"))
+		buffer.WriteString(fmt.Sprintf("RUN %s\n", sanitize(filepath.Join(config.ImageScriptsPath, "assemble"))))
```
gabemontero: Don't see where this is defaulted to `/usr/libexec/s2i` if the user does not explicitly set it ...  maybe `checkValidDirWithContents` does that?

Or should we do a len check on `config.ImageScriptsPath` and give `/usr/libexec/s2i` a go as a last resort?

pkg/api/types.go
bparees: yeah that is better

pkg/build/strategies/dockerfile/dockerfile.go
bparees: it's defaulted when the flag is setup:

`buildCmd.Flags().StringVar(&(cfg.ImageScriptsDir), "image-scripts-dir", "/usr/libexec/s2i", "Specify the absolute path of the directory containing the assemble and run scripts within the builder image")`

bparees:
`gabe 2` pushed.  still working on tests.

gabemontero:
good with `gabe 2`

pkg/build/strategies/dockerfile/dockerfile.go
```diff
@@ -0,0 +1,323 @@
...
+		if injection.Keep {
+			continue
+		}
+		if !wroteRun {
+			buffer.WriteString("RUN ")
+			wroteRun = true
+		} else {
+			buffer.WriteString(" && ")
+		}
+		buffer.WriteString(fmt.Sprintf("rm -rf %s", sanitize(injection.Destination)))
```
adambkaplan: Noting change in behavior here, where secrets are deleted rather than truncated.
It's hard for me to see a use case where a developer relies on a truncated secret file existing, but this xkcd always comes to mind: https://xkcd.com/1172/

pkg/cmd/cli/cmd/build.go
```diff
@@ -196,5 +223,6 @@ $ s2i build . centos/ruby-22-centos7 hello-world-app
 	buildCmd.Flags().StringVar(&(cfg.RuntimeImage), "runtime-image", "", "Image that will be used as the base for the runtime image")
 	buildCmd.Flags().VarP(&(cfg.RuntimeArtifacts), "runtime-artifact", "a", "Specify a file or directory to be copied from the builder to the runtime image")
 	buildCmd.Flags().StringVar(&(networkMode), "network", "", "Specify the default Docker Network name to be used in build process")
+	buildCmd.Flags().StringVarP(&(cfg.AsDockerfile), "as-dockerfile", "", "", "EXPERIMENTAL: Output a Dockerfile to this path instead of building a new image")
```
adambkaplan: /s/as-dockerfile/to-dockerfile/ (?)

pkg/build/strategies/dockerfile/dockerfile.go
```diff
@@ -0,0 +1,323 @@
...
+		return err
+	}
+	glog.V(2).Infof("Wrote custom Dockerfile to %s", builder.config.AsDockerfile)
+	return nil
+}
+
+// Prepare prepares the source code and tar for build.
+// NOTE: this func serves both the sti and onbuild strategies, as the OnBuild
+// struct Build func leverages the STI struct Prepare func directly below.
+func (builder *Dockerfile) Prepare(config *api.Config) error {
```
adambkaplan: Would this be the right place to verify that `AsDockerfile` is valid? Ex - if `AsDockerfile` is a `my/output/path/`, can we assume the user wants the output to be `my/output/path/Dockerfile` and update the `AsDockerfile` value accordingly?

pkg/build/strategies/dockerfile/dockerfile.go
```diff
@@ -0,0 +1,323 @@
...
+	sourcesDestDir := filepath.Join(getDestination(config), "src")
+
+	buffer.WriteString(fmt.Sprintf("FROM %s\n", builder.config.BuilderImage))
+	buffer.WriteString(fmt.Sprintf("USER %s\n", imageUser))
+
+	generatedLabels := util.GenerateOutputImageLabels(builder.sourceInfo, builder.config)
+	if len(generatedLabels) > 0 {
+		buffer.WriteString("LABEL")
+		for k, v := range generatedLabels {
+			buffer.WriteString(fmt.Sprintf(" %s=%s", k, v))
```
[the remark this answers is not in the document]
bparees: we care about it less than we care about creating unnecessary layers...every line in a dockerfile is an additional layer, so that's why i did it this way.

pkg/build/strategies/dockerfile/dockerfile.go
bparees: yeah it's a fair point.  my guess is we'll phase in this new build approach over time so people will be able to fallback to the existing behavior if they want it, at least for some period of time.

pkg/cmd/cli/cmd/build.go
bparees: i went with `as` because in my mind the default behavior is "build as an image" and the new behavior is "build as a dockerfile artifact", but if you can swing @gabemontero or @coreydaley i'll switch it out.

pkg/build/strategies/dockerfile/dockerfile.go
bparees: i think if additional validation logic is needed, i'd put it in build.go where we're already validating the arguments in other ways.

as for what validation/defaulting should be done with the AsDockerfile value...  yes i suppose if it ends in "/" we can append "Dockerfile" by default.

pkg/cmd/cli/cmd/build.go
adambkaplan: full context for voting:
```
$ s2i build . build-image --as-dockerfile my/Dockerfile
```
vs.
``
$ s2i build . build-image --to-dockerfile my/Dockerfile
``

pkg/build/strategies/dockerfile/dockerfile.go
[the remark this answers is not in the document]
adambkaplan: I was thinking of using `\` to keep it all in one layer:
```
LABEL <key>=<value> \
            <key2>=<value2> \
           ...
```

pkg/cmd/cli/cmd/build.go
coreydaley: My vote is for --as-dockerfile
$ s2i build . build-image --as-dockerfile my/Dockerfile

pkg/build/strategies/dockerfile/dockerfile.go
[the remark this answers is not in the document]
bparees: i can take a shot at that.

bparees:
@coreydaley @adambkaplan new commits for you guys.

pkg/build/strategies/dockerfile/dockerfile.go
```diff
@@ -0,0 +1,327 @@
...
+	wroteRun := false
+	for _, injection := range config.Injections {
+		if injection.Keep {
+			continue
+		}
+		if !wroteRun {
+			buffer.WriteString("RUN ")
+			wroteRun = true
+		} else {
+			buffer.WriteString(" && ")
```
adambkaplan: /s/" && "/" \\\n&& "/
interesting note - Docker's ["best practice" example](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#run) doesn't indent the `&&` with multiline

bparees: new commit pushed, here's the latest output:

```
FROM docker.io/centos/nodejs-8-centos7
USER 1001
LABEL io.openshift.s2i.build.image=docker.io/centos/nodejs-8-centos7 \
      io.openshift.s2i.build.commit.author=Honza Horak <hhorak@redhat.com> \
      io.openshift.s2i.build.commit.date=Fri May 4 09:16:28 2018 +0200 \
      io.openshift.s2i.build.commit.id=a2aa814098856b1d935f68db1eb808d091e1a44e \
      io.openshift.s2i.build.commit.ref=master \
      io.openshift.s2i.build.commit.message=Merge pull request #178 from wozniakjan/issue/origin_19379/generalization_of_p.. \
      io.openshift.s2i.build.source-location=https://github.com/openshift/nodejs-ex
ENV FOO=BAR \
    BAR=FOO
COPY --chown=1001:0 upload/src /tmp/src
COPY --chown=1001:0 upload/injections/tmp/s /tmp/s
COPY --chown=1001:0 upload/injections/tmp/s /tmp/s2
# Assemble script sourced from builder image based on user input or image metadata.
# If this file does not exist in the image, the build will fail.
RUN /usr/libexec/s2i/assemble
RUN rm -rf /tmp/s && \
    rm -rf /tmp/s2
# Run script sourced from builder image based on user input or image metadata.
# If this file does not exist in the image, the build will fail.
CMD /usr/libexec/s2i/run
```

hack/test-stirunimage.sh
```diff
@@ -144,6 +144,11 @@ test_debug "s2i build with runtime image"
 s2i build --ref=10.x --context-dir=helloworld https://github.com/wildfly/quickstart docker.io/openshift/wildfly-101-centos7 test-jee-app-thin --runtime-image=docker.io/openshift/wildfly-101-centos7 &> "${WORK_DIR}/s2i-runtime-image.log"
 check_result $? "${WORK_DIR}/s2i-runtime-image.log"
 
+test_debug "s2i build with Dockerfile output"
+s2i build https://github.com/openshift/cakephp-ex docker.io/centos/php-70-centos7 --as-dockerfile=${WORK_DIR}/asdockerfile/Dockerfile --loglevel=5 >& "${WORK_DIR}/s2i-dockerfile.log"
+check_result $? "${WORK_DIR}/s2i-dockerfile.log"
```
gabemontero: Would some examination of the dockerfile be prudent for this particular feature, beyond the rc and file presence ... don't know if an exact file match is possible, but maybe at least grep for some expected entries?

gabemontero: ah - gotcha - sounds good ;-)

bparees:
[test]

bparees:
[test]

dobbymoodge:
[test]

bparees:
@gabemontero @adambkaplan @coreydaley ok i think this is ready for a final review, mostly just the last 2 commits (tests and uid checking)

pkg/build/strategies/dockerfile/dockerfile.go
```diff
@@ -0,0 +1,356 @@
...
+	buffer := bytes.Buffer{}
+
+	if len(config.ImageWorkDir) == 0 {
+		config.ImageWorkDir = "/opt/app-root/src"
+	}
+
+	imageUser := config.AssembleUser
+
+	// where files will land inside the new image.
+	// TODO: get this from the image metadata
```
gabemontero: is this //TODO still relevant given what you have discovered?  If so, was there a reason you didn't do it in this PR?

pkg/build/strategies/dockerfile/dockerfile.go
```diff
@@ -0,0 +1,355 @@
...
+	// copy in the user's source code.
+	buffer.WriteString("# Copying in source code\n")
+	buffer.WriteString(fmt.Sprintf("COPY --chown=%s:0 %s %s\n", sanitize(imageUser), sanitize(filepath.ToSlash(builder.uploadSrcDir)), sanitize(filepath.ToSlash(sourceDestDir))))
+
+	glog.V(2).Infof("Processing secret inputs: %#v", config.Injections)
+	config.Injections = util.FixInjectionsWithRelativePath(config.ImageWorkDir, config.Injections)
+	glog.V(2).Infof("Processed secret inputs: %#v", config.Injections)
+
+	if len(config.Injections) > 0 {
+		buffer.WriteString("# Copying in injected secret/configmap content\n")
```
adambkaplan: rm "secret/configmap" - things besides k8s secrets and configMaps can be injected, and the injections API is agnostic w.r.t. the source.

test/integration/integration_test.go
```diff
@@ -548,6 +551,404 @@ func (i *integrationTest) exerciseIncrementalBuild(tag, imageName string, remove
...
+		ImageScriptsDir: "/usr/libexec/s2i",
+		Source:          git.MustParse("https://github.com/sclorg/nodejs-ex"),
+		ScriptsURL:      "",
+		Injections:      api.VolumeList{},
+		Destination:     "",
+
+		Environment: api.EnvironmentList{},
+		Labels:      map[string]string{},
+
+		AsDockerfile: tempdir + string(os.PathSeparator) + "Dockerfile",
```
adambkaplan: One of these tests should have `AsDockerfile: tempdir` as the input (`Build` should mutate the value to `tmpdir/Dockerfile`)

pkg/build/strategies/dockerfile/dockerfile.go
```diff
@@ -0,0 +1,355 @@
...
+			runProvided = true
+		} else if f.Name() == "assemble" {
+			assembleProvided = true
+		}
+		if runProvided && assembleProvided {
+			break
+		}
+	}
+
+	env := createBuildEnvironment(config.WorkingDir, config.Environment)
```
adambkaplan: Does the build env need to be generated after the scripts check is run? Based on the inputs it seems the ENV directives can be written prior to the scripts check.

pkg/build/strategies/dockerfile/dockerfile.go
```diff
@@ -0,0 +1,355 @@
...
+	}, nil
+}
+
+// Build produces a Dockerfile that when run with the correct filesystem
+// context, will produce the application image.
+func (builder *Dockerfile) Build(config *api.Config) (*api.Result, error) {
+
+	// Handle defaulting of the configuration that is unique to the dockerfile strategy
+	if strings.HasSuffix(config.AsDockerfile, string(os.PathSeparator)) {
+		config.AsDockerfile = config.AsDockerfile + string(os.PathSeparator) + "Dockerfile"
```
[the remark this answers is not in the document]
bparees: "foo//bar" is a valid path (identical to "foo/bar")

but i can go ahead and remove the extra path separator since as you note, it's redundant here.

pkg/build/strategies/dockerfile/dockerfile.go
```diff
@@ -0,0 +1,355 @@
...
+	}
+	if len(config.AssembleUser) == 0 {
+		config.AssembleUser = "1001"
+	}
+	if !user.IsUserAllowed(config.AssembleUser, &config.AllowedUIDs) {
+		return nil, s2ierr.NewUserNotAllowedError(config.AssembleUser, false)
```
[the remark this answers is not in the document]
bparees: i doubt we're going to re-check the dockerfile downstream.  that said:

1) the label logic would be irrelevant here since we have no way to inspect the image to look for one..the assumption is the upstream logic would detect such a label and explicitly pass the assemble-user arg in the config

2) if the base image has ONBUILD commands, those would also need to be validated upstream because we have no way to inspect the image here.  But they do need to be validated because "dockerfile output" is actually a lot like a layered build.

pkg/build/strategies/dockerfile/dockerfile.go
```diff
@@ -0,0 +1,355 @@
...
+	scriptsDestDir := filepath.Join(getDestination(config), "scripts")
+	sourceDestDir := filepath.Join(getDestination(config), "src")
+
+	buffer.WriteString(fmt.Sprintf("FROM %s\n", config.BuilderImage))
+
+	if len(config.AssembleUser) > 0 {
+		buffer.WriteString(fmt.Sprintf("USER %s\n", imageUser))
+	}
+
+	generatedLabels := util.GenerateOutputImageLabels(builder.sourceInfo, config)
```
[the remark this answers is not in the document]
bparees: (also no one is going to read these :)  i highly doubt anyone using the s2i cli invocation is actually going to use this option and when it's used from within openshift no one is going to look at the dockerfile unless we're debugging something)

pkg/build/strategies/dockerfile/dockerfile.go
```diff
@@ -0,0 +1,355 @@
...
+		glog.V(2).Infof("Override scripts are included in directory %q", builder.uploadScriptsDir)
+		buffer.WriteString("# Copying in override assemble/run scripts\n")
+		buffer.WriteString(fmt.Sprintf("COPY --chown=%s:0 %s %s\n", sanitize(imageUser), sanitize(filepath.ToSlash(builder.uploadScriptsDir)), sanitize(filepath.ToSlash(scriptsDestDir))))
+	}
+
+	// copy in the user's source code.
+	buffer.WriteString("# Copying in source code\n")
+	buffer.WriteString(fmt.Sprintf("COPY --chown=%s:0 %s %s\n", sanitize(imageUser), sanitize(filepath.ToSlash(builder.uploadSrcDir)), sanitize(filepath.ToSlash(sourceDestDir))))
+
+	glog.V(2).Infof("Processing secret inputs: %#v", config.Injections)
```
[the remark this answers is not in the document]
bparees: will fix

pkg/build/strategies/dockerfile/dockerfile.go
bparees: will update

test/integration/integration_test.go
bparees: will add

pkg/build/strategies/dockerfile/dockerfile.go
bparees: yeah it can be done either way.  i'll move the env logic up before the script check.

bparees:
@adambkaplan comments responded to or fixed in new commit `adam 3`

bparees:
[merge]

PR #879, opened 2018-05-23
Check Assemble Users Against Allowed UIDs

adambkaplan:
Adding check to ensure the s2i assemble user is allowed if the `--allowed-uids` flag is set.
The assemble user can come from one of two sources:

1. `--assemble-user` flag
2. Builder image `io.openshift.s2i.assemble-user` label.

The assemble user overrides the default image user for an s2i build.
However, if the base image has ONBUILD instructions with USER directives,
all USER directives will be checked to ensure compliance.

Bug [1582976](https://bugzilla.redhat.com/show_bug.cgi?id=1582976)

### DIFF
--- a/pkg/docker/util.go
+++ b/pkg/docker/util.go
@@ -272,56 +272,68 @@ func PullImage(name string, d Docker, policy api.PullPolicy) (*PullResult, error
 	return &PullResult{Image: image, OnBuild: d.IsImageOnBuild(name)}, err
 }
 
-// CheckAllowedUser retrieves the user for a Docker image and checks that user
-// against an allowed range of uids.
+// CheckAllowedUser retrieves the execution users for a Docker image and
+// checks that user against an allowed range of uids.
 // - If the range of users is not empty, then the user on the Docker image
 // needs to be a numeric user
 // - The user's uid must be contained by the range(s) specified by the uids
 // Rangelist
+// - If build image uses an assemble user (via a command override or an
+// image label), that user must be within the allowed range of uids.
 // - If the image contains ONBUILD instructions and those instructions also
-// contain a USER directive, then the user specified by that USER directive
+// contain any USER directives, then all users specified by those USER directives
 // must meet the uid range criteria as well.
-func CheckAllowedUser(d Docker, imageName string, uids user.RangeList, isOnbuild bool) error {
+func CheckAllowedUser(d Docker, imageName string, uids user.RangeList, isOnbuild bool, assembleUserConfig string) error {
 	if uids == nil || uids.Empty() {
 		return nil
 	}
-	imageUserSpec, err := d.GetImageUser(imageName)
-	if err != nil {
-		return err
-	}
-	imageUser := extractUser(imageUserSpec)
-	if !user.IsUserAllowed(imageUser, &uids) {
-		return s2ierr.NewUserNotAllowedError(imageName, false)
-	}
+
+	// OnBuild users always need to be checked for layered builds
+	// Only return error if a user is not allowed, otherwise continue
 	if isOnbuild {
-		cmds, err := d.GetOnBuild(imageName)
+		onBuildUsers, err := extractOnBuildUsers(d, imageName)
 		if err != nil {
 			return err
 		}
-		if !isOnbuildAllowed(cmds, &uids) {
-			return s2ierr.NewUserNotAllowedError(imageName, true)
+		for _, usr := range onBuildUsers {
+			if !user.IsUserAllowed(usr, &uids) {
+				return s2ierr.NewUserNotAllowedError(imageName, true)
+			}
 		}
 	}
+
+	// P1: Assemble user configuration
+	if len(assembleUserConfig) > 0 {
+		if !user.IsUserAllowed(assembleUserConfig, &uids) {
+			// Pass in the override, since assembleUser can come from the image label
+			return s2ierr.NewAssembleUserNotAllowedError(imageName, true)
+		}
+		return nil
+	}
+
+	// P2: Assemble user label in image
+	assembleUser, err := extractAssembleUser(d, imageName)
+	if err != nil {
+		return err
+	}
+	if len(assembleUser) > 0 {
+		if !user.IsUserAllowed(assembleUser, &uids) {
+			// Pass in the override, since assembleUser can come from the image label
+			return s2ierr.NewAssembleUserNotAllowedError(imageName, false)
+		}
+		return nil
+	}
+
+	// Default - image user
+	imageUser, err := extractImageUser(d, imageName)
+	if err != nil {
+		return err
+	}
+	if !user.IsUserAllowed(imageUser, &uids) {
+		return s2ierr.NewUserNotAllowedError(imageName, false)
+	}
+
 	return nil
-}
-
-var dockerLineDelim = regexp.MustCompile(`[\t\v\f\r ]+`)
-
-// isOnbuildAllowed checks a list of Docker ONBUILD instructions for user
-// directives. It ensures that any users specified by the directives falls
-// within the specified range list of users.
-func isOnbuildAllowed(directives []string, allowed *user.RangeList) bool {
-	for _, line := range directives {
-		parts := dockerLineDelim.Split(line, 2)
-		if strings.ToLower(parts[0]) != "user" {
-			continue
-		}
-		uname := extractUser(parts[1])
-		if !user.IsUserAllowed(uname, allowed) {
-			return false
-		}
-	}
-	return true
 }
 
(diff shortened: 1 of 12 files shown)


pkg/errors/errors.go
```diff
@@ -233,7 +233,27 @@ func NewUserNotAllowedError(image string, onbuild bool) error {
...
+// NewAssembleUserNotAllowedError returns a new error that indicates that the build
+// could not run because the build or image uses an assemble user outside of the range
+// of allowed users.
+func NewAssembleUserNotAllowedError(image string, assembleUserOverride string) error {
+	var msg, suggestion string
+	if len(assembleUserOverride) > 0 {
+		msg = fmt.Sprintf("assemble user %q must be numeric and within the range of allowed users", assembleUserOverride)
+		suggestion = "build without the --allowed-uids flag or the --assemble-user flag"
+	} else {
+		assembleLabel := "io.openshift.s2i.assemble-user"
```
adambkaplan: I wish I could use the `AssembleUserLabel` constant, but `errors` is used by every other package in s2i.

pkg/docker/util.go
```diff
@@ -272,64 +272,87 @@ func PullImage(name string, d Docker, policy api.PullPolicy) (*PullResult, error
...
 	if uids == nil || uids.Empty() {
 		return nil
 	}
-	imageUserSpec, err := d.GetImageUser(imageName)
+	assembleUser, err := extractAssembleUser(d, assembleUserOverride, imageName)
+	if err != nil {
+		return err
+	}
+	if len(assembleUser) > 0 {
+		if !user.IsUserAllowed(assembleUser, &uids) {
```
bparees: if the assembleUser is allowed, I don't think we need to proceed to check if the imageUser is allowed?

If anything we probably don't want to because if someone has an image that runs as root, but they've explicitly said "i want to run my assemble script as $randomuser" then we should allow that because it's safe.

the on-build case is trickier since in that case the USER statement will get executed regardless of any overrides.

pkg/docker/util.go
```diff
@@ -344,7 +367,7 @@ func pullAndCheck(image string, docker Docker, pullPolicy api.PullPolicy, config
 		return nil, err
 	}
 
-	err = CheckAllowedUser(docker, image, config.AllowedUIDs, r.OnBuild)
+	err = CheckAllowedUser(docker, image, config.AllowedUIDs, r.OnBuild, config.AssembleUser)
```
bparees: is AssembleUser set to the value of the assemble-user label from the image at this point, if one was present?

pkg/docker/util_test.go
```diff
@@ -117,14 +119,43 @@ func TestCheckAllowedUser(t *testing.T) {
...
+			labels:      map[string]string{AssembleUserLabel: "10"},
+			expectErr:   false,
+		},
+		{
+			name:        "AllowedUIDs is set, numeric user, assemble user label root",
+			allowedUIDs: rangeList("1-"),
+			user:        "200",
+			labels:      map[string]string{AssembleUserLabel: "0"},
+			expectErr:   true,
+		},
```
bparees: need a case where the assembleuser is ok but the imageuser is not ok (and it should be allowed)

pkg/errors/errors.go
```diff
@@ -233,7 +233,27 @@ func NewUserNotAllowedError(image string, onbuild bool) error {
 	return Error{
 		Message:    msg,
 		ErrorCode:  UserNotAllowedError,
-		Suggestion: fmt.Sprintf("modify image %q to use a numeric user within the allowed range or build without the --allowed-uids flag", image),
+		Suggestion: fmt.Sprintf("modify image %q to use a numeric user within the allowed range, or build without the --allowed-uids flag", image),
```
bparees: i think this error gets dumped inside of openshift also, so suggesting passing flags is going to be confusing in that context.

pkg/errors/errors.go
bparees: meaning you're getting a circular dependency when you reference the constant?

pkg/errors/errors.go
```diff
@@ -233,7 +233,27 @@ func NewUserNotAllowedError(image string, onbuild bool) error {
...
+// of allowed users.
+func NewAssembleUserNotAllowedError(image string, assembleUserOverride string) error {
+	var msg, suggestion string
+	if len(assembleUserOverride) > 0 {
+		msg = fmt.Sprintf("assemble user %q must be numeric and within the range of allowed users", assembleUserOverride)
+		suggestion = "build without the --allowed-uids flag or the --assemble-user flag"
+	} else {
+		assembleLabel := "io.openshift.s2i.assemble-user"
+		msg = fmt.Sprintf("image %q includes the %q label whose value is not within the allowed range", image, assembleLabel)
+		suggestion = fmt.Sprintf("modify the %q label in image %q to use a numeric user within the allowed range, or build without the --allowed-uids flag", assembleLabel, image)
```
bparees: same issue w/ these two messages w/ respect to proposing flags. (assuming i'm right that this text gets exposed as is in openshift)

bparees:
[test]

pkg/errors/errors.go
adambkaplan: correct. I'm reluctant to create a standalone `constants` package just for this purpose.

bparees: yeah that's fine

pkg/docker/util.go
adambkaplan: not for the builder image. However, the logic in `extractAssembleUser` (added to CheckAllowedUser) inspects the `assemble-user` label on the image if the override is empty.

bparees: ok.  I feel like there is a hierarchy/priority ordering here that could be managed in a more obvious way..right now it feels like bits of it are scattered throughout the code.

the hierarchy is:

1) if alloweduids isn't set, we don't care what the assemble user is
2) if the config/cli explicitly provided an assemble user, check that value only (break)
3) if the image provides an assemble user label, check that value only (break)
4) if it's not on-build/layeredbuild check the default user of the image (break)
5) if it's a layered build check the USER logic in the on-build image (break)

factoring all that logic out to a single location would make it less fragile and easier to follow i think.

bparees: i should say i *think* that is the hierarchy.

bparees: > Challenge is how to handle OnBuild - do we need to check the onbuilder users as well as the default image user?

As I recall, the on-build logic checks the "on-build" commands in the base image to see if any are "USER" statements, and if any are "USER" statements, then we need to check what user they specify to ensure it is an allowed uid.

the last "USER" statement in the on-build args would become the default user for the image it produces. If there are no "USER" statements, then the default user for the resulting image is going to be the existing default user for the builder image.  (ie it's the same scenario as a non layered build, effectively).

adambkaplan: so it seems our hierarchy needs be reversed a bit (and the logic is a bit more complex):

1. No allowed uids - return nil
2. If layered, check OnBuild users. If ok then **continue**
3. If config has AssembleUser, check & return
4. If image has assemble user label, check & return
5. Fall through - check default image user

pkg/errors/errors.go
adambkaplan: there are other spots where we do this. Created #880 to address.

adambkaplan:
[test]

adambkaplan:
[test]
/assign @csrwng 
/cc @mfojtik while ben is on vacation.

bparees:
Adam can you make a PR against the 3.10 release branch as well? We'll want
that for pulling into origin so it doesn't pick up all my dockerfile
changes.


Ben Parees | OpenShift

pkg/errors/errors.go
```diff
@@ -233,7 +233,27 @@ func NewUserNotAllowedError(image string, onbuild bool) error {
...
+// NewAssembleUserNotAllowedError returns a new error that indicates that the build
+// could not run because the build or image uses an assemble user outside of the range
+// of allowed users.
+func NewAssembleUserNotAllowedError(image string, usesConfig bool) error {
+	var msg, suggestion string
+	if usesConfig {
+		msg = "assemble user must be numeric and within the range of allowed users"
+		suggestion = "build without the allowed-uids or assemble-user configurations set"
+	} else {
+		assembleLabel := "io.openshift.s2i.assemble-user"
```
csrwng: Use constant here

adambkaplan: @bparees and I discussed (comment may have been hidden). Almost all packages import errors, so there's no natural home for this without introducing a circular dependency.

Opened #882 to address.

bparees:
[merge]

PR #884, opened 2018-05-29
[3.9] Check Assemble Users Against Allowed UIDs

adambkaplan:
Adding check to ensure the s2i assemble user is allowed if the --allowed-uids flag is set.
The assemble user can come from one of two sources:

1. --assemble-user flag
2. Builder image io.openshift.s2i.assemble-user label.

The assemble user overrides the default image user for an s2i build.
However, if the base image has ONBUILD instructions with USER directives,
all USER directives will be checked to ensure compliance.

Bug 1582976

Backport of #879 to origin 3.9

/assign @csrwng 
/cc @bparees @gabemontero

### DIFF
--- a/hack/build-test-images.sh
+++ b/hack/build-test-images.sh
@@ -22,6 +22,8 @@ s2i::build_test_image() {
   cd "${S2I_ROOT}"
 
   s2i::build_test_image sti-fake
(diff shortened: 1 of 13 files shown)


bparees:
[merge]

PR #887, opened 2018-05-31
[3.7] prevent save-artifact tar extraction from overwriting files out…

adambkaplan:
…side the working dir

Backporting PR #870 to s2i 3.7 branch.

/assign @csrwng 
/cc @bparees


bparees:
[merge]

PR #889, opened 2018-06-01
[3.7] Revert Missing Vendored Code

adambkaplan:
The `openshift-3.7.0` branch includes code that was never vendored into origin 3.7.
This PR reverts these merges in the `openshift-3.7.0` branch to facilitate future backports.


adambkaplan:
[test]

gabemontero:
ah perhaps the only drawback on working on source-to-image ... deps back into origin :-)

@adambkaplan you should have seen what happened when we toyed with using the k8s client wrapper around docker  ... but don't worry, it got reverted :-)

bparees:
While it would be nice to have s2i and origin aligned in terms of deps, it isn't strictly necessary.  When we update the s2i dep in origin, we should be doing it with the "--strip-vendor" flag meaning that whatever deps s2i has, get ignored in deference to whatever version origin already has, right?

So why did this cause a problem when vendoring into origin/ose?

bparees:
(oh and if it was because 3.7 is pre-glide, we have a "hack/rebase-to-origin.sh" script in s2i that just copies over the s2i code w/o copying any deps).

adambkaplan:
3.7 is pre-glide, and my PR in OSE feels like the "by hand" version of the script. Handling via godep was a mess.

The 3.7 branch included a Docker bump that slightly changed the API used by s2i, resulting in a compile error. I also noticed that the crio change wasn't vendored into OSE on the 3.7 branch.

bparees:
[merge]

PR #899, opened 2018-06-29
Improve cross-platform compatibility in generated test script

joyrex2001:
This pull request contains some adjustments to the test/run script that is created when a new source-2-image project is bootstrapped using ```s2i create``` to make it work on both Linux and OSX. The nginx-centos7 example has been adjusted with the same changes. Additionally, it removes the "file://" uri which breaks the test as well.


bparees:
thanks @joyrex2001 
[merge]

Issue #903, opened 2018-07-17
Infinite Recursion Using --as-dockerfile


adambkaplan:
If the --as-dockerfile argument is in the same directory as the source, s2i errors out with an infinite recursion creating the upload directory.

**Expected result**
s2i creates a Dockerfile in `test/test-app`

**Actual result**
Error: file name too long

**Steps to reproduce**

1. Clone source-to-image
2. `cd examples/nginx-centos7`
3. `s2i build test/test-app nginx-centos7 --as-dockerfile test/test-app/Dockerfile`

PR #905, opened 2018-07-17
disallow copies to target into the source dir of the copy

bparees:
fixes https://github.com/openshift/source-to-image/issues/903


bparees:
[test]

bparees:
@adambkaplan updated:

```
$ s2i build test/test-app/ nginx-centos7 --as-dockerfile test/b/../test-app/Dockerfile
Build failed
ERROR: An error occurred: input source directory contains the target directory for the build, check that your --as-dockerfile path does not reside within your input source path: recursive copy requested, source directory "test/test-app/" contains the target directory "test/test-app/upload/src"
```

PR #907, opened 2018-07-18
Incremental Dockerfile Build

adambkaplan:
Incremental Dockerfile Build

Allow incremental `--as-dockerfile` builds.

Trello card: https://trello.com/c/P6TZg0wO/1582-5-builds-s2i-incremental-build-as-dockerfile

### DIFF
--- a/pkg/build/strategies/dockerfile/dockerfile.go
+++ b/pkg/build/strategies/dockerfile/dockerfile.go
@@ -125,11 +125,62 @@ func (builder *Dockerfile) CreateDockerfile(config *api.Config) error {
 	// where files will land inside the new image.
 	scriptsDestDir := filepath.Join(getDestination(config), "scripts")
 	sourceDestDir := filepath.Join(getDestination(config), "src")
+	artifactsDestDir := filepath.Join(getDestination(config), "artifacts")
+	artifactsTar := sanitize(filepath.ToSlash(filepath.Join(defaultDestination, "artifacts.tar")))
+
+	// only COPY scripts dir if required scripts are present, i.e. the dir is not empty;
+	// even if the "scripts" dir exists, the COPY would fail if it was empty
+	scriptsProvided, fileNames := checkValidDirWithContents(filepath.Join(config.WorkingDir, builder.uploadScriptsDir))
+	assembleProvided := false
+	runProvided := false
+	saveArtifactsProvided := false
+	for _, f := range fileNames {
+		glog.V(2).Infof("found override script file %s", f.Name())
+		if f.Name() == "run" {
+			runProvided = true
+		} else if f.Name() == "assemble" {
+			assembleProvided = true
+		} else if f.Name() == "save-artifacts" {
+			saveArtifactsProvided = true
+		}
+		if runProvided && assembleProvided && saveArtifactsProvided {
+			break
+		}
+	}
+
+	if config.Incremental {
+		imageTag := util.FirstNonEmpty(config.IncrementalFromTag, config.Tag)
+		if len(imageTag) == 0 {
+			return errors.New("Image tag is missing for incremental build")
+		}
+		buffer.WriteString(fmt.Sprintf("FROM %s as cached\n", imageTag))
+		if len(config.AssembleUser) > 0 {
+			buffer.WriteString(fmt.Sprintf("USER %s\n", imageUser))
+		}
+		var artifactsScript string
+		if saveArtifactsProvided {
+			glog.V(2).Infof("Override save-artifacts script is included in directory %q", builder.uploadScriptsDir)
+			buffer.WriteString("# Copying in override save-artifacts script\n")
+			artifactsScript = sanitize(filepath.ToSlash(filepath.Join(scriptsDestDir, "save-artifacts")))
+			uploadScript := sanitize(filepath.ToSlash(filepath.Join(builder.uploadScriptsDir, "save-artifacts")))
+			buffer.WriteString(fmt.Sprintf("COPY --chown=%s:0 %s %s\n", sanitize(imageUser), uploadScript, artifactsScript))
+		} else {
+			buffer.WriteString(fmt.Sprintf("# Save-artifacts script sourced from builder image based on user input or image metadata.\n"))
+			artifactsScript = sanitize(filepath.ToSlash(filepath.Join(config.ImageScriptsDir, "save-artifacts")))
+		}
+		buffer.WriteString(fmt.Sprintf("RUN if [ -s %[1]s ]; then %[1]s > %[2]s; else touch %[2]s; fi\n", artifactsScript, artifactsTar))
+	}
 
 	buffer.WriteString(fmt.Sprintf("FROM %s\n", config.BuilderImage))
 
 	if len(config.AssembleUser) > 0 {
 		buffer.WriteString(fmt.Sprintf("USER %s\n", imageUser))
+	}
+
+	if config.Incremental {
+		buffer.WriteString(fmt.Sprintf("COPY --from=cached --chown=%[1]s:0 %[2]s %[2]s\n", sanitize(imageUser), artifactsTar))
+		buffer.WriteString(fmt.Sprintf("RUN if [ -s %[1]s ]; then mkdir -p %[2]s; tar -xf %[1]s -C %[2]s; fi && \\\n", artifactsTar, sanitize(filepath.ToSlash(artifactsDestDir))))
+		buffer.WriteString(fmt.Sprintf("    rm %s\n", artifactsTar))
 	}
 
 	generatedLabels := util.GenerateOutputImageLabels(builder.sourceInfo, config)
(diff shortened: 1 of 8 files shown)


pkg/build/strategies/dockerfile/dockerfile.go
```diff
@@ -125,13 +125,64 @@ func (builder *Dockerfile) CreateDockerfile(config *api.Config) error {
...
+
+	if config.Incremental {
+		imageTag := util.FirstNonEmpty(config.IncrementalFromTag, config.Tag)
+		if len(imageTag) == 0 {
+			return errors.New("Image tag is missing for incremental build")
+		}
+		buffer.WriteString(fmt.Sprintf("FROM %s as cached\n", imageTag))
+		if len(config.AssembleUser) > 0 {
+			buffer.WriteString(fmt.Sprintf("USER %s\n", imageUser))
+		}
```
adambkaplan: I had the most concerns re: this instruction. `config.AssembleUser` is always non-empty because we default it to 1001. There is always a possibility that this user does not have permissions to run the `save-artifacts` script that is included in the image.

On the other hand, if the user provides an override for `save-artifacts` we need the `--chown` instruction with the `COPY` directive.

pkg/build/strategies/dockerfile/dockerfile.go
```diff
@@ -125,13 +125,64 @@ func (builder *Dockerfile) CreateDockerfile(config *api.Config) error {
...
+	}
 
 	buffer.WriteString(fmt.Sprintf("FROM %s\n", config.BuilderImage))
 
 	if len(config.AssembleUser) > 0 {
 		buffer.WriteString(fmt.Sprintf("USER %s\n", imageUser))
 	}
 
+	if config.Incremental {
+		buffer.WriteString(fmt.Sprintf("COPY --from=cached --chown=%s:0 %s %s\n", sanitize(imageUser), artifactsTar, artifactsTar))
```
gabemontero: Are `--from` and `--chown` newer dockerfile options?  Some albeit quick internet searches were not conclusive for me ...

Ultimately, while --as-dockefile is experimental, the docker version noted at https://github.com/openshift/source-to-image#dependencies seems very very old ...

Not directly specific to this PR, but to --as-dockerfile in general:  do we need to button up the docker dependencies any?

bparees: > Are --from and --chown newer dockerfile options? Some albeit quick internet searches were not conclusive for me ...

yes. newish anyway.  they aren't in the fedora docker version.

> Not directly specific to this PR, but to --as-dockerfile in general: do we need to button up the docker dependencies any?

Since we say "1.6 or greater", i would say not unless we think it no longer works w/ the old versions we indicate for some reason.

Note that s2i itself, when used with --as-dockerfile, has no docker dependency at all.  The dockerfile it generates, of course, will only work with a new enough docker version (or buildah, or some other image building tool that consumes dockerfiles and understands the newer syntax options)

pkg/build/strategies/dockerfile/dockerfile.go
```diff
@@ -125,13 +125,64 @@ func (builder *Dockerfile) CreateDockerfile(config *api.Config) error {
...
+		var artifactsScript string
+		if saveArtifactsProvided {
+			glog.V(2).Infof("Override save-artifacts script is included in directory %q", builder.uploadScriptsDir)
+			buffer.WriteString("# Copying in override save-artifacts script\n")
+			artifactsScript = sanitize(filepath.ToSlash(filepath.Join(scriptsDestDir, "save-artifacts")))
+			uploadScript := sanitize(filepath.ToSlash(filepath.Join(builder.uploadScriptsDir, "save-artifacts")))
+			buffer.WriteString(fmt.Sprintf("COPY --chown=%s:0 %s %s\n", sanitize(imageUser), uploadScript, artifactsScript))
+		} else {
+			buffer.WriteString(fmt.Sprintf("# Save-artifacts script sourced from builder image based on user input or image metadata.\n"))
+			buffer.WriteString(fmt.Sprintf("# If this file does not exist in the image, the build will fail.\n"))
```
bparees: this is going to be an unfortunate difference between how s2i used to work for incremental builds and how it will work.  Previously if the image didn't contain a save artifacts script, but incremental was requested, it would silently do a normal build.  

Anyone who's got a buildconfig that's set to incremental, but doesn't have an incremental capable image(or override script) is going to start failing.

I can't think of much we can do about it at the moment, but we should take note of it.

pkg/build/strategies/dockerfile/dockerfile.go
```diff
@@ -125,13 +125,64 @@ func (builder *Dockerfile) CreateDockerfile(config *api.Config) error {
...
+			artifactsScript = sanitize(filepath.ToSlash(filepath.Join(config.ImageScriptsDir, "save-artifacts")))
+		}
+		buffer.WriteString(fmt.Sprintf("RUN %s > %s\n", artifactsScript, artifactsTar))
```
bparees: maybe we make the run command be "RUN <if exists> runScript <endif>" ?

pkg/cmd/cli/cmd/build.go
```diff
@@ -70,10 +70,6 @@ $ s2i build . centos/ruby-22-centos7 hello-world-app
 			}
 
 			if len(cfg.AsDockerfile) > 0 {
-				if cfg.Incremental {
-					fmt.Fprintln(os.Stderr, "ERROR: --incremental cannot be used with --as-dockerfile")
-					return
-				}
```
bparees: :+1:

pkg/build/strategies/dockerfile/dockerfile.go
bparees: > Is there any standardization around what a Dockerfile supports? 

:rofl:

pkg/build/strategies/dockerfile/dockerfile.go
```diff
@@ -125,13 +125,64 @@ func (builder *Dockerfile) CreateDockerfile(config *api.Config) error {
...
+			glog.V(2).Infof("Override save-artifacts script is included in directory %q", builder.uploadScriptsDir)
+			buffer.WriteString("# Copying in override save-artifacts script\n")
+			artifactsScript = sanitize(filepath.ToSlash(filepath.Join(scriptsDestDir, "save-artifacts")))
+			uploadScript := sanitize(filepath.ToSlash(filepath.Join(builder.uploadScriptsDir, "save-artifacts")))
+			buffer.WriteString(fmt.Sprintf("COPY --chown=%s:0 %s %s\n", sanitize(imageUser), uploadScript, artifactsScript))
+		} else {
+			buffer.WriteString(fmt.Sprintf("# Save-artifacts script sourced from builder image based on user input or image metadata.\n"))
+			artifactsScript = sanitize(filepath.ToSlash(filepath.Join(config.ImageScriptsDir, "save-artifacts")))
+		}
+		buffer.WriteString(fmt.Sprintf("RUN if [ -s %[1]s ]; then %[1]s > %[2]s; else touch %[2]s; fi\n", artifactsScript, artifactsTar))
```
adambkaplan: Now inline check that the `save-artifacts` script exists. If not, then create an empty `artifacts.tar` file.

pkg/build/strategies/dockerfile/dockerfile.go
```diff
@@ -125,13 +125,64 @@ func (builder *Dockerfile) CreateDockerfile(config *api.Config) error {
...
 	buffer.WriteString(fmt.Sprintf("FROM %s\n", config.BuilderImage))
 
 	if len(config.AssembleUser) > 0 {
 		buffer.WriteString(fmt.Sprintf("USER %s\n", imageUser))
 	}
 
+	if config.Incremental {
+		buffer.WriteString(fmt.Sprintf("COPY --from=cached --chown=%[1]s:0 %[2]s %[2]s\n", sanitize(imageUser), artifactsTar))
+		buffer.WriteString(fmt.Sprintf("RUN if [ -s %[1]s ]; then mkdir -p %[2]s; tar -xf %[1]s -C %[2]s; fi && \\\n", artifactsTar, artifactsDestDir))
+		buffer.WriteString(fmt.Sprintf("    rm %s\n", artifactsTar))
```
adambkaplan: Inline check that `artifacts.tar` is non-empty before untar operation. Always delete

pkg/build/strategies/dockerfile/dockerfile.go
bparees: > There is always a possibility that this user does not have permissions to run the save-artifacts script that is included in the image.

i think that possibility already exists in the normal s2i incremental flow today.

bparees:
lgtm, squash and i'll merge if you're ready

adambkaplan:
@bparees squashed

bparees:
[merge]

bparees:
looks like there are some pathing issues on windows:

```
RUN if [ -s /tmp/artifacts.tar ]; then mkdir -p \tmp\artifacts; tar -xf /tmp/artifacts.tar -C \tmp\artifacts; fi && \
		    rm /tmp/artifacts.tar
```

i recall having to work through similar things when i did my PR.

adambkaplan:
[test]

adambkaplan:
[test]

adambkaplan:
[test]

adambkaplan:
[merge]

bparees:
i suspect you don't have merge here.

[merge]

adambkaplan:
[(5) [builds] S2I Incremental Build as Dockerfile](https://trello.com/c/P6TZg0wO/1582-5-builds-s2i-incremental-build-as-dockerfile)

PR #908, opened 2018-07-18
Remove references to cli flags in errors

### LINKED ISSUE #880: Remove References to CLI Flags in Error Messages
s2i is often run in a container and configured via environment variables.
Error messages printed to stderr should not reference command-line flags, as these configuration values can be set by other means.

adambkaplan:
Fixes #880

### DIFF
--- a/pkg/errors/errors.go
+++ b/pkg/errors/errors.go
@@ -74,7 +74,7 @@ func NewPullImageError(name string, err error) error {
 		Message:    fmt.Sprintf("unable to get %s", name),
 		Details:    err,
 		ErrorCode:  PullImageError,
-		Suggestion: "check image name, or if using local image add --pull-policy=never flag",
+		Suggestion: fmt.Sprintf("check image name, or if using a local image set the builder image pull policy to %q", "never"),
 	}
 }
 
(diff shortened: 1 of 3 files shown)


pkg/errors/errors.go
```diff
@@ -74,7 +74,7 @@ func NewPullImageError(name string, err error) error {
 		Message:    fmt.Sprintf("unable to get %s", name),
 		Details:    err,
 		ErrorCode:  PullImageError,
-		Suggestion: "check image name, or if using local image add --pull-policy=never flag",
+		Suggestion: fmt.Sprintf("check image name, or if using a local image set the builder image pull policy to %q", "never"),
```
bparees: is this preferred over escaping quotes in the string? Seems odd.

bparees:
travis is unhappy as a result of my earlier PR:
```
./pkg/scm/downloaders/file/download.go:17:6: exported type RecursiveCopyError should have comment or be unexported
```

can you go ahead and fix it here?

pkg/errors/errors.go
adambkaplan: %q seems to be a short cut, and I've seen it used elsewhere. Under the hood go does additional cleanup, like trimming whitespace.
If this was a performance-intensive, I wouldn't use it (then again, I'd avoid `fmt` for anything high-performance)

bparees: k

adambkaplan:
[test]

bparees:
[merge]

PR #909, opened 2018-07-23
update gitignore for vscode, vscode debug, metadata

gabemontero:
@openshift/sig-developer-experience ptal

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,4 +2,6 @@
 /.atom-build.yml
 /.project
 /.vagrant
+.vscode/
+cmd/s2i/debug
 *~


.gitignore
```diff
@@ -2,4 +2,6 @@
 /.atom-build.yml
 /.project
 /.vagrant
+.vscode/
+cmd/s2i/debug
```
bparees: what's producing cmd/s2i/debug as a folder?

gabemontero: it is actually a compiled binary

if you bring up s2i in the vscode debugger to step through code, set break points, etc, it compiles s2i in debug mode and places the output in cmd/s2i, since the go file with "main" is in cmd/s2i.  It then actually runs `cmd/s2i/debug` with whatever args you specify in the vscode config file (which is under .vscode).

bparees: blech. well let's hope we never want a debug package i guess.

bparees:
[merge]

PR #910, opened 2018-07-25
build: keep symlinks

### LINKED ISSUE #892: Copied repos duplicate symlinks
If an s2i build uses the `--copy` instruction and the source has a symlink, the s2i image produces two full copies of the symlinked file.

Reported in https://github.com/openshift/origin/issues/19809

wozniakjan:
tested with origin master as well. Maintains symlinks as expected if set by default in the `config`
https://github.com/wozniakjan/origin/pull/6/files

fixes #892


pkg/util/fs/fs.go
```diff
@@ -190,25 +190,44 @@ func (h *fs) Copy(source string, dest string) (err error) {
 	return doCopy(h, source, dest)
 }
 
-func doCopy(h FileSystem, source, dest string) error {
-	// Our current symlink behaviour is broken.  We follow symlinks and copy the
-	// referenced file, and fail when a symlink is broken.  In all cases we
-	// should actually copy the symlink as is.  For now, at least catch the
-	// broken symlink case and copy the symlink rather than failing.
+func (h *fs) KeepSymlinks(k bool) {
```
adambkaplan: Needs godoc

pkg/util/fs/fs.go
```diff
@@ -190,25 +190,44 @@ func (h *fs) Copy(source string, dest string) (err error) {
...
+	h.keepSymlinks = k
+}
+
+func (h *fs) ShouldKeepSymlinks() bool {
```
adambkaplan: Needs godoc

pkg/cmd/cli/cmd/build.go
```diff
@@ -214,5 +214,6 @@ $ s2i build . centos/ruby-22-centos7 hello-world-app
 	buildCmd.Flags().VarP(&(cfg.RuntimeArtifacts), "runtime-artifact", "a", "Specify a file or directory to be copied from the builder to the runtime image")
 	buildCmd.Flags().StringVar(&(networkMode), "network", "", "Specify the default Docker Network name to be used in build process")
 	buildCmd.Flags().StringVarP(&(cfg.AsDockerfile), "as-dockerfile", "", "", "EXPERIMENTAL: Output a Dockerfile to this path instead of building a new image")
+	buildCmd.Flags().BoolVarP(&(cfg.KeepSymlinks), "keep-symlinks", "", false, "Copy symlinks as symlinks. Default behavior is to follow symlinks and copy files by content")
```
bparees: this is only relevant when --copy is passed, right?

bparees: (and other cases where we end up in the "copy, not clone" path)

pkg/util/fs/fs.go
wozniakjan: Thanks! Added godocs and also missed the bash completion

pkg/cmd/cli/cmd/build.go
wozniakjan: Yes, only with `--copy`. For completion, I tested old code 'git clone' path and that works well with symlinks, no change needed there.

Thanks for your comment, I added that information to `--keep-symlinks` flag description

bparees:
[merge]

PR #911, opened 2018-07-30
Fix Hang if Injected Dir Does Not Exist

adambkaplan:
- Added a "response file" to report failed injections on s2i
container
- Fixed assemble command to ensure s2i container exits if
injection fails

OpenShift bug [1608714](https://bugzilla.redhat.com/show_bug.cgi?id=1608714)


bparees:
[merge]

PR #912, opened 2018-08-08
Allow URLs for Image Scripts Location

adambkaplan:
* Rename ImageScriptsDir to ImageScriptsURL
* Allow ImageScriptsURL to download scripts from filesystem, web, etc.
* Add default path for image scripts
* Update cli and completions


pkg/build/strategies/dockerfile/dockerfile.go
```diff
@@ -325,18 +329,10 @@ func (builder *Dockerfile) Prepare(config *api.Config) error {
...
-		builder.fs,
-	)
-
-	// all scripts are optional, we assume/hope the image contains scripts if we don't find them
-	// in the source repo.
-	scriptInstaller.InstallOptional(append(scripts.RequiredScripts, scripts.OptionalScripts...), config.WorkingDir)
+	// Install scripts provided by user
+	builder.installScripts(config.ScriptsURL, config)
+	// Install scripts specified by image metadata
+	builder.installScripts(config.ImageScriptsURL, config)
```
bparees: assuming this does what i think it does, the order seems backwards.  We want "scriptsurl" to take precedence, so you need to install those scripts second, so they overwrite anything you got from "imagescriptsurl"

pkg/build/strategies/dockerfile/dockerfile.go
```diff
@@ -128,6 +131,7 @@ func (builder *Dockerfile) CreateDockerfile(config *api.Config) error {
 	sourceDestDir := filepath.Join(getDestination(config), "src")
 	artifactsDestDir := filepath.Join(getDestination(config), "artifacts")
 	artifactsTar := sanitize(filepath.ToSlash(filepath.Join(defaultDestination, "artifacts.tar")))
+	imageScriptsDir := getImageScriptsDir(config)
```
bparees: if "imageScriptsDir" came explicitly from the user (via ScriptsUrl with an image:// prefix), then we need to treat the scenario as "assembleProvided=true" (and run provided, save-artifacts provided, etc).  Basically the user has told us, with the highest level of precedence, "use the scripts from the image, i promise they are there".

pkg/build/strategies/dockerfile/dockerfile.go
```diff
@@ -365,14 +368,48 @@ func getDestination(config *api.Config) string {
 	return destination
 }
 
-// checkValidDirWithContents returns true if the parameter provided is a valid,
-// accessible and non-empty directory.
-func checkValidDirWithContents(name string) (bool, []os.FileInfo) {
+// getImageScriptsDir returns the directory containing the builder image scripts
+func getImageScriptsDir(config *api.Config) (string, bool) {
+	if strings.HasPrefix(config.ScriptsURL, "image://") {
+		return strings.TrimPrefix(config.ScriptsURL, "image://"), true
```
adambkaplan: If the `ScriptsURL` points to an image location, indicate that we blind-trust all scripts are provided

test/integration/integration_test.go
```diff
@@ -957,14 +949,140 @@ func TestDockerfileBuildSourceScriptsRun(t *testing.T) {
...
+		filepath.Join(tempdir, "upload/scripts/run"),
+	}
+	runDockerfileTest(t, config, expected, nil, expectedFiles)
+}
+
+// TestDockerfileBuildScriptsURLImage tests the behavior if the ScriptsURL
+// is set to an image:// URL. In this case we blind trust that the image
+// contains all of the s2i scripts at the given directory, regardless
+// of what is contained in the source.
+func TestDockerfileBuildScriptsURLImage(t *testing.T) {
```
adambkaplan: Updated existing test - it now ensures that we don't upload scripts if ScriptsURL points to an image

test/integration/integration_test.go
```diff
@@ -979,29 +1097,184 @@ func TestDockerfileBuildSourceScriptsRun(t *testing.T) {
...
+		filepath.Join(tempdir, "upload/src/server.js"),
+		filepath.Join(tempdir, "upload/scripts/assemble"),
+	}
+	runDockerfileTest(t, config, expected, nil, expectedFiles)
+}
+
+// TestDockerfileBuildScriptsAndImageURLConflicts tests if both
+// the ScriptsURL and ImageScriptsURL point to a non-image directory.
+// In this event, the ScriptsURL value should take precedence.
+func TestDockerfileBuildScriptsAndImageURLConflicts(t *testing.T) {
```
adambkaplan: Test here if `ScriptsURL` and `ImageScriptsURL` point to non-image directories. Verifies that `ScriptsURL` overrides `ImageScriptsURL`

adambkaplan:
[test]

bparees:
[merge]

PR #913, opened 2018-08-16
Dockerfile Injection Removal

adambkaplan:
* Add script to remove secrets injected into Dockerfile builds
* Ensure dockerfile failure reason is reported in build result

### DIFF
--- a/pkg/build/strategies/dockerfile/dockerfile.go
+++ b/pkg/build/strategies/dockerfile/dockerfile.go
@@ -148,7 +148,8 @@ func (builder *Dockerfile) CreateDockerfile(config *api.Config) error {
 			buffer.WriteString("# Copying in override save-artifacts script\n")
 			artifactsScript = sanitize(filepath.ToSlash(filepath.Join(scriptsDestDir, "save-artifacts")))
 			uploadScript := sanitize(filepath.ToSlash(filepath.Join(builder.uploadScriptsDir, "save-artifacts")))
-			buffer.WriteString(fmt.Sprintf("COPY --chown=%s:0 %s %s\n", sanitize(imageUser), uploadScript, artifactsScript))
+			buffer.WriteString(fmt.Sprintf("COPY %s %s\n", uploadScript, artifactsScript))
+			buffer.WriteString(fmt.Sprintf("RUN chown %s:0 %s\n", sanitize(imageUser), artifactsScript))
 		} else {
 			buffer.WriteString(fmt.Sprintf("# Save-artifacts script sourced from builder image based on user input or image metadata.\n"))
 			artifactsScript = sanitize(filepath.ToSlash(filepath.Join(imageScriptsDir, "save-artifacts")))
@@ -163,8 +164,9 @@ func (builder *Dockerfile) CreateDockerfile(config *api.Config) error {
 	}
 
 	if config.Incremental {
-		buffer.WriteString(fmt.Sprintf("COPY --from=cached --chown=%[1]s:0 %[2]s %[2]s\n", sanitize(imageUser), artifactsTar))
-		buffer.WriteString(fmt.Sprintf("RUN if [ -s %[1]s ]; then mkdir -p %[2]s; tar -xf %[1]s -C %[2]s; fi && \\\n", artifactsTar, sanitize(filepath.ToSlash(artifactsDestDir))))
+		buffer.WriteString(fmt.Sprintf("COPY --from=cached %[1]s %[1]s\n", artifactsTar))
+		buffer.WriteString(fmt.Sprintf("RUN chown %s:0 %s && \\\n", sanitize(imageUser), artifactsTar))
+		buffer.WriteString(fmt.Sprintf("    if [ -s %[1]s ]; then mkdir -p %[2]s; tar -xf %[1]s -C %[2]s; fi && \\\n", artifactsTar, sanitize(filepath.ToSlash(artifactsDestDir))))
 		buffer.WriteString(fmt.Sprintf("    rm %s\n", artifactsTar))
 	}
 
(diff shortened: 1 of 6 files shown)


adambkaplan:
/assign @bparees 

@openshift/sig-developer-experience fyi

adambkaplan:
[test]

test/integration/integration_test.go
```diff
@@ -786,13 +794,15 @@ func TestDockerfileBuildInjections(t *testing.T) {
 
 	// strip the C: from windows paths because it's not valid in the middle of a path
 	// like upload/injections/C:/tempdir/injection1
+	// IS THIS COMPLETELY BROKEN???
```
adambkaplan: nope

pkg/util/injection.go
```diff
@@ -127,6 +129,20 @@ func CreateTruncateFilesScript(files []string, scriptName string) (string, error
 	return f.Name(), err
 }
 
+// CreateDeleteFilesScript creates a shell script that contains removal
+// of the provided files injected into the container. The path to the script is returned.
+func CreateDeleteFilesScript(files []string, dir string) (string, error) {
```
bparees: any reason you ended up going with this approach instead of generating a "RUN rm x && rm y && rm z" command?

bparees: (it's an uglier dockerfile but it seems less fiddly than generating a script, injecting it, running it, and deleting it)

pkg/build/strategies/dockerfile/dockerfile.go
```diff
@@ -79,7 +79,8 @@ func (builder *Dockerfile) Build(config *api.Config) (*api.Result, error) {
 		config.AssembleUser = "1001"
 	}
 	if !user.IsUserAllowed(config.AssembleUser, &config.AllowedUIDs) {
-		return nil, s2ierr.NewUserNotAllowedError(config.AssembleUser, false)
+		builder.setFailureReason(utilstatus.ReasonAssembleUserForbidden, utilstatus.ReasonMessageAssembleUserForbidden)
+		return builder.result, s2ierr.NewUserNotAllowedError(config.AssembleUser, false)
```
bparees: why couldn't (or why wasn't) the caller of Build() interpreting the returned error and handling this?  Seems a bit odd to do it everywhere we are about to return an error.

bparees:
couple questions on your design choices

pkg/build/strategies/dockerfile/dockerfile.go
adambkaplan: the original s2i code relied on the returned `builder.result` to be non-nil. It would use this to record timings for the build.

pkg/util/injection.go
adambkaplan: cleaner Dockerfile. Although in terms of performance, running a bunch of `rm` commands will need one less layer to achieve the same result.

bparees:
lgtm.  want to add one more commit to fix the --chown usage while you're in here?

test/integration/integration_test.go
```diff
@@ -606,17 +607,18 @@ func TestDockerfileBuild(t *testing.T) {
...
 		"\"io.openshift.s2i.build.source-location\"",
 		"\"io.openshift.s2i.build.image\"=\"docker.io/centos/nodejs-8-centos7\"",
 		"\"io.openshift.s2i.build.commit.author\"",
-		"COPY --chown=1001:0 upload/src /tmp/src",
-		"RUN /usr/libexec/s2i/assemble",
-		"CMD /usr/libexec/s2i/run",
+		"(?m)^COPY upload/src /tmp/src",
+		"(?m)^RUN chown -R 1001:0.* /tmp/src",
+		"(?m)^RUN /usr/libexec/s2i/assemble",
+		"(?m)^CMD /usr/libexec/s2i/run",
```
adambkaplan: Updated integration tests to use regex so that I could:

1. Ensure newlines are created
2. Make it easier to verify commands were created.

pkg/build/strategies/dockerfile/dockerfile.go
```diff
@@ -218,8 +221,20 @@ func (builder *Dockerfile) CreateDockerfile(config *api.Config) error {
...
+	}
+
+	// chown directories COPYed to image
+	if len(chownList) > 0 {
+		buffer.WriteString("# Change file ownership to the assemble user. Builder image must support chown command.\n")
+		buffer.WriteString(fmt.Sprintf("RUN chown -R %s:0", sanitize(imageUser)))
+		for _, dir := range chownList {
+			buffer.WriteString(fmt.Sprintf(" %s", dir))
+		}
+		buffer.WriteString("\n")
```
adambkaplan: Use chownList to change ownership of uploaded dirs in one layer

pkg/build/strategies/dockerfile/dockerfile.go
```diff
@@ -153,7 +148,8 @@ func (builder *Dockerfile) CreateDockerfile(config *api.Config) error {
 			buffer.WriteString("# Copying in override save-artifacts script\n")
 			artifactsScript = sanitize(filepath.ToSlash(filepath.Join(scriptsDestDir, "save-artifacts")))
 			uploadScript := sanitize(filepath.ToSlash(filepath.Join(builder.uploadScriptsDir, "save-artifacts")))
-			buffer.WriteString(fmt.Sprintf("COPY --chown=%s:0 %s %s\n", sanitize(imageUser), uploadScript, artifactsScript))
+			buffer.WriteString(fmt.Sprintf("COPY %s %s\n", uploadScript, artifactsScript))
+			buffer.WriteString(fmt.Sprintf("RUN chown %s:0 %s\n", sanitize(imageUser), artifactsScript))
```
adambkaplan: Run `chown` on the `save-artifacts` script that was copied into the incremental build image.

pkg/build/strategies/dockerfile/dockerfile.go
```diff
@@ -168,8 +164,9 @@ func (builder *Dockerfile) CreateDockerfile(config *api.Config) error {
 	}
 
 	if config.Incremental {
-		buffer.WriteString(fmt.Sprintf("COPY --from=cached --chown=%[1]s:0 %[2]s %[2]s\n", sanitize(imageUser), artifactsTar))
-		buffer.WriteString(fmt.Sprintf("RUN if [ -s %[1]s ]; then mkdir -p %[2]s; tar -xf %[1]s -C %[2]s; fi && \\\n", artifactsTar, sanitize(filepath.ToSlash(artifactsDestDir))))
+		buffer.WriteString(fmt.Sprintf("COPY --from=cached %[1]s %[1]s\n", artifactsTar))
+		buffer.WriteString(fmt.Sprintf("RUN chown %s:0 %s && \\\n", sanitize(imageUser), artifactsTar))
+		buffer.WriteString(fmt.Sprintf("    if [ -s %[1]s ]; then mkdir -p %[2]s; tar -xf %[1]s -C %[2]s; fi && \\\n", artifactsTar, sanitize(filepath.ToSlash(artifactsDestDir))))
```
adambkaplan: Run `chown` on the artifacts before they are un-tarred.

adambkaplan:
@bparees removed usage of `--chown` for dockerfile

bparees:
lgtm, thanks @adambkaplan 
[merge]

PR #914, opened 2018-08-23
Dockerfile Run COPY and CHOWN as root

adambkaplan:
Ensure all chown commands run as root.
Assemble user in general does not have permission to do this.

### DIFF
--- a/pkg/build/strategies/dockerfile/dockerfile.go
+++ b/pkg/build/strategies/dockerfile/dockerfile.go
@@ -138,14 +138,14 @@ func (builder *Dockerfile) CreateDockerfile(config *api.Config) error {
 		if len(imageTag) == 0 {
 			return errors.New("Image tag is missing for incremental build")
 		}
+		// Incremental builds run via a multistage Dockerfile
 		buffer.WriteString(fmt.Sprintf("FROM %s as cached\n", imageTag))
-		if len(config.AssembleUser) > 0 {
-			buffer.WriteString(fmt.Sprintf("USER %s\n", imageUser))
-		}
 		var artifactsScript string
 		if _, provided := providedScripts[constants.SaveArtifacts]; provided {
+			// switch to root to COPY and chown content
 			glog.V(2).Infof("Override save-artifacts script is included in directory %q", builder.uploadScriptsDir)
 			buffer.WriteString("# Copying in override save-artifacts script\n")
+			buffer.WriteString("USER root\n")
 			artifactsScript = sanitize(filepath.ToSlash(filepath.Join(scriptsDestDir, "save-artifacts")))
 			uploadScript := sanitize(filepath.ToSlash(filepath.Join(builder.uploadScriptsDir, "save-artifacts")))
 			buffer.WriteString(fmt.Sprintf("COPY %s %s\n", uploadScript, artifactsScript))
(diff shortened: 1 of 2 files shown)


adambkaplan:
/assign @bparees 

@openshift/sig-developer-experience fyi

pkg/build/strategies/dockerfile/dockerfile.go
```diff
@@ -139,13 +139,12 @@ func (builder *Dockerfile) CreateDockerfile(config *api.Config) error {
...
 		buffer.WriteString(fmt.Sprintf("FROM %s as cached\n", imageTag))
-		if len(config.AssembleUser) > 0 {
-			buffer.WriteString(fmt.Sprintf("USER %s\n", imageUser))
-		}
 		var artifactsScript string
 		if _, provided := providedScripts[constants.SaveArtifacts]; provided {
+			// Switch to root to COPY and chown content
 			glog.V(2).Infof("Override save-artifacts script is included in directory %q", builder.uploadScriptsDir)
 			buffer.WriteString("# Copying in override save-artifacts script\n")
+			buffer.WriteString("USER root\n")
```
adambkaplan: use `USER 0` instead?

pkg/build/strategies/dockerfile/dockerfile.go
```diff
@@ -154,41 +153,31 @@ func (builder *Dockerfile) CreateDockerfile(config *api.Config) error {
...
 			first = false
 		}
-		for k, v := range config.Labels {
-			if !first {
-				buffer.WriteString(fmt.Sprintf(" \\\n      "))
-			}
-			buffer.WriteString(fmt.Sprintf("%q=%q", k, v))
-			first = false
-		}
-
```
adambkaplan: Noted that if `config.Labels` had duplicate keys, we would have the same ENV variable set twice.

pkg/build/strategies/dockerfile/dockerfile.go
bparees: this should be fine i think, especially given our intent to remove it.

pkg/build/strategies/dockerfile/dockerfile.go
```diff
@@ -154,41 +153,31 @@ func (builder *Dockerfile) CreateDockerfile(config *api.Config) error {
...
+		// Switch to the image user if it is not root
+		if len(imageUser) > 0 && imageUser != "root" {
+			buffer.WriteString(fmt.Sprintf("USER %s\n", imageUser))
+		}
 		buffer.WriteString(fmt.Sprintf("RUN if [ -s %[1]s ]; then %[1]s > %[2]s; else touch %[2]s; fi\n", artifactsScript, artifactsTar))
 	}
 
 	buffer.WriteString(fmt.Sprintf("FROM %s\n", config.BuilderImage))
+	// Run as root to COPY and chown source content
+	buffer.WriteString("USER root\n")
```
bparees: can we move this closer to the copy?  right now it looks like it'll come out before the label+env commands.

pkg/build/strategies/dockerfile/dockerfile.go
bparees: you mean the same LABEL twice?

that's ok, the latter takes precedence i believe, so as long as we set the user's value last so they can overwrite our generated label, it's fine.

bparees: debating whether it's better as it was so we can see the values we generated in the dockerfile, or better the way you have it..   i guess i can go either way, so let's go with this.  less layers anyway.

adambkaplan:
@bparees updated per comments

adambkaplan:
[test]

bparees:
no offense but i can't wait to undo this :)

[merge]

PR #916, opened 2018-09-04
Remove trailing slash from secret destinationDir

coreydaley:
Fixes: https://github.com/openshift/origin/issues/18690

### DIFF
--- a/pkg/docker/docker.go
+++ b/pkg/docker/docker.go
@@ -367,7 +367,7 @@ func (d *stiDocker) UploadToContainer(fs fs.FileSystem, src, dest, container str
 // If the source is a single file, then the file copied into destination (which
 // has to be full path to a file inside the container).
 func (d *stiDocker) UploadToContainerWithTarWriter(fs fs.FileSystem, src, dest, container string, makeTarWriter func(io.Writer) s2itar.Writer) error {
-	destPath := filepath.Dir(dest)
+	destPath := filepath.Dir(strings.TrimRight(dest, "/"))
 	r, w := io.Pipe()
 	go func() {
 		tarWriter := makeTarWriter(w)


coreydaley:
@bparees ptal

pkg/docker/docker.go
```diff
@@ -367,7 +367,7 @@ func (d *stiDocker) UploadToContainer(fs fs.FileSystem, src, dest, container str
 // If the source is a single file, then the file copied into destination (which
 // has to be full path to a file inside the container).
 func (d *stiDocker) UploadToContainerWithTarWriter(fs fs.FileSystem, src, dest, container string, makeTarWriter func(io.Writer) s2itar.Writer) error {
-	destPath := filepath.Dir(dest)
+	destPath := filepath.Dir(filepath.Clean(dest))
```
bparees: this fundamentally changes the behavior of UploadToContainer that could break other consumers of it. I don't think it's the right fix.

(an upload that used to go to "/tmp" because "/tmp/" was passed in, is now going to go to "/")

coreydaley:
@bparees ptal

bparees:
i'm still nervous about the potential change in behavior here.

going back to the original issue:

https://github.com/openshift/origin/issues/18690

if i set "/tmp" as the destpath, where does the secret content get injected today (w/o your change)?
if i set it as "/tmp/" where does it get injected?

I assume the first one injects it to "/myinjection" and the second one to "/tmp/myinjection".

bparees:
fundamentally i think the bug in https://github.com/openshift/origin/issues/18690 is not how the upload behaves, it's how the truncate path is being defined.

bparees:
so rather than changing the upload behavior (and potentially breaking people), fix how truncate determines the path to truncate (so that it actually truncates the right path based on where upload put the content).

PR #922, opened 2018-09-14
build: only check if docker is reachable when not using --as-dockerfile

pkubatrh:
afaict there is no need to have docker accessible for using `s2i build --as-dockerfile`

### DIFF
--- a/pkg/cmd/cli/cmd/build.go
+++ b/pkg/cmd/cli/cmd/build.go
@@ -152,10 +152,12 @@ $ s2i build . centos/ruby-22-centos7 hello-world-app
 				glog.Fatal(err)
 			}
 
-			d := docker.New(client, cfg.PullAuthentication)
-			err = d.CheckReachable()
-			if err != nil {
-				glog.Fatal(err)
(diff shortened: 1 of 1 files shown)


pkg/cmd/cli/cmd/build.go
```diff
@@ -154,7 +154,7 @@ $ s2i build . centos/ruby-22-centos7 hello-world-app
 
 			d := docker.New(client, cfg.PullAuthentication)
 			err = d.CheckReachable()
-			if err != nil {
+			if len(cfg.AsDockerfile) == 0 && err != nil {
```
bparees: you can wrap the d.CheckReachable() call with inside this if block, no?  No point in running it if we are going to ignore the result.

pkubatrh:
Makes sense, modified the PR accordingly.

Not sure how to get rid of the CI failure - I have run `gofmt` against the files as the logs told me to but seems like its failing anyway.

bparees:
[merge]

bparees:
i'll look into the fmt issues separately.

PR #924, opened 2018-09-17
Update example repos that were moved to sclorg

liangxia:
github.com/openshift/django-ex has been moved to github.com/sclorg/django-ex
github.com/openshift/cakephp-ex has been moved to github.com/sclorg/cakephp-ex


bparees:
[merge]

thanks @liangxia

PR #927, opened 2018-09-19
add add-hosts

MagicSong:
Add a `--add-host` flag，same as `--add-host` in `docker run` . It is very useful when user in LAN or in dev environment.

### DIFF
--- a/hack/test-stirunimage.sh
+++ b/hack/test-stirunimage.sh
@@ -136,6 +136,12 @@ docker run test >& "${WORK_DIR}/s2i-override-run.log"
 grep "Running custom run" "${WORK_DIR}/s2i-override-run.log"
 check_result $? "${WORK_DIR}/s2i-override-run.log"
 
+test_debug "s2i build with add-host option"
+set +e
+s2i build https://github.com/openshift/ruby-hello-world centos/ruby-23-centos7 --add-host rubygems.org:0.0.0.0 test-ruby-app &> "${WORK_DIR}/s2i-add-host.log"
+grep "Gem::RemoteFetcher::FetchError: Errno::ECONNREFUSED" "${WORK_DIR}/s2i-add-host.log"
+check_result $? "${WORK_DIR}/s2i-add-host.log"
+set -e
 test_debug "s2i build with remote git repo"
 s2i build https://github.com/sclorg/cakephp-ex docker.io/centos/php-70-centos7 test --loglevel=5 &> "${WORK_DIR}/s2i-git-proto.log"
 check_result $? "${WORK_DIR}/s2i-git-proto.log"
--- a/pkg/cmd/cli/cmd/build.go
+++ b/pkg/cmd/cli/cmd/build.go
@@ -215,5 +215,6 @@ $ s2i build . centos/ruby-22-centos7 hello-world-app
 	buildCmd.Flags().StringVar(&(networkMode), "network", "", "Specify the default Docker Network name to be used in build process")
 	buildCmd.Flags().StringVarP(&(cfg.AsDockerfile), "as-dockerfile", "", "", "EXPERIMENTAL: Output a Dockerfile to this path instead of building a new image")
 	buildCmd.Flags().BoolVarP(&(cfg.KeepSymlinks), "keep-symlinks", "", false, "When using '--copy', copy symlinks as symlinks. Default behavior is to follow symlinks and copy files by content")
+	buildCmd.Flags().StringArrayVar(&cfg.AddHost, "add-host", []string{}, "Specify additional entries to add to the /etc/hosts in the assemble container, multiple --add-host can be used to add multiple entries")
 	return buildCmd
 }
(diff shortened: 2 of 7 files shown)


MagicSong:
fix #927

bparees:
please add a test that confirms the hosts are added to the container at assemble time (which i assume is your goal).

bparees:
[test]

bparees:
> example: s2i build https://github.com/openshift/ruby-hello-world centos/ruby-23-centos7 --add-host rubygems.org:172.0.0.1 test-ruby-app. With --add-host, this sample will fail because of "rubygems.org" has been led to local. Without --add-host, it will work as expected.

great, please add that scenario as a test to https://github.com/openshift/source-to-image/blob/master/hack/test-stirunimage.sh

so we can make sure it continues to work.

MagicSong:
@bparees The scenario has been added to `test-stirunimage.sh`. I also change `172.0.0.1` to `1.1.1.1` because it takes too long to wait the job to be failed. The job stops immediately when using public ip.

bparees:
thanks @MagicSong, can you squash your commits and i'll merge?

hack/test-stirunimage.sh
```diff
@@ -136,6 +136,11 @@ docker run test >& "${WORK_DIR}/s2i-override-run.log"
 grep "Running custom run" "${WORK_DIR}/s2i-override-run.log"
 check_result $? "${WORK_DIR}/s2i-override-run.log"
 
+test_debug "s2i build with add-host option"
+s2i build https://github.com/openshift/ruby-hello-world centos/ruby-23-centos7 --add-host rubygems.org:1.1.1.1 test-ruby-app &> "${WORK_DIR}/s2i-add-host.log"
+grep "Build failed" "${WORK_DIR}/s2i-add-host.log"
```
coreydaley: Can this check for a more specific failure reason related to the rubygems.org host not being found?

MagicSong: yes, but the error will change if 1.1.1.1 is down or close its http/https ports. I change the test scenario testing both cases----with and without --add-host flag---to make sure the failure is caused by --add-host flag

hack/test-stirunimage.sh
```diff
@@ -136,6 +136,12 @@ docker run test >& "${WORK_DIR}/s2i-override-run.log"
 grep "Running custom run" "${WORK_DIR}/s2i-override-run.log"
 check_result $? "${WORK_DIR}/s2i-override-run.log"
 
+test_debug "s2i build with add-host option"
```
coreydaley: I think that this should be two separate tests, less one give the wrong result and they both pass.
Also, it should give some kind of specific error if rubygems.org can not be found.

coreydaley: You could set rubygems.org:127.0.0.1 and it should fail (since localhost is not a rubygems server) and you should get an error message about rubygems.org not being found

MagicSong: i have change to 0.0.0.0 because error of 172.0.0.1 is much more unclear and it takes more time because it is the LOOPBACK internet interface. I also squash all my commits.  @bparees

bparees:
[merge]

hack/test-stirunimage.sh
```diff
@@ -136,6 +136,11 @@ docker run test >& "${WORK_DIR}/s2i-override-run.log"
...
+s2i build https://github.com/openshift/ruby-hello-world centos/ruby-23-centos7 --add-host rubygems.org:0.0.0.0 test-ruby-app &> "${WORK_DIR}/s2i-add-host.log"
```
bparees: since you expect this to fail you need an || true at the end, because we run our script with exit on error.

pkg/cmd/cli/cmd/build.go
```diff
@@ -215,5 +215,6 @@ $ s2i build . centos/ruby-22-centos7 hello-world-app
 	buildCmd.Flags().StringVar(&(networkMode), "network", "", "Specify the default Docker Network name to be used in build process")
 	buildCmd.Flags().StringVarP(&(cfg.AsDockerfile), "as-dockerfile", "", "", "EXPERIMENTAL: Output a Dockerfile to this path instead of building a new image")
 	buildCmd.Flags().BoolVarP(&(cfg.KeepSymlinks), "keep-symlinks", "", false, "When using '--copy', copy symlinks as symlinks. Default behavior is to follow symlinks and copy files by content")
+	buildCmd.Flags().StringArrayVar(&cfg.AddHost, "add-host", []string{}, "Specify additional entry added to /etc/hosts in container, multiple --add-host can be used to add multiple entry")
```
bparees: grammar.  should be:
"Specify additional entries to add to the /etc/hosts in the assemble container, multiple --add-host arguments can be used to add multiple entries"

hack/test-stirunimage.sh
MagicSong: ok, I added a `set +e` before run my test.

bparees:
[merge]

Issue #929, opened 2018-10-11
failure to restore cache produces corrupted files passed to incremental builds


asottile:
there appear to be two problems here, first that a valid tar is being rejected and second that the partially written files are still being passed into subsequent builds.

I've had difficulty finding a minimal testcase yet so I'll include my logs, observations, and in parallel work on trying to make a minimal reproduction.

The command we're using to build is:

```bash
s2i build \
    --pull-policy=never --incremental-pull-policy=never \
    --incremental --copy \
    "$DIRECTORY" "$BUILDER" "$TARGET_TAG"
```

The log outputs (I've prefixed our helper tool's output with `~~~ `)

```
Checking if image "$BUILDER" is available locally ...
Checking if image "$BUILDER" is available locally ...
Checking if image "$TARGET_TAG" is available locally ...
~~~ Saving pre-commit-cache...
~~~ --> sha256 of inputs adf5190953
ERROR: Error extracting tar stream
ERROR: Error extracting file "pre-commit-cache.tar": unexpected EOF
WARNING: Clean build will be performed because of error saving previous build artifacts
~~~ Attempting to restore pre-commit-cache...
~~~ --> sha256 of inputs adf5190953
~~~ tar: Skipping to next header
~~~ tar: Skipping to next header
~~~ tar: Skipping to next header
~~~ tar: Skipping to next header
~~~ tar: Skipping to next header
~~~ tar: Unexpected EOF in archive
~~~ tar: Error is not recoverable: exiting now
```

(see below for the structure of our artifacts, this error happens as it tries to untar `pre-commit-cache.tar` -- we're using tar-of-tar to work around the IO bottlenecks alluded to in #832)

If I manually run the `save-artifacts` myself it appears to succeed:

```
$ docker run --rm -ti 926f31bc68fb bash
lyftsvc@88d6197c4ad3:/code$ time /usr/local/s2i/save-artifacts > /tmp/artifact.tgz
Saving pre-commit-cache...
--> sha256 of inputs adf5190953

real	0m10.680s
user	0m0.152s
sys	0m5.376s
```

And that tarfile is fine:

```
$ tar --list -f /tmp/artifact.tgz 
pre-commit-cache.s4i.json
pre-commit-cache.tar
```

Is it possibly a file size thing -- the artifact involved is quite large:

```
$ du -hs /tmp/artifact.tgz 
678M	/tmp/artifact.tgz
```

[...]

PR #930, opened 2018-10-12
increase timeout on tar extraction and make error clearer

bparees:
fixes https://github.com/openshift/source-to-image/issues/929

### DIFF
--- a/pkg/tar/tar.go
+++ b/pkg/tar/tar.go
@@ -23,7 +23,7 @@ var glog = utilglog.StderrLog
 // defaultTimeout is the amount of time that the untar will wait for a tar
 // stream to extract a single file. A timeout is needed to guard against broken
 // connections in which it would wait for a long time to untar and nothing would happen
-const defaultTimeout = 30 * time.Second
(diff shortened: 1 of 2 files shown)


bparees:
[merge]

pkg/tar/tar.go
```diff
@@ -395,7 +395,7 @@ func (t *stiTar) ExtractTarStreamFromTarReader(dir string, tarReader Reader, log
 	})
 
 	if err != nil {
-		glog.Error("Error extracting tar stream")
+		glog.Error("Error extracting tar stream: %v", err)
```
asottile: the %v doesn't appear to do what you intend (note: I modified the timeout to hit this case in my clone)?

```
ERROR: Error extracting tar stream: %vfunction timed out after 30s
```

bparees: https://github.com/openshift/source-to-image/pull/932

asottile: dangit lol https://github.com/openshift/source-to-image/pull/933

PR #933, opened 2018-10-12
Format the tar error

asottile:


bparees:
already open as https://github.com/openshift/source-to-image/pull/932

Issue #934, opened 2018-10-13
S2I and Minikube Docker - Build Hangs


cliveseldon:
With the latest minikube 0.30.0, if one uses ```eval (minikube docker-env)``` to use the minikube docker daemon and then run an s2i build, the build hangs at:

```
I1013 08:23:57.590425    1449 sti.go:681] ---> Installing application source...
I1013 08:23:57.614246    1449 docker.go:1055] Waiting for container "81385c5e9f3e6071069208b535ebed99b251dc749b352239262f28118279dcd0" to stop ...
```

With 0.25.2  version of minikube this works:

```
I1013 08:11:34.572651   32515 sti.go:681] ---> Installing application source...
I1013 08:11:34.599362   32515 docker.go:1055] Waiting for container "1d8e9a0719134f4706cf0db0a81ca8f1c4001a8ef3c7042b07f03fae4498a61e" to stop ...
I1013 08:11:34.639881   32515 docker.go:1080] Invoking PostExecute function
I1013 08:11:34.639896   32515 postexecutorstep.go:67] Skipping step: store previous image
I1013 08:11:34.639901   32515 postexecutorstep.go:116] Executing step: commit image
I1013 08:11:34.642032   32515 postexecutorstep.go:521] Checking for new Labels to apply... 
I1013 08:11:34.642048   32515 postexecutorstep.go:529] Creating the download path '/tmp/s2i230013349/metadata'
I1013 08:11:34.642099   32515 postexecutorstep.go:463] Downloading file "/tmp/.s2i/image_metadata.json"
I1013 08:11:34.700951   32515 postexecutorstep.go:537] unable to download and extract 'image_metadata.json' ... continuing
[...]

PR #935, opened 2018-10-15
check for container termination, not state transition

bparees:
fixes https://github.com/openshift/source-to-image/issues/934


bparees:
[merge]

oshev:
Hi, when are you planning to release this? I have the same issue as @cliveseldon and ti blocks me. :(

bparees:
just published it in 1.1.13 for you, but of course you can always build your own binary from master.

oshev:
Amazing, thank you for the quick response! I did build my own binary from master but then everyone in our team would have to be forced to install Go (we don't use it) and do the same or I would need to distribute binaries for several OS... Besides, it's not only for us but for the everyone in Seldon community who use MiniKube. ;-) It takes time and skills to understand the bug, find, and apply the solution.

PR #940, opened 2018-10-30
increase timeout on tar extraction and make error clearer

bparees:


bparees:
[merge]

bparees:
[merge]

bparees:
looks like the windows CI has bit-rotted in a way i don't have time to investigate right now, manually merging.  Definitely not a failure caused by this change.

PR #945, opened 2018-12-29
Clarify that one feature of S2I is performance/speed.

ElijahLynn:
The gist of this PR is to communicate to users that S2I is a faster option than using `docker build`. Maybe this can be said a better way but hopefully this can get it started.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@
 [![Travis](https://travis-ci.org/openshift/source-to-image.svg?branch=master)](https://travis-ci.org/openshift/source-to-image)
 [![License](https://img.shields.io/github/license/openshift/source-to-image.svg)](https://www.apache.org/licenses/LICENSE-2.0.html)
 
-Source-to-Image (S2I) is a toolkit and workflow for building reproducible Docker images from source code. S2I produces
+Source-to-Image (S2I) is a toolkit and workflow for building reproducible Docker images from source code. One big advantage of this is that it is much faster than `docker build` if all you are needing is getting your source code into the image. S2I produces
 ready-to-run images by injecting source code into a Docker container and letting the container prepare that source code for execution. By creating self-assembling **builder images**, you can version and control your build environments exactly like you use Docker images to version your runtime environments.
 
 For a deep dive on S2I you can view [this presentation](https://www.youtube.com/watch?v=flI6zx9wH6M).


README.md
```diff
@@ -7,7 +7,7 @@
 [![Travis](https://travis-ci.org/openshift/source-to-image.svg?branch=master)](https://travis-ci.org/openshift/source-to-image)
 [![License](https://img.shields.io/github/license/openshift/source-to-image.svg)](https://www.apache.org/licenses/LICENSE-2.0.html)
 
-Source-to-Image (S2I) is a toolkit and workflow for building reproducible Docker images from source code. S2I produces
+Source-to-Image (S2I) is a toolkit and workflow for building reproducible Docker images from source code. One big advantage of this is that it is much faster than `docker build` if all you are needing is getting your source code into the image. S2I produces
```
bparees: this is already called out in the "goals" section:
https://github.com/openshift/source-to-image#goals

perhaps "goals" should be renamed to "benefits" or something?  but i don't think this is the right place to list one specific advantage away from the others.

ElijahLynn: Thanks, I put it up front because it seems like something to highlight as a primary feature and to give users the "why" very early on and I specifically searched the readme for `docker build`. I feel like the current "speed" section in Goals is being too humble. I get that there may be a reason to not specifically name `docker build` but I think that does the user a disservice if that is the case.

bparees: yes, if you're writing a dockerfile like:

FROM java
ADD . /src
RUN javac /src

you're a strong candidate for s2i (though if your dockerfile is that simple you also likely won't see speed benefits.  But if your dockerfile is also installing dependencies and such, you might, especially if you've written the dockerfile inefficiently).

Anyway i'd still prefer not to call out this one specific aspect, i'd rather raise the visibility of all the items in the "goals" section, either by teasing them in a bullet list in the overview, and/or by renaming the "goals" section to "advantages of s2i" or something.

and there's no necessary a problem with calling out "docker build" explicitly, but there are definitely numerous image building tools that consume dockerfiles at this point, so it's easier/more flexible to just refer to image building/dockerfile building, or "image building from dockerfile".

Note that the waters are also getting muddier as s2i can now generate a dockerfile you can build, but again it attempts to create an efficient dockerfile when doing so.

PR #963, opened 2019-05-14
Fixes to release process

adambkaplan:
- Use podman instead of docker
- Use UBI as base image
- Bump golang

### DIFF
--- a/images/release/Dockerfile
+++ b/images/release/Dockerfile
@@ -3,9 +3,9 @@
 #
 # The standard name for this image is openshift/sti-release
 #
-FROM openshift/origin-base
+FROM registry.redhat.io/ubi8/ubi
 
-ENV VERSION=1.7.4 \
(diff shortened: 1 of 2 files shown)


images/release/Dockerfile
```diff
@@ -3,9 +3,9 @@
 #
 # The standard name for this image is openshift/sti-release
 #
-FROM openshift/origin-base
+FROM registry.access.redhat.com/ubi8/ubi
```
bparees: registry.redhat.io?  (assuming you're only running this locally where you can expect to have proper creds).

otherwise this will break on you someday.

adambkaplan: went with the location published on Scott's announcement. I understand why that was done (no pull secret needed, lowers the barrier to entry).

adambkaplan:
[TEST]

bparees:
[merge]

PR #969, opened 2019-05-21
Bug 1712245: ConvertEnvironmentToDocker: escape "$", `\`, and `"`

nalind:
When values that are passed to ConvertEnvironmentToDocker() include the `$`, `\`, or `"` characters, escape them using a `\`.

I _think_ this will produce the same results when a passed-in value for the `-e` flag includes one or more of these characters, whether or not `--as-dockerfile` is used to generate an intermediate `Dockerfile` step, but could really use another pair of eyes to verify that.  I think this triggers bugs in ~~imagebuilder~~buildah as well, so any manual checks should probably use `docker build`.


adambkaplan:
@nalind please add a unit test for the escape logic

nalind:
Added a unit test.  Testing it properly pulls in some new dependencies, and updates a few others that we pinned to a range of versions.

pkg/scripts/environment_test.go
```diff
@@ -40,3 +42,38 @@ func equalArrayContents(a []string, b []string) bool {
...
+	}
+	var charValues []string
+	for ch := rune(32); ch < 127; ch++ {
+		charValues = append(charValues, "AB"+string(ch)+"CD")
+	}
+	for _, testValue := range append(testValues, charValues...) {
+		list := api.EnvironmentList{{Name: "TEST", Value: testValue}}
+		converted := ConvertEnvironmentToDocker(list)
+		builder := imagebuilder.NewBuilder(nil)
+		parsed, err := imagebuilder.ParseDockerfile(bytes.NewBufferString(converted + "\n"))
```
adambkaplan: @nalind I'm reluctant to add a dependency to `imagebuilder` when we also have a dep on docker/docker - especially for just a test.
Can you refactor to use https://github.com/openshift/moby-moby/blob/openshift-3.10-docker-b68221c/builder/dockerfile/parser/parser.go#L259 ?

nalind:
Switched to using the `dockerfile` package from the vendored version of `github.com/docker/docker`.  This seems to be more lax about whether or not we should escape `\` (the test passed whether `escape()` did so or not), but I left it out of the list of characters we escape because I know that openshift/builder is using imagebuilder, which would have produced an error if we started escaping `\`.

nalind:
Adding additional tests to suss out what docker's machinery expects turned up failures when we weren't escaping `\\`, so that's back in the list of characters that we escape.

glide.lock
```diff
@@ -1,71 +1,315 @@
...
 - name: github.com/docker/distribution
   version: edc3ab29cdff8694dd6feb85cfeb4b5f1b38ed9c
   subpackages:
+  - context
   - digestset
   - reference
+  - uuid
 - name: github.com/docker/docker
-  version: fcdab531a6d9b133eaa3f2a3f6c11fd02a97de75
+  version: da810a85109d7bd7ccd8f2d622556cbcfc00386b
```
adambkaplan: ok bump - @nalind adds in your SIGTERM patch.

glide.yaml
```diff
@@ -80,7 +80,7 @@ import:
   subpackages:
   - libcontainer/user
 - package: github.com/pkg/errors
-  version: a22138067af1c4942683050411a841ade67fe1eb
+  version: 839d9e913e063e28dfd0e6c7b7512793e0a48be9
```
adambkaplan: I assume this is to address conflicts with moby/buildkit?

glide.yaml
```diff
@@ -102,3 +102,6 @@ import:
   subpackages:
   - unix
   - windows
+- package: github.com/moby/buildkit
+  version: da2b9dc7dab99e824b2b1067ad7d0523e32dd2d9
+  repo: https://github.com/dmcgowan/buildkit.git
```
adambkaplan: Why are we using a fork instead of upstream?

pkg/scripts/environment_test.go
```diff
@@ -3,6 +3,8 @@ package scripts
 import (
 	"testing"
 
+	"github.com/docker/docker/api/types/container"
+	"github.com/docker/docker/builder/dockerfile"
```
adambkaplan: A bit confused why we need to pin our dependency on `moby/buildkit` when we don't reference it in the test.

glide.yaml
nalind: Without this, the test fails to compile with the error messages:
```
vendor/github.com/docker/docker/plugin/v2/plugin_linux.go:33:15: undefined: "github.com/openshift/source-to-image/vendor/github.com/pkg/errors".WithStack
vendor/github.com/docker/docker/plugin/v2/plugin_linux.go:105:16: undefined: "github.com/openshift/source-to-image/vendor/github.com/pkg/errors".WithStack
```

glide.yaml
nalind: It's the version that the docker fork specified.  When I didn't pin it, glide wasn't able to resolve compatible versions of its dependencies.

pkg/scripts/environment_test.go
nalind: It's a dependency of the dockerfile module.

glide.yaml
adambkaplan: @nalind please add a comment to the `glide.yaml` that these pins resolve dependencies in the `openshift/moby-moby` branch.

nalind: Added.

adambkaplan:
[TEST]

adambkaplan:
[MERGE]

bparees:
[merge]

PR #980, opened 2019-07-08
remove func GetStrategy()

soulseen:

func GetStrategy() is deprecated, use Strategy() instead.

### DIFF
--- a/pkg/build/strategies/strategies.go
+++ b/pkg/build/strategies/strategies.go
@@ -13,18 +13,13 @@ import (
 	utilstatus "github.com/openshift/source-to-image/pkg/util/status"
 )
 
-// GetStrategy decides what build strategy will be used for the STI build.
-// TODO: deprecated, use Strategy() instead
-func GetStrategy(client docker.Client, config *api.Config) (build.Builder, api.BuildInfo, error) {
-	return Strategy(client, config, build.Overrides{})
-}
-
 // Strategy creates the appropriate build strategy for the provided config, using
 // the overrides provided. Not all strategies support all overrides.
-func Strategy(client docker.Client, config *api.Config, overrides build.Overrides) (build.Builder, api.BuildInfo, error) {
+func Strategy(client docker.Client, config *api.Config) (build.Builder, api.BuildInfo, error) {
 	var builder build.Builder
 	var buildInfo api.BuildInfo
 	var err error
+	var overrides = build.Overrides{}
 
 	fileSystem := fs.NewFileSystem()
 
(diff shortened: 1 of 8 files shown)


adambkaplan:
/ok-to-test

Code is no longer referenced in openshift/builder, this is safe to merge with a subsequent minor version bump on the next release.

pkg/build/strategies/strategies.go
```diff
@@ -13,18 +13,13 @@ import (
...
-// GetStrategy decides what build strategy will be used for the STI build.
-// TODO: deprecated, use Strategy() instead
-func GetStrategy(client docker.Client, config *api.Config) (build.Builder, api.BuildInfo, error) {
-	return Strategy(client, config, build.Overrides{})
-}
-
 // Strategy creates the appropriate build strategy for the provided config, using
 // the overrides provided. Not all strategies support all overrides.
-func Strategy(client docker.Client, config *api.Config, overrides build.Overrides) (build.Builder, api.BuildInfo, error) {
+func Strategy(client docker.Client, config *api.Config) (build.Builder, api.BuildInfo, error) {
```
adambkaplan: Please revert - this will break openshift/builder

pkg/build/strategies/strategies.go
```diff
@@ -13,18 +13,13 @@ import (
...
 	var builder build.Builder
 	var buildInfo api.BuildInfo
 	var err error
+	var overrides build.Overrides
```
adambkaplan: Ditto

pkg/cmd/cli/cmd/build.go
```diff
@@ -162,7 +162,7 @@ $ s2i build . centos/ruby-22-centos7 hello-world-app
 
 			glog.V(2).Infof("\n%s\n", describe.Config(client, cfg))
 
-			builder, _, err := strategies.GetStrategy(client, cfg)
+			builder, _, err := strategies.Strategy(client, cfg)
```
adambkaplan: Pass an empty `build.Overrides` for all invocations

test/integration/docker/integration_test.go
```diff
@@ -298,7 +298,7 @@ func (i *integrationTest) exerciseCleanAllowedUIDsBuild(tag, imageName string, e
 		ExcludeRegExp:     tar.DefaultExclusionPattern.String(),
 	}
 	config.AllowedUIDs.Set("1-")
-	_, _, err := strategies.GetStrategy(engineClient, config)
+	_, _, err := strategies.Strategy(engineClient, config)
```
adambkaplan: Pass empty `build.Overrides`

pkg/build/strategies/strategies.go
soulseen: How about define an empty build.Overrides{} in this function?

adambkaplan: This function is referenced by another project - changing signature breaks the API.

PR #982, opened 2019-07-12
Move from glog to klog

adambkaplan:
glog is no longer being actively maintained, and conflicts with klog used in downstream projects (openshift/builder in particular).

/assign @coreydaley 

/cc @bparees


adambkaplan:
/assign @bparees 

Note that we may need a refresh to get the docker-based integration tests running. CI is broken due to an invalid cert.

bparees:
you should also rename refs like this:
`var glog = utilglog.StderrLog`

and for that matter rename `github.com/openshift/source-to-image/pkg/util/glog` itself.

pkg/cmd/cli/cli.go
```diff
@@ -16,16 +15,12 @@ import (
...
+// log is a placeholder until the builders pass an output stream down
+// client facing libraries should not be using log
+var log = utillog.StderrLog
 
 // NewCmdCLI implements the S2I command line functionality.
 func NewCmdCLI() *cobra.Command {
-	// Applying partial glog flag initialization workaround from: https://github.com/kubernetes/kubernetes/issues/17162
-	// Without this fake command line parse, glog will compain its flags have not been interpreted
-	flag.CommandLine.Parse([]string{})
-
```
adambkaplan: Don't need this any more with klog

PR #984, opened 2019-07-20
Add OS-agnostic path style in files to ignore list

itscaro:
This is because in other functions path.Join() is used. This patch corrects a bug on Windows, a better solution could be implemented here. Since I do not fully know this project, I propose here a solution which impacts least.

Signed-off-by: Minh-Quan TRAN <account@itscaro.me>

### DIFF
--- a/pkg/ignore/ignore.go
+++ b/pkg/ignore/ignore.go
@@ -5,6 +5,7 @@ import (
 	"io"
 	"os"
 	"path/filepath"
+	"runtime"
 	"strings"
 
 	"github.com/openshift/source-to-image/pkg/api"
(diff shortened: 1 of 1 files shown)


itscaro:
z

adambkaplan:
/ok-to-test

Unfortunately we don't have CI for windows any more, so we can't verify that this PR doesn't break the cli on Windows.

I'd also rather us use `filepath.FromSlash`: https://golang.org/pkg/path/filepath/#FromSlash

itscaro:
> /ok-to-test
> 
> Unfortunately we don't have CI for windows any more, so we can't verify that this PR doesn't break the cli on Windows.
> 
> I'd also rather us use `filepath.FromSlash`: https://golang.org/pkg/path/filepath/#FromSlash

This PR converts Separators into slash because in fs.go path.join is used which uses slash instead of separator, this causes the mismatch in filesToIgnore list on windows.

PR #986, opened 2019-07-28
remove func GetStrategy()

soulseen:


.gitignore
```diff
@@ -4,4 +4,5 @@
 /.vagrant
 .vscode/
 cmd/s2i/debug
+.idea/
```
coreydaley: If you are using intellij products to do development, you should really put the .idea in your global gitignore file, as it's not a project specific file/folder that is being ignored.

https://help.github.com/en/articles/ignoring-files#create-a-global-gitignore

coreydaley:
looks like you need to run `make verify` and fix any issues that it shows, probably with `gofmt -w pkg test `

.gitignore
soulseen: ok

PR #989, opened 2019-08-20
fix get SourceInfo

soulseen:


bparees:
@soulseen what issue is this PR fixing?  can you describe the scenario where the current code is not behaving as desired/expected?

PR #994, opened 2019-09-14
Go Modules

otaviof:
Migrating `s2i` to use Go Modules. Documentation and scripts are updated to reflect Go Modules usage.

Packages described in Glide are migrated to `go.mod`/`go.sum` files using exactly the same versions, however, since Go Modules are managing packages in a sightly different way, it has tainted vendor directory as well.


sbose78:
Good to have this use modules from 1.12 !
Did you run `go mod vendor`?

otaviof:
> Good to have this use modules from 1.12 !
> Did you run `go mod vendor`?

Indeed :-) And yes, I have ran `go mod vendor`!

DhritiShikhar:
@otaviof Adding a comment here as well.

These two files should be removed:

[1] https://github.com/openshift/source-to-image/blob/master/hack/godep-restore.sh
[2] https://github.com/openshift/source-to-image/blob/master/hack/godep-save.sh

DhritiShikhar:
I can't refer to the exact diff in the github UI because of too many file changes. So, I am posting here.

I saw these godep references locally which should be removed:

[1] https://github.com/openshift/source-to-image/blob/master/hack/install-tools.sh#L11
[2] https://github.com/openshift/source-to-image/blob/master/hack/common.sh#L211
[3] https://github.com/openshift/source-to-image/blob/master/hack/rebase-to-origin.sh#L62

adambkaplan:
/retitle WIP - Go Modules

This is great stuff - openshift/builder will need this for 4.3 when it too switches to use go modules

otaviof:
@DhritiShikhar just removed the scripts you've listed, please consider. Thanks for pointing them out :-)

otaviof:
@adambkaplan What's your take on removing `hack/rebase-to-origin.sh`? I could not identify where it's being used.

DhritiShikhar:
Looks good @otaviof

adambkaplan:
> What's your take on removing `hack/rebase-to-origin.sh`? I could not identify where it's being used.

Pretty sure that is obsolete - we can drop it.

adambkaplan:
Echoing a conversation I had with @otaviof on Slack:

We try to adhere to OpenShift's conventions when it comes to updating dependencies. In general the rule is that an "update deps" PR should have up to 3 commits:

1. Commit titled `bump(*):`, with a note detailing the reason to update the dependencies. Only changes here should be related to code in the `vendor/` tree and any change to files that define the dependencies (ex: `glide.yaml`).
2. (optional) Updates to generated code, if any. Typically marked as `generated`
3. (optional) An additional commit if any authored code needed to be changed to compile/run properly. In some cases this may include generated code.

This PR is a bit more complex since we're changing the machinery that updates dependencies. I would squash and isolate the changes as follows:

1. Commit with the changes to set up `go mod`
2. `bump(*):` commit with actual dep updates
3. Any "downstream" changes that needed to be made after deps were updated.

otaviof:
Thanks for the pointers, @adambkaplan. Changes applied, please consider it again :-)

sbose78:
@otaviof , can you remove the WIP ?

CONTRIBUTING.md
```diff
@@ -87,17 +87,18 @@ Run the integration tests with:
...
+[Go Modules](https://github.com/golang/go/wiki/Modules#how-to-use-modules).
 
-    $ go get github.com/Masterminds/glide
-
-If you are not updating packages you should not need glide installed.
+The basic usage of `go mod` in this project is:
 
+    $ go mod vendor
+    $ go mod verify
+    $ go mod tidy -v
```
adambkaplan: Shouldn't the order be

1. `go mod tidy -v`
2. `go mod vendor`
3. `go mod verify`

.gitignore
```diff
@@ -4,4 +4,6 @@
 /.vagrant
 .vscode/
 cmd/s2i/debug
+contrib/completions/bash
+contrib/completions/zsh
```
adambkaplan: why are we ignoring these dirs? Code is still there, so we can't .gitignore without deleting them as well.

hack/update-deps.sh
```diff
@@ -6,4 +6,8 @@ set -o pipefail
 
 cd "$(dirname "${BASH_SOURCE}")/.."
 
-glide update --strip-vendor
+# https://github.com/golang/go/wiki/Modules#how-to-upgrade-and-downgrade-dependencies
+#
+# Updating all direct and indirect dependencies to latest minor or patch upgrades, where pre-releases
+# are ignored. To list possible updates before installing, execute: "$ go list -u -m all"
+go get -u=patch
```
adambkaplan: I'm not as big a fan of this since updating deps can get hairy with go mod. We should probably drop this script.

I'd rather go with the approach of updating as per the wiki:

>  A common mistake is thinking go get -u foo solely gets the latest version of foo. In actuality, the -u in go get -u foo or go get -u foo@latest means to also get the latest versions for all of the direct and indirect dependencies of foo. A common starting point when upgrading foo is instead to do go get foo or go get foo@latest without a -u (and after things are working, consider go get -u=patch foo, go get -u=patch, go get -u foo, or go get -u).

.travis.yml
```diff
@@ -1,7 +1,7 @@
 language: go
 
 go:
-  - 1.12.x
+  - 1.13.x
```
adambkaplan: Why can't 1.12 work with go.mod?

go.mod
```diff
@@ -0,0 +1,92 @@
+module github.com/openshift/source-to-image
+
+go 1.13
+
+replace (
+	github.com/containerd/containerd => github.com/containerd/containerd v0.2.10-0.20170808145631-06b9cb351610
+	github.com/docker/docker => github.com/openshift/moby-moby v1.4.2-0.20190308215630-da810a85109d
+	github.com/moby/buildkit => github.com/dmcgowan/buildkit v0.0.0-20170731200553-da2b9dc7dab9
+	google.golang.org/genproto => google.golang.org/genproto v0.0.0-20190708153700-3bdd9d9f5532
+	google.golang.org/grpc => google.golang.org/grpc v1.3.0
```
adambkaplan: Why replace with v1.3.0?

.travis.yml
isutton: More a question than a comment: what is the policy for Go compilers when building OpenShift artifacts?

adambkaplan: Our CI system has imagestream tags which list supported golang versions. I believe 1.12 is the highest version we currently support (following upstream k8s).

.gitignore
otaviof: I've added during running the tests, it's a left over to be removed from this PR. Thanks.

hack/update-deps.sh
otaviof: Good suggestion on this matter, the best course of action is to handle dependencies one by one.

go.mod
otaviof: This is the tag corresponding to the commit-id used in `glide.lock`:

```
$ cd ${GOPATH}/src/google.golang.org/grpc
$ git name-rev --tags --name-only d2e1b51f33ff8c5e4a15560ff049d200e83726c5
v1.3.0
```

CONTRIBUTING.md
otaviof: Indeed, sequence wise it makes more sense what you suggest. Thanks.

.travis.yml
otaviof: @adambkaplan it can work with `1.12`, and since that's the current version in other systems it reverting this change from my PR.

sbose78:
@adambkaplan @gabemontero Does this PR look good now? :)

gabemontero:
I'll defer to @adambkaplan ... it is his call @sbose78

go.mod
```diff
@@ -0,0 +1,92 @@
+module github.com/openshift/source-to-image
+
+go 1.13
```
adambkaplan: Should be `go 1.12` to align with what we build

otaviof: @adambkaplan that's auto-generated, it depends in the local Golang version.

adambkaplan:
/lgtm

/hold

We're in the process of adding a prior commit to openshift/builder. Once that is in we can land this safely.

sbose78:
Bump.

adambkaplan:
/hold cancel

Bumps landed 😄

sbose78:
Haha! Thank you Adam!

Issue #973, opened 2019-06-03
--as-dockerfile doesn't read builder image labels labels


vdemeester:
It seems that `s2i build --as-dockerfile` doesn't read the builder image labels when generating the `Dockerfile`, especially the `io.openshift.s2i.*` ones.

As an example, see this [gist](https://gist.github.com/siamaksade/ae0b33c5b82b16cfe2dff5a46adb667f). The image use in there `registry.access.redhat.com/redhat-openjdk-18/openjdk18-openshift`.

```
skopeo inspect docker://registry.access.redhat.com/redhat-openjdk-18/openjdk18-openshift | grep io.openshift.s2i
        "io.openshift.s2i.destination": "/tmp",
        "io.openshift.s2i.scripts-url": "image:///usr/local/s2i",
```

When trying to build the generated `Dockerfile` from this builder, it fails as it is looking at it tries to execute a `RUN` on `/usr/libexec/s2i/assemble`, which doesn't exists in that image.

```
STEP 6: USER 1001
--> 496bfb3d9130701e154304c2b444788fc6110780b26adc71620315b3d440a8b2
STEP 7: RUN /usr/libexec/s2i/assemble
/bin/sh: /usr/libexec/s2i/assemble: No such file or directory
subprocess exited with status 127
subprocess exited with status 127
error building at STEP "RUN /usr/libexec/s2i/assemble": exit status 127
```

Related to https://github.com/openshift/source-to-image/issues/966#issuecomment-495169765.

cc @siamaksade @rhuss

rhuss:
Looking into the source at https://github.com/openshift/source-to-image/blob/3bdb78d004cc1f012b0eb4cb8aca41273bb6e6cf/pkg/build/strategies/dockerfile/dockerfile.go#L415-L423

it's really only about the options `--scripts-url` and `--image-scripts-url` (with a fallback to `/usr/libexec/s2i/`)

In fact, it would be super awesome if this detection could be done automatically as all information is already there. No need to bug the user to know where within the builder image the s2i directory is located.

bparees:
The challenge is how to discover the information.  The intent of the --as-dockerfile invocation is that we have no reliance on any specific tool/library that would be used to inspect the image in question, thus we can't look at the labels to determine this information.  (If we assume the presence of a docker daemon and use that to inspect the image, we might as well just run s2i in the traditional mode where it builds the whole image for you).

vdemeester:
[...]

PR #996, opened 2019-09-26
Execute External Build Command

otaviof:
Extending `s2i build` command to build images using an external command, in this PR we are adding `buildah`, `docker` and `podman`. Therefore, users can run:

```bash
s2i build app-src nginx:latest app-dst --with-builder buildah
s2i build app-src nginx:latest app-dst --with-builder docker
s2i build app-src nginx:latest app-dst --with-builder podman
```

The idea is to improve user experience to in a single action, generate the `Dockerfile` and use a external command to build the container image, therefore the build environment would allow to bring OCI toolset.


sbose78:
Good work, let's ensure that this is tested without the docker socket active on the host.

otaviof:
> Good work, let's ensure that this is tested without the docker socket active on the host.

Indeed, without Docker socket exposed it was not able to reach the point of invoking the external builder. I've made a few more changes, making sure it's only calling `CreateDockerfile` instead of `Build`, and the Docker integration is not enabled either when using `--as-dockerfile` or `--with-builder` options.

sbose78:
If a user uses,

```
s2i build app-src nginx:latest app-dst --with-builder docker
```

Let's ensure we don't return an error. 

It could internally take any of the following code paths:

- invoke the docker API ( as it does today ). This touch less code.
- invoke the external command. I see you have https://github.com/openshift/source-to-image/pull/996/files#diff-e4469a02d9eb983856b288f828175cb8R35  :)

sbose78:
1. Turn off docker daemon
2. user does an `s2i build --with-builder=buildah`
3. `s2i rebuild --with-builder=buildah`

This would fail because (2) does a Pull Image using the docker API

https://github.com/openshift/source-to-image/pull/996/https://github.com/openshift/source-to-image/blob/0b950e4c21af8838c0f523e77114a19b061deaba/pkg/api/types.go#L48
```
                         client, err := docker.NewEngineAPIClient(cfg.DockerConfig)
			s2ierr.CheckError(err)
			dkr := docker.New(client, cfg.PullAuthentication)
			pr, err := docker.GetRebuildImage(dkr, cfg)
			s2ierr.CheckError(err)
			err = build.GenerateConfigFromLabels(cfg, pr)
			s2ierr.CheckError(err)
```

sbose78:
Another thing: 

If `s2i build --with-builder=docker` takes the route of using the external docker command instead of the existing docker API route, then https://github.com/openshift/source-to-image/issues/973 gets exposed here. 

`s2i build ...` and `s2i build --with-builder=docker` should not be taking two different code paths. That could cause inconsistent behavior.

In light of the above, my preference would be to invest in :

1. Refactor all workflows to make all invocations using external tools ( docker / podman / buildah). When the `--with-builder` is not specified, we use `docker build ..` for now.
2. Fix #973 as part of that.

otaviof:
> If a user uses,
> 
> ```
> s2i build app-src nginx:latest app-dst --with-builder docker
> ```
> 
> Let's ensure we don't return an error.
> 
> It could internally take any of the following code paths:
> 
>     * invoke the docker API ( as it does today ). This touches less code.
> 
>     * invoke the external command. I see you have https://github.com/openshift/source-to-image/pull/996/files#diff-e4469a02d9eb983856b288f828175cb8R35  :)

@sbose78 Please consider latest commit. Now it support `--with-builder=docker` triggering the original path, so using Docker API.

sbose78:
Summarizing the conversation on slack
https://coreos.slack.com/archives/CHK22LM6Y/p1569950216332100?thread_ts=1569948800.330600&cid=CHK22LM6Y

➡️  s2i build --with-builder=docker. We invoke external tool. That's what is being done now in the PR.
➡️  Let's not support --with-builder in s2i rebuild ...., it's going to bring up other issues with image inspection. rebuild would be a deprecated feature in the next minor release.
➡️  Removing all docker API calls in valuable codepaths isn't a priority but a nice goal to have.

sbose78:
> @sbose78 Please consider latest commit. Now it support --with-builder=docker triggering the original path, so using Docker API.

Not that I don't like this ;) I like it very much. However, this is opposite to what we summarized the other day :) It is fine to fall back to your older approach of calling external `docker` since we are no longer going to invest in `rebuild`.

> ➡️ s2i build --with-builder=docker. We invoke external tool. That's what is being done now in the PR.

pkg/build/strategies/strategies.go
```diff
@@ -24,6 +25,18 @@ func Strategy(client docker.Client, config *api.Config, overrides build.Override
...
+			buildInfo.FailureReason = utilstatus.NewFailureReason(
+				utilstatus.ReasonGenericS2IBuildFailed,
+				utilstatus.ReasonMessageGenericS2iBuildFailed,
+			)
+			return nil, buildInfo, err
+		}
+		return builder, buildInfo, nil
+	}
+
 	if len(config.AsDockerfile) != 0 {
```
sbose78: So, if I did a 
```
s2i build --as-dockerfile=path/to/write/dockerfile --with-builder=podman
```
first the control enters

```
if len(config.WithBuilder) != 0 {

}
```
and then the control again enters
```
if len(config.AsDockerfile) != 0 {

}
```

otaviof:
> Summarizing the conversation on slack
> https://coreos.slack.com/archives/CHK22LM6Y/p1569950216332100?thread_ts=1569948800.330600&cid=CHK22LM6Y
> 
> ➡️ s2i build --with-builder=docker. We invoke external tool. That's what is being done now in the PR.

Done. Reverting the changes we had before, where on using `--with-builder` you always invoke a external tool, including `docker` in this case.

> ➡️ Let's not support --with-builder in s2i rebuild ...., it's going to bring up other issues with image inspection. rebuild would be a deprecated feature in the next minor release.

Okay. The parameter `--with-builder` is not introduced over there, so it's "done".

> ➡️ Removing all docker API calls in valuable codepaths isn't a priority but a nice goal to have.

This topic I would like to tackle in a different PR.

pkg/build/strategies/strategies.go
otaviof: In this case, we would like to short-circuit it first on `--with-builder`. By using this option we will generate the `Dockerfile` on behalf of the user, and thus `if len(config.WithBuilder) != 0 {` will return the builder.

Notice if option `--as-dockerfile` is informed without `--with-builder` it works like before (like in `master` branch).

sbose78:
Sounds great, @otaviof .

pkg/build/strategies/strategies.go
sbose78: I understand. 
Would the dockerfile generation happen twice in this flow?

otaviof: Nope, since each of those `if`s are executing `return`.

sbose78: Ah right, I missed that! thank you.

sbose78:
@otaviof could you move this out of `draft` ?

otaviof:
> @otaviof could you move this out of `draft` ?

Sure! :-)

pkg/build/strategies/external/external.go
```diff
@@ -0,0 +1,154 @@
...
+	builders := []string{}
+	for k := range commands {
+		builders = append(builders, k)
+	}
+	sort.Strings(builders)
+	return builders
+}
+
+// ValidBuilderName returns a boolean based in keys of global commands map.
+func ValidBuilderName(name string) bool {
```
sbose78: Thinking out aloud, I would have implemented it this way :) 

```
# external_builder.go

type ExternalBuilder interface{
         Build(config *api.Config) (*api.Result, error)
}


# buildah_builder.go


// validate at compile type if interface contract is met
var _ BuildahBuilder = (ExternalBuilder)(nil)

type BuildahBuilder struct{
	dockerfile *dockerfile.Dockerfile
}

func NewBuildahBuilder(config *api.Config){
	// initialize everything needed from config
}

func (*BuildahBuilder) Build() (*api.Result, error) {
	// use `buildah bud --tag {{ .Tag }} --file {{ .AsDockerfile }} {{ or .WorkingDir "." }}`
}


```

Avni-Sharma:
looks good to me :)

sbose78:
Validate that --assemble-user value is passed on.

sbose78:
Hold off on inspecting the image to get the  --assemble-user option.

sbose78:
Do not depend on docker socket for inspecting image.

otaviof:
> Validate that --assemble-user value is passed on.

@adambkaplan, @sbose78, the behavior regarding `--assemble-user` is kept the same. In the new external builder we use original `CreateDockerfile` at https://github.com/otaviof/source-to-image/blob/issue-966-with-builder/pkg/build/strategies/external/external.go#L139 .

sbose78:
> Validate that --assemble-user value is passed on.

https://github.com/otaviof/source-to-image/blob/issue-966-with-builder/pkg/build/strategies/dockerfile/dockerfile.go#L122

```
       // run remaining commands as the image user
	if len(imageUser) > 0 && imageUser != "root" {
		buffer.WriteString(fmt.Sprintf("USER %s\n", imageUser))
	}
```

Looks like we are good.

pkg/build/strategies/external/external.go
```diff
@@ -0,0 +1,154 @@
...
+func (e *External) Build(config *api.Config) (*api.Result, error) {
+	config.AsDockerfile = e.asDockerfile(config)
+
+	externalCommand, err := e.renderCommand(config)
+	if err != nil {
+		return nil, err
+	}
+
+	// generating dockerfile following AsDockerfile directive
+	err = e.dockerfile.CreateDockerfile(config)
```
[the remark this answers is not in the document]
otaviof: > The more I think about it, why are we not using e.dockerfile.Build here? 

@adambkaplan using `e.dockerfile.Build` would require Docker, right? The idea is to have a sidekick of `--as-dockerfile`.

> There is a whole host of use cases which occur to assemble the dockerfile (script download, injections, etc.). It may mean that all builds run in a temporary directory, and not in the user's current directory containing the source code.

Okay, understood. In this PR we are using [`config.WorkingDir`](https://github.com/otaviof/source-to-image/blob/f75c941c92b0c02e98ff794c6be1ff77b6c762eb/pkg/build/strategies/external/external.go#L33) as the context directory, would that be applicable considering the location of the source-code and temporary directory?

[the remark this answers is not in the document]
sbose78: @otaviof & @adambkaplan could we resolve this, please, soon? :) We should be taking this PR in, and making it part of the productized image.

DhritiShikhar:
Fails for me at step 4.
Before COPY, we should ensure that the directory exists.

```
➜  source-to-image git:(external-build-command) ./_output/local/bin/linux/amd64/s2i build app-src nginx:latest app-dst --with-builder buildah              
WARNING: Unable to access directory "upload/scripts": open upload/scripts: no such file or directory
Executing external build command: 'buildah bud --tag app-dst --file Dockerfile.s2i .'
STEP 1: FROM nginx:latest
Getting image source signatures
Copying blob eb22865337de done
Copying blob bee5d581ef8b done
Copying blob 000eee12ec04 done
Copying config 231d40e811 done
Writing manifest to image destination
Storing signatures
STEP 2: LABEL "io.openshift.s2i.build.image"="nginx:latest"       "io.k8s.display-name"="app-dst"
STEP 3: USER root
STEP 4: COPY upload/src /tmp/src
error building at STEP "COPY upload/src /tmp/src": no files found matching "/home/dhritishikhar/go/src/github.com/openshift/source-to-image/upload/src": no such file or directory
ERRO exit status 1                                
External command return-code: 1
Build failed
ERROR: An error occurred: exit status 1

➜  source-to-image git:(external-build-command) ✗ ./_output/local/bin/linux/amd64/s2i build app-src nginx:latest app-dst --with-builder docker
WARNING: Unable to access directory "upload/scripts": open upload/scripts: no such file or directory
Executing external build command: 'docker build --tag app-dst --file Dockerfile.s2i .'
Sending build context to Docker daemon  37.77MB
Step 1/7 : FROM nginx:latest
 ---> 231d40e811cd
Step 2/7 : LABEL "io.k8s.display-name"="app-dst"       "io.openshift.s2i.build.image"="nginx:latest"
 ---> Running in 7a59ae444bb8
Removing intermediate container 7a59ae444bb8
 ---> 0df0a7c9dafc
Step 3/7 : USER root
 ---> Running in f11cf45bb3d8
Removing intermediate container f11cf45bb3d8
 ---> 5c6513cbe678
Step 4/7 : COPY upload/src /tmp/src
COPY failed: stat /var/lib/docker/tmp/docker-builder972331229/upload/src: no such file or directory
External command return-code: 1
Build failed
ERROR: An error occurred: exit status 1

➜  source-to-image git:(external-build-command) ✗ ./_output/local/bin/linux/amd64/s2i build app-src nginx:latest app-dst --with-builder podman
WARNING: Unable to access directory "upload/scripts": open upload/scripts: no such file or directory
Executing external build command: 'podman build --tag app-dst --file Dockerfile.s2i .'
STEP 1: FROM nginx:latest
STEP 2: LABEL "io.k8s.display-name"="app-dst"       "io.openshift.s2i.build.image"="nginx:latest"
2391974afb1f26d8ea0ccc75d42acf275e861bfd92757b40ad42425344ccf86c
STEP 3: USER root
4226006dcf997e4074f07cf242e08c49425edbb351601f456d99dd5a2fe9cfd4
STEP 4: COPY upload/src /tmp/src
Error: error building at STEP "COPY upload/src /tmp/src": no files found matching "/home/dhritishikhar/go/src/github.com/openshift/source-to-image/upload/src": no such file or directory
External command return-code: 125
Build failed
ERROR: An error occurred: exit status 125
➜  source-to-image git:(external-build-command) ✗ 
```

pkg/build/strategies/external/external.go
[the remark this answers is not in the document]
adambkaplan: @otaviof `e.dockerfile.Build` does not require a docker socket - that flow is invoked in OCP 4 builds.

pkg/cmd/cli/cmd/build.go
```diff
@@ -218,6 +225,7 @@ $ s2i build . centos/ruby-22-centos7 hello-world-app
 	buildCmd.Flags().VarP(&(cfg.RuntimeArtifacts), "runtime-artifact", "a", "Specify a file or directory to be copied from the builder to the runtime image")
 	buildCmd.Flags().StringVar(&(networkMode), "network", "", "Specify the default Docker Network name to be used in build process")
 	buildCmd.Flags().StringVarP(&(cfg.AsDockerfile), "as-dockerfile", "", "", "EXPERIMENTAL: Output a Dockerfile to this path instead of building a new image")
```
adambkaplan: Separately we can graduate `--as-dockerfile` from alpha to GA.

pkg/build/strategies/external/external.go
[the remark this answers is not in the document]
otaviof: @adambkaplan check, thanks for the information. I think for the feature intended on this PR we can use only `CreateDockerfile`, right?

adambkaplan:
@otaviof this needs an integration test in the `test/integration/docker` directory. Golang tests here run in a VM that has a docker binary and socket, so you should be able to invoke the new strategy and run a build end to end.

pkg/build/strategies/external/external.go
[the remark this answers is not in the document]
otaviof: @adambkaplan indeed, you're correct on this. Changing the implementation to build `Build` instead. Thanks!

otaviof:
> @otaviof this needs an integration test in the `test/integration/docker` directory. Golang tests here run in a VM that has a docker binary and socket, so you should be able to invoke the new strategy and run a build end to end.

Please consider latest commit.

pkg/cmd/cli/cmd/build.go
sbose78: Do you recommend this be done before we cut a release of 1.3 ?

sbose78:
@adambkaplan , wondering what's causing the build failure ?

adambkaplan:
@sbose78 looks like go.mod is trying to download dependencies in the `dockerfile` test suite. Need to see which version of golang we are running - we should only use dependencies that are vendored in.

adambkaplan:
Opened https://github.com/openshift/release/pull/6983 to hopefully fix the CI issue encountered here.

otaviof:
@adambkaplan, @sbose78 the build failures are addressed, please consider.

pradeepto:
@sbose78 @adambkaplan @otaviof What is pending for this PR to be merged? There is no action on this since last 14 days. Can one of you please have a look and unblock this? Thanks.

test/integration/docker/with_builder_test.go
```diff
@@ -0,0 +1,39 @@
...
+		Tag:          "test:tag",
+		WithBuilder:  "docker",
+	}
+
+	// simple probe of dockerfile, it's already extensively tested
+	expected := []string{"^FROM.*"}
+	// expect to find generated Dockerfile with s2i extension
+	expectedFiles := []string{filepath.Join(tempdir, "Dockerfile.s2i")}
+
+	dockerfile.RunDockerfileTest(t, config, expected, nil, expectedFiles, false)
```
adambkaplan: @otaviof why run the `Dockerfile` test here? Since this is running on our Jeninks VM, why not run an actual Docker build (just like the other docker-based tests)?

otaviof: Indeed, @adambkaplan, the integration test is on the wrong location. Thanks!

otaviof: In fact, it does run the actual Docker file build, but via the `--with-builder` external builder command-line. Therefore, we `os/exec` this builder command and wait for the outcome.

adambkaplan:
/lifecycle frozen

We will want to resurrect this PR once we are ready to implement the node cache feature.

coreydaley:
@otaviof Is this pull request still relevant?

otaviof:
> @otaviof Is this pull request still relevant?

Definitely not relevant. I'm closing it.

PR #999, opened 2019-10-24
Honor builder image labels when --as-dockerfile is enabled

isutton:
This PR consults the builder image labels to compose the resulting Dockerfile.

Similar to what happens in the `rebuild` command, the builder image metadata will be read and relevant labels will be used when composing the resulting Dockerfile.

If `--as-dockerfile` is expected to be used in environments where a Docker socket is not available, then this PR might break that expectation.

### DIFF
--- a/pkg/build/config.go
+++ b/pkg/build/config.go
@@ -12,12 +12,12 @@ import (
 
 // GenerateConfigFromLabels generates the S2I Config struct from the Docker
 // image labels.
-func GenerateConfigFromLabels(config *api.Config, metadata *docker.PullResult) error {
+func GenerateConfigFromApplicationImageLabels(config *api.Config, metadata *docker.PullResult) error {
 	if config == nil {
-		return errors.New("config must be provided to GenerateConfigFromLabels")
+		return errors.New("config must be provided to GenerateConfigFromApplicationImageLabels")
 	}
 	if metadata == nil {
-		return errors.New("image metadata must be provided to GenerateConfigFromLabels")
+		return errors.New("image metadata must be provided to GenerateConfigFromApplicationImageLabels")
 	}
 
 	labels := metadata.Image.Config.Labels
--- a/pkg/build/config.go
+++ b/pkg/build/config.go
@@ -10,7 +10,20 @@ import (
 	"github.com/openshift/source-to-image/pkg/scm/git"
 )
 
-// GenerateConfigFromLabels generates the S2I Config struct from the Docker
+// GenerateConfigFromBuilderImageLabels generates the S@I Config struct from the Docker
+// image labels.
+func GenerateConfigFromBuilderImageLabels(config *api.Config, metadata *docker.PullResult) error {
+	if config == nil {
+		return errors.New("config must be provided to GenerateConfigFromBuilderImageLabels")
+	}
+	if metadata == nil {
+		return errors.New("image metadata must be provided to GenerateConfigFromBuilderImageLabels")
+	}
+
+	return nil
+}
+
+// GenerateConfigFromApplicationImageLabels generates the S2I Config struct from the Docker
 // image labels.
 func GenerateConfigFromApplicationImageLabels(config *api.Config, metadata *docker.PullResult) error {
 	if config == nil {
--- a/pkg/api/describe/describer.go
+++ b/pkg/api/describe/describer.go
@@ -108,7 +108,7 @@ func describeBuilderImage(client docker.Client, config *api.Config, out io.Write
 	dkr := docker.New(client, c.PullAuthentication)
 	builderImage, err := docker.GetBuilderImage(dkr, c)
 	if err == nil {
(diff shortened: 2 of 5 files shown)


bparees:
> If --as-dockerfile is expected to be used in environments where a Docker socket is not available, then this PR might break that expectation.

that's definitely the expectation (the primary use case for --as-dockerfile is for envs that don't have a docker socket but have some other mechanism for turning a Dockerfile into an image, such as `podman build` or `buildah`).

cc @otaviof this might be of interest to you if there's a way to make it work when the docker socket is available and no-op otherwise (or fallback to using podman or something else to inspect the image when the docker socket isn't available).  I think it relates to some of the other work you're doing to enable non-docker-socket flows.

pkg/build/config.go
```diff
@@ -10,14 +10,27 @@ import (
...
-		return errors.New("image metadata must be provided to GenerateConfigFromLabels")
+		return errors.New("image metadata must be provided to GenerateConfigFromBuilderImageLabels")
+	}
+
+	return nil
+}
+
+// GenerateConfigFromApplicationImageLabels generates the S2I Config struct from the Docker
+// image labels.
+func GenerateConfigFromApplicationImageLabels(config *api.Config, metadata *docker.PullResult) error {
```
sbose78: What is the difference between a builder image and application image in this context ?

sbose78:
How about we do this, @bparees :

1. Get this PR merged after review comments are addressed, with expectation of availability of docker socket - ie, the existing paradigm in the codebase. That addresses the bug.

2. @isutton to open a second PR with code changes to have `container/images` being used for inspecting images - this PR is expected to be large and a bit more complex.

We shouldn't hold up the basic functionality/fix in (1) till (2) is done.

adambkaplan:
@sbose78 per @bparees 's comment, the `--as-dockerfile` flow explicitly does _not_ expect there to be a docker socket present. Wiring that assumption in breaks OpenShift 4.

I'm afraid option 2 is the only real option.

isutton:
> > If --as-dockerfile is expected to be used in environments where a Docker socket is not available, then this PR might break that expectation.
> 
> that's definitely the expectation (the primary use case for --as-dockerfile is for envs that don't have a docker socket but have some other mechanism for turning a Dockerfile into an image, such as `podman build` or `buildah`).

There are some expectations on how a builder image influences the configuration of a build (including --as-dockerfile). From what I understood, Destination and ScriptsURL could be specified in the builder image and, if overwritten, should be present in the generated Dockerfile. If this is indeed the case, then --as-dockerfile requires access to container images, and without @otaviof's work, the only way to do so is using the existing Docker implementation. 

Since the intention of --as-dockerfile is, as you've mentioned, to provide a Dockerfile on environments Docker socket is not available, the documentation and expectations for the current s2i implementation around Destination and ScriptsURL are wrong and should be amended, stating that builder images will not be consulted when using --as-dockerfile.

It should be clear that coordination between builders and --as-dockerfile is required, and it should be also clear that @otaviof's work is trying to elegantly shoehorn external builders at a very specific place to make it work, but it is visible that some structural changes should be performed in the code base to properly support it; for example, abstracting the concept of a builder for the whole program as a core architecture goal, instead of branching based on a selected builder inside s2i's business logic. 

Since this can't be a simple bug fix because of the unwanted side-effects of implementing this fix, I consider the right thing to do to close this PR, and improve on @otaviof's work once #994 and #996 are finalized and merged.

What do you think?

isutton:
> @sbose78 per @bparees 's comment, the `--as-dockerfile` flow explicitly does _not_ expect there to be a docker socket present. Wiring that assumption in breaks OpenShift 4.
> 
> I'm afraid option 2 is the only real option.

Like I put in another comment, then the documentation should be amended to mention that --as-dockerfile won't honour Builder Image labels.

Am I right to assume that?

pkg/build/config.go
isutton: There are different expectations for labels in the builder image and application image. In that particular case, Application Image's labels are reused in `rebuild`, where `build` should honour Builder Image's labels.

sbose78:
Yes, the scale of the work is indeed big and is beyond just being a bug fix. 

Let's get #994 and #996 merged asap since that is blocking your work.

bparees:
> Since the intention of --as-dockerfile is, as you've mentioned, to provide a Dockerfile on environments Docker socket is not available, the documentation and expectations for the current s2i implementation around Destination and ScriptsURL are wrong and should be amended, stating that builder images will not be consulted when using --as-dockerfile.

agreed, the immediate short term action needed is to improve the docs to make it clear where the fidelity between "s2i build" and "s2i build --as-dockerfile" breaks down.  the `--scripts-url` and `--destination` args can be used to overcome this limitation for now.

sbose78:
We can close this.

isutton:
Closing this PR since we're following the work in #1002 and #1003.

PR #1000, opened 2019-10-24
Bug 1758305: support .s2i/image_metadata.json with --as-dockerfile

gabemontero:
/assign @adambkaplan 

@bparees @otaviof @sbose78 @openshift/openshift-team-developer-experience  FYI

### DIFF
--- a/pkg/util/labels_test.go
+++ b/pkg/util/labels_test.go
@@ -1,0 +1,62 @@
+package util
+
+import (
+	"github.com/openshift/source-to-image/pkg/api"
+	"github.com/openshift/source-to-image/pkg/api/constants"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+
+	"testing"
+)
+
+func TestImageMetadataLabels(t *testing.T) {
+	tests := []struct {
+		json  string
+		count int
+	}{
+		{
+			json:  "{\"labels\": [{\"org.tenkichannel/service\":\"rain-forecast-process\"}]}",
+			count: 1,
+		},
+		{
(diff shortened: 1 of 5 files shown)


gabemontero:
My first s2i PR in 15 months :-)

pkg/util/labels_test.go
```diff
@@ -0,0 +1,35 @@
...
+
+import (
+	"io/ioutil"
+	"os"
+	"path/filepath"
+
+	"testing"
+)
+
+func TestImageMetadataLabels(t *testing.T) {
```
adambkaplan: @gabemontero can we get an e2e in our integration tests - ideally for both the `docker` and `dockerfile` flows.

gabemontero: ah ... those are new since I last traversed these waters

sure, I'll add an `image_metadata.json` file in the `.s2i` folder and add the labels there to the expected contents

gabemontero: quick update @adambkaplan - there appears to be a image_metadata.json test via https://github.com/openshift/source-to-image/blob/master/test/integration/scripts/.s2i/bin/assemble#L19-L24 but based on my examinations that only applies to the `docker` integration tests, and not the `dockerfile` ones (which I think explains the gap here).

gabemontero: @adambkaplan and I talked and since even with dockerfile, we have to git clone a repo with a image metadata file and there is no injection path for it like the scripts, I'm just going to broaden the new unit test a bit.

gabemontero:
@adambkaplan PTAL

gabemontero:
CI tests are passing @adambkaplan

PR #1001, opened 2019-10-29
Container Image

otaviof:
Building project to run as a container image.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,23 @@
+FROM openshift/origin-release:golang-1.13 AS builder
+
+ENV S2I_GIT_VERSION="" \
+    S2I_GIT_MAJOR="" \
+    S2I_GIT_MINOR=""
+
+ENV GOARCH="amd64"
+
+COPY . $GOPATH/src/github.com/openshift/source-to-image
+
+RUN cd $GOPATH/src/github.com/openshift/source-to-image && \
+    make && \
+    install _output/local/bin/linux/${GOARCH}/s2i /usr/local/bin
+
+#
+# Runner Image
+#
+
+FROM registry.redhat.io/ubi8/ubi
+
+COPY --from=builder /usr/local/bin/s2i /usr/local/bin
+
+ENTRYPOINT [ "/usr/local/bin/s2i" ]
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,9 @@
+.*
+_output/
+docs/
+examples/
+images/
+*.md
+AUTHORS
+LICENSE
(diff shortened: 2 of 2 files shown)


Dockerfile
```diff
@@ -0,0 +1,29 @@
+FROM registry.redhat.io/ubi8/ubi AS builder
```
sbose78: Go for `FROM openshift/origin-release:golang-1.13 AS builder`

Dockerfile
```diff
@@ -0,0 +1,29 @@
...
+    S2I_GIT_MAJOR="" \
+    S2I_GIT_MINOR=""
+
+ENV GOPATH="/go" \
+    GOARCH="amd64"
+
+RUN yum install -y \
+    make \
+    git \
+    go
```
gabemontero: all three of these should be unnecessary once you base off of the image @sbose78 noted above

Dockerfile
```diff
@@ -0,0 +1,29 @@
+FROM registry.redhat.io/ubi8/ubi AS builder
+
+ENV S2I_GIT_VERSION="" \
+    S2I_GIT_MAJOR="" \
+    S2I_GIT_MINOR=""
+
+ENV GOPATH="/go" \
+    GOARCH="amd64"
+
+RUN yum install -y \
```
sbose78: If you use the above builder base image, you wouldn't need to do `yum install`.

Dockerfile
```diff
@@ -0,0 +1,29 @@
...
+
+#
+# Runner Image
+#
+
+FROM registry.redhat.io/ubi8/ubi
+
+COPY --from=builder /usr/local/bin/s2i /usr/local/bin
+
+ENTRYPOINT [ "/usr/local/bin/s2i" ]
```
sbose78: This looks good!

Dockerfile
sbose78: > Using the CI imagestream means folks will need a pull secret to our CI registry.

Not really, it gets pulled from
https://hub.docker.com/r/openshift/origin-release/dockerfile/

Is there a similar builder UBI image? I'm fine with UBI, just want to avoid `yum install..`

otaviof: @sbose78 I think that's a good suggestion, thanks.

sbose78:
@bparees  @adambkaplan  Could you get this harmless commit in? :)

bparees:
@sbose78 needs to be taken out of draft state

otaviof:
> @sbose78 needs to be taken out of draft state

In this case, just out of draft. Please consider.

PR #1002, opened 2019-10-31
Introduce "generate" command

isutton:
This PR create a new s2i top level command, `generate`. 

The choice to create this command instead of just implementing the same functionality in the `build --as-dockerfile` flow, is that the semantics of builder image would change between `build` and `build --as-dockerfile`; the former would reflect changes made to local images, where the latter would ignore those. This means `s2i build my-image .` and `s2i build --as-dockerfile my-image . | buildah` commands could produce different results.

This new command expects the given builder image name to be available on a remote registry, and is inspected through the infrastructure provided by `containers/image`.

Once OCI support is implemented in s2i, this command could benefit and provide transparent access to local images as well, since it would delegate the inspection task to the container runtime itself and even deprecate `build --as-dockerfile`.

This PR requires some dependencies to be updated, since it depends on containers/image, and both depended (and still depend) on Docker.

In [cli.go](https://github.com/openshift/source-to-image/pull/1002/files#diff-b9503498c06e1c2a08f7576382f1ae1e) the `gen-dockerfile` command is registered.

In [dockerfile.go](https://github.com/openshift/source-to-image/pull/1002/files#diff-c9b9ece11522f4925d11d344afaf8c3c), the `CreateDockerfile()` method was modified to inherit both `scripts-url` and `destination` labels.

[gen-dockerfile.go](https://github.com/openshift/source-to-image/pull/1002/files#diff-0611cbb736f2b7d4529fd35f2e038b95) is a new file, where the command is implemented.

In [go.mod](https://github.com/openshift/source-to-image/pull/1002/files#diff-37aff102a57d3d7b797f152915a6dc16), dependency to containers/image was added, together with replacements for indirect dependencies based on containers/image `go.mod` file.


otaviof:
Naming wise I suggest calling this new sub-command `generate` instead, to follow the names of the other existing sub-commands.

sbose78:
Thanks @isutton and @otaviof for this, and thank you @bparees for taking a quick look.

Even though I find `s2i build --as-dockerfile` super-confusing because it doesn't actually build, I feel there's a need to fix image inspection for it https://github.com/openshift/source-to-image/issues/973 instead of having a new command.

My worry is:

1. `s2i build` : inspects image labels, builds images
2. `s2i build --as-dockerfile`: doesn't inspect image labels, doesn't build image, generates dockerfile
3. `s2i gen-dockerfile` inspects registry images image labels , doesn't build images, generates dockerfile

I would love to remove (2) entirely to avoid this. Or move (3) into (2).


@siamaksade  , thoughts?

bparees:
>   1)  s2i build : inspects image labels, builds images
>    2) s2i build --as-dockerfile: doesn't inspect image labels, doesn't build image, generates dockerfile
>    3) s2i gen-dockerfile inspects registry images image labels , doesn't build images, generates dockerfile

> I would love to remove (2) entirely to avoid this. Or move (3) into (2).

why not get rid of (2) entirely to avoid this?

bparees:
(long term, after a deprecation period)

sbose78:
:+1:  to removing (2) . I think we should do it as part of an version 1.x release?

siamaksade:
@bparees @sbose78 I agree with deprecation and removal following its grace period

adambkaplan:
+1 to removal.

Now that we are switching to go modules, v2 may come sooner than we think. Any refactor that removes or alters an exported method/constant requires a major version bump.

https://github.com/golang/go/wiki/Modules#semantic-import-versioning

isutton:
The past couple of days @otaviof and I have been enabling buildah in s2i from start to end (#1003), and since this effort is being quite successful I'd like to propose a strategy for aligning that PR and this one.

As stated previously, the `gen-dockerfile` command was proposed because the same functionality present in `s2i build --as-dockerfile`:

* Changed the semantics of the builder image, where without `--as-dockerfile` it would consult both local and remote builder images, but with it would consider only remote builder images. This alone makes the functionality incoherent and error prone for the end user.

* Doesn't correctly represent the same actions s2i would do for the given input parameters, it means that images built by s2i and images built by a Dockerfile produced by s2i are not equivalent in some scenarios, making it harder for both the end user and s2i or builder image developers to correctly assess whether the right results are being yielded.

There are a couple of points I'm still unsure:

* In the case Buildah is a first class citizen in s2i, I would expect that generating a Dockerfile might be not as important as before, since s2i build would work in OCP 4 with its Docker socket restrictions.
* To produce a Dockerfile equivalent to the process s2i applies in its build process, the majority of build arguments and switches are required in this command to properly build the Dockerfile, since both commands would share the build process. I personally dislike this fact, and IMO it points back in the direction of `--as-dockerfile` in the developer experience front, since the result of s2i build would be equivalent to `s2i build --as-dockerfile | buildah bud -f -`. 

My take on this is to finalize #1003, since it will enable s2i to build an application image without requiring a Docker socket, making it suitable to run in Tekton, Travis CI and other build pipelines. Additionally, it also gives us the foundation to produce an accurate Dockerfile based on the existing s2i build workflow.

sbose78:
https://github.com/openshift/builder does use the Dockerfile generation flow and then does a lot of other things. So, I wouldn't want to comment on the future of Dockerfile generation. 

Dockerfile generation is handy because it lets us separate the concern of 'define strategy' v/s 'build the image using a strategy'.

So, irrespective of #1003 , we would need this to be in ASAP.

bparees:
> Dockerfile generation is handy because it lets us separate the concern of 'define strategy' v/s 'build the image using a strategy'.

+1.  i think there is a lot of value in avoiding tying s2i to runtime libraries both on the inspection/generation side as well as the image construction side.

bparees:
@adambkaplan jenkins-s2i is failing..I thought we didn't have any more jenkins-s2i jobs?  Looks like maybe that's not a merge blocking job anyway?

adambkaplan:
@bparees it clearly isn't blocking merge...job is broken because we're using golang 1.13 in our go.mod, and the build steps are breaking in that suite.

There's a separate issue where Prow doesn't think the Jenkins jobs are completing that we need to fix.

DhritiShikhar:
This commit a8f31a2 renames `gen-dockerfile` command to `generate`.
but I can see the filename is still `pkg/cmd/cli/cmd/gen-dockerfile.go`

pkg/cmd/cli/cmd/gen-dockerfile.go
```diff
@@ -0,0 +1,57 @@
...
+func NewCmdGenDockerfile(cfg *api.Config) *cobra.Command {
+
+	genDockerfileCmd := &cobra.Command{
+		Use:   "gen-dockerfile <image> [<dockerfile>]",
+		Short: "Generate a Dockerfile based on the provided builder image",
+		Example: `
+# Generate a Dockerfile from a remote builder image:
+$ s2i gen-dockerfile docker://quay.io/redhat-developer/app-binding-operator
+`,
+		RunE: func(cmd *cobra.Command, args []string) error {
```
DhritiShikhar: @isutton Is there a reason to use `RunE:` instead of `Run:`?

```
// Run: Typically the actual work function. Most commands will only implement this
Run func(cmd *Command, args []string)
// RunE: Run but returns an error
RunE func(cmd *Command, args []string) error
```

I can see other s2i top level commands like `build`, `rebuild`, `completion`, `create` use `Run:`

DhritiShikhar:
Since, its difficult to add a comment on github UI because of so many file changes, I will add my comment here.

In file `/pkg/cmd/cli/cmd/gen-dockerfile.go`,

`args` parameter is never used anywhere in the function.

Screenshot from 2019-11-20 18-41-50

DhritiShikhar:
@isutton Are there any tests for this new `s2i generate` command?

DhritiShikhar:
In `/pkg/cmd/cli/cmd/gen-dockerfile.go` file

The following 4 lines can be replaced by `cmdutil.AddCommonFlags(generateCmd, cfg)`:

Screenshot from 2019-11-21 02-13-21

There is already a function to add common flags to a command which can be reused for `generate` command also. 
Refer: https://github.com/openshift/source-to-image/blob/master/pkg/cmd/cli/cmd/build.go#L193

sbose78:
@DhritiShikhar Please add inline comments ( if needed ) here:
- https://github.com/openshift/source-to-image/pull/1002/files#diff-0611cbb736f2b7d4529fd35f2e038b95
- https://github.com/openshift/source-to-image/pull/1002/files#diff-c9b9ece11522f4925d11d344afaf8c3c

sbose78:
Hi @adambkaplan , This PR is complete, considering that @isutton will work on @DhritiShikhar 's review comments shortly. The primary files changed specified in the PR description here  https://github.com/openshift/source-to-image/pull/1002#issue-334953881 . Could you do a high-level review of this please, to ensure there's nothing significantly off?

pkg/cmd/cli/cmd/gen-dockerfile.go
isutton: To propagate the error to the caller, if any.

Now looking closely at the referred commands, their implementations differ regarding surfacing the errors to the user-just highlighting this as a reference to the intent of using `RunE`. 

Perhaps this should be changed in another work unit, where we could make use of the existing `errors` pkg from source-to-image, where it implements a concrete `Error` type and has supporting functions to treat and display an error message in all commands.

isutton:
> In `/pkg/cmd/cli/cmd/gen-dockerfile.go` file
> 
> The following 4 lines can be replaced by `cmdutil.AddCommonFlags(generateCmd, cfg)`:
> 
> Screenshot from 2019-11-21 02-13-21
> 
> This is a function to add common flags to a command which can be reused for `generate` command also.
> Refer: https://github.com/openshift/source-to-image/blob/master/pkg/cmd/cli/cmd/build.go#L193

I noticed that function but chose to extract from it those that were applicable to the new command.

isutton:
> @isutton Are there any tests for this new `s2i generate` command?

I considered all the important logic was already being tested, and the logic that creates the Dockerfile was already being tested, so I didn't do it at the time.

Will look at that.

adambkaplan:
Test is currently not compiling with the following:

```
# github.com/openshift/source-to-image/test/integration/docker [github.com/openshift/source-to-image/test/integration/docker.test]
test/integration/docker/integration_test.go:154:6: undefined: client.IsErrImageNotFound
```

Note that I believe our CI vms use golang 1.12

adambkaplan:
@isutton note that I'm working on a fix to our CI job so that the artifacts and status are being reported correctly.

isutton:
> Test is currently not compiling with the following:
> 
> ```
> # github.com/openshift/source-to-image/test/integration/docker [github.com/openshift/source-to-image/test/integration/docker.test]
> test/integration/docker/integration_test.go:154:6: undefined: client.IsErrImageNotFound
> ```
> 
> Note that I believe our CI vms use golang 1.12

Fixed in cbcc36a.

isutton:
> > @isutton Are there any tests for this new `s2i generate` command?
> 
> I considered all the important logic was already being tested, and the logic that creates the Dockerfile was already being tested, so I didn't do it at the time.
> 
> Will look at that.

All `generate` does is either encoded in very straightforward and simple code, or is already tested through either SBO's or vendored libraries unit tests, so tests specific to the generate command have not been implemented.

isutton:
Is there anything I can do to move this PR and finalize this piece of work?

go.mod
```diff
@@ -17,76 +18,39 @@ exclude (
...
-	github.com/coreos/go-systemd v0.0.0-20190719114852-fd7a80b32e1f // indirect
-	github.com/coreos/pkg v0.0.0-20180928190104-399ea9e2e55f // indirect
-	github.com/deckarep/golang-set v1.7.1 // indirect
-	github.com/dgrijalva/jwt-go v3.2.0+incompatible // indirect
-	github.com/docker/distribution v2.6.0-rc.1.0.20170726174610-edc3ab29cdff+incompatible
-	github.com/docker/docker v0.0.0-20190404075923-dbe4a30928d4
+	github.com/containerd/fifo v0.0.0-20190816180239-bda0ff6ed73c // indirect
+	github.com/containers/image/v5 v5.0.0
+	github.com/coreos/etcd v3.3.17+incompatible // indirect
+	github.com/docker/distribution v2.7.1-0.20190205005809-0d3efadf0154+incompatible
```
adambkaplan: this should be ok - docker/distribution on openshift/builder is 2.7.1

go.mod
```diff
@@ -17,76 +18,39 @@ exclude (
...
-	github.com/coreos/etcd v3.3.13+incompatible // indirect
-	github.com/coreos/go-semver v0.3.0 // indirect
-	github.com/coreos/go-systemd v0.0.0-20190719114852-fd7a80b32e1f // indirect
-	github.com/coreos/pkg v0.0.0-20180928190104-399ea9e2e55f // indirect
-	github.com/deckarep/golang-set v1.7.1 // indirect
-	github.com/dgrijalva/jwt-go v3.2.0+incompatible // indirect
-	github.com/docker/distribution v2.6.0-rc.1.0.20170726174610-edc3ab29cdff+incompatible
-	github.com/docker/docker v0.0.0-20190404075923-dbe4a30928d4
+	github.com/containerd/fifo v0.0.0-20190816180239-bda0ff6ed73c // indirect
+	github.com/containers/image/v5 v5.0.0
```
adambkaplan: this is ok - containers/image uses strict semvers so we won't have transitive dep conflicts in openshift/builder

pkg/cmd/cli/cmd/generate.go
```diff
@@ -0,0 +1,99 @@
...
+}
+
+// NewCmdGenerate implements the S2I cli generate command.
+func NewCmdGenerate(cfg *api.Config) *cobra.Command {
+	generateCmd := &cobra.Command{
+		Use:   "generate <image> <dockerfile>",
+		Short: "Generate a Dockerfile based on the provided builder image",
+		Example: `
+# Generate a Dockerfile from a remote builder image:
+$ s2i generate docker://quay.io/redhat-developer/app-binding-operator Dockerfile.gen
```
adambkaplan: I'd use one of our canonical s2i images as an example - `docker.io/centos/nodejs-10-centos7`

pkg/cmd/cli/cmd/generate.go
```diff
@@ -0,0 +1,99 @@
...
+
+}
+
+// NewCmdGenerate implements the S2I cli generate command.
+func NewCmdGenerate(cfg *api.Config) *cobra.Command {
+	generateCmd := &cobra.Command{
+		Use:   "generate <image> <dockerfile>",
+		Short: "Generate a Dockerfile based on the provided builder image",
+		Example: `
+# Generate a Dockerfile from a remote builder image:
```
adambkaplan:
```suggestion
# Generate a Dockerfile from a builder image:
```

pkg/cmd/cli/cmd/generate.go
```diff
@@ -0,0 +1,99 @@
...
+`,
+		RunE: func(cmd *cobra.Command, _ []string) error {
+			if cmd.Flags().NArg() != 2 {
+				return cmd.Help()
+			}
+
+			cfg.BuilderImage = cmd.Flags().Arg(0)
+			cfg.AsDockerfile = cmd.Flags().Arg(1)
+
+			ctx := context.TODO()
```
adambkaplan: `context.Background()` is acceptable here, since this is effectively the main entrypoint to the command.

go.mod
```diff
@@ -3,9 +3,10 @@ module github.com/openshift/source-to-image
 go 1.13
 
 replace (
-	github.com/containerd/containerd => github.com/containerd/containerd v0.2.10-0.20170808145631-06b9cb351610
-	github.com/docker/docker => github.com/openshift/moby-moby v1.4.2-0.20190308215630-da810a85109d
-	github.com/moby/buildkit => github.com/dmcgowan/buildkit v0.0.0-20170731200553-da2b9dc7dab9
+	github.com/Sirupsen/logrus => github.com/sirupsen/logrus v1.2.0
+	github.com/containerd/containerd => github.com/containerd/containerd v1.3.0
+	github.com/docker/docker => github.com/docker/docker v0.0.0-20180522102801-da99009bbb11
```
adambkaplan: Docker has been downgraded here - why?

go.mod
```diff
@@ -3,9 +3,10 @@ module github.com/openshift/source-to-image
...
+	golang.org/x/crypto => golang.org/x/crypto v0.0.0-20180904163835-0709b304e793
```
adambkaplan: `golang.org/x/crypto` has been added in the replace stanza, but is not referenced in the `require` stanza. Why is this necessary?

isutton:
> @isutton General logic here looks good. I think we can refactor the `getImageLabels` function in a follow-up PR, since openshift/builder implements its own logic to set image labels.

Sounds fine.
 
> I'm worried a bit about the go mod bumps. From what I can gather all we need is `containers/image/v5` and its dependencies.

`containers/image/v5` adds a new Docker version as a transitive dependency, being reflected on the `replace` section in `go.mod` (some dependencies have been added from Docker's dependencies and other libraries as well). 

> 
> Finally, these commits need to be squashed. I'd like to see the following:
> 
>     1. A clean commit with the dependencies bumped in `go.mod` and `go.sum`
> 
>     2. A `bump(*)` commit with the vendor updates.
> 
>     3. A commit with the new functionality

Will do.

go.mod
isutton: To use the same Docker version `containers/image` is using:

https://github.com/containers/image/blob/v5.0.0/go.mod#L13

go.mod
isutton: After I did my first changes, I noticed that `make test` didn't work. After I dug into it, I've noticed that buildkit and swarmkit should be updated as well, and this lead to all those dependencies  

Tests use buildkit and swarmkit, which also had to be bumped and adjusted (see 9f3ead4fc4a53bdcfd515e5e9e6e6a2fa7b66c6d for reference).

pkg/cmd/cli/cmd/generate.go
isutton: Done in 7eb02e7e1af6913e74a7f8599f551ffabf24abed.

adambkaplan:
@isutton note that we recently merged a `go.mod` bump that addressed some of the issues with buildkit.

pkg/cmd/cli/cmd/generate.go
sbose78: @isutton Could you please make the above change?

sbose78:
I think this is now ready?

isutton:
I've updated the branch according to the project's guidelines.

/cc @adambkaplan @sbose78

sbose78:
@adambkaplan Should we take this in?

pkg/cmd/cli/cmd/generate.go
adambkaplan: @isutton please update the example here.

adambkaplan: Also a bit of a nit - is it worth switching the order around? Thoughts on the following semantics?

```
$ s2i generate DOCKERFILE [IMAGE]
```

or 

```
$ s2i generate DOCKERFILE --from=[IMAGE]
```

isutton: From what I understood, IMAGE is always required in this use case since there's no information regarding that; additionally, the protocol currently must be informed by the user to refer to the image since `containers/image` is responsible for retrieving its contents and inspecting its labels.

I believe the following example is closer to what we want, where DOCKERFILE could have the `Dockerfile` value by default:

```
$ s2i generate IMAGE [DOCKERFILE]
```

During implementation, I opted for DOCKERFILE to be mandatory and not having defaults to avoid unwillingly Dockerfile overwrites.
---
The requested example modification has been made.

sbose78: Right, IMAGE is always required since it is the s2i builder image. I think Adam is OK with that.

I do like `s2i generate DOCKERFILE --from=quay.io/builder/image`

 instead of 

`s2i generate quay.io/builder/image Dockerfile.gen`

adambkaplan: > From what I understood, IMAGE is always required in this use case since there's no information regarding that; 

Correct

> During implementation, I opted for DOCKERFILE to be mandatory and not having defaults to avoid unwillingly Dockerfile overwrites.

I agree with this choice for now. We can always opt to make it optional later; I have other ideas on how to arrange the arguments to this command that are worth discussing in a follow-up PR.

adambkaplan:
/approve

/lgtm

Will follow up on my "other ideas" I have for the cli args.

Issue #966, opened 2019-05-16
Can we get this to work with Buildah/podman rather then being Docker Centric


rhatdan:
There is no Docker in RHEL8 so this package will not work there.

s2i in OpenShift 4.1 is based on Buildah, shouldn't this package be based on it also?

bparees:
s2i builds in openshift 4.1 are based on the `--as-dockerfile` flow.  The s2i code (as vendored into/used in openshift) doesn't actually invoke any buildah libraries directly.

We don't currently have any plans to invest in re-implementing the "traditional" s2i flow to directly communicate with additional container runtimes because the "generate a dockerfile, then build the dockerfile w/ your image building tool of choice" gives us more insulation than having to implement a "container runtime driver" for each additional container runtime we want s2i to be able to drive.

If someone else has the time/resources to contribute the work, we'd probably accept it, but it's a significant refactor to:

1) abstract all the places where s2i directly talks to docker today (creating containers, starting them, injecting content, monitoring container progress, committing images w/ modified metadata)
2) implement the buildah/podman based equivalents of that logic

/cc @adambkaplan @smarterclayton

smarterclayton:
Seems reasonable Ben

rhatdan:
Can we deprecate this repo then?

adambkaplan:
Feels like it is worth doing a survey of the image building tools to see which ones support a `Dockerfile` as input. That could motivate, say, an s2i v2 that produces a `Dockerfile` and build context that is consumed by the chosen image builder.

jmtd:
> s2i builds in openshift 4.1 are based on the `--as-dockerfile` flow

That's a useful hint, thanks. My images (openjdk-?-rhel?) don't have `/usr/libexec/s2i/assemble` so fail for this. but last I checked a basic s2i worked quickstart worked in OCP 4.1. 

> s2i in OpenShift 4.1 is based on Buildah, 

TIL *this* s2i code isn't in OCP 4.1.

bparees:
> That's a useful hint, thanks. My images (openjdk-?-rhel?) don't have /usr/libexec/s2i/assemble so fail for this. but last I checked a basic s2i worked quickstart worked in OCP 4.1.

[...]

PR #1003, opened 2019-11-06
[WIP] Optionally use buildah to build images using s2i

otaviof:
This PR adds `buildah` as a first class citizen in s2i workflows. It implements `docker.Docker` interface using `buildah` via `os/exec` calls, and it's able to produce container images with the same contents than using Docker.  For instance:

```
$ docker images |egrep 'test/sti-fake-app\s.*(buildah|latest)'
test/sti-fake-app                                    latest              e53bb38a9a56        8 seconds ago       1.24MB
test/sti-fake-app                                    buildah             f31bdf26d972        2 minutes ago       1.24MB
```

Users can enable `buildah` or `docker` backends using `S2I_CONTAINER_MANAGER` environment variable, or global `--container-manager` (`-m`) parameter:

```sh
# using buildah
s2i --container-manager=buildah build ...
# using Docker, default option
s2i build ... 
```

Additionally, use `s2i` image (temporarily stored [here](https://quay.io/repository/olemefer/source-to-image?tab=tags)) to test it out.

For example, building [nodejs-ex](https://github.com/sclorg/nodejs-ex.git):

```sh
docker run --privileged quay.io/olemefer/source-to-image:latest \
    build \
    https://github.com/sclorg/nodejs-ex.git \
    otaviof/nodejs-12-rhel7:latest \
    otaviof/nodejs-ex:latest
```

## Motivation

This PR addresses issue #966, and take over the effort of #996.

## Testing

On this PR the integration tests are extended to run business logic against `buildah` container manager as well. Therefore, export `S2I_CONTAINER_MANAGER` as `buildah` or `docker`, in order to choose the container manager backend.

To run the integration tests use:

```sh
S2I_CONTAINER_MANAGER="buildah" make test-docker
```

Additionally, Travis-CI setup is extended to have Docker service and `buildah` available, and therefore integration tests are also executed over there.

Since `s2i` workflows are preserved, please execute this PR against existing image builds, and report your findings.

### `--run`

Buildah is a daemon-less application, and therefore is not well suited as Docker to execute a container in background.

### `ONBUILD`

Since `buildah` is using [OCI Container Image](https://github.com/opencontainers/image-spec) format, `ONBUILD` directive is not supported.

## TODOs

- [x] do not allow `--run` and buildah at the same time;
- [x] command line parameter and environment variable to use `buildah` or `docker`;


isutton:
All the workflows specified in [this README](https://github.com/openshift/source-to-image/blob/master/examples/nginx-centos7/README.md) seem to work.

otaviof:
I've got good results on running builds with `--runtime-image` options. It does copy over the artifacts in the new image. However, the runtime image was lacking `io.openshift.s2i.scripts-url` setting, which required a manual intervention. Other than that it works as expected using `buildah` as container-manager.

```sh
s2i --loglevel=99 build . otaviof/nodejs-12-rhel7:latest otaviof/nodejs-ex:latest --runtime-image=nodejs:latest --runtime-artifact="/opt/app-root" --runtime-artifact="/usr/libexec/s2i"
```

isutton:
Good work @otaviof! I've tested the following with success:

```
➜  ~ docker volume create containers-storage
containers-storage

➜  ~ docker run --privileged --rm \
    --mount="source=containers-storage,target=/var/lib/containers/storage" \
    quay.io/olemefer/source-to-image:latest \
        build \
        https://github.com/sclorg/nodejs-ex.git \
        otaviof/nodejs-12-rhel7:latest \
        otaviof/nodejs-ex:latest

Creating application image based on '"otaviof/nodejs-12-rhel7:latest"'
---> Installing application source ...
---> Installing all dependencies

> ejs@2.7.4 postinstall /opt/app-root/src/node_modules/ejs
> node ./postinstall.js

Thank you for installing EJS: built with the Jake JavaScript build tool (https://jakejs.com/)

added 120 packages from 342 contributors and audited 205 packages in 7.053s
found 9 vulnerabilities (2 low, 2 moderate, 4 high, 1 critical)
run `npm audit fix` to fix them, or `npm audit` for details
---> Building in production mode
---> Pruning the development dependencies
audited 205 packages in 4.899s
found 9 vulnerabilities (2 low, 2 moderate, 4 high, 1 critical)
run `npm audit fix` to fix them, or `npm audit` for details
/opt/app-root/src/.npm is not a mountpoint
---> Cleaning the npm cache /opt/app-root/src/.npm
/tmp is not a mountpoint
---> Cleaning the /tmp/npm-*
Build completed successfully

➜  ~ docker run --privileged --rm \
    --mount="source=containers-storage,target=/var/lib/containers/storage" \
    quay.io/buildah/stable:latest \
        buildah images

REPOSITORY                          TAG      IMAGE ID       CREATED              SIZE
localhost/otaviof/nodejs-ex         latest   6b3d74bd184c   About a minute ago   595 MB
docker.io/otaviof/nodejs-12-rhel7   latest   ce17a227c92c   5 weeks ago          587 MB

➜  ~ docker volume rm containers-storage
containers-storage
```

sbose78:
/retitle Optionally use buildah to build images using s2i

rhatdan:
With these changes we can run s2i without being root?  And not having to open a huge security hole on the system.
Would love to see more Podman references in here as well.

otaviof:
> With these changes we can run s2i without being root? And not having to open a huge security hole on the system.
> Would love to see more Podman references in here as well.

Good catch! Indeed, {{buildah}} is prepared to run without being {{root}}, therefore you can run {{s2i}} in a unprivileged user.

gabemontero:
ok we have all green tests ... taking a peek

Dockerfile
```diff
@@ -4,26 +4,72 @@ ENV S2I_GIT_VERSION="" \
...
+gpgcheck=1\n\
+enabled=1\n\
+gpgkey=https://www.centos.org/keys/RPM-GPG-KEY-CentOS-Official\n\
+" > /etc/yum.repos.d/appstream.repo && \
+    cat /etc/yum.repos.d/appstream.repo && \
+    yum -y reinstall shadow-utils && \
+    yum -y install buildah fuse-overlayfs fuse3 git && \
+    rm -rf /var/cache /var/log/dnf* /var/log/yum.*
+
+RUN sed -i \
```
gabemontero: any sort of documentation you all referenced when you worked this up?  ... if so, let's add it as a reference in a comment that precedes this line

this entire set of bash feels very specific/detailed and perhaps fragile as underling ubi evolves

contrib/completions/bash/s2i
```diff
@@ -324,6 +324,8 @@ _s2i_build()
     local_nonpersistent_flags+=("--volume=")
     flags+=("--ca=")
     flags+=("--cert=")
+    flags+=("--container-manager=")
```
gabemontero: names can be hard .... coming up with something that captures what `docker` vs. `buildah` vs. the other tools in this space certainly falls in that category

I'm pondering the use of `container-manager` .... perhaps "safe" but feels overreaching to me

My best alternative is `oci-tool`  ... we are talking about producing OCI images, right?

hack/build-test-images.sh
```diff
@@ -6,14 +6,19 @@ set -o errexit
 set -o nounset
 set -o pipefail
 
+readonly S2I_CONTAINER_MANAGER="${S2I_CONTAINER_MANAGER:-docker}"
```
gabemontero: I won't make the comment everywhere, but again on my container manager vs. oci tool point

pkg/api/constants/env.go
```diff
@@ -10,4 +10,8 @@ const (
 	//
 	// DEPRECATED - use ScriptsURLLabel instead.
 	ScriptsURLEnvironment = "STI_SCRIPTS_URL"
+
+	// ContainerManagerEnv is the environment variable name that specifies the container runtime S2I will use to
+	// build the application's image.
+	ContainerManagerEnv = "S2I_CONTAINER_MANAGER"
```
gabemontero: OK I'll make the comment here as well ;-) ... container manager -> oci tool

pkg/buildah/buildah.go
```diff
@@ -0,0 +1,620 @@
...
+	s2ierr "github.com/openshift/source-to-image/pkg/errors"
+	s2itar "github.com/openshift/source-to-image/pkg/tar"
+	"github.com/openshift/source-to-image/pkg/util/fs"
+	"github.com/openshift/source-to-image/pkg/util/interrupt"
+	utillog "github.com/openshift/source-to-image/pkg/util/log"
+)
+
+var log = utillog.StderrLog
+
+// Buildah implements docker.Docker interface using buildah as a backend.
```
gabemontero: OK this is good / was going to look for that.

That said, as a result, I would have expected a buildah_test.go here with some unit tests a la https://github.com/openshift/source-to-image/blob/master/pkg/docker/docker_test.go

pkg/cmd/cli/cli.go
```diff
@@ -60,6 +61,19 @@ func NewCmdCLI() *cobra.Command {
 	return s2iCmd
 }
 
+// defaultContainerManager returns the default container manager name, based on environment variable.
+func defaultContainerManager() string {
+	if name := os.Getenv(constants.ContainerManagerEnv); isValidContainerManager(name) {
+		return name
+	}
+	return "docker"
```
gabemontero: make constant

pkg/cmd/cli/cli.go
```diff
@@ -60,6 +61,19 @@ func NewCmdCLI() *cobra.Command {
...
+}
+
+// isValidContainerManager validates the given container manager name.
+func isValidContainerManager(s string) bool {
+	return s == "docker" || s == "buildah"
```
gabemontero: constants

pkg/docker/docker.go
```diff
@@ -491,9 +488,18 @@ func (d *stiDocker) CheckAndPullImage(name string) (*api.Image, error) {
 	return image, nil
 }
 
+// ImageShortName shorten image name when tag longer than 18 characters.
```
gabemontero: feels like there would already be a k8s or docker utility function around this but my quick searches could not find one

Where did the 18 limit come from ?  Why not 20 or 12?

test/integration/buildah/buildah_test.go
```diff
@@ -0,0 +1,271 @@
...
+	"github.com/openshift/source-to-image/pkg/docker"
+	"github.com/openshift/source-to-image/pkg/util/fs"
+)
+
+const (
+	busyboxImage = "docker.io/busybox:latest"
+	exampleImage = "nginx-centos7:testing"
+)
+
+func TestBuildah(t *testing.T) {
```
gabemontero: cool ... that said, I do not think we should refrain from unit tests for pkg/buildah/buildah.go just cause these are here

rajivml:
sry this is not a PR comment but just want to check, so once this PR is merged and if switch to use buildah as my container manager then the issue that am facing here will also be resolved ?https://github.com/openshift/source-to-image/issues/1016,  the issue that am facing in 1016 is, the s2i containers that are getting spawned are repeatedly getting killed on Kubernetes when the Nodes are under memory pressure and am not able to figure out how to avoid them being killed

Dockerfile
otaviof: That's a direct copy of [`buildah` Dockerfile](https://github.com/containers/buildah/blob/11ad04e83d5206ccd367eed65becb7d0908d3f95/contrib/buildahimage/stable/Dockerfile#L17).

otaviof: I end up copying buildah upstream, however, I'm also inclined to hold the whole configuration file, instead of editing during container build process. What do you think?

contrib/completions/bash/s2i
otaviof: In the community we see Docker being referred as a "container-manager", and to complete the task of building an container image requires traits of the "container-manager". And, indeed, the term is overreaching, can imply a number of different tools.

So, maybe we would like to have the creation of OCI images as a flag, and therefore, we could add `--oci` boolean flag to turn on this behavior. What do you think about enabling this as a flag, instead of a string parameter?

pkg/buildah/buildah.go
otaviof: Since the unit-tests of this component would require shell-out to call `buildah`, I end moving those tests [here](https://github.com/otaviof/source-to-image/blob/oci/test/integration/buildah/buildah_test.go). Therefore, we can make sure in that level of testing we can enable `buildah`, and find it installed.

pkg/docker/docker.go
otaviof: That was already present in s2i, I only moved that to a common place. I'm not sure why this limitation is in place, but in the case of `buildah` it should produce a similar user experience, hence the reuse.

otaviof: Maybe we would like to open a new issue with this subject, track it separately.

test/integration/buildah/buildah_test.go
otaviof: The tests are moved to this location, as commented before. This was done because of CI settings, where we only have `buildah` and `docker` for integration tests, and only in a few selected locations.

Dockerfile
gabemontero: Not sure I know what `hold the whole configuration file` would entail @otaviof .... is it possible to elaborate.

Otherwise, I would add the https://github.com/containers/buildah/blob/11ad04e83d5206ccd367eed65becb7d0908d3f95/contrib/buildahimage/stable/Dockerfile#L17 link you posted here as part of a comment in this file, explaining where it came from.

contrib/completions/bash/s2i
gabemontero: Yeah regardless of how Docker may have referred to itself upstream (and I suspect overreaching may have been purposeful), I like `--oci` better.

pkg/buildah/buildah.go
gabemontero: In addition to the actual callouts you have in the integration test you noted, your unit test version of buildah_test.go could have a `fakeBuildah` just like the `fakeDocker` we have at https://github.com/openshift/source-to-image/blob/master/pkg/docker/docker_test.go#L145-L147, no?

contrib/completions/bash/s2i
otaviof: Okay, I open to have it as a flag. @sbose78, @isutton, what's your take on this?

Dockerfile
otaviof: Okay, I meant having the actual configuration file in this repository, and later `COPY` in the image. But sure, it indeed deserves a reference comment registering the origin.

contrib/completions/bash/s2i
isutton: If having it as a flag is merely deciding one or the other, and the choices are Docker (non OCI) and Buildah (OCI), I'm fine with a boolean flag to enable or disable it in the CLI domain.

jmtd: so is `--oci` meant to enable buildah, and the default would still be docker? I think this is confusing. After all Docker has some kind of relationship to OCI too, and that also could evolve as time goes on. `--container-manager` is more explicit if overreaching. Also what if a third (or fourth) option is added later? E.g. Podman — `--container-manager` is not a binary switch at least.

PR #1004, opened 2019-11-21
update restore-artifacts script

sdarwin:
Hi,

When testing the code in the example 'assemble' script, the syntax didn't seem to work:

```
mv /tmp/artifacts/. ./
```
resulted in the error

```
mv: cannot move '/tmp/artifacts/.' to './.': Device or resource busy
```
Actually, I would have expected it to be:

```
mv /tmp/artifacts/* ./
```

Any thoughts on this?  Could you reproduce the error? 

A more thorough modification would be the following. This catches hidden files also.   

```
shopt -s dotglob
mv /tmp/artifacts/* ./
shopt -u dotglob
```

### DIFF
--- a/pkg/create/templates/scripts.go
+++ b/pkg/create/templates/scripts.go
@@ -19,7 +19,9 @@ fi
 #
 if [ "$(ls /tmp/artifacts/ 2>/dev/null)" ]; then
   echo "---> Restoring build artifacts..."
+  shopt -s dotglob
   mv /tmp/artifacts/* ./
+  shopt -u dotglob
 fi
 
 echo "---> Installing application source..."
--- a/pkg/create/templates/scripts.go
+++ b/pkg/create/templates/scripts.go
@@ -19,7 +19,7 @@ fi
 #
 if [ "$(ls /tmp/artifacts/ 2>/dev/null)" ]; then
   echo "---> Restoring build artifacts..."
-  mv /tmp/artifacts/. ./
+  mv /tmp/artifacts/* ./
 fi
 
 echo "---> Installing application source..."


bparees:
@sdarwin can you give us more thorough recreate steps?

sdarwin:
@bparees , sure.  The issue is the trailing dot "/." in combination with the mv command. 
 Again, the original was:

```
mv /tmp/artifacts/. ./
```
To recreate, try this:

```
[ec2-user@ip-172-31-6-107 ~]$ mkdir test1
[ec2-user@ip-172-31-6-107 ~]$ mkdir test2
[ec2-user@ip-172-31-6-107 ~]$ touch test1/test1.txt
[ec2-user@ip-172-31-6-107 ~]$ touch test1/test2.txt
[ec2-user@ip-172-31-6-107 ~]$ touch test1/test3.txt
[ec2-user@ip-172-31-6-107 ~]$ mv test1/. test2/
mv: cannot move 'test1/.' to 'test2/.': Device or resource busy
[ec2-user@ip-172-31-6-107 ~]$ mv test1/. ./
mv: cannot move 'test1/.' to './.': Device or resource busy
[ec2-user@ip-172-31-6-107 ~]$ 
```

bparees:
@sdarwin yeah it looks like we previously had this as `*` and (w/o testing I guess) switch it to `.`:
https://github.com/openshift/source-to-image/commit/dd9ba6571183ca78800196e401e5a3cdbda1a7e6

I think the dotglob option is probably best to ensure we continue to catch hidden files.  Can you update your PR to that?

sdarwin:
> I think the dotglob option is probably best to ensure we continue to catch hidden files. Can you update your PR to that?

updated.

bparees:
/lgtm
thanks @sdarwin

PR #1011, opened 2019-12-10
Revert "support .s2i/image_metadata.json with --as-dockerfile"

gabemontero:
This reverts commit 38d2b729fd1d1358397f538775c84074f48dadea.

/assign @adambkaplan 
/assign @bparees


gabemontero:
fyi used `git revert 38d2b729fd1d1358397f538775c84074f48dadea`

bparees:
we are reverting this because it did not actually implement the same behavior as the normal s2i build flow provided (dynamic labelling of the output image based on a file that was created during the assemble script execution)

/lgtm

gabemontero:
> we are reverting this because it did not actually implement the same behavior as the normal s2i build flow provided (dynamic labelling of the output image based on a file that was created during the assemble script execution)

In hindsight the "need to be created during the assemble script processing" requirement is clearly stated in https://github.com/openshift/source-to-image/blob/master/docs/new_labels.md#creating-the-file

> 
> /lgtm

PR #1012, opened 2019-12-10
bump(*) - Use buildkit for dockerfile parsing

adambkaplan:
1. Move parser tests to use `moby/buildkit`'s parser
2. Use upstream version of `buildkit` instead of an outdated fork.
3. `go mod tidy && go mod vendor`


pkg/scripts/environment_test.go
```diff
@@ -43,30 +41,30 @@ func equalArrayContents(a []string, b []string) bool {
 	return true
 }
 
-func TestConvertEnvironmentToDocker(t *testing.T) {
```
adambkaplan: This test seems to run a pseudo-Docker build. It pulls in a lot of deps from `docker/docker` which seem to break things.

Moving this out for the time being.

pkg/build/strategies/onbuild/onbuild_test.go
```diff
@@ -7,7 +7,7 @@ import (
 	"strings"
 	"testing"
 
-	"github.com/docker/docker/builder/dockerfile/parser"
+	"github.com/moby/buildkit/frontend/dockerfile/parser"
```
adambkaplan: Use dockerfile parser from moby/buildkit

test/integration/dockerfile/dockerfile_test.go
```diff
@@ -11,7 +11,8 @@ import (
 	"strings"
 	"testing"
 
-	"github.com/docker/docker/builder/dockerfile/parser"
+	"github.com/moby/buildkit/frontend/dockerfile/parser"
```
adambkaplan: Use dockerfile parser from moby/buildkit

pkg/scripts/environment_test.go
```diff
@@ -43,30 +41,30 @@ func equalArrayContents(a []string, b []string) bool {
...
+// 		"`Value1`",
+// 		`"`,
+// 	}
+// 	var charValues []string
+// 	for ch := rune(32); ch < 127; ch++ {
+// 		charValues = append(charValues, "AB"+string(ch)+"CD", "AB\\"+string(ch)+"CD")
+// 	}
+// 	for _, testValue := range append(testValues, charValues...) {
+// 		list := api.EnvironmentList{{Name: "TEST", Value: testValue}}
+// 		converted := ConvertEnvironmentToDocker(list)
```
gabemontero: Seems like you could do some validation of the return from `ConvertEnvironmentToDocker` without calling dockerfile functions

adambkaplan: Yeah - the actual function is pretty trivial. I'll add a true unit test.

adambkaplan: @gabemontero restored coverage via a simple unit test. PTAL

gabemontero: cool

PR #1019, opened 2020-02-04
Allow releases with podman or docker

adambkaplan:
Fix release script to allow an override build command.
Works with `podman` or `docker`.


otaviof:
/lgtm

Thanks!

PR #1020, opened 2020-02-04
Bump travis to golang 1.13

adambkaplan:


.travis.yml
```diff
@@ -1,14 +1,14 @@
...
+  - 1.13.x
 
 install:
   - make install-travis
 
 script:
-  - make verify
-  - make check TESTFLAGS="-p=4" TESTS="''" # empty quotes are because hack/test-go.sh requires 2 args
+  - GOFLAGS="-mod=vendor" make verify
+  - GOFLAGS="-mod=vendor" make check TESTFLAGS="-p=4" TESTS="''" # empty quotes are because hack/test-go.sh requires 2 args
```
adambkaplan: `GOFLAGS` env var is needed because `go.mod` by default won't use the vendor tree. OpenShift's CI images for golang 1.13 export this by default.

sbose78: Right.

PR #1023, opened 2020-02-07
Update documentation for generate and --as-dockerfile

isutton:
- [x] Graduate the `--as-dockerfile` flag in s2i build from "EXPERIMENTAL" to a normal flag.
- [x] Update our docs on Github with the new command and syntax.
- [x] Mark the new generate command as "Tech Preview" so that we're consistent with the rest of the OpenShift ecosystem.

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -237,6 +237,13 @@ Usage:
 $ s2i generate <builder image> <output file>
 ```
 
+#### Example usage
+
+Generate a Dockerfile for the `centos/nodejs-10-centos7` builder image:
+```
+$ s2i generate docker://docker.io/centos/nodejs-10-centos7 Dockerfile.gen
+```
+
 # s2i usage
 
 The `s2i usage` command starts a container and runs the `usage` script which prints
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -9,6 +9,7 @@ following sections of this document:
 * [create](#s2i-create)
 * [build](#s2i-build)
 * [rebuild](#s2i-rebuild)
+* [generate](#s2i-generate)
 * [usage](#s2i-usage)
 * [version](#s2i-version)
 * [help](#s2i-help)
@@ -218,6 +219,19 @@ Usage:
 $ s2i rebuild <image name> [<new-tag-name>]
 ```
 
+# s2i generate
+The `s2i generate` command produces a Dockerfile using an existing S2I base 
+image that can be used to produce an image by any application supporting the
+format.
+
+The *image name* argument is in the URL-like, for example 
+`docker://docker.io/centos/nodejs-10-centos7`.
+
+Usage:
+
+```
+$ s2i generate <image name> <output file>
+```
 
 # s2i usage
 
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -228,13 +228,13 @@ The `s2i generate` command generates a Dockerfile using an existing S2I builder
 image that can be used to produce an image by any application supporting the
 format.
 
-The *image name* argument is in the URL-like, for example 
-`docker://docker.io/centos/nodejs-10-centos7`.
+The *builder name* is a reference to the builder image that serves as the base 
+of the generated Dockerfile. For example, 
+`docker://docker.io/centos/nodejs-10-centos7:latest`.
 
 Usage:
-
 ```
-$ s2i generate <image name> <output file>
+$ s2i generate <builder image> <output file>
 ```
 
 # s2i usage
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -220,7 +220,11 @@ $ s2i rebuild <image name> [<new-tag-name>]
 ```
 
 # s2i generate
-The `s2i generate` command produces a Dockerfile using an existing S2I base 
+
+This command is a Technology Preview feature, and is subject to change or be 
+removed without notice in a future release.
+
+The `s2i generate` command generates a Dockerfile using an existing S2I builder 
 image that can be used to produce an image by any application supporting the
 format.
 
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -230,7 +230,7 @@ format.
 
 The *builder name* is a reference to the builder image that serves as the base 
 of the generated Dockerfile. For example, 
-`docker://docker.io/centos/nodejs-10-centos7:latest`.
+`docker.io/centos/nodejs-10-centos7:latest`.
 
 Usage:
 ```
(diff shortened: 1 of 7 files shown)


docs/cli.md
```diff
@@ -218,6 +219,19 @@ Usage:
 $ s2i rebuild <image name> [<new-tag-name>]
 ```
 
+# s2i generate
+The `s2i generate` command produces a Dockerfile using an existing S2I base
```
sbose78: Add the text "This is a Tech Preview feature" before this line, and we should be good.

docs/cli.md
```diff
@@ -218,6 +219,19 @@ Usage:
...
+image that can be used to produce an image by any application supporting the
+format.
+
+The *image name* argument is in the URL-like, for example 
+`docker://docker.io/centos/nodejs-10-centos7`.
+
+Usage:
+
+```
+$ s2i generate <image name> <output file>
```
sbose78: Could you keep it consistent with the `s2i build..` docs and call it `<builder image>` ?

docs/cli.md
```diff
@@ -218,6 +219,19 @@ Usage:
...
+```
```
sbose78: Could you follow it up with an example, please:

```
s2i generate docker://docker.io/centos/nodejs-10-centos7 Dockerfile.gen
```

docs/cli.md
```diff
@@ -73,7 +74,7 @@ that image and add them to the tar streamed to the container into `/artifacts`.
 |:----------------------------|:--------------------------------------------------------| 
 | `-u (--allowed-uids)`       | Specify a range of allowed user ids for the builder and runtime images. Ranges can be bounded (`1-10001`) or unbounded (`1-`). |
 | `-n (--application-name`)   | Specify the display name for the application (default: output image name) |
-| `--as-dockerfile`           | EXPERIMENTAL: Output a Dockerfile to this path instead of building a new image |
```
sbose78: Looks good 👍

docs/cli.md
```diff
@@ -218,6 +219,19 @@ Usage:
 $ s2i rebuild <image name> [<new-tag-name>]
 ```
 
+# s2i generate
+The `s2i generate` command produces a Dockerfile using an existing S2I base 
+image that can be used to produce an image by any application supporting the
+format.
+
+The *image name* argument is in the URL-like, for example 
+`docker://docker.io/centos/nodejs-10-centos7`.
```
adambkaplan: In my experience the `docker://` scheme is not used often. I would propose the following:

```
The *image name* is a reference to the builder image that serves as the base of the generated Dockerfile. For example, `docker.io/centos/nodejs-10-centos7:latest`.
```

docs/cli.md
adambkaplan: As with my previous comment, we should drop the `docker://` scheme. I rarely see examples of this in the field.

docs/cli.md
```diff
@@ -218,6 +219,19 @@ Usage:
 $ s2i rebuild <image name> [<new-tag-name>]
 ```
 
+# s2i generate
+The `s2i generate` command produces a Dockerfile using an existing S2I base 
+image that can be used to produce an image by any application supporting the
+format.
```
adambkaplan: I would add a "tech preview" note:

```
This command is a Technology Preview feature, and is subject to change or be removed without notice in a future release.
```

The CLI usage should also be updated. In general we try to keep the CLI help docs and the text here identical.

docs/cli.md
isutton: Removing the scheme now would require another PR and would also limit the `containers/image` benefits by limiting the registry type it can fetch metadata from to only Docker (assuming all build images are indeed stored in a Docker registry).

Since this is a Tech Preview, can we move on with this as is for now and discuss how the introduction of OCI registries would look like?

docs/cli.md
isutton: Should the CLI usage include the "tech preview" note as well?

isutton: Done in https://github.com/openshift/source-to-image/pull/1023/commits/a9c3f7ff062161628a57f1c5948b09a4584e1f5d

docs/cli.md
isutton: Done in https://github.com/openshift/source-to-image/pull/1023/commits/cb94360b477113940ab8efd47411f12cb791bc84

docs/cli.md
isutton: Done in https://github.com/openshift/source-to-image/pull/1023/commits/38d86ba41a23fc0cdd094fcb12e5c08bd602ae52

docs/cli.md
isutton: Done in https://github.com/openshift/source-to-image/pull/1023/commits/a9c3f7ff062161628a57f1c5948b09a4584e1f5d

sbose78:
Made a quick PR to remove need for `docker://..`
https://github.com/openshift/source-to-image/pull/1024

docs/cli.md
sbose78: #1024

docs/cli.md
otaviof: @isutton, @sbose78 should we address the documentation bits in #1024?

docs/cli.md
isutton: Solved in https://github.com/openshift/source-to-image/pull/1023/commits/fb0be78c7f00a88b37cea07d4fc8e9e6ce59f9e9 and https://github.com/openshift/source-to-image/pull/1023/commits/888560cf0d71289ad9f706773c40c8974847f339

docs/cli.md
isutton: Solved in https://github.com/openshift/source-to-image/pull/1023/commits/fb0be78c7f00a88b37cea07d4fc8e9e6ce59f9e9 and https://github.com/openshift/source-to-image/pull/1023/commits/888560cf0d71289ad9f706773c40c8974847f339

sbose78:
Thanks for the test, Igor.

Also, validated 👍 

- _output/local/bin/darwin/amd64/s2i generate docker://docker.io/centos/nodejs-10-centos7 Dockerfile.gen

-  _output/local/bin/darwin/amd64/s2i generate docker.io/centos/nodejs-10-centos7 Dockerfile.gen

sbose78:
I see we do `exclude_graphdriver_btrfs exclude_graphdriver_devicemapper` https://github.com/openshift/builder/search?q=OS_GOFLAGS_TAGS&unscoped_q=OS_GOFLAGS_TAGS as well, though I am not quite sure why.

sbose78:
Also could you squash commits, please?

PR #1024, opened 2020-02-07
Remove the need for "docker://" in the "generate" command

sbose78:
Both

`s2i generate centos/nodejs-10-centos7 Dockerfile.gen`
and
`s2i generate docker.io/centos/nodejs-10-centos7 Dockerfile.gen`

work now.

Given that 
`s2i build https://github.com/openshift/ruby-hello-world docker://docker.io/centos/ruby-22-centos7 hello-world-app` doesn't work.

I've ensured `s2i generate docker://docker.io/centos/nodejs-10-centos7 Dockerfile.gen` also consistently doesn't work :)

### DIFF
--- a/pkg/cmd/cli/cmd/generate.go
+++ b/pkg/cmd/cli/cmd/generate.go
@@ -2,6 +2,7 @@ package cmd
 
 import (
 	"context"
+
 	"github.com/containers/image/v5/transports/alltransports"
 	"github.com/containers/image/v5/types"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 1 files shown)


pkg/cmd/cli/cmd/generate.go
```diff
@@ -62,21 +63,21 @@ func NewCmdGenerate(cfg *api.Config) *cobra.Command {
...
 		RunE: func(cmd *cobra.Command, _ []string) error {
 			if cmd.Flags().NArg() != 2 {
 				return cmd.Help()
 			}
 
-			ref, err := alltransports.ParseImageName(cmd.Flags().Arg(0))
+			cfg.BuilderImage = cmd.Flags().Arg(0)
+			cfg.AsDockerfile = cmd.Flags().Arg(1)
+
+			ref, err := alltransports.ParseImageName("docker://" + cfg.BuilderImage)
```
adambkaplan: What does `ParseImageName` do if there is no protocol? Ideally we should be able to support both forms.

`BuilderImage` needs to be a valid pull spec that is used in a Dockerfile.

sbose78: > What does ParseImageName do if there is no protocol?

It gets us the `ref` for doing the following later on
```
img, err := ref.NewImage(ctx, &types.SystemContext{})
	if err != nil {
		return nil, err
	}

	imageMetadata, err := img.Inspect(ctx)
	if err != nil {
		return nil, err
	}
```

> Ideally we should be able to support both forms.

We don't support the following at the moment, but we could support `docker://..` in `s2i generate.. ` of course ?
```
s2i build https://github.com/openshift/ruby-hello-world docker://docker.io/centos/ruby-22-centos7
```

otaviof: I think we should assume `docker://` as default, it's the default approach in `s2i` in general. Therefore, we need a parser to split up container image address and protocol, on which it would assume `docker://` when protocol is empty.

sbose78:
Closing this, will let Igor do a fresh one.

PR #1025, opened 2020-02-10
Add option to set TLS from environment variable

mtatheonly:
Add option to set TLS in default docker config from environment variable DOCKER_TLS, same as docker cli (https://github.com/docker/cli/pull/863)


thiagoarrais:
I'm interested in this too. Anything else we need to get this merged, @adambkaplan?

thiagoarrais:
@adambkaplan: I was revisiting this and it seems like 1.3 is already out? I'm not 100% sure how the release process works here, but it seems like we can now remove the `hold` label here?

PR #1033, opened 2020-04-14
Update of new_labels.md on a bug and corresponding workaround in OpenShift 4.x

ka-zo:
A new section is added to new_labels.md on a bug in OpenShift 4.x. The method of updating output image labels during the s2i build process is not possible currently in OpenShift 4.x and a corresponding workaround is described.


bparees:
thanks!

PR #1036, opened 2020-04-23
Workaround for wrong command created in the output docker image under…

ka-zo:
… OSE 4.x.

During the s2i build process neither the labels of the output Docker image are updated correctly under OpenShift 4.x, nor the command field. A workaround is shared here for this faulty behavior.


bparees:
thank you @ka-zo

PR #1040, opened 2020-06-01
Add ppc64le support

Siddhesh-Ghadi:
Add ppc64le support for ci job on travis-ci.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,6 @@
+arch:
+  - amd64
+  - ppc64le
 language: go
 
 go:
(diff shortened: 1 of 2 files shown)


adambkaplan:
/ok-to-test

@gabemontero @coreydaley @otaviof I don't see an issue adding Power to the list of supported arches, as I believe the release script generates ppc64le binaries for Linux. Down the road we can consider adding an integration suite to run on Power if/when the OpenShift on Power CI systems are open to the entire org.

ghatwala:
@adambkaplan - thanks for the review here , would be good to have `ppc64le` tarball showing up in releases post this PR merger 👍 .

PR #1041, opened 2020-06-01
Changed dockerfile to create container for Power

snehlatamohite:
Hi ,

I have updated dockerfile to create container for Power as well. while building container below "arch" parameter can be used to select architecture. 

Command to build for Power - 
docker build --no-cache --build-arg arch=ppc64le -t s2i .

Command to build for Intel -  
docker build --no-cache --build-arg arch=amd64 -t s2i .

Thanks,
Snehlata Mohite,

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,7 @@
+ARG arch=amd64
 FROM openshift/origin-release:golang-1.13 AS builder
 
+ARG arch
 ENV S2I_GIT_VERSION="" \
     S2I_GIT_MAJOR="" \
     S2I_GIT_MINOR=""
@@ -8,7 +10,7 @@ ENV S2I_GIT_VERSION="" \
 WORKDIR /tmp/source-to-image
 COPY . .
 
-ENV GOARCH="amd64"
+ENV GOARCH=$arch
 
 USER root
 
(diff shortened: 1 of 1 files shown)


adambkaplan:
/ok-to-test

Let's see if this works with our CI (powered by OpenShift). At one point there was a bug preventing build args from being used in builds, but that issue should be fixed.

IMO we shouldn't be setting `GOARCH` in the Dockerfile - this should be set in the underlying base builder image, and our scripts should copy the built binary to a platform-agnostic location.

snehlatamohite:
@adambkaplan - Yeah , Ideally that is the correct way however in multistage dockerfile its not able to select ENV variable in COPY statement hence I have used ARG which default points to amd64 hence GOARCH will also point to amd64 by default. For creating power container, we need to override it by passing build arg to docker command like docker build --no-cache --build-arg arch=ppc64le -t s2i . 

Error when underlying arch is used to GOARCH/ARCH - 
---> 0a8f09f43143
Step 12/14 : COPY --from=builder /tmp/source-to-image/_output/local/bin/linux/${ARCH}/s2i  /usr/local/bin/s2i
COPY failed: stat /var/lib/docker/overlay2/80d8577ae9c01c4bb4cfca8ca081fa0e97397097cc7197e68afbe3d7653fda47/merged/tmp/source-to-image/_output/local/bin/linux/s2i: no such file or directory

snehlatamohite:
@adambkaplan - Thanks , I will implement potential alternate solution and will revert.

yselkowitz:
/do-not-merge

Please do not merge, this is not how multi-arch enablement works, and will only make downstream productization more difficult.

snehlatamohite:
@yselkowitz  - Community suggested this approach, we will implement it in coming sprint  - https://github.com/openshift/source-to-image/issues/1044

yselkowitz:
Can this be closed with #1046 merged?

snehlatamohite:
No longer needed.

PR #1055, opened 2020-10-09
Use volume mount in release build

adambkaplan:
- Use bind mount instead of tar stream to add source code to the release
  container.
- Update release Dockerfile to declare volume for the s2i source.
  Command updated to remove tar stream in
- Add sti-version-defs file to .gitignore

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,6 @@
 .vscode/
 cmd/s2i/debug
 *~
+
+# Version definition file used in release build
+sti-version-defs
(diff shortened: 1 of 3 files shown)


adambkaplan:
/assign @gabemontero 

Seems that every time I try to release, I find something wrong with the s2i release script.
This go around, it seems that podman with cgroups v2 doesn't like the tar stream in. This seemed like a silly aspect of the release script anyway, since nowadays it's much easier to bind mount the source code into the container. Bind mounting also saves us the extra `podman cp` call - the resulting client binaries sit in the `_output` directory at the end of the build.

PR #1060, opened 2020-12-17
Build with Golang 1.14

sbose78:
Use Golang 1.14 for building the s2i image.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM openshift/origin-release:golang-1.13 AS builder
+FROM openshift/origin-release:golang-1.14 AS builder
 
 ENV S2I_GIT_VERSION="" \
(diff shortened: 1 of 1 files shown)


coreydaley:
The change you made does not make this build with 1.14 for CI.
https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_source-to-image/1060/pull-ci-openshift-source-to-image-master-verify/1339701565425979392

You need to change https://github.com/openshift/release/blob/master/ci-operator/config/openshift/source-to-image/openshift-source-to-image-master.yaml#L6

sbose78:
Ah base image replacements, thank you!

This Dockerfile as far as I remember is used for internal brew builds. It needs to go through CI first, anyway. I will make the change in the place suggested by you! Thank you.

PR #1072, opened 2021-10-12
Windows ignore path fix

seeker25:
Path fix for CopyContents, the ignore files use filepath (https://github.com/openshift/source-to-image/blob/30d81a9440f30b472bb32e592b12c1a83a396edd/pkg/ignore/ignore.go#L58), Ignore probably should too.
Untested on linux.


gabemontero:
/assign @bparees 

@adambkaplan is unavailable this week @bparees ... if you have cycles ptal to consider adding the approved label; if you don't, no worries, we'll engage with @adambkaplan next week when he becomes available again

thanks

PR #1073, opened 2021-10-13
eol php 5.5 image no longer functional so move off of it in cmd line tests

gabemontero:
discovered while processing https://github.com/openshift/source-to-image/pull/1072

/assign @bparees 

cross ref - that image was removed from the 4.x sample imagestreams a while ago; see https://github.com/sclorg/s2i-php-container/blob/master/imagestreams/php-centos.json and https://github.com/sclorg/s2i-php-container/blob/master/imagestreams/php-rhel.json


gabemontero:
Hey hey all green

/assign @coreydaley 

for lgtm in case @bparees is too busy (we already have the approve from him)

PR #1088, opened 2022-04-27
Migrate to go 1.17

coreydaley:


nalind:
Any chance I can convince you to merge the two `require` sections in `go.mod`?  Is it safe to unpin some of the versions in the `replace` section?  I'm mainly interested in the `golang.org/x/crypto` entry there, but the fewer we need to pin, I think, the better.

adambkaplan:
> Any chance I can convince you to merge the two `require` sections in `go.mod`?

This is the standard behavior for golang 1.17 and higher - one `require` section with direct, declared dependencies, then a second `require` with the full graph of indirect deps. See the go 1.17 release note for "Pruned module graphs": https://go.dev/doc/go1.17

>  Is it safe to unpin some of the versions in the replace section?

I would address that in a follow-up PR.

adambkaplan (changes requested):
It looks like a lot of the test code relies on golang build constraints (aka build tags). In go 1.17, the syntax of these changed from `// +build <constraint>` to `//go:build <constraint>`. See the "//go:build lines" section of the 1.17 release notes: https://go.dev/doc/go1.17

Because these build constraints aren't being respected, our Dockerfile "integration" test suite is not running any tests.

nalind:
> `go mod tidy` automatically separates them into two `require` sections, I even tried manually mixing them up and it re-separated them.

Oh, wow, when I run it I'm seeing the same thing.  I don't think I've seen it do that before.

> When I unpin `golang.org/x/crypto` and run `go mod tidy` i get:
> 
> ```
> go: github.com/moby/buildkit@v0.6.3 requires
> 	github.com/tonistiigi/fsutil@v0.0.0-20190819224149-3d2716dd0a4d requires
> 	golang.org/x/crypto@v0.0.0-20190129210102-0709b304e793: invalid pseudo-version: does not match version-control timestamp (expected 20180904163835)
> ```
> 
> Is there a better version that I should pin it to?

I'm tempted to avoid pinning it to a specific version unless we absolutely have to.  I get a self-consistent set of dependencies if I start with
```
module github.com/openshift/source-to-image

go 1.17

require (
        github.com/containers/image/v5 v5.21.0
        github.com/docker/distribution v2.8.1+incompatible
        github.com/docker/docker v20.10.14+incompatible
        github.com/docker/go-connections v0.4.0
        github.com/moby/buildkit v0.10.1
        github.com/spf13/cobra v1.4.0
        github.com/spf13/pflag v1.0.5
        golang.org/x/net v0.0.0-20220225172249-27dd8689420f
        k8s.io/klog/v2 v2.30.0
)
```
and run `go mod tidy -compat=1.17` – the `golang.org/x/crypto` package actually gets dropped as an indirect dependency.  The `Client` interface definition in `pkg/docker/docker.go` ends up needing an update to add a new `github.com/opencontainers/image-spec/v1.Platform` pointer argument to its `CreateContainer` method, and we need to start passing `nil` in when we call it, but after that it compiles here.

coreydaley:
Updated in next push

> It looks like a lot of the test code relies on golang build constraints (aka build tags). In go 1.17, the syntax of these changed from `// +build <constraint>` to `//go:build <constraint>`. See the "//go:build lines" section of the 1.17 release notes: https://go.dev/doc/go1.17
> 
> Because these build constraints aren't being respected, our Dockerfile "integration" test suite is not running any tests.

PR #1090, opened 2022-06-28
Update dependencies; update to Go 1.18

nalind:
* Update our module definition to use Go 1.18.
* Update our direct dependencies.
* Regenerate TLS certificates used for TLS connectivity tests, and add a script that can be re-run to do that.
* Stop clearing `S2I_BUILD_TAGS` in `hack/test-go.sh` so that integration tests will run when it's invoked from `hack/test-dockerfile.sh` and from `hack/docker.sh`.
* Add unit tests for `pkg/docker.GetAssembleUser()`.


coreydaley:
@nalind We may have a chicken/egg issue here where we need to force-merge the pr for openshift/release first so that we can make this use go 1.18
```
build io/fs: cannot load io/fs: open /go/src/github.com/op...ift/source-to-image/vendor/io/fs: no such file or directory
```

nalind:
@coreydaley it sure looks that way to me.  I've manually checked that the "test-unit", "test-dockerfile", and "verify" makefile targets pass when I run them manually on my Fedora 36 box with Go 1.18.3, so I'm reasonably confident that they'll pass once we do that.  (The "verify" target caught some scripts I needed to run, so I've re-pushed with its suggested changes added in.)

OpenShift CI doesn't appear to be configured to check the `test-docker` target, which fails for me when I try it on my system with this PR, so there's probably still some work to do here.

coreydaley:
/retest
https://github.com/openshift/release/pull/29928 merged but it may take a couple of hours until it is available for these tests

coreydaley:
Looks like it's using the new go 1.18 now at least.
```
INFO[2022-06-30T01:58:08Z] Tagging openshift/release:golang-1.18 into pipeline:root.
```

coreydaley:
/retitle BUILD-485: Update dependencies; update to Go 1.18

coreydaley:
Looks like the `jenkins-s2i` test is failing since it is trying to use an older go version, i'm going to try to track down how to fix that.
```
 export BUILD_GO_VERSION=go1.12.5
+ BUILD_GO_VERSION=go1.12.5
+ wget -q https://dl.google.com/go/go1.12.5.linux-amd64.tar.gz
+ sudo tar -C /usr/local -xzf go1.12.5.linux-amd64.tar.gz
+ sudo rm /bin/go /bin/gofmt
+ sudo ln -s /usr/local/go/bin/go /usr/local/go/bin/godoc /usr/local/go/bin/gofmt /bin
+ make build
hack/build-go.sh
++ Building go targets for linux/amd64: cmd/s2i
build github.com/openshift/source-to-image/cmd/s2i: cannot load io/fs: open /data/src/github.com/openshift/source-to-image/vendor/io/fs: no such file or directory
make: *** [build] Error 1
++ export status=FAILURE
++ status=FAILURE 
```

nalind:
Found it: https://github.com/openshift/aos-cd-jobs/blob/master/sjb/config/test_cases/test_pull_request_s2i_master.yml.  Opened https://github.com/openshift/aos-cd-jobs/pull/3129.

coreydaley:
/retest
```
ERROR: Clone failed: source https://github.com/sclorg/nodejs-ex, target /tmp/s2i-dockerfiletest-dir2495762830/upload/src, with output "error: RPC failed; result=6, HTTP code = 0\nfatal: The remote end hung up unexpectedly\n"
I0707 12:31:45.611217    3152 clone.go:49] error: git clone failed: exit status 128
    dockerfile_test.go:1053: An error occurred during the build: exit status 128
```

PR #1105, opened 2023-03-09
fix: honor --exclude when using --as-dockerfile

matejvasek:


matejvasek:
Maybe this removes need for filtering in tar `type Tar interface` ?

matejvasek:
@coreydaley any idea why the test is failing?

coreydaley:
looking

pkg/scm/downloaders/file/download.go
```diff
@@ -54,9 +55,19 @@ func (f *File) Download(config *api.Config) (*git.SourceInfo, error) {
 		return nil, lerr
 	}
 
+	exclude, err := regexp.Compile(config.ExcludeRegExp)
```
coreydaley, line 58: My first thought is that if `config.ExlcludeRegExp` is empty, the compiled regular expression will match EVERYTHING, so everything will get excluded?
I think you should check it's length first to make sure it's not empty before you compile/use it. And below on line 65 you will probably have to split that into separate checks, one for the `filesToIgnore[path]` and one for the regexp.

matejvasek: Ah, yes, thanks. Empty string may be indeed problem.
Event with the bug locally most of the integration test passed.
And all unittest passed.

matejvasek: The failure looked more like some flake.

matejvasek:
@coreydaley ok, thanks, now test are passing.

matejvasek:
PTAL @coreydaley

pkg/scm/downloaders/file/download.go
```diff
@@ -54,9 +55,27 @@ func (f *File) Download(config *api.Config) (*git.SourceInfo, error) {
...
+			_, ok := filesToIgnore[path]
+			return ok || exclude.MatchString(path)
+		}
+	} else {
+		isIgnored = func(path string) bool {
+			_, ok := filesToIgnore[path]
+			return ok
+		}
+	}
+
```
coreydaley, line 75: how about something like this? (warning, untested)
```suggestion
isIgnored := func(path string) bool {
	if _, ok := filesToIgnore[path]; ok {
			  return ok
	}
	if len(config.ExcludeRegExp) != 0 {
			  return regexp.MustCompile(config.ExcludeRegExp).MatchString(path)
	}
}


```

matejvasek: @coreydaley  That's quite a pesimisation. It compiles regexp for each check.

If I wanted to optimise code size I would do:
```go
	isIgnored := func(path string) bool {
		_, ok := filesToIgnore[path]
		return ok
	}

	if config.ExcludeRegExp != "" {
		exclude, err := regexp.Compile(config.ExcludeRegExp)
		if err != nil {
			return nil, err
		}
		old := isIgnored
		isIgnored = func(path string) bool {
			return old(path) || exclude.MatchString(path)
		}
	}
```
but that introduces one more indirect call via the closure.

matejvasek: TBH  I am not huge fan of:
```go
filesToIgnore, lerr := di.GetListOfFilesToIgnore(copySrc)
```
It basically introduces one more traverse on top of the traverse done by `Copy()`.

IMHO the `ignore.DockerIgnorer` struct should have method `Matches(path string) bool` that could we used withing the `Copy()` traversal.

matejvasek: Not that it made any visible difference on M.2 drives today.

matejvasek: What I mean:
https://github.com/knative/func/blob/ac575e078f436df13839465d3fb31b2b97ab4a5c/pkg/pipelines/tekton/pipeplines_provider.go#L211-L219

there is even a library like that for .gitignore format.

matejvasek: >It basically introduces one more traverse on top of the traverse done by `Copy()`.

Actually, maybe it's even worse, it traverse FS for each positive glob in `.s2iignore`.

PR #1113, opened 2023-03-15
Clean up Makefile, remove outdated tools, setup goio

coreydaley:


hack/verify-imports.sh
apoorvajagtap, line 1: `hack/verify-imports.sh` errors out with:
`hack/verify-imports.sh
go: no packages loaded from ./vendor/github.com/go-imports-organizer/goio`

Having the `go-imports-organizer` vendored helps to overcome this error. It might be useful to have the `go-imports-organizer/goio` available in the `vendor`?

coreydaley: Good catch! For some reason on my workstation i have to do `git add -f vendor/` or else it will ignore newly vendored files, not sure why :( I have pushed an update with the go-imports-organizer library vendored.

PR #1132, opened 2023-10-27
Improve stacktrace to fix tech debts

divyansh42:


pkg/docker/docker.go
```diff
@@ -995,7 +995,14 @@ func (d *stiDocker) RunContainer(opts RunContainerOptions) error {
 		if signal == syscall.SIGQUIT {
 			buf := make([]byte, 1<<16)
 			runtime.Stack(buf, true)
-			fmt.Printf("%s", buf)
+            f, err := os.Create("/var/log/s2i_docker_stack_trace.log")
+            if err != nil {
+                return
+            }
+            defer f.Close()
+            w := bufio.NewWriter(f)
```
coreydaley, line 1003: I think you forgot to import this package ...

divyansh42: Updated.

.gitignore
```diff
@@ -3,6 +3,7 @@
 /.project
 /.vagrant
 .vscode/
+.idea/
```
coreydaley, line 6: This should be in your global gitignore on your workstation, not here. Go ahead and remove the .vscode one also.

https://gist.github.com/subfuzion/db7f57fff2fb6998a16c

coreydaley (changes requested):

.gitignore
divyansh42: Thanks @coreydaley for providing the information

PR #1140, opened 2024-02-13
chore: update docker to v25

matejvasek:
update docker to v25


matejvasek:
@coreydaley

pkg/docker/docker.go
```diff
@@ -310,14 +318,20 @@ func NewEngineAPIClient(config *api.DockerConfig) (*dockerapi.Client, error) {
 			},
 		}
 	}
-	return dockerapi.NewClient(config.Endpoint, os.Getenv("DOCKER_API_VERSION"), httpClient, nil)
+
+	return dockerapi.NewClientWithOpts(
+		dockerapi.WithHost(config.Endpoint),
+		dockerapi.WithHTTPClient(httpClient),
```
matejvasek, line 324: Are these even need today? AFAIK just using `dockerapi.FromEnv` should ensure everything is set up just all right.

matejvasek:
@coreydaley is the security check failure really related to my PR?

coreydaley:
@matejvasek No, and the security check is also option, so it will not stop the pull request from merging.

matejvasek:
@coreydaley We need this to get rid off some CVE reports in RH jira. The CVEs do not really affects us but we do not want to see the warnings.

pkg/docker/docker.go
```diff
@@ -310,14 +318,20 @@ func NewEngineAPIClient(config *api.DockerConfig) (*dockerapi.Client, error) {
...
+		dockerapi.WithAPIVersionNegotiation(),
```
adambkaplan, line 325: My understanding is that `APIVersionNegotiation` is more or less incompatible with `FromEnv` if the `DOCKER_API_VERSION` env var is set.

See https://github.com/moby/moby/blob/24.0/client/client_deprecated.go#L13-L17

adambkaplan (changes requested):
The update to the new client constructor is not a clean port from v24 to v25. I'm concerned that this is exercised in the `s2i build` code paths, which we don't have great test coverage for at present. OpenShift and the s2i container images for Tekton, Shipwright, and KNative use the `s2i generate` code paths, which we can verify in CI.

pkg/docker/docker.go
matejvasek: I think `WithAPIVersionNegotiation()` is necessary. It does not work on my system with older moby.
Of course if `DOCKER_API_VERSION` is set that will take precedence.

pkg/docker/docker.go
```diff
@@ -310,14 +318,20 @@ func NewEngineAPIClient(config *api.DockerConfig) (*dockerapi.Client, error) {
...
+		dockerapi.FromEnv,
```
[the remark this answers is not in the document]
matejvasek: FromEnv probably takes more envvars into account witch is better, or not?

pkg/docker/docker.go
matejvasek: so `DOCKER_API_VERSION` beats negotiation, but from my experience we need negotiation.

matejvasek:
@adambkaplan wort case: I can use the old deprecated ctor and override linter error.

matejvasek:
PTAL @adambkaplan

pkg/docker/docker.go
matejvasek: IMO this is expected and desired behaviour.

pkg/docker/docker.go
```diff
@@ -310,14 +318,20 @@ func NewEngineAPIClient(config *api.DockerConfig) (*dockerapi.Client, error) {
 			},
 		}
 	}
-	return dockerapi.NewClient(config.Endpoint, os.Getenv("DOCKER_API_VERSION"), httpClient, nil)
+
+	return dockerapi.NewClientWithOpts(
```
adambkaplan, line 322: It took me a long time to fully understand how the CLI options interact - I'm recommend we add a comment for future maintainers.

```suggestion
        // Create a new docker client with the provided host endpoint and HTTP Client.
        // By default, this client will negotiate the Docker API version to use with the backend engine on the first request.
        // The API version to use can be fixed by setting the DOCKER_API_VERSION environment variable.
        // See https://pkg.go.dev/github.com/docker/docker/client#WithAPIVersionNegotiation and
        // https://pkg.go.dev/github.com/docker/docker/client#WithVersionFromEnv for more information
	return dockerapi.NewClientWithOpts(
```

adambkaplan (changes requested):
One last suggestion, otherwise looks good.

What started out as a chore feels like a net new feature! I don't think it warrants any additional documentation beyond a release note, but after this lands we should increase the minor version number on the next release (to 1.4.0).

adambkaplan:
/assign @sayan-biswas 

Feel free to use `/approve` to green-light merge - I can add the LGTM once the comment suggestion is added commits are squashed.

@matejvasek - in your commit message please explain why we need API version negotiation for docker. This makes it easier for future maintainers to understand why the change was introduced, without having to chase down this pull request and unroll the comments.

PR #1145, opened 2024-05-09
BUILD-960: Update Containers to UBI9 go-toolset

adambkaplan:
- Update Dockerfiles to use the ubi9 go-toolset images. For the releaser image, this avoids downloading the go sdk as part of the relase process.
- Add `build-container` make target for convenience.
- Update ci-operator build root to use rhel-9 golang 1.20 image for OCP 4.15

### DIFF
--- a/.ci-operator.yaml
+++ b/.ci-operator.yaml
@@ -1,4 +1,4 @@
 build_root_image:
   name: release
   namespace: openshift
-  tag: rhel-8-release-golang-1.19-openshift-4.14
+  tag: rhel-9-release-golang-1.20-openshift-4.15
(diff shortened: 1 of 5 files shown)


adambkaplan:
/assign @ayushsatyam146 

We discussed building s2i with Konflux earlier today. I had previously started onboarding, but ran into an issue pulling the base images in this Dockerfile (which reference the OpenShift CI registry). This updates the Dockerfiles to use UBI images, which can be pulled from the Red Hat container catalog. The only limitation is creating a Red Hat account, which anyone can do using the Developer subscription.

PR #1172, opened 2024-09-28
Use UBI8 based base image and builder image

sayan-biswas:
Changes:
- Use UIB8 based base imgae
- Use go-toolset from UBI8

Currently comet repo exists for rhel-8 based images, so to onboard to konflux easily we'll go ahead with rehl-8 image now. When comet repo for rhel-9 is configured, we'll switch to rehl-9 based images.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@
 /.atom-build.yml
 /.project
 /.vagrant
+/.idea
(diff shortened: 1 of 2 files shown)


ramessesii2 (approved):
/lgtm

PR #1191, opened 2024-11-07
BUILD-1165: Updating s2i to golang 1.22

adambkaplan:
Updating the Dockerfile and go.mod to require golang 1.22. Changing the go runtime also requires a bump to the s2i minor version.

Dockerfile has also been updated to use the "patch" version of the UBI base images. This ensures we have reproducible builds, since the "patch" tags from Red Hat are effectively immutable. Konflux should also provide updates to the patch version through Renovate.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,8 +1,8 @@
-FROM registry.redhat.io/ubi8/go-toolset:1.21 AS builder
+FROM registry.redhat.io/ubi8/go-toolset:1.22.7-5 AS builder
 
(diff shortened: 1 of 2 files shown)


adambkaplan:
/hold

Frankly, our unit and integration tests should be moved to Konflux so we don't have to do the silly dance of updating openshift/release.

ayushsatyam146:
@sayan-biswas this can be closed now

sayan-biswas:
Closing this since #1197 includes these changes.

PR #1197, opened 2025-01-03
BUILD-1165: Updating Dockerfile to us golang 1.22 image

ayushsatyam146:
Mirror of https://github.com/openshift/source-to-image/pull/1191
Updating the Dockerfile to require golang 1.22. Golang version was already updated to 1.22.5

Dockerfile has also been updated to use the "patch" version of the UBI base images. This ensures we have reproducible builds, since the "patch" tags from Red Hat are effectively immutable. Konflux should also provide updates to the patch version through Renovate.


sayan-biswas (approved):

sayan-biswas:
/override "Red Hat Konflux / source-to-image-enterprise-contract / source-to-image"

outdate tasks

PR #1242, opened 2025-06-13
feat: Default to Podman Socket

### LINKED ISSUE #1163: s2i is using docker and not podman
Is there a way to get s2i to use podman instead of docker?

adambkaplan:
Update s2i's defaults to inspect the host system for Podman sockets. Podman socket provides support for the Docker v1.40 API, which meets s2i's needs. The logic looks for the podman socket in the following order:

1. Rootless podman socket
2. Root podman socket
3. Docker socket

If none are found, the default falls back to the docker socket. As before, the `--url` and `DOCKER_HOST` environment variables can override the container engine socket location.

Fixes #1163

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,20 @@ ready-to-run images by injecting source code into a container image and letting 
 
 For a deep dive on S2I you can view [this presentation](https://www.youtube.com/watch?v=flI6zx9wH6M).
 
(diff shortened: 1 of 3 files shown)


adambkaplan:
@psrvere please take a look - this will let us (_finally_) remove the hard dependency on Docker!

PR #1250, opened 2025-07-01
chore: use minor version tags for docker images

avinal:
- using minor version tags for dockerfile ensures that the images
are built using latest update

Signed-off-by: Avinal Kumar <avinal@redhat.com>


sayan-biswas:
correction: minor version tag

PR #1405, opened 2026-06-01
BUILD-992: Use `COPY --chown`

adambkaplan:
Use `COPY --chown` when generating the s2i Dockerfile. This feature has been part of the Dockerfile "standard" since v1.0. Nearly all tools that process a Dockerfile are capable of supporting this flag - any prior versions of tools such as docker, podman, or buildah are long past end of life.

Using `COPY --chown` has multiple advantages, particularly if the container is built by Podman or buildah:

- No need to switch user to `root` and execute `RUN chown`
- Fewer layers in the output image due to fewer steps
- Podman/buildah makes the appropriate syscalls to chown the copied files outside of the "build container" sandbox.

Assisted-by: Claude Code


adambkaplan:
/hold

I want to validate that this doesn't break OCP BuildConfig

adambkaplan:
/hold cancel

Backporting to BuildConfig proved more difficult because we are using a newer version of golang.

adambkaplan:
Holy shirtballs, this is effectively a revert of https://github.com/openshift/source-to-image/pull/913!

I have zero memory of doing this.

